Issue #9, opened 2016-01-12
Badges don't disappear when clicking on the chat


stefuNz:
Hi,

the badges in the Mac OS X dock (didn't try windows) don't disappear when i click on the chat which has new messages, only when i leave this chat (switch to another room)

stefuNz:
Same goes for windows as my colleague has tested it just now.

yuya-oc:
It's a side effect of solution for #2. As you said, badges disappear by switching to another channel.

I think v1.0.1 behavior is better than '<=v1.0.0'. Because it has been important that "Are there new messages while I 'm away from the PC?" in my daily work. 

If you have any idea for this, please let me know. Of course PR is welcome.

Johnich:
Same issue here. This 'fake unread message' makes my teammates throw out lots of complaints. Hope it will be rolled back in next version.

yuya-oc:
The fix is WIP at #22. For now, I think you can overwrite `src/browser/webview/mattermost.js` of v1.0.2 with v1.0.0 version (I can't test it now).

yuya-oc:
v1.0.3 clears badges for active channel when the tab(or window) is focused.

jobspierings:
@yuya-oc this fix works great on OS X.

PR #12, opened 2016-01-14
Count for active channel only when window is not focused.

mid0111:
Make unread count of active channel disappear when the channel is focused. (issue #9)
Because I don't have test environment, I test only at Mac OSX.

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -23,7 +23,8 @@ var MainPage = React.createClass({
   getInitialState: function() {
     return {
       key: 0,
-      unreadCounts: new Array(this.props.teams.length)
+      unreadCounts: new Array(this.props.teams.length),
+      unreadAtActive: new Array(this.props.teams.length)
     };
   },
   componentDidMount: function() {
@@ -31,6 +32,8 @@ var MainPage = React.createClass({
     var focusListener = function() {
       var webview = document.getElementById('mattermostView' + thisObj.state.key);
       webview.focus();
+
+      this.handleOnTeamFocused(key);
     };
 
     var currentWindow = remote.getCurrentWindow();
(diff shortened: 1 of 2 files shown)


yuya-oc:
Why did you add `active` to the arg for `getElementsByClassName()` ?
I checked with Mattermost 1.4 RC2, but `active` class was not contained in the separator.

mid0111:
Sorry, it's my mistake. I am ashamed of myself.

I fixed the mistake, and checked unread count of active channel appeared when it was not focused, and disappeared when it was focused.
But appeared again, after it was not focused again.

So, don't be care about previous commit.
I'll check this later ...

yuya-oc:
JFYI, circleci failed because apt-get failed to download `zip` in `docker build`. It's not your mistake.

yuya-oc:
If the app manages multiple teams, unread counts of active channel disappear at inactive tabs when it's focused.
I think we have to resolve this issue to add features of the PR.

mid0111:
I agree with you.


I don't have any idea how to count unread title for active channel from DOM.

I'm trying to count unread title from notification event of active channel.
But, this approach will make the application complex.
I wonder if there is another good way.

mid0111:
I changed to count unread message of active channel from the desktop notification event.

Then, check about belows at Mac OSX and windos 7.
* Active channel badge disappear when the active channel is focused.
* Active channel badge does not disappear when the another tab is focused.

yuya-oc:
Thanks a lot, however I want to merge #18 first. Because this PR makes large changes for code structure.

mid0111:
OK, I think so too.

yuya-oc:
@mid0111 @ttyniwa Could you check whether app performs as you expect?

ttyniwa:
ok. I'll check it.

mid0111:
Thank you for marge PR.
The unread count is appeared I expected.

However it is not the interest on this issue, I’m worried about the mention count was not disappeared when clicking on the application.

ttyniwa:
I fixed bugs, please see #12 for details.

PR #21, opened 2016-01-23
Allow hiding menu bar on Linux

alerque:
The UI feature to auto-hide the menu bar works on Linux as well as
Windows. Don't ignore it if the config value is set.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -61,7 +61,7 @@ app.on('window-all-closed', function() {
 
 // For win32, auto-hide menu bar.
 app.on('browser-window-created', function(event, window) {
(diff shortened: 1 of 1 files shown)


yuya-oc:
Thank you! I wasn't able to decide adding it in previous verions because Linux has some different windows managers.

alerque:
The default config option is off and the user will need to manually poke their config to turn it on anyway, so I think it's safe to leave this up to the user and obey the config value if present. In my testing it behaves about the same as option does in Atom: the bar is hidden until you hit <kbd>Alt</kbd>

PR #33, opened 2016-02-15
Add default Mac application menu

pascalw:
I noticed the default app menu on Mac was missing so I added it.

All Mac apps generally have this menu, allows to hide the app with the
default shortcut.


yuya-oc:
Thanks! Could you run `npm run prettify`? Codes will be formatted.

pascalw:
@yuya-oc done :)

PR #43, opened 2016-02-24
Minimize to tray only and hide the Taskbar Icon

sinky:


yuya-oc:
Please provide information about environment you tested.

Besides, I can't merge this now because there is no way to know whether there are new messages/mentions.

sinky:
Tested on Windows 7 x64.
The Ballon Notification works. Or what did you mean with "new messages"

yuya-oc:
The balloon is one-time, it will be disappear. Currently, the app shows a red overlay on the taskbar icon when there are new messages. It works permanently.

For example, on Windows, the tray icon should indicate the state.

sinky:
Ok, i think we could just change the tray icon or add an second trayicon.
I Think the trayicon is too small to add an overlay badge counter like the taskbar icon.

This changes the trayicon to the red badge, but i dont know how to reset of count goes 0.
```
// Set overlay icon from dataURL
ipc.on('win32-overlay', function(event, arg) {
  var overlay = electron.nativeImage.createFromDataURL(arg.overlayDataURL);
  mainWindow.setOverlayIcon(overlay, arg.description);
+++  trayIcon.setImage(overlay);
});
```

yuya-oc:
It's too easy way. It has a problem that users can't know the badge is related to this app.
And sorry for forget to check, I think following behavior is better. Slack app is same.
* when minimize: minimize window (same as current)
* when close: hide window and taskbar icon (current: minimize window)

Sure, the tray icon is too small, so I'm considering add a small dot to indicate "there are new messages". Could you wait this feature?

sinky:
Of course, the my obove code is a really simple example.

Slack like "Minimize to Taskbar" and "Close to Tray" (and quit through traymenu) is good compromise. Personally i would prefer a "tray only" messenger. Perhaps through an option when you added a trayicon indication.

yuya-oc:
Yes, I need to prepare an option. I like "taskbar icon" messenger e.g. skype.

jmreynolds:
Ah! You beat me to this by just a few hours!

Oh well, your implementation looks better. Mine just had a checkbox for selecting the configuration setting - it didn't watch to see if you had minimized it away and couldn't get back to it.

sinky:
I think i close this PR an add an issue.

PR #46, opened 2016-02-25
Added a skipTaskbar configuration

jmreynolds:
One of my clients is using this as their primary means of communication, and I noticed that I couldn't hide it to the notification area. So I had a little time free today due to a storm knocking out my internet, and decided I should fix it.

I think I stayed within your coding standards. I added a test for the feature. I tried to add a test around whether the BrowserWindow was correctly reading the skipTaskbar config, but there seems not to be any such property. I may take another stab this weekend, and come at it from another direction.

Let me know if there's anything I can do to help out or improve!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ npm-debug.log
 
 test-results.xml
 test_config.json
+.idea/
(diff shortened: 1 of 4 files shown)


jmreynolds:
And of course, right after I submitted this, I saw that someone else had done it a couple hours after I forked earlier today. I'll close my pull request.

PR #69, opened 2016-03-11
Fixing link to GIF in README

it33:
I accidentally linked it to the branch in the first PR :p


yuya-oc:
I had overlooked, too. How about using relative path?

it33:
I'm not sure if relative path will work with embedded image files because we need to reference the raw version?

yuya-oc:
9d5b1a8f986e0f86c849f06f10e2ed8956bce96d looks fine. It can be also previewed by atom editor.

it33:
Thanks @yuya-oc. It does work, I've updated the link. Thanks for the tip!

yuya-oc:
The test looks still unstable...

PR #84, opened 2016-03-17
Consistent icon

### LINKED ISSUE #82: OS X: icon doesn't respect OS grid
OS X has design guidelines that use consistent sizes and appearance (subtle to-down gradient). See [Inspecting Yosemite's Icons](http://martiancraft.com/blog/2014/07/inspecting-yosemite-icons/)

`electron-mattermost` doesn't respect those and jarringly stands out as unpolished.

screen shot 2016-03-17 at 10 13 02

Feel free to use, hack or take inspiration from Matterfront's icon (which respects both internal and external rings as well as the ambient lightning gradient). Alternatively, use the [icon grid](http://martiancraft.com/downloads/20140703_yosemite_Grid-PSD.zip).

lloeki:
I don't want to overstep and just arbitrarily jump into the project, so this is to get the ball rolling on the subject because this one is nagging me to no end each time I pull up the Dock!

This gets the icon respecting the OS X grid so it's sized properly on the dock, and also consistent with the iOS and Android icons. 

Fixes #82.


yuya-oc:
Thank you, I also feel it's worse. But I couldn't find out the basis of the grid. 
However the Mattermost brand guidelines request enough space around the logo. So I think these icons should not be used as it is.

lloeki:
> However the Mattermost brand guidelines request enough space around the logo

This is something that I considered carefully, but the end decision is definitely not mine :smile:.

For the sake of completeness, here's my reasoning:

[Brand guideline](https://www.mattermost.org/brand-guidelines/ ) says `x/5` whitespace where `x` is the solid part width. Guideline doesn't specify whether "whitespace" counts to painted area or to bounding box (i.e includes or excludes unpainted area).

guidelinesi

OS X grid has either a ratio of ~22 instead on the 3rd ring from center (~8.2 when counting to the box), or a factor of ~3.3 on the second ring from center (~2.3 when counting to box).

screen shot 2016-03-17 at 13 45 47

The one that was the most aesthetically pleasing, the most consistent at various scales, and the most identifiable to the Mattermost logo was the larger one, as the smaller one was just having too much whitespace, and also due to the additional unpainted space to the bounding box.

Additionally, many reference icons [such as favicons](https://pre-release.mattermost.com/static/images/favicon/favicon-16x16.png) don't follow the guideline to the letter (which is fair since after all, it's "merely" a guideline). As for platform consistency, the larger one sits very well alongside Safari and iTunes icons.

I can redo one that fits the 2nd ring as a comparison, and also one with x/5 whitespace.

lloeki:
The above pertains to OS X, here are the current [Windows guidelines](https://msdn.microsoft.com/en-us/windows/uwp/controls-and-patterns/tiles-and-notifications-app-assets).

yuya-oc:
Thank you for description and feedback at core channel. The icon is replaced with #85.

BTW, the Windows guidelines should be applied for UWP app (and Modern UI app), but this app is a traditional desktop application. So I think we can't follow all of the guidelines. We have to create new icons when want to follow some of it.

Issue #56, opened 2016-03-09
Add support for basic HTTP authentication


PiotrZierhoffer:
Electron supports it, as in https://github.com/atom/electron/pull/3250

I added this configuration to app/main.js and it worked, but I'd love having a default with a popup window. Keeping my password unencrypted on hdd is no fun.

Is that even possible with electron?

yuya-oc:
Maybe possible, the API has been just tested at dca9b62cda34025124195de6c31f0f961f19267a.
But Electron doesn't support input dialog such as `window.prompt`, so the app has to implement on its own.

mgielda:
Let me jump in here, I had solved the problem in a not-so-graceful but effective way once for a different app, perhaps this experience can come in handy (did spend some hours testing various libraries which refused to work before coming up with this):

```
var app = require('app');
var BrowserWindow = require('browser-window');

var prompt = require('sync-prompt').prompt;

var name = prompt('Username: ');
var password = prompt.hidden('Password: ');

app.on('login', function (event, webContents, request, authInfo, callback) {
  event.preventDefault();
  console.log("Replacing default auth behaviour.");
  callback(name, password);
});
```

yuya-oc:
Thank you for feedback. I think It will work, however users have to use command-line and keep it opened.

Note: `callback()` can be called asynchronously, so `sync-prompt` might not be necessary. Then, `sync-prompt` is a native module, so it's a little hard to compile release version. (In my environment, Windows 10 + Node 4.2.2 + VS2015 Update1, I could not compile `sync-prompt`.)

yuya-oc:
`readline-sync` might be nice for @mgielda's example. It looks JavaScript.

mgielda:
@yuya-oc indeed it works (sorry it took some time to answer) - so I agree, the updated snippet is:

```js
var readlineSync = require('readline-sync'); //tested on readline-sync 1.4.1

var username = readlineSync.question('Username: ');
var password = readlineSync.question('Password: ', {
  hideEchoBack: true // The typed text on screen is hidden by `*` (default). 
});

app.on('login', function (event, webContents, request, authInfo, callback) {
  event.preventDefault();
  console.log("Replacing default auth behaviour.");
  callback(username, password);
});
```

How should we proceed with that? There needs to be some config option for this I suppose. I agree that doing a graphical version later is a good idea but for now this would be good enough. Should we perhaps propose a pull request?

yuya-oc:
[...]

PR #100, opened 2016-03-28
Add "httpBasicAuth" option

mgielda:
As per discussion in #56 

Tested on Linux 64-bit (Ubuntu 15.10 to be precise).
Mattermost Desktop version 1.07, Mattermost server version 1.4 
I did not create any tests since there aren't any for the other options. I would appreciate an example :)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -58,7 +58,7 @@ if (config.httpBasicAuth) {
     hideEchoBack: true // The typed text on screen is hidden by `*` (default). 
   });
 
-  app.on('login', function (event, webContents, request, authInfo, callback) {
+  app.on('login', function(event, webContents, request, authInfo, callback) {
     event.preventDefault();
     console.log("Replacing default auth behaviour.");
     callback(username, password);
--- a/src/main.js
+++ b/src/main.js
@@ -49,6 +49,23 @@ var mainWindow = null;
 var trayIcon = null;
 var willAppQuit = false;
 
+if (config.httpBasicAuth) {
+
+  var readlineSync = require('readline-sync');
+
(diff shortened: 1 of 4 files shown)


yuya-oc:
Thanks. Please note that these codes will be changed in future release because this is an ad hoc feature as I said at #56.

src/main.js
```diff
@@ -49,6 +49,23 @@ var mainWindow = null;
...
+  var readlineSync = require('readline-sync');
+
+  var username = readlineSync.question('Username: ');
+  var password = readlineSync.question('Password: ', {
+    hideEchoBack: true // The typed text on screen is hidden by `*` (default). 
+  });
+
+  app.on('login', function(event, webContents, request, authInfo, callback) {
+    event.preventDefault();
+    console.log("Replacing default auth behaviour.");
```
yuya-oc: Why don't you use the prompt in the 'login' event? It appears when it's needed.

In this commit, the app doesn't show anything when the app is executed by double-click. Users must know that command-line is necessary to use this feature.

yuya-oc: like:
```js
app.on('login', function(event, webContents, request, authInfo, callback){
  event.preventDefault();
  var username = readlineSync.question('Username: ');
  var password = readlineSync.question('Password: ', {
    hideEchoBack: true // The typed text on screen is hidden by `*` (default). 
  });
  callback(username, password);
});
```

Besides, in the case, the config might not be necessary.

mgielda:
I will create a new pull request as your solution is obviously better. It did not work for me with previous, asynchronous solutions but obviously as it is now just adding those few lines works much better.

PR #114, opened 2016-04-20
Tell users where to place the application directory

jeremycook:
Using "Extract All..." with the default options will extract the files to the Downloads folder. Many users will do exactly that and end up running Mattermost from there downloads folder. Not a good thing. These instructions tell users where to extract the files.

Perhaps in a future release these instructions can be reduced to run a batch script or installer.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -8,11 +8,16 @@ To set up the Mattermost desktop application for Windows 7, 8 and 10:
 
    Find the file ending in `-win64.zip` if you're running a x64 version of Windows and `-win32.zip` if you're running an older 32-bit version.
 
-2. From the `/Downloads` directory right-click on the file and select "Extract All..."
+2. From the `\Downloads` directory right-click on the file and select "Extract All..."
 
-   A new directory should be created on your PC.
+   - Change the extract folder from `C:\Users\(currentuser)\Downloads\mattermost-desktop-1.1.1-win64` to `C:\Users\(currentuser)\AppData\Local` where `(currentuser)` will be the name of your user account.
+   - Check the "Show extracted files when complete" checkbox.
+   - Click the "Extract" button.
+   - Look for the new application directory at `C:\Users\(currentuser)\AppData\Local\mattermost-desktop...`.
+   - Remove the version number by renaming the `mattermost-desktop...` application directory to `mattermost-desktop`.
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -8,11 +8,16 @@ To set up the Mattermost desktop application for Windows 7, 8 and 10:
 
    Find the file ending in `-win64.zip` if you're running a x64 version of Windows and `-win32.zip` if you're running an older 32-bit version.
 
-2. From the `/Downloads` directory right-click on the file and select "Extract All..."
+2. From the `\Downloads` directory right-click on the file and select "Extract All..."
 
-   A new directory should be created on your PC.
+   - Change the extract folder from `C:\Users\(currentuser)\Downloads\mattermost-desktop-1.1.1-win64` to `C:\Users\(currentuser)\AppData\Local` where `(currentuser)` will be the name of your user account.
```
yuya-oc: It's specific for v1.1.1 and win64. So could you use a description like `...`?

GhostLyrics:
Are you recommending installing to `AppData` instead of `…Programs` due to the fact that you don't need admin privileges?

yuya-oc:
I don't care, so no. The app should work at the desired directory.
btw, I usually create `D:\Program Files\...` for this kind of apps. Where is an appropriate directory for the app which has no installer?

jeremycook:
@yuya-oc This commit is too master so I hadn't considered it specific to anyone version, which is why I continued using the pre-existing convention of dot-dot-dot (...). Providing an example folder name isn't a bad idea.

@GhostLyrics and @yuya-oc Since there is no installer at this time, the target audience I have in mind for these instructions are users who may not know where to place application files and likely have no idea why you might choose one folder over another. This is why I opted for \AppData\Local over Program Files or Program Files (x86). With Program Files and Program Files (x86) I would have to explain if you are on a 64-bit machine installing the 32-bit version then place it in Program Files (x86), if you are on a 64-bit machine installing the 64-bit version then use the Program Files folder, if you are installing the 32-bit version...blah blah blah. Too much info. Using \AppData\Local also means if multiple users are sharing a machine one user messing up their install (for whatever reason) won't break another user's install of Mattermost. Install problems will be localized.

More advanced users or administrators will feel comfortable ignoring the advice here and use Program Files or Program Files (x86) when that makes sense, and take steps to add shortcuts to everyone's start menu, desktop, etc. which I believe does require admin privileges.

yuya-oc:
@jeremycook I see.

Issue #88, opened 2016-03-19
Feature request: show tray icon and unread new message badge on Linux tray


smetj:
Should show a tray icon on the bar, and show the number of unread new messages on the tray icon when new messages arrived.

Similar request to #45 for for Linux/Gnome instead.

yuya-oc:
Could you provide some example in Gnome?

greenik:
@yuya-oc Any chance that this will be done?

smetj:
@yuya-oc 
![blah](https://cloud.githubusercontent.com/assets/1007219/14811822/56fee4f6-0b9a-11e6-9935-4e2f1a172a13.png)

edurenye:
It must also work in ubuntu unity, I guess should be a way to add it and make it work everywhere.

rpglover64:
For the record, I'm using i3 with no desktop environment and with i3bar, and I don't see any tray icon.

Great-Antique:
+1 for KDE

PR #128, opened 2016-05-08
Linux tray icon

jnugh:
This adds support for Tray icons on linux (issue #88).
Tested with Unity, KDE and Gnome.

On KDE plasma 5, the icon looks pretty messed up [because electron uses a legacy approach to render the icons](http://blog.davidedmundson.co.uk/blog/xembed_back), instead of SNIs.

I added two different themes, dark and light, one of them should look good on most desktop environments. The dark version is simply a copy of the Mac icons.

To support themes, I added a select field in the settings dialog and added a new property to the config.
I didn't bump the version number - don't know about your policy here - I wrote my code so that it silently defaults to light when `trayIconTheme` is undefined.

In `showUnreadBadgeLinux` I commented the 'badge code' because those badges are [not yet supported by electon](https://github.com/electron/electron/issues/4193) (I plan to work on that).


yuya-oc:
Thanks! I have some questions.

- Could you show me screenshots? It's a little hard to prepare all environment.
- On Ubuntu 14.04, I couldn't see the tray icon. How can I confirm it?

Then, now preparing v1.2.0 release, so this might keep you waiting.

jnugh:
Sure, I can provide some screens:
Gnome 3: Mentioned
Gnome mentioned
Gnome 3: standard icon
Gnome standard icon

Ubuntu 14.04 in this case, but it looks the same on 16.04:
Mentioned:
Ubuntu mentioned
Unread:
Ubuntu unread
Standard:
Ubuntu

It didn't work for you because you need to install `libappindicator1` for this to work. (A lot of applications depend on this lib, so I didn't notice).

KDE Plasma 5.6, 5.5 looks pretty bad, because it has problems handling alpha values in legacy tray icons:
KDE
Mention:
KDE Mention
Unread:
KDE unread

PR #132, opened 2016-05-12
Propose updating text for mixed content option

it33:
When this option is enabled using Mattermost on desktop and web has about the same risks. 

Therefore, propose we use the same defaults and also similar language? 

It would make the application potentially more usable by non-technical people. 

In future, this option might be pre-configured by an IT admin, and they can choose which default to use if it's an important part of their system security.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ This release contains a security update and it is highly recommended that users 
 
 ### Improvements
 - Improve the style for tab badges.
-- Add **Allow insecure contents** option to render images with `http://`.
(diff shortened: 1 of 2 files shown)


it33:
@yuya-oc, wondering your thoughts on this change?

yuya-oc:
I see . But it seems that the option is always "on". So could you modify only `src/common/settings.js`? To keep user's choice, the jsx should not be changed.

jwilander:
Thanks @yuya-oc, I'll submit a PR for this.

yuya-oc:
@jwilander Just in case, of course the message in jsx can be changed.

PR #138, opened 2016-05-13
Implement installers via electron-builder

Razzeee:
Hey,

this might be totally wrong, as I'm a fist time node user. And so far I could only test win.
The installation of the exe seems to work fine, but there is no shortcut anywhere to be found. Uninstall also seems to work fine.

Would be very cool, if other people could build it for osx/linux and let the maintainer know if it works.

Cheers

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "package:osx": "gulp package:osx",
     "package:linux": "gulp build && build --platform linux --arch all -d deb",
     "package:all": "gulp package:all",
-    "prettify": "gulp prettify"
+    "prettify": "gulp prettify",
+    "installer": "node ./script/installer.js"
   },
   "devDependencies": {
     "babel-core": "^6.7.5",
@@ -33,10 +34,12 @@
     "babel-preset-react": "^6.5.0",
     "chromedriver": "^2.20.0",
     "del": "^2.2.0",
-    "electron-builder": "^3.11.0",
+    "electron-builder": "3.20.0",
     "electron-connect": "^0.3.7",
     "electron-packager": "^7.0.1",
     "electron-prebuilt": "0.37.8",
+    "electron-squirrel-startup": "^1.0.0",
+    "electron-winstaller": "^2.2.0",
     "esformatter": "^0.9.3",
     "esformatter-jsx": "^5.0.0",
     "gulp": "^3.9.0",
@@ -47,6 +50,7 @@
     "json-loader": "^0.5.4",
     "mocha": "^2.3.4",
     "mocha-circleci-reporter": "0.0.1",
+    "rimraf": "^2.5.2",
     "should": "^8.0.1",
     "style-loader": "^0.13.0",
     "through2": "^2.0.1",
@@ -56,11 +60,13 @@
     "webpack-stream": "^3.1.0"
   },
   "build": {
+    "app-bundle-id": "com.mattermost.desktop",
+    "app-category-type": "public.app-category.productivity",
     "linux": {
       "synopsis": "Mattermost Desktop"
     }
   },
-  "directories":{
+  "directories": {
     "buildResources": "resources",
     "app": "dist",
     "output": "release"
--- a/script/installer.js
+++ b/script/installer.js
@@ -1,0 +1,38 @@
+#!/usr/bin/env node
+
+const createWindowsInstaller = require('electron-winstaller').createWindowsInstaller
+const path = require('path')
+const rimraf = require('rimraf')
+
+deleteOutputFolder()
+  .then(getInstallerConfig)
+  .then(createWindowsInstaller)
+  .catch((error) => {
+    console.error(error.message || error)
+    process.exit(1)
+  })
+
+function getInstallerConfig() {
+  const rootPath = path.join(__dirname, '..')
+  const outPath = path.join(rootPath, 'release')
+
+  return Promise.resolve({
+    appDirectory: path.join(outPath, 'Mattermost-win32-x64'),
+    iconUrl: 'https://raw.githubusercontent.com/mattermost/desktop/master/resources/icon.ico',
+    //loadingGif: path.join(rootPath, 'assets', 'img', 'loading.gif'),
+    noMsi: true,
+    outputDirectory: path.join(outPath, 'windows-installer'),
+    setupExe: 'Mattermost.exe',
+    setupIcon: path.join(rootPath, 'resources', 'icon.ico'),
+    skipUpdateIcon: true,
+    exe: 'Mattermost.exe'
+  })
+}
+
+function deleteOutputFolder() {
+  return new Promise((resolve, reject) => {
+    rimraf(path.join(__dirname, '..', 'out', 'windows-installer'), (error) => {
+      error ? reject(error) : resolve()
+    })
+  })
+}
--- a/src/main.js
+++ b/src/main.js
@@ -2,6 +2,9 @@
 
 const electron = require('electron');
 const app = electron.app; // Module to control application life.
+
+if (require('electron-squirrel-startup')) app.quit();
+
 const BrowserWindow = electron.BrowserWindow; // Module to create native browser window.
 const Menu = electron.Menu;
 const Tray = electron.Tray;
--- a/README.md
+++ b/README.md
@@ -86,5 +86,10 @@ $ npm run package:linux
 $ npm run package:all (Packages for all platform)
 ```
 
+Create a windows installer with the following command. It will appear in the `release\windows-installer` directory.
+```
+$ npm run installer
+```
+
 ## Contributing
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md).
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -56,9 +59,11 @@
     "webpack-stream": "^3.1.0"
   },
   "build": {
-    "linux": {
-      "synopsis": "Mattermost Desktop"
-    }
+  "app-bundle-id": "org.mattermost.desktop",
```
Razzeee: @yuya-oc this might be something you want me to change

package.json
```diff
@@ -25,15 +24,19 @@
...
+    "dist:win64": "build --platform win32 --arch x64"
   },
   "devDependencies": {
     "babel-core": "^6.7.5",
     "babel-loader": "^6.2.4",
     "babel-preset-react": "^6.5.0",
     "chromedriver": "^2.20.0",
     "del": "^2.2.0",
-    "electron-builder": "^3.11.0",
+    "electron-builder": "3.20.0",
```
Razzeee: Fixed version due to https://github.com/electron-userland/electron-builder/issues/402

package.json
yuya-oc: Please keep indent.

@it33 Mattermost has two kind of domains. Which is better, `.org` or `.com`?

package.json
```diff
@@ -8,13 +8,12 @@
...
     "email": "yuya0321@gmail.com"
   },
-  "license": "MIT",
   "engines": {
     "node": ">=4.2.0"
   },
   "scripts": {
     "install": "cd src && npm install",
-    "postinstall": "npm run build",
+    "postinstall": "install-app-deps",
```
yuya-oc: This one is not necessary because dependencies are prepared by `npm run build`.

yuya-oc:
The builder seems to work with my Ubuntu 14.04. But it doesn't meet with our CI process (You don't have to take care yet). I saw following behavior about installation.

1. The installer launches the app.
2. There are no shortcuts as you said.
3. The uninstaller also launches the app.

I'm also beginner for `electron-builder`, but it seems that some codes are necessary to handle shortcuts. So could you check below?
- https://github.com/electron-userland/electron-builder#auto-update
- Where is the actual app binary installed into?

Then, if you want to continue this PR, please check Mattermost CLA. http://www.mattermost.org/mattermost-contributor-agreement/

Razzeee:
Sorry, I forgot to add the build folder. We would also need a background.png for osx I think? Not sure what to do there.

Yes, I'm aware of it not taking into account your ci setup right now.

3.) Yes uninstaller also opens the app, takes a few seconds then it closes and is uninstalled. That's whats happening with windows.

I'll try to look into the shortcuts again.
On windows it's installed to the userspace so C:\Users\razze\AppData\Local\mattermost for example.

CLA signed, thanks for pointing me towards it.

Razzeee:
The problematic part about me doing the autoupdate stuff is, that I don't know if or how you will handle the server part. If I can assume, that we're using https://github.com/GitbookIO/nuts and heroku, I can probably come up with something.

Razzeee:
Well it seems like you will need to do the server stuff. I don't have these data :)
image

yuya-oc:
Thanks. For now, we don't have to think about the update server. I think that the point of handling Squirrel event is create/remote shortcuts on the event. If there is no server, the updater merely fails. So I think that's not important matter. Of course, we will have to handle updater and server to provide auto-update.

script/installer.js
```diff
@@ -0,0 +1,38 @@
...
+  })
+
+function getInstallerConfig () {
+  const rootPath = path.join(__dirname, '..')
+  const outPath = path.join(rootPath, 'release')
+
+  return Promise.resolve({
+    appDirectory: path.join(outPath, 'Mattermost-win32-ia32'),
+    iconUrl: 'https://raw.githubusercontent.com/mattermost/desktop/master/resources/icon.ico',
+    //loadingGif: path.join(rootPath, 'assets', 'img', 'loading.gif'),
```
Razzeee: It would be nice if you could get somebody to provide a now installing splash screen. See https://github.com/electron/electron-api-demos/blob/master/assets/img/loading.gif as an example

Razzeee:
Is it correct that this PR is against master? Or should it be dev?

Finally got it to work on windows, haven't tested any other OS. Implementing auto update should be trivial when we get a server. This work is heavily based on: https://github.com/electron/electron-api-demos

You might want to review this with whitespaces off:
https://github.com/mattermost/desktop/pull/138/files?w=0

package.json
it33: .com please since desktop apps will eventually be hosted at releases.mattermost.com, sorry for late response for some reason I didn't see this mention

Razzeee: Thank you. Changed.

src/main.js
```diff
@@ -9,293 +9,316 @@ const ipc = electron.ipcMain;
...
+      if (index !== 0) {
+        event.preventDefault();
+      }
+    });
   });
-});
+}
+
+// Handle Squirrel on Windows startup events
+switch (process.argv[1]) {
```
yuya-oc: electron-winstaller recommends that the event handler should be placed at the top of `main.js`. https://github.com/electron/windows-installer#handling-squirrel-events

Then, please `return` after `app.quit()` to suppress the other events of `app`.

package.json
```diff
@@ -56,9 +59,11 @@
     "webpack-stream": "^3.1.0"
   },
   "build": {
-    "linux": {
-      "synopsis": "Mattermost Desktop"
-    }
+  "app-bundle-id": "com.mattermost.desktop",
+  "app-category-type": "public.app-category.productivity",
+  "linux": {
```
yuya-oc: Could you keep original indent?

script/installer.js
```diff
@@ -0,0 +1,38 @@
...
+    console.error(error.message || error)
+    process.exit(1)
+  })
+
+function getInstallerConfig() {
+  const rootPath = path.join(__dirname, '..')
+  const outPath = path.join(rootPath, 'release')
+
+  return Promise.resolve({
+    appDirectory: path.join(outPath, 'Mattermost-win32-ia32'),
```
yuya-oc: Please consider also win32-x64.

src/main.js
yuya-oc: Did you check [mongodb-js/electron-squirrel-startup](https://github.com/mongodb-js/electron-squirrel-startup)? It's introduced on README.md of electron-winstaller.

yuya-oc:
> Is it correct that this PR is against master? Or should it be dev?

Good question. master is okay. 

JFYI, now there are two active branches, master and dev. They almost follows git-flow. However recently I feel that it's not simple for pull-requests. So the branching model will be a coming issue.

script/installer.js
Razzeee: I somewhere read, that win64 is not supported. But I might mix up stuff, as I have read pretty much every readme of every installer project. 
Edit: Found it and remembered it the wrong way around. Only 64 bit it is. https://www.npmjs.com/package/electron-builder#auto-update

Razzeee:
@yuya-oc well you might want to update contributing.md to help with the branch choice

package.json
Razzeee: sure, if this wasn't the prettify task

src/main.js
Razzeee: Well as everything above is a function, this will be called as the first thing. But if you prefere, I can move it to the top, just let me know.

script/installer.js
yuya-oc: It seems that electron-winstaller can handle x64 app because of the sample written on README.md. I'm using x64 Windows, so I will check later.

src/main.js
yuya-oc: Please move to the top.

Sure, this part will be called at the first, but events are already registered into app module. So I'm afraid of that some issues happen by depending its timing or other reason. The top of `main.js` looks proper to avoid any other processing.

script/installer.js
Razzeee: Well it works fine, I changed back to 32 bits as one of the last steps.

Razzeee:
1) Updated to use `mongodb-js/electron-squirrel-startup`
2) Fixed indent
3) Only build installer from 64bit
4) Small update for the copyright year

package.json
```diff
@@ -25,18 +25,21 @@
...
     "babel-loader": "^6.2.4",
     "babel-preset-react": "^6.5.0",
     "chromedriver": "^2.20.0",
     "del": "^2.2.0",
-    "electron-builder": "^3.11.0",
+    "electron-builder": "3.20.0",
     "electron-connect": "^0.3.7",
     "electron-packager": "^7.0.1",
     "electron-prebuilt": "0.37.8",
+    "electron-squirrel-startup": "^1.0.0",
```
yuya-oc: This must be installed with `src/package.json` to bundle.

yuya-oc:
Just curious, what does your "works fine" mean? I have never seen any shortcuts, is it correct?

Environment:
  Windows 10 x64
  Node v4.2.2, npm 2.14.7 (maybe this is not matter)

Razzeee:
image


Shortcuts work fine for me, but its under the Author name so in the folder Yuya Ochiai.

Enviroment is Win10 x64

yuya-oc:
I had overlooked the folder, thank you!

PR #145, opened 2016-05-20
Use different colors for unread/mentioned in windows task bar. 

Razzeee:
Like we do for tray icons.

As it's very confusing for windows users, why the trayicon can have a different colored dot, then the taskbar icon.
I also tried to mimic the black boarder around the dot.

One thing that's not matching is the color. Should I use the color you use in the taskbar dots or the material colors, like you used for the red before?

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -461,10 +461,10 @@ var showUnreadBadgeWindows = function(unreadCount, mentionCount) {
   };
 
   if (mentionCount > 0) {
-    const dataURL = badge.createDataURL(mentionCount.toString());
+    const dataURL = badge.createDataURL(mentionCount.toString(), "#FF1744"); // Material Red A400
     sendBadge(dataURL, 'You have unread mention (' + mentionCount + ')');
   } else if (unreadCount > 0) {
-    const dataURL = badge.createDataURL('•');
+    const dataURL = badge.createDataURL('•', "#00e5ff"); // Material Cyan A400
     sendBadge(dataURL, 'You have unread channels');
   } else {
     sendBadge(null, 'You have no unread messages');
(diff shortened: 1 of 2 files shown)


yuya-oc:
Originally, there are no tray icon and the feature is brought from OS X badge. So there is red circle only. But the tray icon is too small to indicate number, so I used two different colors.

> One thing that's not matching is the color. Should I use the color you use in the taskbar dots or the material colors, like you used for the red before?
Sorry for my poor understanding about English, what answer do you expect? Just in my first impression, cyan is too light and the black border makes mismatching.


If there is screenshot, it's very helpful to discuss because everyone can know the result.

yuya-oc:
Before: before

After: after

Razzeee:
The trayicon color dots are:
```
#FF154E red
#00E4FA blue/green
```
Versus task bar color dots:
```
#FF1744 Material Red A400
#00e5ff Material Cyan A400
```

I think we should align those to one another, but which is the "correct" one?

I don't get the mismatching part about the border :) As in I don't understand what your trying to say.

yuya-oc:
Thank you for description, and it's interesting. First, material color is correct. The border is just for coloring, so I think that it's not matter at this PR.

JFYI, material colors are used for tray icon. So I think I could not understand what you said.



```
0, 229, 255 -> #00e5ff: Cyan A400
```

Razzeee:
Your absolutely right. Paint.net says the same for me. Yesterday when I did this, I had no pc with a colorpicker and used some browser plugin, that now appears to be quiet crappy :D

So we're fine color wise.

yuya-oc:
Thank you!

Razzeee:
What exactly would I do to the color?

yuya-oc:
In my thought, border line might be fine by using daker color (A700?).

If you clould, please try other colors for badges. There are no reason to use Cyan.
When you change the color from cyan, you don't have to change the tray icons. They will be modified to solve other issues.

PR #148, opened 2016-05-21
Use a fitting border color for the taskbar badge status

Razzeee:
Looks like this for unread.
example

Couldn't test mentions, as I can't mention myself :/


yuya-oc:
:+1:

Issue #146, opened 2016-05-21
Introduce windows installer into CI process


yuya-oc:
for #138

Issue #137, opened 2016-05-13
Future build, documentation and download location changes to desktop project


it33:
FYI, here are some future steps for build, documentation and download location changes for the desktop project being discussed:

1. Move Desktop to Jenkins-based build process and have it in a process similar to /platform. 
2. Add code-signing to Windows and Mac versions to confirm end users are on the right build. 
3. Move documentation on desktop apps to docs.mattermost.com to make it easier to find the apps
3. Change the major version number of Desktop apps to 3.x.x, to be aligned with /platform releases after we have code-signing complete
4. Drop final releases and release candidates for Desktop into releases.mattermost.com/X.X.X/ so IT admins can see all downloads at once by version. We'd also remove the downloads from GitHub releases to avoid confusion (if you're writing scripts to download, you can monitor this issue to know when the change is coming. We'll have at least a week or two of binaries being in both places before we remove them from GitHub releases).
5. Add UI in web server UI to make it easier to find and download and use the desktop apps

Razzeee:
We might want to move this content to #171 and close this issue?

jasonblais:
Thanks @Razzeee, makes sense. I'll close this issue as the content is in the [plan shared with community](https://docs.google.com/document/d/19cyP_phXzPQikB3WBKTzX-1A0PHpVUGhTDUIau1lY9w/edit#bookmark=kix.bmy6h0mu75ew) in #171

PR #149, opened 2016-05-21
Introduce windows installer into CI process

Razzeee:
Is this how you envisioned it in https://github.com/mattermost/desktop/issues/146 @yuya-oc ?

I also took the liberty to remove some duplicated code and simplify it a little bit. It's untested, as I can't run it from here.


yuya-oc:
Unfortunately no. I thought about `circle.yml`.

Due to #137 , I think we don't have to focus to CircleCI deployment now.
Then, please DON'T remove Windows x86 support.

PR #152, opened 2016-05-25
Edit teams in settings window

MetalCar:
The users can now edit their configured teams in settings window. I only tested it on windows. I also did changes on the UI for adding teams, because I think on this way it's much clearer how to add a team and how to edit a team. This related to #3.

`npm test`ran successfully except this `prettify`. I don't know if I got it right, but I called the command `npm run prettify`and I doesn't get an error there.

Cheers

### DIFF
--- a/src/browser/settings.jsx
+++ b/src/browser/settings.jsx
@@ -39,8 +39,10 @@ var SettingsPage = React.createClass({
     this.setState({
       teams: teams
     });
+
+    this.handleSave(false);
   },
-  handleSave: function() {
+  handleSave: function(backToIndex) {
     var config = {
       teams: this.state.teams,
       hideMenuBar: this.state.hideMenuBar,
@@ -55,7 +57,10 @@ var SettingsPage = React.createClass({
       currentWindow.setAutoHideMenuBar(config.hideMenuBar);
       currentWindow.setMenuBarVisibility(!config.hideMenuBar);
     }
-    backToIndex();
+
+    if (typeof backToIndex == 'undefined' || backToIndex) {
+      backToIndex();
+    }
   },
   handleCancel: function() {
     backToIndex();
@@ -151,7 +156,12 @@ var SettingsPage = React.createClass({
 var TeamList = React.createClass({
   getInitialState: function() {
     return {
-      showTeamListItemNew: false
+      showTeamListItemNew: false,
+      team: {
+        url: '',
+        name: '',
+        index: false
+      }
     };
   },
   handleTeamRemove: function(index) {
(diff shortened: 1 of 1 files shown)


MetalCar:
Found that `prettify` thing by myself. Did some weird jsx-code there.

MetalCar:
I forgot to put an image there how my changes looks like.
editteamsettings

MetalCar:
I found out that when I first edit a team and them add a new one, that I still edit the last selected team. The problem were the states. Hope that I get it right now.

yuya-oc:
Thanks! Just curious, is it possible to move '+' button to near the 'Team' label? For example, right of the label. The button is far from the label, so I think the user can't notice the button when adding a new team.

@CarmDam Could you complete Mattermost CLA?

MetalCar:
yes i will move the button next to the label when i'm back home.

I did the mattermost cla as MetalCar @CarmDam is also my account. it seems that i pushed my last commit with the wrong github account

MetalCar:
@yuya-oc Is this OK? Or should I place the button nearer to the `Teams`-Headline? 
image

yuya-oc:
Looks better. Could you try resizing into narrow width?

MetalCar:
Do you mean resize the complete window? Like that?
image

yuya-oc:
Sorry for few description, more narrow. I want to know whether Bootstrap correctly layouts the button in too narrow window (like mobile devices).

MetalCar:
Ah, now I understand what you want. Hope when i push my changes, that all is OK....first time I used bootstrap with reactJS
image

yuya-oc:
Thanks. That looks OK!
Please let me confirm, you haven't pushed the change yet, right?

MetalCar:
Yep, i will push it now.

src/browser/settings.jsx
```diff
@@ -28,14 +28,21 @@ var SettingsPage = React.createClass({
     } catch (e) {
       config = settings.loadDefault();
     }
+
+    this.setState({
```
Razzeee: This causes an warning in the console log.
```
settings.js:982 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the SettingsPage component.warning @ settings.js:982
```
Removing this code, solves it, not sure if it has any drawbacks/problems/how to do it properly.

Razzeee: This might be as simple as ```config.showAddTeamForm = false;``` instead of this.

yuya-oc: Should be `config.showAddTeamForm = false;`. As of warning in the console log, `setState()` can't be used in `getInitialState()` of React.js.

magicmonty:
This seems not to work correctly on OS X

screenflow

yuya-oc:
Does [this build](https://circleci.com/gh/mattermost/desktop/456#artifacts) fix problems?

magicmonty:
I have tested this build on OSX, and the problem is gone

yuya-oc:
Thanks, I will merge later.

jasonblais:
Thanks @magicmonty, @yuya-oc!!

Issue #83, opened 2016-03-17
OS X: naming and menu consistency with platform


lloeki:
Platform conventions indicate the application name should be `Electron Mattermost.app`, together with the app menu being `Electron Mattermost`.

Also, the `Settings` entry is to be separated from `About <app-name>`, labeled `Preferences...`, and have a `CMD`+`,` shortcut.

electron-mattermost:
<img width="289" alt="screen shot 2016-03-17 at 10 25 19" src="https://cloud.githubusercontent.com/assets/75445/13841734/eb7cb066-ec2a-11e5-95c8-e13587b88056.png">
Chrome:
<img width="321" alt="screen shot 2016-03-17 at 10 25 35" src="https://cloud.githubusercontent.com/assets/75445/13841739/edbe12e8-ec2a-11e5-90d4-e9ba74676a97.png">

lloeki:
Tied to #77, which solves the app name part.

lloeki:
Found out that menus are lacking standard entries on OS X, so standard shortcuts like `CMD`+`W` are not working.

it33:
Hi @lloeki, is there a straight forward way to add the shortcuts you mentioned?

lloeki:
Yes, just create appropriately named entries, and OS X will do some magic.
There's an example in Matter front code. Link pending.

lloeki:
Here it is, in [menu-template.js](https://github.com/lloeki/matterfront/blob/master/src/menu-template.js).  Just naming the things the right way makes OS X have the shortcuts, but they're included for Windows to behave. As for "Preferences...", it's lacking in the current code but naming it correctly shouldn't be a problem.

yuya-oc:
Maybe I had overlooked the separator.

Then, the wording was difficult for me when I had implemented the menu. (Both of "Preferences" and "Settings" become "設定" in Japanese...). On Windows and Linux, which is better?

lloeki:
Forgot to say that the `Edit` and `Window` entries are the one expected to be present on OS X.

lloeki:
@yuya-oc haha localisation sure is hard in many subtle ways :-)

leipert:
A few missing shortcuts on OS-X:
- `Command-M` Minimize Window
- (optional) `Alt-Command-Right` Next (unread?) Channel? (Like Next Tab in Chrome)
- (optional) `Alt-Command-Left` Prev (unread?) Channel? (Like Prev Tab in Chrome
[...]

PR #154, opened 2016-05-26
Added seperator in view menu, added new help menu and moved version i…

Razzeee:
…nto it.

This is also nice to have in preparation of the auto update feature, as it needs a home and help is probably, where we would place it.

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -26,17 +26,6 @@ var createTemplate = function(mainWindow) {
   template.push({
     label: '&' + first_menu_name,
     submenu: [{
-      label: 'About ' + app_name,
-      role: 'about',
-      click: function(item, focusedWindow) {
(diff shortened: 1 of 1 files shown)


src/main/menus/app.js
```diff
@@ -26,17 +26,6 @@ var createTemplate = function(mainWindow) {
   template.push({
     label: '&' + first_menu_name,
     submenu: [{
-      label: 'About ' + app_name,
```
yuya-oc: Please keep the 'About Mattermost' item for OS X. #83

Razzeee: Will do, just recently discovered that it's done that way on osx

Razzeee:
Just updated the PR, as I have no mac, this is more or less untested and needs to be verified. Screenshot would also be nice.

yuya-oc:
I have checked on Mac. Thanks!

Issue #39, opened 2016-02-18
Windows - Feature Idea: "Blink" Windows Taskbar icon on new messages


i23098:
In skype, when a message is received and it doesn't have the focus, the icon on the taskbar will "blink" until the user goes to the app. I think the same behavior should be implemented (and configurable, so some people could enable it while others disable it)

Johnich:
The better way is to show a tray icon on the bar, and show the number of unread new messages on the tray icon when new messages arrived. That will help a lot when we set the OS X Dock to auto-hide. Hope to see this new feature in the next version.

yuya-oc:
I have understood this is issue for Windows.

@Johnich, could you make a new issue to split topic? (yours is for osx)

btw, please indicate platform to understand the point of topic. It's very helpful because I have to think about all platforms(win, osx, linux). :(

i23098:
I had raised it with Windows in mind as I don't know if other platforms support that behavior on the icons in the taskbar. If they do, and it's easy, can't see why not implement it in all platforms. If it's only possible to implement it in Windows, then it's a Windows only request :)

Johnich:
Got it. I will new an issue.

lunawesley9:
Yes, we recently switched to mattermost, 10% of users complain that without blinking they can't notice new messages.

lunawesley9:
![image](https://cloud.githubusercontent.com/assets/14857117/15045810/2f76aeb0-130f-11e6-89f8-966f56412125.png)

MetalCar:
Currently on Windows 7 this blinking notification stuff isn't working, as I understand this right: https://github.com/electron/electron/blob/master/docs/tutorial/desktop-environment-integration.md

So, that I only has win7 on my computer, I can't really help out here :disappointed: 
Otherwise we should wait til electron add a full support for blink notifications.

MetalCar:
Seems that this type of notification doesn't work on windows. Maybe electron isn't so far yet?

yuya-oc:
I had thought [browserWindow.flashFrame()](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#winflashframeflag) might realize this, however I have never tested this function yet.

MetalCar:
Tried it in a quick test. You're right, it works. I will make a Pull Request to implement this.

Razzeee:
Jep, works but maybe something we would want a setting for?

MetalCar:
[...]

PR #159, opened 2016-06-07
Blink notification in taskbar

MetalCar:
Here is the blink notification in the taskbar after being mentioned by someone.
I only tested it on windows, since I doesn't own a macOS or linux distribution.
More information see #39 

taskicon
taskiconnotificationsettings

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -190,6 +190,7 @@ app.on('ready', function() {
         content: arg.options.body
       });
 
+      /* Todo: add idle here */
       if (config.notifications.flashWindow == 2) {
         mainWindow.flashFrame(true);
       }
@@ -213,8 +214,8 @@ app.on('ready', function() {
       }
       else {
         trayIcon.setImage(trayImages.normal);
-        mainWindow.flashFrame(false);
         trayIcon.setToolTip(app.getName());
+        mainWindow.flashFrame(false);
       }
     });
   }
--- a/src/main.js
+++ b/src/main.js
@@ -189,6 +189,10 @@ app.on('ready', function() {
         title: arg.title,
         content: arg.options.body
       });
+
+      if (config.notifications.flashWindow == 2) {
+        mainWindow.flashFrame(true);
+      }
     });
 
     // Set overlay icon from dataURL
@@ -209,6 +213,7 @@ app.on('ready', function() {
       }
       else {
         trayIcon.setImage(trayImages.normal);
+        mainWindow.flashFrame(false);
         trayIcon.setToolTip(app.getName());
       }
     });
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -189,6 +189,11 @@ app.on('ready', function() {
         title: arg.title,
         content: arg.options.body
       });
+
+      /* Todo: add idle here */
```
Razzeee: Not sure if this is a good place for this, since ```ipc.on('notified',``` is only a "hack" to enable baloon tooltips on windows < 8.1 if I understood it correctly.
This means, we should try to get rid of it in the log run. And I'm wondering if this does even get called for non win32?
If we refactor notifications, which isn't hopefully isn't unlikely, we might want to drop this.

MetalCar: When I saw it correct, that's the only place where the baloon tooltip is shown.

Razzeee: Exactly, and this is probably why this is the wrong place ;)
As we might at least want it for the other windows distributions.
Look at https://github.com/mattermost/desktop/blob/108cf91bf6023b7c13c58f6434ba36025debed9e/src/browser/webview/mattermost.js#L110 which should make it clear.

yuya-oc: Yes, `notified` message is sent on Windows <= 8.1 only. So this doesn't work with Windows 10.

src/main.js
```diff
@@ -210,6 +215,7 @@ app.on('ready', function() {
       else {
         trayIcon.setImage(trayImages.normal);
         trayIcon.setToolTip(app.getName());
+        mainWindow.flashFrame(false);
```
yuya-oc: It seems that the blinking is disabled when all unread messages have been read. Is this your expected behavior?

src/main.js
yuya-oc: "add idle here" might make confusing later. Could you change the comment? for example, "add detection for idle state".

yuya-oc:
Please refer to my comments.

yuya-oc:
@MetalCar Do you have any updates for this?

yuya-oc:
I refactored codes at 722807a. The blinking stops when the window gets focus.
Is this working correctly? https://circleci.com/gh/mattermost/desktop/405#artifacts

Razzeee:
Either it's not working for me or the blink feature is not supported on windows 7. Will test with win10 at home.

MetalCar:
@yuya-oc sorry for late response, I didn't got the time to respond on this. Now that I had a look again at the code, I understand what Razzeee tried to say. So, your refactored code looks ok, but I didn't tried it, if this works for me on win7.

Razzeee:
@yuya-oc @MetalCar 
Ok, here is what's happening. It seems like you need to restart the app to get it to load the config correctly.

1. Start mattermost
2. Check settings.
3. If it's on never, change it to always.
4. Get a message, don't restart. It won't blink.
5. Shutdown the app and start it again.
6. Get a message, it will blink correctly.

yuya-oc:
Sure, it seems correct. For now, I think it's enough.

Razzeee:
Well if it's expected we should mention that in the option tags, as we will get issues regarding it otherwise.

yuya-oc:
I will implement dynamic re-configuring.

PR #163, opened 2016-06-10
Hide 'Toggle Developer Tools' when running the app as normal user. St…

Razzeee:
…ill show it to a dev running the app.
Also hides the seperator. Not sure if you want this, if not, just close it :)


yuya-oc:
Developer Tools is very important feature to understand issues when something happens on renderer process. Please don't remove.

PR #164, opened 2016-06-11
Adds a setting to launch the app on system login

Razzeee:
login

Is this something you would accept as a PR? It's done via https://github.com/Teamwork/node-auto-launch which should work on all three platforms, I only tested windows so far.

However, this has one problem. if you uninstall the app, you will end up with the autostart entry abandoned. (at least on windows)
I wrote code, that should handle it, but it doesn't work and I don't see why.

### DIFF
--- a/src/browser/settings.jsx
+++ b/src/browser/settings.jsx
@@ -6,6 +6,7 @@ const settings = require('../common/settings');
 const React = require('react');
 const ReactDOM = require('react-dom');
 const ReactBootstrap = require('react-bootstrap');
+var AutoLaunch = require('auto-launch');
 
 const Grid = ReactBootstrap.Grid;
 const Row = ReactBootstrap.Row;
@@ -15,6 +16,10 @@ const Button = ReactBootstrap.Button;
 const ListGroup = ReactBootstrap.ListGroup;
 const ListGroupItem = ReactBootstrap.ListGroupItem;
 const Glyphicon = ReactBootstrap.Glyphicon;
+
+var appLauncher = new AutoLaunch({
+  name: 'Mattermost'
+});
 
 function backToIndex() {
   remote.getCurrentWindow().loadURL('file://' + __dirname + '/index.html');
@@ -34,6 +39,16 @@ var SettingsPage = React.createClass({
     });
 
     return config;
+  },
+  componentDidMount: function() {
+    if (process.platform === 'win32' || process.platform === 'linux') {
+      var self = this;
+      appLauncher.isEnabled().then(function(enabled) {
+        self.setState({
+          autostart: enabled
+        });
+      });
+    }
   },
   handleTeamsChange: function(teams) {
     this.setState({
@@ -56,6 +71,15 @@ var SettingsPage = React.createClass({
       var currentWindow = remote.getCurrentWindow();
       currentWindow.setAutoHideMenuBar(config.hideMenuBar);
       currentWindow.setMenuBarVisibility(!config.hideMenuBar);
+
+      var autostart = this.state.autostart;
+      appLauncher.isEnabled().then(function(enabled) {
+        if (enabled && !autostart) {
+          appLauncher.disable();
+        } else if (!enabled && autostart) {
+          appLauncher.enable();
+        }
+      });
     }
 
     if (typeof toIndex == 'undefined' || toIndex) {
@@ -83,6 +107,11 @@ var SettingsPage = React.createClass({
   handleChangeTrayIconTheme: function() {
     this.setState({
       trayIconTheme: this.refs.trayIconTheme.getValue()
+    });
+  },
+  handleChangeAutoStart: function() {
+    this.setState({
+      autostart: this.refs.autostart.getChecked()
     });
   },
   handleShowTeamForm: function() {
@@ -127,6 +156,10 @@ var SettingsPage = React.createClass({
     }
     options.push(<Input key="inputDisableWebSecurity" ref="disablewebsecurity" type="checkbox" label="Allow mixed content (Enabling allows both secure and insecure content, images and scripts to render and execute. Disabling allows only secure content.)"
                    checked={ this.state.disablewebsecurity } onChange={ this.handleChangeDisableWebSecurity } />);
+    //OSX has an option in the tray, to set the app to autostart, so we choose to not support this option for OSX
+    if (process.platform === 'win32' || process.platform === 'linux') {
+      options.push(<Input key="inputAutoStart" ref="autostart" type="checkbox" label="Start app on login." checked={ this.state.autostart } onChange={ this.handleChangeAutoStart } />);
+    }
     var options_row = (options.length > 0) ? (
       <Row>
         <Col md={ 12 }>
--- a/src/main.js
+++ b/src/main.js
@@ -3,7 +3,21 @@
 const electron = require('electron');
 const app = electron.app; // Module to control application life.
 
-if (require('electron-squirrel-startup')) app.quit();
+if (process.platform === 'win32') {
+  var cmd = process.argv[1];
+  if (cmd === '--squirrel-uninstall') {
+    var AutoLaunch = require('auto-launch');
+    var appLauncher = new AutoLaunch({
+      name: 'Mattermost'
+    });
+    appLauncher.isEnabled().then(function(enabled) {
+      if (enabled)
+        appLauncher.disable();
+    });
+  }
+}
+
+require('electron-squirrel-startup');
 
 const BrowserWindow = electron.BrowserWindow; // Module to create native browser window.
 const Menu = electron.Menu;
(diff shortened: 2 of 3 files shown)


src/main.js
```diff
@@ -3,6 +3,18 @@
 const electron = require('electron');
 const app = electron.app; // Module to control application life.
 
+var cmd = process.argv[1];
+if (cmd === '--squirrel-uninstall') {
```
Razzeee: This is the part, that should act upon uninstall, but doesn't

yuya-oc: First, how about checking the part which is working properly? Sure, the debugging of installation seems difficult, but for example, you can inspect by writing simple log file with `fs` module.

yuya-oc:
Please refer to my comments.

yuya-oc:
BTW, on OS X, users can easily setup an application as `Open on login` by right-clicking on Dock. So I think this feature is not necessary on OS X. How do you think?

open_on_login

src/common/settings.js
```diff
@@ -26,6 +26,7 @@ var loadDefault = function(version) {
         showTrayIcon: false,
         trayIconTheme: '',
         disablewebsecurity: true,
+        autostart: false,
```
yuya-oc: node-auto-launch uses registry entry. So I think that the app should not keep the config by itself. How about read the state by node-auto-launch when showing the settings page?

yuya-oc:
Just in my personal opinion, the registry is not good way. How do you feel?
For example, Slack creates a shortcut on `C:\Users\%USER%\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup`.

src/common/settings.js
Razzeee: Yes, that was what I originally tried to do, but ended up with the config. Will try to do the other way again.

Razzeee:
Yes, which I tried some weeks ago, but it either didn't work or wasn't as neat. Not sure about the details why. Maybe I was fooled by the not working uninstall behavior there.

src/main.js
```diff
@@ -3,6 +3,18 @@
...
+  var appLauncher = new AutoLaunch({
+    name: 'Mattermost'
+  });
+  appLauncher.isEnabled().then(function(enabled) {
+    if (enabled)
+      appLauncher.disable();
+  });
+}
+
 if (require('electron-squirrel-startup')) app.quit();
```
Razzeee: I don't think it will change anything or matter, looking at https://github.com/mongodb-js/electron-squirrel-startup/blob/master/index.js
I'm not sure why they are saying that there should be an if around it in the docs

Razzeee: Removing the if/app.Quit seems to fix the behaviour.

yuya-oc:
Checked on Windows 10 x64 only. Do you have any updates?

Razzeee:
Yes, I can't get it to load on intialize, for some reason
```
settings.js:982 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the SettingsPage component.warning @ settings.js:982

settings.js:982 Warning: DeprecatedComponent is changing a uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

Also I start to think, that it might not be worth to exclude OSX from this. If it works without problems with OSX, like it does on windows (if we can figure the other stuff out) it might make sense to still add it on OSX. Just so that we don't have if cases everywhere.

Razzeee:
Okay, the first error is also happening on master :(

Razzeee:
Can't get/block the promise from node-auto-launch, so it only fills up way after we loaded everything for the settings menu, which then leads to a wrong checkbox.

src/browser/settings.jsx
```diff
@@ -33,6 +38,12 @@ var SettingsPage = React.createClass({
       showAddTeamForm: false
     });
 
+    appLauncher.isEnabled().then(function(enabled) {
```
yuya-oc: > Can't get/block the promise from node-auto-launch, so it only fills up way after we loaded everything for the settings menu, which then leads to a wrong checkbox.

You can use `didComponentMounted` API to reflect the initial state asynchronously instead of here. Would this be solution for your comment?

Razzeee: I did try that yesterday, without luck. But I was thinking to go back and retry it, so thank you for getting me going in the right direction.

Razzeee: @yuya-oc doesn't seem to work :(
image
I'll push that iteration of the code, anyway

yuya-oc:
webpack raises warning about node-auto-launch. I'm not sure what is vertx, but it seems the promise from node-auto-launch is not working.

```
WARNING in ./~/auto-launch/~/es6-promise/dist/es6-promise.js
Module not found: Error: Cannot resolve module 'vertx' in D:\Users\yuya-oc\git\mattermost-desktop\node_modules\auto-launch\node_modules\es6-promise\dist
 @ ./~/auto-launch/~/es6-promise/dist/es6-promise.js 131:20-30
```

Probably https://github.com/stefanpenner/es6-promise/issues/100 is related.

package.json
```diff
@@ -29,6 +29,7 @@
     "installer": "node ./script/installer.js"
   },
   "devDependencies": {
+    "auto-launch": "^2.0.1",
```
yuya-oc: auto-launch is a dependency for the actual app. So please move into `dependencies` of `src/package.json`.

yuya-oc:
Sorry, the previous comment was my misunderstanding. The promise was returned.

src/browser/settings.jsx
```diff
@@ -33,8 +38,22 @@ var SettingsPage = React.createClass({
       showAddTeamForm: false
     });
 
+
     return config;
   },
+  componentDidMount: function() {
+    if (process.platform === 'win32' || process.platform === 'linux') {
+      appLauncher.isEnabled().then(function(enabled) {
+        this.setState({
```
yuya-oc: The context of `this` is changed by function object. To use `this` here, an arrow function should be used for this callback.

Razzeee:
Thank you working fine now.
Your still sure about leaving OSX out of the mix?

yuya-oc:
Sure, however when I read node-auto-launch, it seems that the module internally uses the same way which I wrote. But I haven't checked actual behavior yet. If the behavior is same, I agree that the feature is also added to OS X.

Razzeee:
Should I change it then and you test or will you test and I will change it afterwards?

src/browser/settings.jsx
```diff
@@ -56,6 +71,14 @@ var SettingsPage = React.createClass({
       var currentWindow = remote.getCurrentWindow();
       currentWindow.setAutoHideMenuBar(config.hideMenuBar);
       currentWindow.setMenuBarVisibility(!config.hideMenuBar);
+
+      var autostart = this.state.autostart;
+      appLauncher.isEnabled().then(function(enabled) {
```
yuya-oc: Just curious, is this confirmation necessary? I think that simple enable/disable are enough in this part.

src/browser/settings.jsx
```diff
@@ -56,6 +71,14 @@ var SettingsPage = React.createClass({
...
+        if (enabled && !autostart)
```
yuya-oc: Please use brackets. Not only readability for me, `npm run prettify` incorrectly beautified this.

yuya-oc:
Just checked on my Mac by simply removing `if` statements and small fix. The feature doesn't work.
Although, I think the PR is enough for merging. Please check my comments to fix CI.

src/browser/settings.jsx
Razzeee: Well if you do that, it would even change, when you don't change the checkbox, as it will just toggle each time you hit save.

yuya-oc: I see. Maybe it's a matter of taste.

Razzeee: Don't think so, if would be very confusing if it gets set of you don't click the option.

Razzeee:
Should be updated now.

src/browser/settings.jsx
yuya-oc: Sorry for my poor English, I couldn't understand.
Point of my question was "Is there any problem for executing enable/disable every time when clicking save button?" `if` condition looks a little complex because of `isEnabled()`

Razzeee: Well we could do
```
        if (!autostart) {
           appLauncher.disable();
         } else {
           appLauncher.enable();
         }
```
but that might fail if somebody changes the auto-start outside of our app and already has opend the settings page. Probably a cornercase ;)

yuya-oc: I see. Thank you for many times.

Issue #98, opened 2016-03-24
Win64 Desktop v1.1.0-rc2 on Win7: Icon in top left of Windows Desktop app seems blurry. 


it33:
![image](https://cloud.githubusercontent.com/assets/177788/14034559/85c20c8c-f1e1-11e5-89c9-c757ef734a5e.png)

yuya-oc:
It seems that it's a side effect of 7e7109413307336af211c4d19bbb4a280a470644. I think a too high resolution image is assigned, and v1.0.7 is same.

yuya-oc:
Reopen to track this. It's a side effect by fixing #192.

yuya-oc:
Closed by a0453b5. Thanks @asaadmahmood for preparing icons!

PR #166, opened 2016-06-13
Bump prebuild to 1.2.2

Razzeee:
The icon seemed to be lost, at least when running it on windows just for dev purpose.
It might probably get injected by the build scripts, but not sure if that's good enough for us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "electron-builder": "3.20.0",
     "electron-connect": "^0.3.7",
     "electron-packager": "^7.0.1",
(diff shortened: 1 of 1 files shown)


yuya-oc:
The icon is added when the app is packaged. So the icon is missing in dev mode only. Then, the way as to assign png to `window_options.icon` makes the app icon blurry (#98).

Razzeee:
Dropped the second commit, just thought it was weird, as master has the icon even in dev mode.

PR #162, opened 2016-06-10
Implement basic autoupdate

Razzeee:
This will add autoupdate for windows and possibly osx via [electron-gh-releases](https://github.com/jenslind/electron-gh-releases)

I can't really work on the OSX side of things.

Windows is done and works, you just need to build an installer and upload them to the github releases. Like I did [here](https://github.com/Razzeee/desktop/releases)

OSX is missing the 'auto-updater.json' and probably the signing, see documentation [here](https://github.com/jenslind/electron-gh-releases/blob/master/docs/2.x/how-to.md#publishing-a-new-release-on-github)
There is also an package to simplify the osx part of the release: https://github.com/jenslind/electron-release

update

PR #168, opened 2016-06-16
Update build scripts and gh draft

Razzeee:
This will hopefully build and create drafts with the needed files for the auto installer for windows. I settled for x64 as x32 shouldn't be needed in most cases.
See https://github.com/mattermost/desktop/pull/162


yuya-oc:
Due to #137, github release will be obsoleted. Then, we should consider users who don't want auto update or have a specific version.

Just curious, is there any problem if there is ia32 installer? I could not understand the reason.

Razzeee:
Well #137 has no handling for auto updates, isn't that something we want?

No there is no problem, but I think we can only autoupdate one of those. And as far as I know the auto update PR I did is using x64 for now. So offering 32 isn't a good practice as it can't be updated and should probably only be used in corner cases, by user that know what they are doing.

Issue #178, opened 2016-06-27
Closing/Quitting Mattermost desktop should minimize it to the system tray (i.e. notification area)


jasonblais:
Closing/Quitting Mattermost desktop should minimize it to the system tray (i.e. notification area).

Currently, closing/quitting Mattermost desktop simply minimizes the application to the taskbar.

jasonblais:
Would anyone from the community be open to helping with a PR to address this?

jasonblais:
In progress for Windows by @magicmonty ([#179]https://github.com/mattermost/desktop/pull/179)

Any takers for OS X or Linux?

jasonblais:
Thanks @magicmonty!

magicmonty:
The hide on close for OSX is now part of PR #179

yuya-oc:
It's done for Windows and OS X at #179.

jasonblais:
Huge thanks again @magicmonty!! I'll mark this for `Linux` and remove from v1.3 for now

jasonblais:
Closing the issue as done

PR #179, opened 2016-06-27
Hide window to system tray on minimize/close in Windows/OSX

magicmonty:
Added an option to hide the Window on minimize/close to the system tray.
if this option is enabled, a click on the system tray would also toggle the window.

This addresses #178 at least partially (just for Windows and OSX currently)
---

  - Tested on Windows 10
  - Tested on OS X 10.11.6 Beta

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -27,6 +27,7 @@ var loadDefault = function(version) {
         trayIconTheme: '',
         disablewebsecurity: true,
         minimizeToTray: false,
+        toggleWindowOnTrayIconClick: false,
         version: 1,
         notifications: {
           flashWindow: 0 // 0 = flash never, 1 = only when idle (after 10 seconds), 2 = always
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -26,6 +26,7 @@ var loadDefault = function(version) {
         showTrayIcon: false,
         trayIconTheme: '',
         disablewebsecurity: true,
+        minimizeToTray: false,
         version: 1,
         notifications: {
           flashWindow: 0 // 0 = flash never, 1 = only when idle (after 10 seconds), 2 = always
@@ -38,7 +39,7 @@ var upgradeV0toV1 = function(config_v0) {
   var config = loadDefault(1);
   config.teams.push({
     name: 'Primary team',
-    url: config_v0.url
+    url: config_v0.url    
   });
   return config;
 };
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -26,7 +26,6 @@ var loadDefault = function(version) {
         showTrayIcon: false,
         trayIconTheme: '',
         disablewebsecurity: true,
-        minimizeToTray: true,
         toggleWindowOnTrayIconClick: false,
         version: 1,
         notifications: {
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -26,7 +26,7 @@ var loadDefault = function(version) {
         showTrayIcon: false,
         trayIconTheme: '',
         disablewebsecurity: true,
-        minimizeToTray: false,
+        minimizeToTray: true,
         toggleWindowOnTrayIconClick: false,
         version: 1,
         notifications: {
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -39,7 +39,7 @@ var upgradeV0toV1 = function(config_v0) {
   var config = loadDefault(1);
   config.teams.push({
     name: 'Primary team',
-    url: config_v0.url    
+    url: config_v0.url
   });
   return config;
 };
--- a/src/main.js
+++ b/src/main.js
@@ -241,10 +241,22 @@ app.on('ready', function() {
           mainWindow.focus();
         }
       }
+      else if (process.platform === 'darwin') {
+        if (mainWindow.isHidden || mainWindow.isMinimized()) {
+          mainWindow.show();
+          mainWindow.isHidden = false;
+          mainWindow.focus();
+          app.dock.show();
+        }
+        else {
+          mainWindow.focus();
+        }
+      }
       else {
         mainWindow.focus();
       }
     });
+
     trayIcon.on('right-click', () => {
       trayIcon.popUpContextMenu();
     });
@@ -380,7 +392,14 @@ app.on('ready', function() {
           mainWindow.minimize();
           break;
         case 'darwin':
-          mainWindow.hide();
+          if (config.minimizeToTray) {
+            mainWindow.hide();
+            app.dock.hide();
+            mainWindow.isHidden = true;
+          }
+          else {
+            mainWindow.minimize();
+          }
           break;
         default:
       }
--- a/src/main.js
+++ b/src/main.js
@@ -234,7 +234,7 @@ app.on('ready', function() {
             mainWindow.show();
             mainWindow.isHidden = false;
           }
-          else {
+          else if (config.toggleWindowOnTrayIconClick) {
             mainWindow.hide();
             mainWindow.isHidden = true;
           }
--- a/src/main.js
+++ b/src/main.js
@@ -261,12 +261,15 @@ app.on('ready', function() {
       trayIcon.popUpContextMenu();
     });
     trayIcon.on('balloon-click', function() {
-      if (process.platform === 'win32') {
-        if (config.minimizeToTray) {
-          mainWindow.show();
-          mainWindow.isHidden = false;
-        }
+      if (process.platform === 'win32' || process.platform === 'darwin') {
+        mainWindow.show();
+        mainWindow.isHidden = false;
       }
+
+      if (process.platform === 'darwin') {
+        app.dock.show();
+      }
+
       mainWindow.focus();
     });
     ipcMain.on('notified', function(event, arg) {
@@ -380,26 +383,16 @@ app.on('ready', function() {
       event.preventDefault();
       switch (process.platform) {
         case 'win32':
-          if (config.minimizeToTray) {
-            mainWindow.hide();
-            mainWindow.isHidden = true;
-          }
-          else {
-            mainWindow.minimize();
-          }
+          mainWindow.hide();
+          mainWindow.isHidden = true;
           break;
         case 'linux':
           mainWindow.minimize();
           break;
         case 'darwin':
-          if (config.minimizeToTray) {
-            mainWindow.hide();
-            app.dock.hide();
-            mainWindow.isHidden = true;
-          }
-          else {
-            mainWindow.minimize();
-          }
+          mainWindow.hide();
+          app.dock.hide();
+          mainWindow.isHidden = true;
           break;
         default:
       }
--- a/src/main.js
+++ b/src/main.js
@@ -391,8 +391,10 @@ app.on('ready', function() {
           break;
         case 'darwin':
           mainWindow.hide();
-          app.dock.hide();
-          mainWindow.isHidden = true;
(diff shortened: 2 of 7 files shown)


src/main.js
```diff
@@ -227,12 +228,30 @@ app.on('ready', function() {
 
     trayIcon.setToolTip(app.getName());
     trayIcon.on('click', function() {
+      if (process.platform === 'win32') {
+        if (config.minimizeToTray) {
+          if (mainWindow.isHidden) {
```
yuya-oc: It seems to work as 'toggle'. In my feel, clicking should always show the window, would you hear me your opinion?
(Slack always opens/focuses the window when clicking the tray icon.)

magicmonty: Yes this was intended as a toggle. As currently the tray icon has no 'use' for the left click, if the window is open, I find it very convenient to click again to close the window. 

Especially if I just want to check the currently open channel (I use it currently mainly for Build-Server-Notifications), I just click on the Icon, and if there is nothing that much interesting I just click again to close. Else I  have to go explicitely on the X-Button or the Minimize-Button of the window and that would be much more way with the mouse.

yuya-oc: > However, the result of a single click with the left mouse button will vary with the function of the icon. It should display what the user would expect to see in the form best suited to that content—a popup window, a dialog box or the program window itself. For instance, it could show status text for a status icon, or a slider for the volume control.

On [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ee330740(v=vs.85).aspx), it seems that a single click is expected as "show" function.

jasonblais:
Thanks @magicmonty for your pull request, super-excited!

Would you be able to do a local release so that we can test it out before we merge it?

We're trying out a new process where at least two members test pull requests so we've done some testing before a merge

yuya-oc:
Until introducing the new CI, CircleCI hosts the new release. Please use it.

https://circleci.com/gh/mattermost/desktop/430#artifacts

magicmonty:
@yuya-oc Ok, I see your point. What if I change the close to happen on a double click? Another possibility would be a "Minimize to tray" menu entry in the context menu of the tray icon, although this would not as much as convenient. A third possibility would be, to enable this behaviour explicitely via the settings, so everyone can decide by itself, if this feature should be used or not. Whats your opinion here?

src/main.js
magicmonty: Ok, I see your point. What if I change the close to happen on a double click? Another possibility would be a "Minimize to tray" menu entry in the context menu of the tray icon, although this would not as much as convenient. A third possibility would be, to enable this behaviour explicitely via the settings, so everyone can decide by itself, if this feature should be used or not. Whats your opinion here?

magicmonty: Also at least in Windows 10 there are some examples, where the click on the tray icon toggles the visibility of the respective window: e.g. the Notification center or the Volume control

yuya-oc: I think they are temporary views to just know info. But this tray icon shows the main window of app. So the purpose is different between them.

Assuming your research, it seems that the behavior should be selectable. If possible, I hope the default behavior is "just show". Then, the entry in the context menu is a nice idea because users don't have to think about its configuration.

magicmonty: Ok, then I add a new Setting for the toggle

src/common/settings.js
```diff
@@ -26,6 +26,8 @@ var loadDefault = function(version) {
         showTrayIcon: false,
         trayIconTheme: '',
         disablewebsecurity: true,
+        minimizeToTray: false,
```
Razzeee: From reading https://github.com/mattermost/desktop/issues/178 I think this should be ```true``` and therefor the default behavior

magicmonty: As discussed above, the default behaviour should be "non" toggling on windows and therefore should be explicitely enabled.

Razzeee:
For reference, how slack words their option
image

magicmonty:
Yeah, this wording is certainly better.

yuya-oc:
FYI, Slack's one is not only for wording, the behavior is different.

- On minimize: Minimize the window
- On close: Hide to tray

jasonblais:
thanks @magicmonty! This will be an excellent feature to add!

Along the lines of @yuya-oc 's comment above, these are the use cases I'm envisioning - the app should collapse to the system tray only if a user closes the app:
 1. Clicking the [-] Minimize button on the top right corner should minimize it to the taskbar
 2. Clicking on the icon, either in the taskbar or the system tray, should toggle the app open/minimized to task bar
 3. Clicking the [X] Close button on the top right corner should close the app to the system tray

^ Currently 1. and 2. actually closes it directly to the system tray. Would you be able to make a change, essentially reverting to their previous behaviour? This is more standard with running apps.

magicmonty:
What would be the use case (apart from being more standard) to minimize to the taskbar. 
I don't see an additional value on the task bar icon currently.

yuya-oc:
In my feel, task tray is far from to switch from other task bar buttons. Then, some users might not notice the tray icon at first time.
At least, I choose the config to minimize the window on the `minimize` event if the option is provided.

jasonblais:
Thanks @magicmonty for the update!

Could we have two community members testing the latest submission on Windows and OS X, then respond back with either a 'pass' or any issues they may have found?

@yuya-oc, not sure if you'll need to generate a new CircleCI for it?

magicmonty:
Ok, as there is more consent to hide to tray only on "Close", I have changed the behaviour here, so the click on the tray icon just toggles the "Minimize". A click on "Minimize" also minimizes the Window to the task bar and only a "Close" hides the window

yuya-oc:
This is the latest. https://circleci.com/gh/mattermost/desktop/443#artifacts

Note: the PR is only for Windows, so the OS X app should not have any changes.

@magicmonty Would you update CHANGELOG.md and docs/setup.md?

jasonblais:
Thanks @yuya-oc, updated my comment above to only test on Windows

magicmonty:
@yuya-oc Done. I removed also the dependency on the "Minimize to tray" setting from the "Toggle minimize" setting, so they can now set independently

yuya-oc:
Sorry for my few description, what I meant is "need to confirm there is no changes on OS X".

jasonblais:
Sorry for the confusion @yuya-oc , made an update

magicmonty:
Just added the "Hide window on close" functionality for OSX

magicmonty:
The latest artifacts should now be availble under https://circleci.com/gh/mattermost/desktop/448#artifacts

jasonblais:
Perfect, thanks @magicmonty !! Love your work here,

Anyone from the community open to test [latest artifacts](https://circleci.com/gh/mattermost/desktop/448#artifacts) on Windows and OSX?

jasonblais:
Verified on Windows 10, @jnugh verified on OS X

jasonblais:
I was testing the latest artifact on Windows 10, and it appears I'm unable to edit a team in the Settings page, although the feature was added by #152 . Wondering if this is related to a bug introduced by this pull request, or whether it's something introduced earlier? 

Also, proposing we remove both options in the settings menu
 a) `Leave app running in notification area when the window is closed` actually means that if it's 
  - `selected`, closing the desktop app collapses it to the system tray
  - `not selected`, closing the desktop app quits the application and doesn't hide it in the system tray

In my mind, since we're not implementing the `not selected` option, there isn't a need for a setting. If a user minimizes the app, it goes to the taskbar; if the user closes the app, it hides in the system tray. Hence, I don't think we need to add an option for it.

 b) `Toggling window visibility when clicking on the tray icon` is fine in my opinion and doesn't need an option. It doesn't remove any functionality.

Thoughts @magicmonty, @yuya-oc?

Otherwise loving the new feature, solid work here!!

image

magicmonty:
@jasonblais I think you are right with the first option, I don't see a real use case here for really closing the app with the Close-Button. Although I would leave the "Toggle" option, as this has a side effect, which may not everyone wants. If the window is hidden behind another window (but open), then the toggle would trigger a minimize anyway, as I currently haven't found a way to check, if the window is behind another window. In an ideal world this click would focus the window and bring it into the foreground (what the functionality without the toggle does). So here someone might prefer the other behaviour over the  toggle.

magicmonty:
@jasonblais For the bug, that you can't edit a team, I tested this on OS X with both the master and this branch. If I just added a new team, then I'm not be able to edit the other team. If I save, after I added the team then I can edit both teams normally. But this behaviour appears also on the master branch. This should have nothing to do with this PR. 
Here is the behaviour on OSX (from the master branch):

screenflow

yuya-oc:
Sorry for late.

@jasonblais @magicmonty As @magicmonty wrote, the edit function is not broken. Sure, it might not be good behavior, but it's not related this PR.

@jasonblais 
a.) The app doesn't quit on `not selected` when closing the window. But, I'm agree.
b.) Ideally, the latest magicmonty's idea seems the best, but currently it's hard. So I think this would be left, and the reason is same as magicmonty.

src/main.js
```diff
@@ -346,11 +380,26 @@ app.on('ready', function() {
...
           break;
         case 'darwin':
-          mainWindow.hide();
+          if (config.minimizeToTray) {
+            mainWindow.hide();
+            app.dock.hide();
+            mainWindow.isHidden = true;
+          }
+          else {
+            mainWindow.minimize();
```
yuya-oc: This should be `mainWindow.hide()`. This is `close` event, so minimizing is incorrect behavior.

magicmonty:
@jasonblais @yuya-oc, so I would remove the first option and make it the default behaviour

magicmonty:
@yuya-oc @jasonblais I removed the option, and made it the default behaviour. The Window now closes always into the tray on Windows and OSX

magicmonty:
Oh, ok, then I revert the change for OSX

yuya-oc:
Tested on Win10 and OS X 10.11.5. And I found a issue on OS X.

- If both of options are enabled at the same time, the window can't be reopen because the menu icon doesn't appear until restarting the app.

I think it's a little difficult to fix for now. Currently the app can't enable/disable the menu icon without restarting. Maybe it can be fixed, but it might affect to main.js structure. So is it possible to note this as a known issue?

magicmonty:
@yuya-oc Thanks for the hint. I found a way to fix this behaviour. The option is now only after a restart available (if the tray icon was not visible before). The text on the icon reflects this too.

jasonblais:
Thanks @yuya-oc  and @magicmonty !!

Just to clarify:
1. `Leave app running in notification area when the window is closed` has been 
  - removed from the Windows app (with 'close' button always collapsing the app to the system tray)
  - kept for OS X as the menu icon might be hidden
2. `Toggling window visibility when clicking on the tray icon` is kept as a setting as otherwise if the window is hidden behind another window (but open), the toggle could trigger the app to be minimized

Both of the above would make sense, thank you for both of you! Many of our users have asked for the ability to collapse the app to the system tray, so I'm excited to get this in for v1.3.

yuya-oc:
Tested two cases on OS X and simply on Win10. It seems fine.

a. Enable `Leave app running in notification area when the window is closed` only.
b. Disable `Leave app running in notification area when the window is closed` when both config are enabled.

jasonblais:
Tested the latest changes on Windows 10 as well and @jnugh helped test it on OS X on Friday.

Looking great!! Love your contribution @magicmonty!

yuya-oc:
@magicmonty Thanks for many times!

magicmonty:
You're welcome. Glad to help

Issue #193, opened 2016-07-07
Bug Hunt Competition (Desktop App v1.3)


jasonblais:
Huge thanks to everyone who helped submit [pull requests for Desktop App v1.3](https://github.com/mattermost/desktop/pulls?utf8=%E2%9C%93&q=merged%3A%3E2016-05-13), excited to ship the new release soon!

We're trying out something new: a **Bug Hunt Competition**! Top 5 point scorers will win a special prize (if your mailing address is reachable by our merchandise vendor, we'd like to send you a personalized award with your name and Bug Hunt Champion achievement date.).

Anyone who scores at least one point will be recognized on the Mattermost page.

Here are steps to participate in the competition:
 1. Download [release candidate for v1.3](https://github.com/mattermost/desktop/releases/tag/v1.3.0-rc1)
 2. Find an issue with the product which is
   - [ ] not a feature idea (feature ideas go here: http://www.mattermost.org/feature-ideas/)
   - [ ] not a troubleshooting question (troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
   - [ ] is not already reported by someone else in this thread or in [GitHub issues](https://github.com/mattermost/desktop/issues)
   - [ ] is a product defect issue that can be reproduced using the steps listed below.
 3. Reply to this thread with an issue you have found, containing
  - the operating system 
  - Mattermost server version and database (from Main Menu > About Mattermost)
  -  clear repro steps

You are welcome to join our [Desktop App](https://pre-release.mattermost.com/core/channels/desktop-app) discussion channel to meet the Desktop App community and use the Mattermost server for testing.

The point system that will be used: 
   - 1 point: Issue filed satisfies all three steps above
   - 2 bonus points: Issue reproduces on Windows
   - 3 bonus points: Issue is deemed critical and fixed for v1.3
   - 4 bonus points: The issuer submits a pull request, which is merged to the /desktop repo

**The competition will run from Thursday July 7th until Monday July 11th 23:59 GMT**

Good luck, let the bug hunt begin!
![image pasted at 2016-7-5 18-09](https://cloud.githubusercontent.com/assets/13119842/16640754/b973d220-43c8-11e6-8422-7f33391be348.png)

jasonblais:
Update: [Release candidate for Desktop App v1.3 has been posted](https://github.com/mattermost/desktop/releases/tag/v1.3.0-rc1).

The competition has now begun, good luck everyone!

DavidLu1997:
[...]

PR #195, opened 2016-07-09
Fixed typo in settings (task icon)

Razzeee:
As mentioned in
https://github.com/mattermost/desktop/issues/193#issuecomment-231553779

### DIFF
--- a/src/browser/settings.jsx
+++ b/src/browser/settings.jsx
@@ -231,7 +231,7 @@ var SettingsPage = React.createClass({
       notifications_row = (
         <Row id="notificationsRow">
           <Col md={ 12 }>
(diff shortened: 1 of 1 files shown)


jasonblais:
Thanks @Razzeee, propose updating the help text to

`When a new message is received, flash the taskbar icon`

instead? 

It helps convey to users what the setting does a little better, by having the condition `when a new message ...` before the result `flash the taskbar icon`

jasonblais:
Thanks @Razzeee, propose updating the help text to

`Flash the taskbar icon when a new message is received`

instead, to have the setting conveyed more concisely

Razzeee:
Changed. Also added a commit to align the casing on some settings.

jasonblais:
+1 from me, propose we merge this change for v1.3

PR #200, opened 2016-07-12
Urgent hint on linux

jnugh:
This enables the *window flashing in the taskbar on messages* feature for linux.

---

  - Ubuntu 16.04 running unity

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -261,7 +261,7 @@ app.on('ready', function() {
       mainWindow.focus();
     });
     ipcMain.on('notified', function(event, arg) {
-      if (process.platform === 'win32' || process.platform == 'linux') {
+      if (process.platform === 'win32' || process.platform === 'linux') {
         if (config.notifications.flashWindow === 2) {
           mainWindow.flashFrame(true);
         }
--- a/src/main.js
+++ b/src/main.js
@@ -261,10 +261,13 @@ app.on('ready', function() {
       mainWindow.focus();
     });
     ipcMain.on('notified', function(event, arg) {
-      if (process.platform === 'win32') {
+      if (process.platform === 'win32' || process.platform == 'linux') {
         if (config.notifications.flashWindow === 2) {
           mainWindow.flashFrame(true);
         }
+      }
+
+      if (process.platform === 'win32') {
         // On Windows 8.1 and Windows 8, a shortcut with a Application User Model ID must be installed to the Start screen.
         // In current version, use tray balloon for notification
         if (osVersion.isLowerThanOrEqualWindows8_1()) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Improvements
 
+#### Linux
+ - Added an option to make the taskbar icon flash on new messages
+
+### Improvements
+
 ### Bug Fixes
 
 
--- a/src/browser/settings.jsx
+++ b/src/browser/settings.jsx
@@ -202,7 +202,7 @@ var SettingsPage = React.createClass({
       ) : null;
 
     var notifications_row = null;
-    if (process.platform === 'win32') {
+    if (process.platform === 'win32' || process.platform === 'linux') {
       var notificationSettings = [
         {
           label: 'Never',
--- a/test/specs/browser/settings_test.js
+++ b/test/specs/browser/settings_test.js
@@ -172,8 +172,8 @@ describe('browser/settings.html', function() {
     });
 
     describe('Notifications', function() {
-      it('should appear on win32', function() {
-        const expected = (process.platform === 'win32');
+      it('should appear on win32 and linux', function() {
+        const expected = (process.platform === 'win32' || process.platform === 'linux');
         env.addClientCommands(this.app.client);
         return this.app.client
           .loadSettingsPage()


timroes:
Would be awesome, if this could get merged. This is really one of the larger pains of mattermost, that there is no persistent state/signal after you have received some message.

jasonblais:
Big thnaks @jnugh for the pull request!

Would we be able to have two community members testing this pull request with Linux/Ubuntu? We would like to have at least two reviewers before it's merged.

The artifacts to test can be found here: https://circleci.com/gh/mattermost/desktop/497#artifacts

Razzeee:
You forgot to add the changelog entry @jnugh

timroes:
@jasonblais I tested the `mattermost-desktop-linux-x64.tar.gz` from the link you posted above and unfortunately it's not working.

I checked, that I had `flashWindow` set to `2` in the config, but even after receiving personal messages, while the Mattermost window wasn't focused, it didn't get the urgent hint. 

*How do I know?* I use awesome WM as window manager and have scripted my config to highlight urgent windows in a special way and I also cross checked with `xwininfo`. It would show something like the following when urgent hint is set:

```
 Window manager hints:
      Displayed on desktop 1
      Window type:
          Normal
      Window state:
          Demands Attention
```

Mattermost didn't had the last two lines when it should have. I also tried using `wmctrl` to set the flag manually on the Mattermost window and cross check again (if my checking methods work) and if I manually set the urgent flag, it highlights correctly, and xwininfo outputs the right info. Unfortunately also with this build it doesn't do this on new messages. (I also tried restarting Mattermost after changing the settings.)

src/main.js
```diff
@@ -236,10 +236,13 @@ app.on('ready', function() {
       mainWindow.focus();
     });
     ipcMain.on('notified', function(event, arg) {
-      if (process.platform === 'win32') {
+      if (process.platform === 'win32' || process.platform == 'linux') {
```
timroes: Why not using `===` here (but only `==`)?

jasonblais:
Big thanks @timroes for testing this pull request! 

@jnugh can you help take a look at Tim's note above?

jnugh:
The API uses `gtk_window_set_urgency_hint` internally.
I'm not sure if awesome handles this correctly. I'm actually not sure where to find documentation on how to set `Demands Attention`.

src/main.js
jnugh: No particular reason :wink: Should not be a problem semantically.
But you are right, it does not match the surrounding code style, will be updated.

timroes:
I've looked through the source code of pidgin and it also uses `gtk_window_set_urgency_hint` and it works fine for pidgin. I will try to stick together a simple electron sample on weekend and see if it works at all in a plain simple electron sample on my desktop.

If it works for you - the commit itself looks perfectly fine and logically for me - I would also say it could be merged already and I look why it doesn't work on my system (even though wmctrl and other applications just work fine), by building a simple electron sample and maybe a simple C gtk sample.

timroes:
I just used the (electron-quick-start)[https://github.com/electron/electron-quick-start] project and added:

```
mainWindow.on('blur', function() {
    console.log("Blur!");
    mainWindow.flashFrame(true);
});
```

The window gets correctly highlighted. Which brings me to the question: was the above link **really** the correct version, that included this pull request? :-) @jasonblais

jnugh:
Seems to be the right one.
What kind of notification did you test? A mention should work.

timroes:
So I tested again, private messages and mentions. Both trigger a notification and none of it set the urgent hint correctly (even though setting itself is active). I also build from @jnugh git branch directly and tried that version, but same behavior. So very strange...

- `flashFrame(true)` itself works on my system, in the minimal electron sample
- Notifications work fine
- setting the urgent hint manually via `wmctrl` on the window works
- But this branch doesn't work (even though it seem logically and apparently works for @jnugh )

I also tried to put some debug output to `src/browser/webview/mattermost.js` and it seems it never reaches the `notified` ipc callback. But it might be that I've done debugging here wrong. If I open the debug console in mattermost and enter `require('electron').ipcRenderer.send('notified', { title: "Test", options: {} });` should that actually do anything or isn't it that simple to get a hold on the correct  ipc? Because in my case that doesn't do anything.

jnugh:
Added a changelog entry and added the equals as described above.

@timroes:
```
require('electron').ipcRenderer.send('notified', { title: "Test", options: {} });
```
triggers a flash for me.

Maybe the problem is somewhere else. Did you try to find out if the code gets executed? As this does not run on the render process we can't use the build in devtools to add a breakpoint or something but when you add a `console.log` you should see something in the terminal you are running `npm start` or `npm watch` from whenever you send `notified`.

timroes:
I've just found the error (and it is not really related to this issue, but just another bug :D).

The registration to the `notified` event in the `src/main.js` is only done if `shouldShowTrayIcon()` is set to `true`. Most likely because inside the notified-callback on win32 platforms the trayIcon is used to show some baloon. But flashing the frame has nothing to do with the tray icon and so it should also be done if the user disabled the tray icon.

The pull request for this should be pretty easy, but I would wait for this one to be merged, otherwise you just end up resolving merge conflicts afterwards.

I also tested that enabling the tray icon, also enables the flashing of the window perfectly fine on my system. So I second that pull request :-)

jasonblais:
Excellent teamwork guys!! Love it :) 

Sounds like this is good to go @yuya-oc

yuya-oc:
Thank you all. The PR looks enough excluding the point @timroes said.

PR #201, opened 2016-07-15
Make unread badge configurable

Razzeee:
Adds a new setting that shows the red dot indicatior on the task bar for unread messages when activated. Default is false.

We may want to check the wording I've choose, we might be able to come up with something better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@
 #### Linux
  - Added an option to make the taskbar icon flash on new messages
 
+#### OS X
+- Added an option to toggle the red dot icon for unread messages (default is on).
+
+#### Windows
+- Added an option to toogle the red dot icon for unread messages (default is on).
+
+
 ### Bug Fixes
 
 
@@ -53,7 +60,7 @@ Release date: 2016-07-18
     - Ctrl+Tab or Alt+Command+Right (Menu Bar -> Window -> Select Next Team): Switch to the next window.
     - Ctrl+Shift+Tab or Alt+Command+Left (Menu Bar -> Window -> Select Previous Team): Switch to the previous window.
     - Right click on the tray item, to see an overview of all your teams. You can also select one and jump right into it.
-  - Added **Help** to the Menu Bar, which includes
+- Added **Help** to the Menu Bar, which includes
     - Link to [**Mattermost Docs**](docs.mattermost.com)
     - Field to indicate the application version number.
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -128,6 +128,8 @@ The Settings Page is available from the **File** menu under **Settings** (Click 
       - This option hides the window from the dock, if the window is closed
    - **Toggle window visibility when clicking on the tray icon** (Windows)
       - If checked, then a click on the system tray icon leads to a toggling of the minimized/maximized state of the window
+   - **Show the red badge for unread messages.** (Windows, OS X)
+      - If this is checked it will show the red dot on your task bar when you have unread messages.
 
 
 ## Menu Bar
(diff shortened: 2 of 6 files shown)


Razzeee:
Rebased and changed targeted version to 1.4.0

jasonblais:
Great, thanks @Razzeee ! I will test this out on Windows!

If anyone else in the community would like to test it on Mac OSX or Linux Ubuntu, you can do so from the circleci artifacts: https://circleci.com/gh/mattermost/desktop/502#artifacts

jasonblais:
I've tested this and looks good to me.

I'm running through a couple of options on the settings text, we might be able to make it a little more precise.

I'll post my suggestion here on 7/25 when I'm back in office.

jasonblais:
For anyone wanting to help test this change in OS X, [you can find the builds here]( https://circleci.com/gh/mattermost/desktop/502#artifacts)

yuya-oc:
Works fine in OS X.

OS X 10.11.5
Mattermost server 3.2

@jasonblais In this PR, the feature is disabled as default even if users upgrade the app from older version. So some users might not notice this change. Is that okay?

jasonblais:
Tested it out once more and propose 

1) an update to text: `[X] Show red badge on taskbar icon to indicate unread messages. Regardless of this setting, mentions are always indicated with a red badge and item count on the taskbar icon.`

@Razzeee, please help update the text in the `changelog` and `setup` as well?

2) agree with @yuya-oc that it would be better to have the badge on by default and give an option to disable it, if the user wants to

Razzeee:
@2) I think that only makes sense for mac, as it seems to be standard there? I feel like windows users will be confused. As most would expect it also to be green, like in the tray icon.

jasonblais:
I think we can fall back to [our three principles](https://github.com/mattermost/desktop/issues/207) - trusted, connected and easy-to-use.

If the red badge is off by default, there is a risk of missing important messages, which would be against our 'connected' principle. We don't know how many users change the options in the settings menu, or would be aware that a setting for the red badge would exist.

If the red badge is on by default, it could be confusing to some initially, but I don't see it affecting usability, or avoid people being connected. It's also used across our mobile apps, so in that way it's consistent.

Razzeee:
Hrm, never seen that on mobile. And some at my work are definitely confused by one app seemingly existing in two different states at once. I feel like it is against the easy to use principle.

Razzeee:
> It's also used across our mobile apps, so in that way it's consistent.

Not sure if you've seen it, but it seems like at least one user is confused about it.
https://mattermost.uservoice.com/forums/306457-general/suggestions/14775780-only-show-the-number-of-mentions-in-the-red-circle

yuya-oc:
I'm also not sure about mobile apps. I think it shows only for mention (at least on iOS). Probably @jasonblais says about the badge of mobile layout in mobile apps?

jasonblais:
Sorry for the confusion regarding mobile: There are plans to update the behaviour but the changes aren't in.

Once they are, Mattermost would have the following for mobile and desktop taskbar icon
 - unread messages: red badge (on by default)
 - unread mentions: red badge with a number (on by deafult)
thus making the experience consistent.

Therefore, [propose making the red badge on by default](https://github.com/mattermost/desktop/pull/201#issuecomment-234992165), giving an option to turn it off from the settings menu?

@Razzeee and others in the community, feel free to share feedback on the topic?

Razzeee:
Updated settings string as suggested. 
Changed default to activated.

jasonblais:
Thanks @Razzeee, appreciate your work here. 

Tested and all looks great.

Anyone want to help test the PR on OS X? Artifacts here: https://circleci.com/gh/mattermost/desktop/533#artifacts

yuya-oc:
Tested on OS X 10.11.5 and it looks okay.

jasonblais:
Great, thanks @yuya-oc, ready to merge

PR #214, opened 2016-07-21
Bump app version and dependencies

Razzeee:
Updated dependencies and app version, as it seems to be the right time for this, as we just released 1.3.0


jasonblais:
Nice - big thanks @Razzeee!

Razzeee:
Some more infos, I tested this on win7 and everything seems to be alright.
I skipped updating to the next major yargs version, as there are breaking changes. Will look into that somewhere down the line.

jasonblais:
"I skipped updating to the next major yargs version, as there are breaking changes."

How do you mean?

Razzeee:
> "I skipped updating to the next major yargs version, as there are breaking changes."

I bumped every dependency to the newest available. I only bumped "yargs" one minor version, as the next major version for it introduced breaking changes. So we would need to update our code to work with that version. 
See: https://github.com/yargs/yargs/blob/master/CHANGELOG.md#v400-2016214-127-0700

Razzeee:
For anyone wanting to help out and test, builds are here: https://circleci.com/gh/mattermost/desktop/507#artifacts

jasonblais:
Excellent, good call, agree. I'm planning to test all Windows-related pull requests this weekend.

But we're also looking for testers on OS X and Linux

jasonblais:
I've been testing the artifacts and are not receiving any issues. I'll report back here if I find any later.

If anyone would like to test OS X and Linux, would be highly appreciated!

Razzeee:
@yuya-oc can you have a quick look on osx? I don't think this will break anything, but it would be nice to get it in.

yuya-oc:
Tested on OS X 10.11.5 and it seems okay. And Electron's changelog seems it works in Linux, and CircleCI is also run on Linux.

In my thought, there is no need to hurry upgrading Electron if there is no strong reason (new features, important fixes, etc). But it's welcome as you read changes of dependencies. I had wrote tests in order to check the whole of app when upgrading Electron and others.

Razzeee:
Well we should go for 1.3.1 as it has a bugfix for windows. Which seems to have been introduced by 1.2.7 or somewhere there. https://github.com/electron/electron/issues/6574

yuya-oc:
I could not see inside of webview in both of Windows 10 and OS X 10.11.5 with Server 3.2.0. It seems successfully loaded the initial page(gray background), but the contents are not rendered. 
And just downgraded only electron-prebuilt to 1.2.8, it looks correct. @Razzeee would you check again?

In Electron 1.3.x, Chromium and Node.js are also upgraded. So perhaps there are some difference we don't know.

Razzeee:
The lastest artifact 1.3.1 is working just fine. Tested with win7 and the 64bit executable. Also using Server 3.2.0.
Will check at home, what windows 10 does.

jasonblais:
Thanks @Razzeee, did you have a chance to take a look in Windows 10?

yuya-oc:
I found an issue for this. After I upgraded Chrome to v52, the URL including team like `https://example.com/team` doesn't show anything. And the host-only URL, `https://example.com` correctly shows UI. Electron v1.3.x also uses

Razzeee:
Had those too, but they resolved itself after a refresh, without removing the team from the url

jasonblais:
@yuya-oc I haven't seen any issues around URLs like `pre-release.mattermost.com/core/` on the server side, and this works with the current release (V1.3)

I think we should fix it though, any thoughts @Razzeee?

yuya-oc:
I tried again and observed following behavior in Chrome.

1. Enter the team URL `https://example.com/team` into the address bar.
2. After loading, the bar shows `https://example.com/team/channels/town-square`.
3. But Chrome shows only gray background.
4. Reload the tab, then the town-square of Mattermost properly appears.

My setting uses team URLs, so I think the same behavior happens in the Desktop app which uses Electron v1.3. As the result, I can't see anything in the tab when launching the app.

Razzeee:
I actually changed my settings to ````https://example.com/team/channels/town-square```` thinking that's how it's supposed to be now.

jasonblais:
@yuya-oc  did the reloading happen automatically on Chrome? It did for me when I tried

If a user enters `https://example.com/team` as their team URL for the desktop app, they shouldn't be forced to refresh it every time. That's a bad user experience.

yuya-oc:
@jasonblais Need to reload manually. The description was missing.

jasonblais:
Does it reproduce on this site too, which is now locked on v3.3 (RC1): `https://pre-release.mattermost.com/core/`

Razzeee:
@jasonblais yes, reproduces on RC1 and on 3.2 (not tested more versions)

I've rebased the PR and bumped electron again, as there have been more fixes.
See: https://github.com/electron/electron/releases/tag/v1.3.2

jasonblais:
Got it, thanks @Razzeee. Will test this during my workday tomorrow.

If any OS X or Linux users would like to help, you can download the artifacts here: https://circleci.com/gh/mattermost/desktop/544#artifacts.

jasonblais:
Update: The platform team is working on a fix [where joining an open team from the team landing page or through team selection page gets stuck on the loading screen](https://mattermost.atlassian.net/browse/PLT-3846).

It is planned for Mattermost v3.3-RC4.

Given the issue is related to what's been discussed here, I will wait until the fix has been merged, likely later today.

Sorry for the delay in completing the test, I haven't found other issues so far.

jasonblais:
@Razzeee I tested with Mattermost v3.3-rc5 and the loading issue no longer seems to reproduce for me. If I enter `https://pre-release.mattermost.com/core/` as the team URL, the page loads successfully (albeit after a few seconds, but the same happens for me with the official v1.3 desktop app).

@yuya-oc Can you help test if the issue still reproduces for you?

I've been testing this version for a couple of days now, so this seems ready to merge.

yuya-oc:
I still need to reload manually on Windows 10... It seems I could not get websocket connection before reloading. Chrome 52 shows same behavior.

Razzeee:
@yuya-oc yep, same is happening for me on win10

src/package.json
```diff
@@ -11,16 +11,16 @@
...
+    "auto-launch": "^2.1.0",
+    "bootstrap": "^3.3.7",
     "electron-squirrel-startup": "^1.0.0",
     "os-locale": "^1.4.0",
-    "react": "^15.0.1",
-    "react-bootstrap": "~0.29.0",
-    "react-dom": "^15.0.1",
-    "yargs": "^3.31.0"
+    "react": "^15.3.0",
+    "react-bootstrap": "~0.30.2",
```
yuya-oc: Did you check the updated app? I could not see anything in the settings page.

In my understanding, `react-bootstrap` is a very unstable library. In 0.28 to 0.29, it was definitely changed when I had written the settings page. So I think there is something wrong by using `react-bootstrap@0.30.x`.

Razzeee: Your right, I removed the offending part. Sorry, for missing that.

Razzeee:
I removed the electron bump to make this more likely to actually get merged ;)

yuya-oc:
I see. I want to be carefully about the Chrome 52 issue because it affects to user experience.

Just tested on OS X 10.11.6, it works well.

Razzeee:
Just tested windows 10 x64. No obvious problems.

jasonblais:
Have tested it as well on Windows 10, and everything seems to be working.

@yuya-oc I think we're ready to merge this?

jasonblais:
Thanks for your patience @Razzeee, but I think the steps we took were important to ensure everything is working as expected.

@yuya-oc As for the Chrome issue -- I was able to reproduce the issue on one of our test sites and have reported it to the platform team.

yuya-oc:
Thanks all.

To upgrade Electron would be great improvement, but also we should not miss the Chrome issue now.

PR #216, opened 2016-07-21
Bold unread tabitem if there are unread messages

Razzeee:
Not sure if this is something we want, but I feel like we should speak the same language everywhere we try to let the user know, that he has unread messages.

tabbar

Tested on Win10

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -219,30 +219,16 @@ var TabBar = React.createClass({
     var tabs = this.props.teams.map(function(team, index) {
       var unreadCount = 0;
       var badgeStyle = {
-        unread: {
-          background: '#383838',
-          float: 'right',
-          color: 'white',
-          textAlign: 'center',
-          marginTop: '5px',
-          width: '10px',
-          height: '10px',
-          marginLeft: '5px',
-          borderRadius: '50%',
-        },
-
-        mention: {
-          background: '#f1342c',
-          float: 'right',
-          color: 'white',
-          minWidth: '19px',
-          fontSize: '12px',
-          textAlign: 'center',
-          lineHeight: '20px',
-          height: '19px',
-          marginLeft: '5px',
-          borderRadius: '50%',
-        }
+        background: '#FF1744',
+        float: 'right',
+        color: 'white',
+        minWidth: '19px',
+        fontSize: '12px',
+        textAlign: 'center',
+        lineHeight: '20px',
+        height: '19px',
+        marginLeft: '5px',
+        borderRadius: '50%',
       };
 
       if (thisObj.props.unreadCounts[index] > 0) {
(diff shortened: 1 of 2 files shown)


jasonblais:
Yes, 100% agree we should make these badges/icons consistent. The black dot is a bit confusing 

Thanks @Razzeee, I've run this through our UX team for feedback, and will share their response here too.

jasonblais:
FYI - an update on this. The UX team in Mattermost is considering an alternate option which is bolding the tab when the team contains unread messages. This would be similar to bolding unread channels and is less flashy than having red dots/badges across the tabs.

Will share here once we have a mock-up ready

Razzeee:
That sounds smart, I will change that later. Should be trivial.

Razzeee:
This is how it looks right now:
tabbar
- The mentions don't have a bold number.
- We can show both new unread and mentions at once, which is pretty cool

jasonblais:
Great, thanks @Razzeee! How would this look like when the `Mattermost` tab has unread messages? Would it be 'bold' enough to distinguish when there is/isn't unread messages? I could then share this with the UX team for feedback.

`We can show both new unread and mentions at once, which is pretty cool` -- This would be inconsistent with our channel bolding though.. If there is an unread message (including a mention), the channel name is bolded in the left hand sidebar, given the mention is also an unread message.

Razzeee:
> Great, thanks @Razzeee! How would this look like when the Mattermost tab has unread messages? Would it be 'bold' enough to distinguish when there is/isn't unread messages? I could then share this with the UX team for feedback.

Would be nice, if you could just grab the artifact and check it. https://circleci.com/gh/mattermost/desktop/519#artifacts

> We can show both new unread and mentions at once, which is pretty cool -- This would be inconsistent with our channel bolding though.. If there is an unread message (including a mention), the channel name is bolded in the left hand sidebar, given the mention is also an unread message.

Yeah, now that I'm reading that, I'm not sure anymore if that's really a case that can happen. Need to check it and if it's really inconsistant, change it.

jasonblais:
Tested on Windows 10 when
 - unread messages only,
 - unread mentions only (which will also bold the tab, as expected),
both when the tab is and isn't in focus

All looks good and UX designers have given it a go!

Really great work here @Razzeee!

jasonblais:
Would be looking for Linux and OS X testers as well, the artifacts can be found here: https://circleci.com/gh/mattermost/desktop/519#artifacts

src/browser/index.jsx
```diff
@@ -219,30 +219,16 @@ var TabBar = React.createClass({
...
-          color: 'white',
-          minWidth: '19px',
-          fontSize: '12px',
-          textAlign: 'center',
-          lineHeight: '20px',
-          height: '19px',
-          marginLeft: '5px',
-          borderRadius: '50%',
-        }
+        background: '#FF1744',
```
yuya-oc: It might be a little thing, the `background` was changed to Material Red A400 (#FF1744).
@asaadmahmood @jasonblais In other part (Red badge in taskbar and tray), Red A400 is used now. Would you hear opinion?

Razzeee: Yep, changed to match https://github.com/mattermost/desktop/blob/master/src/browser/js/badge.js#L12 which we show as an task bar overlay

jasonblais: @yuya-oc Checked with @asaadmahmood and this change is fine

yuya-oc: I see. Thanks.

yuya-oc:
Tested on OS X 10.11.5. The code is common in all platform and it is just for CSS styling about appearance. So I think it's enough.

jasonblais:
Sounds good @yuya-oc -- it's been tested by two people on Windows and one on OS X so ready to merge

Issue #209, opened 2016-07-20
Custom Emoji do not appear properly on Windows and OS X


jasonblais:
I confirm (by marking "x" in the [ ] below): 
- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---
#### Summary

Custom Emoji do not appear properly on Windows and OS X
#### Steps to reproduce
1. Create custom emoji from **Main Menu** > **Custom Emoji**, named `:emoji:`
2. Type `:emoji:` in Town Square.
3. Logout from your team
4. Log back in to your team

**Observed:**
- Step 2: `:emoji:` rendered as an image
- Step 4: `:emoji:` renders as text. The user must go to **Main Menu** > **Custom Emoji**, then back to Town Square for the custom emoji to render as an image.

**Expected:** `:emoji:` renders as an image in both steps 2 and 4.

Reproduced on
- Mattermost v3.2
- Windows 10 Desktop v1.2.1 
- Mac OS X Desktop v.1.3.0

jasonblais:
Wondering if anyone would be able to help investigate this issue?

jasonblais:
@jnugh has offered to help investigate the issue and whether it is related to #212 and #213 -- big thanks!

yuya-oc:
Reproduced in Chrome(Windows 10), and reloading solved this. So it seems that it's definitely caused by cache handling.

jnugh:
Yes, no cache control or expires header. Cache will be invalidated by chance only. 
My pr does a reload as a workaround...

jasonblais:
I have a team site where Custom Emoji didn't work with Mattermost Server v3.2 (Windows 10) and tried to clear cache and reload without success.

They only appeared after going to **Main Menu > Custom Emoji**, and it appears to happen sporadically when I launch the app (not consistently every time).

This is also what a few others have reported. So it might not be fixed with #219

jnugh:
*IF* this is related to a cache hit, #219 should fix it. But it might also be something totally different we didn't think of.

Did you test the 219 build to see if it makes a difference?

jasonblais:
[...]

PR #219, opened 2016-07-22
Workaround: Invalidate cache of the main page on load

jnugh:
This PR will add a cache invalidating page reload to each webview as a workaround.
Mattermost platform does not send an `Expires`/`Cache-Control` header, therefore the cache is not being re validated upon load in most cases.
If the cache is beeing invalidated depends on heuristics https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.2.2

In a browser session, the user is used to hit the reload button when something does not work. In the mattermost app, reload does not reload the embedded webviews.

I added a call to `webview.reloadIgnoringCache` right after the `did-start-loading` event is being fired to force the webview to re fetch the main page, as this results in a load cancel and therefore in a `did-fail-load` I extended the handler for that, so that no error is shown.

---

  - Ubuntu 16.04

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -313,7 +313,7 @@ var MattermostView = React.createClass({
     webview.addEventListener('did-fail-load', function(e) {
       console.log(thisObj.props.name, 'webview did-fail-load', e);
       if (e.errorCode === -3 || // An operation was aborted (due to user action).
-        webview.cacheInvalidated) { //The operation was aborted to invalidate application cache
+        e.errorCode === -300) { //The operation was aborted to invalidate application cache
         return;
       }
 
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -312,7 +312,8 @@ var MattermostView = React.createClass({
 
     webview.addEventListener('did-fail-load', function(e) {
       console.log(thisObj.props.name, 'webview did-fail-load', e);
-      if (e.errorCode === -3) { // An operation was aborted (due to user action).
+      if (e.errorCode === -3 || // An operation was aborted (due to user action).
+        webview.cacheInvalidated) { //The operation was aborted to invalidate application cache
         return;
       }
 
@@ -346,6 +347,13 @@ var MattermostView = React.createClass({
       } else {
         // if the link is external, use default browser.
         shell.openExternal(e.url);
+      }
+    });
+
+    webview.addEventListener("did-start-loading", function() {
+      if (!webview.cacheInvalidated) {
+        webview.reloadIgnoringCache();
+        webview.cacheInvalidated = true;
       }
     });
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Release date: TBD
 - Changed display of unread messages on the team tabbar, they are now shown as bold text
 - Reload only the selected tab and keep its URL on "Reload" and "Clear Cache and Reload".
 - Disabled `eval()` function for security improvements.
+- Invalidate cache before load, to make server upgrades easy
 
 #### Windows
 - Update Mattermost icon for desktop notifications in Windows 10.


jasonblais:
Huge thanks @jnugh, great work on investigating and submitting a potential fix.

This should also resolve another issue where [Custom Emoji do not appear properly on Windows and OS X](https://github.com/mattermost/desktop/issues/209#issuecomment-234707807).

We'd need to test this carefully to make sure it doesn't cause unforeseen issues. Therefore, community's help would be greatly appreciated, and I will be testing this next week on Windows while attempting to test #209 with this PR as well.

yuya-oc:
@jnugh I think the code probably works. Do you have any ideas to check whether web browsers use cache?

jnugh:
You mean a way to check if a page has been loaded from cache? I'm not sure if there is a way that works consistently, probably not.
But as the http response does not specify something different, the page is almost always pulled from cache.

jasonblais:
FYI, I tested [the issue with Custom Emoji not appearing](https://github.com/mattermost/desktop/issues/209#issuecomment-234707807) with the [artifacts](https://circleci.com/gh/mattermost/desktop/511#artifacts), and it appears to not fix it. I added a separate note to the issue.

I didn't find any issues with this PR itself however, and it seems to be working as far as I can tell

jasonblais:
@yuya-oc Any thoughts on this PR -- would you have a chance testing nothing appears to break on OS X?

I think otherwise it's ready to merge as no issues were found when I tested it on Windows.

yuya-oc:
It looks no breaking on OS X. But I don't have confidence whether there is effect by this.

Just for note:
`reloadIgnoreCache()` would ignores caches of the js/css/html. But perhaps the server uses ajax to load everything. So I'm not sure whether `reloadIgnoringCache` makes sense in ajax.

jnugh:
Okay, so let me explain whats actually the problem here:

The Browser / Electron Webview has a cache hit on e.g. pre-release.mattermost.com/core/channels/desktop-app because this *HTML* pages does not have any headers, meta tags, ... to signal that it should not be cached.

So that HTML page includes the main runtime JS which has a name that chanegs from release to release. By chaning the name the cache is busted for this file, no problem with that.

But when you do not reload the page that contains the path to the main runtime js, the new file will not be fetched.

When talking about `ajax`: there are not that many ajax requests which could be cached, as most communication is done through websockets. So there is no need to fix anything.

yuya-oc:
Thanks. Assuming that, it looks okay. And I tested on Ubuntu 14.04, there is no breaking.

jasonblais:
Thanks @jnugh , @yuya-oc, 

I will test it carefully once `/platform` repo is upgraded to v3.3, which will be tomorrow. I'll come back with an update

jasonblais:
@jnugh @yuya-oc 

No issues found either, seems to be working fine.

@yuya-oc Would this be ready to merge?

yuya-oc:
It looks enough for merging and workaround. Thanks!

@jnugh Sorry for late reply. Would you update `CHANGELOG.md`?

jasonblais:
@jnugh Great work here, love it!

@yuya-oc Changelog updated, ready to merge,

yuya-oc:
Thanks!

src/browser/index.jsx
```diff
@@ -312,7 +312,8 @@ var MattermostView = React.createClass({
 
     webview.addEventListener('did-fail-load', function(e) {
       console.log(thisObj.props.name, 'webview did-fail-load', e);
-      if (e.errorCode === -3) { // An operation was aborted (due to user action).
+      if (e.errorCode === -3 || // An operation was aborted (due to user action).
+        e.errorCode === -300) { //The operation was aborted to invalidate application cache
```
Razzeee: This is unfortunatly causing a regression. Please see https://github.com/mattermost/desktop/issues/253#issuecomment-240816346

PR #220, opened 2016-07-22
Added two security improvements as mentioned at http://electron.atom.…

Razzeee:
Tested on win10, but just a little. More tests would be good I guess.

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -1,5 +1,9 @@
 'use strict';
 
+window.eval = global.eval = function() {
+  throw new Error("Sorry, Mattermost does not support window.eval() for security reasons.");
+}
+
 const React = require('react');
 const ReactDOM = require('react-dom');
 const ReactBootstrap = require('react-bootstrap');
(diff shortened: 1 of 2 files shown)


jasonblais:
Thanks @Razzeee, once again!

Here are the artifacts, can we help from the community testing this pull request: https://circleci.com/gh/mattermost/desktop/512

@Razzeee Is there anything specific we should pay more attention to?

Also, can you post the entire URL as a comment, just so that we have it for reference. It's partially cut in the title.

Razzeee:
http://electron.atom.io/docs/all/#checklist

Not sure how to test, as you would have to try to inject code seems run out via eval

yuya-oc:
The best way to test is to actually execute `window.eval()`.
`test/specs/security_test.js` checks whether Node.js API is actually disabled by injecting `require('child_process')`. It might help writing tests for this.

jasonblais:
I see --- @Razzeee have you tested the above yet? 

It's also a good suggestion from @yuya-oc to write tests for this PR, something we should start doing more often. Would set a good example for other similar pull requests from the community in the future.

jasonblais:
I've removed the tester tags here, since they're not really feasible in this case.

Razzeee:
Yep, I'm planning to look at tests as soon as I find some time.

Razzeee:
I've added some code that resemble a test, but it's my first javascript test and my virusscanner is picking on the testsuite as soon as I execute 'npm test' so it might be easier, for someone else to fix.

yuya-oc:
Perhaps the latest spectron is detected? I met before.
Spectron is integrated with WebdriverIO APIs. So please see [`execute`](http://webdriver.io/api/protocol/execute.html) for this test.

Razzeee:
I had a look, but no clue how I would do that.

jasonblais:
Maybe @yuya-oc, can you help take a look when you get the chance

Razzeee:
Thank you @yuya-oc

PR #221, opened 2016-07-25
Flash Window also without systray icon

timroes:
As mentioned in pull request [#200](https://github.com/mattermost/desktop/pull/200#issuecomment-234721964) flashing the mattermost window only worked right now, when you have the systray icon enabled.

Since the setting (and also logically) the systray has nothing to do with the flashing window, this pull request now fixes this behavior, so that the window will flash as long as the setting is turned on (no matter what the systray setting is).

*PS: Sorry for the typo in the commit message, only saw it after I opened the pull request.*

---

  - Archlinux x64

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -218,6 +218,13 @@ allowProtocolDialog.init(mainWindow);
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 app.on('ready', function() {
+  ipcMain.on('notified', function(event, arg) {
+    if (process.platform === 'win32' || process.platform === 'linux') {
+      if (config.notifications.flashWindow === 2) {
+        mainWindow.flashFrame(true);
+      }
+    }
+  });
   if (shouldShowTrayIcon()) {
     // set up tray icon
     trayIcon = new Tray(trayImages.normal);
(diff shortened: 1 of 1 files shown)


Razzeee:
Thanks! If you want to correct the type, just amend the commit and force push it against this branch.

src/main.js
```diff
@@ -218,6 +218,13 @@ allowProtocolDialog.init(mainWindow);
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 app.on('ready', function() {
+  ipcMain.on('notified', function(event, arg) {
```
Razzeee: I don't think we can have ```ipcMain.on('notified', function(event, arg)``` two times. If this even works, it's likely that one overrides the other, which isn't really what we want here. As that would mean, the window will not flash, when you set shouldShowTrayIcon to true.

Razzeee: Take that with a grain of salt, as I haven't tested it and I'm not a node/js expert.

timroes:
@Razzeee it worked for me also with systray switched on. The window still flashes.

src/main.js
timroes: https://nodejs.org/api/events.html#events_emitter_on_eventname_listener You CAN add multiple event listeners to an `EventEmitter`, which ipcMain is :-)

yuya-oc:
Thanks @timroes ! Would you update the commits to fix typo?

@jasonblais The changes are clear, so I think we can merge this without further tests. Any thoughts?

timroes:
Updated the commit message!

jasonblais:
Thanks @timroes!! 

@yuya-oc Given this has been tested by @timroes, and you feel this is good to go, we can merge it.

Issue #228, opened 2016-08-02
Windows: If "Start app on login." is set to True, start the app automatically, but keep the window minimized in the taskbar until the user has set focus to the app


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

#### Summary

If "Start app on login." is set to True, start the app automatically, but keep the window minimized until the user has set focus to the app

#### Steps to reproduce

On
- Windows 10,
- Mattermost Desktop v1.3,
- Mattermost Server v3.2

1. In **Desktop Settings**, set `Start app on login.` to `True`.
2. Restart your machine.

#### Expected behavior

App launches automatically, but remains minimized in the taskba until the user has set focus on the app.

#### Observed behavior

App launches automatically, and is focused on launch.

This means that if the user has opened another program while Mattermost desktop is launching, the desktop app will open on top. This can be annoying and interrupt user's work.

#### Possible fixes

yuya-oc:
Auto-starting was implemented with `node-auto-launch`. @Razzeee Would you check this?

PR #230, opened 2016-08-02
Hide on autostart

Razzeee:
This should hopefully address #228
Untested so far, will grab the artifacts from ci, as soon as they are up and test those.

Let's hope https://github.com/Teamwork/node-auto-launch/issues/31 is solved.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -20,7 +20,8 @@ if (process.platform === 'win32') {
   if (cmd === '--squirrel-uninstall') {
     var AutoLaunch = require('auto-launch');
     var appLauncher = new AutoLaunch({
-      name: 'Mattermost'
+      name: 'Mattermost',
+      isHidden: true
     });
     appLauncher.isEnabled().then(function(enabled) {
       if (enabled)
@@ -40,7 +41,8 @@ var certificateStore = require('./main/certificateStore').load(path.resolve(app.
 var appMenu = require('./main/menus/app');
 const allowProtocolDialog = require('./main/allowProtocolDialog');
 
-var argv = require('yargs').argv;
+var argv = require('yargs')
+  .parse(process.argv.slice(1));
 
 var client = null;
 if (argv.livereload) {
@@ -50,6 +52,12 @@ if (argv.livereload) {
   });
 }
 
+var hideOnStartup;
+if (argv.hidden) {
+  hideOnStartup = true;
+}
+
+// TODO: We should document this if that hasn't been done already
 if (argv['config-file']) {
   global['config-file'] = argv['config-file'];
 }
@@ -364,11 +372,16 @@ app.on('ready', function() {
   });
 
   mainWindow.setFullScreenable(true); // fullscreenable option has no effect.
-  if (window_options.maximized) {
-    mainWindow.maximize();
+  if (hideOnStartup) {
+    mainWindow.minimize();
   }
-  if (window_options.fullscreen) {
-    mainWindow.setFullScreen(true);
+  else {
+    if (window_options.maximized) {
+      mainWindow.maximize();
+    }
+    if (window_options.fullscreen) {
+      mainWindow.setFullScreen(true);
+    }
   }
 
   // and load the index.html of the app.
--- a/src/browser/settings.jsx
+++ b/src/browser/settings.jsx
@@ -22,7 +22,8 @@ const ListGroupItem = ReactBootstrap.ListGroupItem;
 const Glyphicon = ReactBootstrap.Glyphicon;
 
 var appLauncher = new AutoLaunch({
-  name: 'Mattermost'
+  name: 'Mattermost',
+  isHidden: true
 });
 
 function backToIndex() {
--- a/src/package.json
+++ b/src/package.json
@@ -14,7 +14,7 @@
     "electron-connect": "^0.4.4"
   },
   "dependencies": {
-    "auto-launch": "^2.1.0",
+    "auto-launch": "^4.0.1",
     "bootstrap": "^3.3.7",
     "electron-squirrel-startup": "^1.0.0",
     "os-locale": "^1.4.0",


Razzeee:
Doesn't seem to work for me on windows 10.
Can other platforms please try?
https://circleci.com/gh/mattermost/desktop/539#artifacts

Razzeee:
Also doesn't work on win7

jasonblais:
Any help from the community here? 

@Razzeee [Teamwork/node-auto-launch#31](https://github.com/Teamwork/node-auto-launch/issues/31) seems to reference `isHidden` function. Would that be of any use?

Razzeee:
That's what we're trying to use. I'm just not sure if our exe understands --hidden or if they have a bug not even setting that

jasonblais:
@yuya-oc Would you have any thoughts?

yuya-oc:
Not applicable for OS X. In my thoughts, apps should handle `--hidden` flag because `electron` doesn't understand it, too.

Razzeee:
Well the mac implementation isn't trying to use --hidden
https://github.com/Teamwork/node-auto-launch/blob/master/src/AutoLaunchMac.coffee

src/main.js
```diff
@@ -19,8 +19,10 @@ if (process.platform === 'win32') {
   var cmd = process.argv[1];
   if (cmd === '--squirrel-uninstall') {
     var AutoLaunch = require('auto-launch');
+    //BUILD
     var appLauncher = new AutoLaunch({
-      name: 'Mattermost'
+      name: 'Mattermost',
+      isHidden: true
```
yuya-oc: I misunderstood the changelog. This part is called when uninstalling the application. So you should use `isHidden` in `settings.jsx` which enables auto-start. Then, possibly `main.js` should handle the `--hidden` flag.

Razzeee:
Changed and now it's writing the registry key correctly. But it's like you said, electron doesn't understand --hidden for windows

jasonblais:
Any  thoughts from the community here?

Razzeee:
Would be nice to know if this implementation works on osx and Linux. Then we just need to figure out why windows doesn't.

yuya-oc:
Not need to test on OS X because auto-start is not implemented. Please see #164.

jasonblais:
I see, thanks @yuya-oc , 

@timroes @jnugh Would love to have your help checking if the desktop app stays minimized after autostart when logging in to a Linux machine.

If either of you would like to try, you can download an artifact here: https://circleci.com/gh/mattermost/desktop/539#artifacts

timroes:
Doesn't seem to work on my system (Arch x64, pretty custom window manager - so perhaps someone with a regular Desktop Environment should test).

But looking at the generated desktop file:

```
[Desktop Entry]
Type=Application
Vestion=1.0
Name=Mattermost
Comment=Mattermost startup script
Exec=/home/timroes/apps/mattermost-desktop-linux-x64/Mattermost
StartupNotify=false
Terminal=false
```

and looking at the [source code](https://github.com/Teamwork/node-auto-launch/blob/master/src/AutoLaunchLinux.coffee#L16) of `auto-launch` when setting `isHidden` there should actually be the `--hidden` flag in the end of the Exec command, which isn't in my case.

So there might be some problems with the generation of this desktop file. Also I tried starting the above mattermost with `--hidden` directly which doesn't cause it to be hidden on my system either.

Could you point me to the source parts, where `--hidden` is read in mattermost and done somethign appropriate?

Razzeee:
I thought the hidden would be handled by electron packager as that's responsible for building the executeables etc.  But it seems like it doesn't or is not responsible for the handling.

jnugh:
Does not work for me (Unity).

auto-launch generates a desktop file in `~/.config/autostart` which should contains the `--hidden` param. When the default desktop file, which is being generated by `electron-(builder/packager?)`, contains the hidden flag, the app would start minimized even if it wasn't auto started.

timroes:
Yeah. It should :-)

But apparently for me the hidden isn't in the generated desktop file. If the electron wrapper doesn't handle the hidden flag, and you don't do it, that might some the mystery why it isn't working.

jnugh:
https://github.com/Teamwork/node-auto-launch/blob/master/src/index.coffee#L14
`isHidden` is actually mapped to `isHiddenOnLaunch`.

When running this directly from your branch, without using the artifacts, things work like expected. 3.0 is out now. maybe it's been fixed already?

Razzeee:
Version 3.0 got rid of the ES6-Promise dependency, not sure how to fix that for us and get it building again. Changelog doesn't seem to involve a fix for our problem.

jnugh:
https://github.com/Teamwork/node-auto-launch/issues/37 Seems like the current npm package has a bug.

yuya-oc:
Not working in Windows 10. `--hidden` flag is passed to the app, so you have to handle the flag in `main.js`. `mainWindow` and other windows should be managed by js, so Electron should not handle this automatically. The application code should decide the behavior.

Then, please take care probably `yargs` doesn't parse all flags properly in packaged apps because the argc is different in the cases:

- electron-prebuilt: `$ electron APP_PATH args...`
- packaged-app: `$ APP args...`

Razzeee:
Latest code should fix all but the windows installer. That will need https://github.com/Teamwork/node-auto-launch/pull/39 merged and bumped.

yuya-oc:
Tested on Windows 10. As you said, the zip version is working properly, but the installer version is not working yet.

jasonblais:
Adding a `pending` label until https://github.com/Teamwork/node-auto-launch/pull/39 is merged

Razzeee:
@timroes @jnugh 
Would be nice to know if it already works for linux like this: https://circleci.com/gh/mattermost/desktop/649#artifacts

jnugh:
Minimized autostart works for me :)

timroes:
@Razzeee works for me too (custom window manager, Archlinux x64), but brings me back to another problem, which I will open another ticket for (skip taskbar hints).

Razzeee:
Upstream just merged and release https://github.com/Teamwork/node-auto-launch/pull/39
I've updated this PR to include it, so maybe we can still make it into 1.4 :rocket: 

Let's wait for the artifacts

jasonblais:
Awesome! I tried looking for the artifacts earlier today [in CircleCI](https://circleci.com/gh/mattermost/desktop) but couldn't find them.. 

@yuya-oc are they located elsewhere?

Added for v1.4 so we can speed up the testing.

Razzeee:
https://circleci.com/gh/Razzeee/desktop/17#artifacts

jasonblais:
Thanks!

Just tested (Windows 10, 64-bit) and it didn't auto-start for me when I restarted my machine.

I have `Start app on login.` checked.

yuya-oc:
I have tested .zip and setup.exe on Windows 10 64-bit, and .deb on Ubuntu 14.04 64-bit. They are working well.

@jasonblais Would you check again?

jasonblais:
Hmm okay, maybe the issue was on my end.

jasonblais:
Tested; works on Windows 10 after disabling and re-enabling the `start app on login` setting.

@yuya-oc helped troubleshoot and likely related to me running a v1.3 version where auto-start might have been disabled. 

The setting is not overwritten unless you disable and re-enable it in the settings UI (if you simply click 'Save' on the settings page without a change, the option will remain to what it was before in the registry). Investigating whether an easy fix for v1.4/3.4 or if we leave it for the next version.

For this particular PR however, we're good to go. Thanks everyone!

yuya-oc:
So I will change the behavior when saving options in settings. It would always update the registry info. Thanks!

Razzeee:
Please do that in a pr and give me some heads up, as I have some problems
with other apps that work like that. Due to macaffee doing stupid stuff.

yuya-oc:
@Razzeee I created new PR #283. Would you write about the problem you met in detail?

jasonblais:
Thanks @Razzeee, agree that this will need some careful testing.  Let's take it in the next release.

Thanks @yuya-oc for the pull request! We'll test it out

PR #187, opened 2016-07-04
Improve menus due to improvements in electron

Razzeee:
Just updated to the newest version of electron and used some of the new features, to streamline our menus.

See each commit for details.

Tested on windows 10 so far.

PR #240, opened 2016-08-11
Use default electron implementation for zoomin, zoomout and resetzoom

Razzeee:
This is a possible followup to https://github.com/mattermost/desktop/pull/187

The three zoom related menu items are now handled by electron itself.

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -386,9 +386,6 @@ var MattermostView = React.createClass({
     webview.addEventListener("dom-ready", function() {
       // webview.openDevTools();
 
-      // In order to apply the zoom level to webview.
-      webFrame.setZoomLevel(parseInt(localStorage.getItem('zoomLevel')));
-
       // Use 'Meiryo UI' and 'MS Gothic' to prevent CJK fonts on Windows(JP).
       if (process.platform === 'win32') {
         var applyCssFile = function(cssFile) {
@@ -661,22 +658,6 @@ var showUnreadBadge = function(unreadCount, mentionCount) {
     default:
   }
 }
-
-if (!localStorage.getItem('zoomLevel')) {
-  localStorage.setItem('zoomLevel', 0);
-}
-webFrame.setZoomLevel(parseInt(localStorage.getItem('zoomLevel')));
-
-ipcRenderer.on('zoom-in', (event, increment) => {
-  const zoomLevel = webFrame.getZoomLevel() + increment
-  webFrame.setZoomLevel(zoomLevel);
-  localStorage.setItem('zoomLevel', zoomLevel);
-});
-
-ipcRenderer.on('zoom-reset', (event) => {
-  webFrame.setZoomLevel(0);
-  localStorage.setItem('zoomLevel', 0);
-});
 
 ReactDOM.render(
   <MainPage teams={ config.teams } onUnreadCountChange={ showUnreadBadge } />,
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -112,30 +112,16 @@ var createTemplate = function(mainWindow, config) {
     }, {
       role: 'togglefullscreen'
     }, separatorItem, {
-      label: 'Actual Size',
-      accelerator: 'CmdOrCtrl+0',
-      click: () => {
-        mainWindow.webContents.send('zoom-reset');
-      }
+      role: 'resetzoom'
     }, {
-      label: 'Zoom In',
-      accelerator: 'CmdOrCtrl+Plus',
-      click: () => {
-        mainWindow.webContents.send('zoom-in', 1);
-      }
+      role: 'zoomin'
     }, {
       label: 'Zoom In (hidden)',
       accelerator: 'CmdOrCtrl+=',
       visible: false,
-      click: () => {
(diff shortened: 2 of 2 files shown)


jasonblais:
Great, thank you @Razzeee!! 

Would we have two people in the community to help test the pull request? You can download the builds below:
 - [64-bit Windows](https://552-43964563-gh.circle-artifacts.com/0//tmp/circle-artifacts.x04fsxm/mattermost-desktop-win64.zip)
 - [32-bit Windows](https://552-43964563-gh.circle-artifacts.com/0//tmp/circle-artifacts.x04fsxm/mattermost-desktop-win32.zip)
 - [Mac OS X](https://552-43964563-gh.circle-artifacts.com/0//tmp/circle-artifacts.x04fsxm/mattermost-desktop-osx.tar.gz)
 - [Linux](https://552-43964563-gh.circle-artifacts.com/0//tmp/circle-artifacts.x04fsxm/mattermost-desktop-linux-ia32.tar.gz)

[Other artifacts can also be found here](https://circleci.com/gh/mattermost/desktop/552#artifacts).

yuya-oc:
On OS X, the shortcuts are working. But this PR requires Electron v1.3. Because of discussion in #214, we should not use v1.3 now. So unfortunately we should not merge this.

jasonblais:
Once this is merged, will it resolve some of the zoom in/out issues on the desktop app? Or would that be a separate issue?

image
_Above: When increasing font size, the chat area somehow shrinks to the top left corner. When decreasing font size, it expands towards bottom right causing the scrollbar to be hidden_

^Tested on 1.3 and it reproduced there too, so not a new issue. Doesn't always repro, maybe 30% of the time for me on Windows 10

yuya-oc:
@Razzeee Now we can consider this PR. Possibly does this resolve the issue reported on the previous comment and #334?

Razzeee:
It might, we'll need to test it

Razzeee:
Rebased, but doesn't seem to solve #334

jasonblais:
Tested on Windows, can confirm that it doesn't solve #334. 

For the PR itself, only the following shortcuts work: 
 - zoom in: `CTRL + =`
 - zoom out: `CTRL + MINUS`
 - reset zoom: `CTRL+0`

Specifically, `CTRL+SHIFT+=` and `CTRL+PLUS` no longer work for zooming in.

yuya-oc:
Probably need to add some hidden accelerators for zooming in.

jasonblais:
If we add two accelerators for `CTRL+SHIFT+=` and `CTRL+PLUS`, propose we also add [`CTRL+SHIFT+MINUS` for zooming out](https://github.com/mattermost/desktop/issues/301) at the same time

Razzeee:
This PR wasn't about adding featues/keystrokes. It should be exactly the same we already have, see https://github.com/electron/electron/blob/9c19b4e3d5bd9a17d46b09e7ee1ebd6fc2fefc30/lib/browser/api/menu-item-roles.js#L119

jasonblais:
`CTRL+SHIFT+=` and `CTRL+PLUS` already work though, so we shouldn't remove that functionality.

I'm fine adding `CTRL+SHIFT+MINUS` on a separate PR in the future.

Razzeee:
- `CTRL+SHIFT+=` never did work for me, probably due to german keyboard 
- `CTRL+PLUS` does still works for me

yuya-oc:
Tested on Windows 10 with japanese keyboard.

- `CTRL+SHIFT+=` is not applicable because the combination is impossible on the keyboard.
- `CTRL+PLUS` works as `CTRL+SHIFT+;`. `CTRL+;` also works.

yuya-oc:
On english keyboard, `CTRL+PLUS` works as `CTRL+SHIFT+=`, and `CTRL+=` also works.
 I noticed one thing, the behavior is different depends on focus inside of the window.

zoom

When clicking a tab bar, the behavior is same with original one. But when clicking the webview, only it is zoomed, and the region is correct.

jasonblais:
I started with a completely fresh install, and it seems to be working now.. 

Thanks again @Razzeee and @yuya-oc, ready to merge.

One interesting note: I tried with [win64.zip](https://53-58761399-gh.circle-artifacts.com/0/tmp/circle-artifacts.foOszCe/mattermost-desktop-win64.zip) and #334 wasn't reproducing. It only reproduced on the installer.

yuya-oc:
@jasonblais Please let me confirm, did you met the behavior like my previous gif?

jasonblais:
> When clicking a tab bar, the behavior is same with original one. But when clicking the webview, only it is zoomed, and the region is correct.

Sorry for missing the note below the GIF.

I just tested and you're correct, the issue only reproduces after clicking on the tab bar. Tried on both the .exe and .zip files for 64-bit Windows.

Note: in v3.4.1 the issue described in #334 occurs whether you've clicked on the webview or the tab bar, at least on 64-bit Windows installer. So this would be an improvement to it.

Razzeee:
For reference https://github.com/electron/electron/issues/7375

yuya-oc:
Thanks all! Yeah, #334 is improved compared to v3.4.1.
From the result, we have to try to keep the focus of webview as possible.

PR #241, opened 2016-08-12
Change window title to the one thats set from the platform for the cu…

Razzeee:
As specified here: https://mattermost.uservoice.com/forums/306457-general/suggestions/14974134-windows-mattermost-client-make-us-of-the-window-c

## Known bug 
(I would like some help here)
~~- If you change your team, via the tabs, it doesn't change the title. You first need to select some channel in the new team.~~

~~I fall back to the app name now in this case
Possible solution is to go grab the title on a switch, but might need some custom code.~~

~~- If you have multiple teams and and are in team a, if a channel title in team b changes, it will be set as window title~~

### DIFF
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -201,8 +201,9 @@ var MainPage = React.createClass({
         thisObj.handleSelect(index);
       }
       var id = 'mattermostView' + index;
-      return (<MattermostView key={ id } id={ id } style={ thisObj.visibleStyle(thisObj.state.key === index) } src={ team.url } name={ team.name } onUnreadCountChange={ handleUnreadCountChange }
-                onNotificationClick={ handleNotificationClick } ref={ id } />)
+      var is_active = thisObj.state.key === index;
+      return (<MattermostView key={ id } id={ id } style={ thisObj.visibleStyle(is_active) } src={ team.url } name={ team.name } onUnreadCountChange={ handleUnreadCountChange }
+                onNotificationClick={ handleNotificationClick } ref={ id } active={ is_active } />)
     });
     var views_row = (<Row>
                        { views }
@@ -409,9 +410,11 @@ var MattermostView = React.createClass({
     });
 
     webview.addEventListener('page-title-updated', function(event) {
-      ipcRenderer.send('update-title', {
-        title: event.title
-      });
+      if (thisObj.props.active) {
+        ipcRenderer.send('update-title', {
+          title: event.title
+        });
+      }
     });
 
     webview.addEventListener('console-message', (e) => {
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -94,6 +94,11 @@ var MainPage = React.createClass({
       key: newKey
     });
     this.handleOnTeamFocused(key);
+
+    var webview = document.getElementById('mattermostView' + key);
+    ipcRenderer.send('update-title', {
+      title: webview.getTitle()
+    });
   },
   handleUnreadCountChange: function(index, unreadCount, mentionCount, isUnread, isMentioned) {
     var unreadCounts = this.state.unreadCounts;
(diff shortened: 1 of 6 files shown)


src/browser/index.jsx
```diff
@@ -150,6 +150,12 @@ var MainPage = React.createClass({
   handleOnTeamFocused: function(index) {
     // Turn off the flag to indicate whether unread message of active channel contains at current tab.
     this.markReadAtActive(index);
+
+    // TODO handle titles when switching teams correctly
+    ipcRenderer.send('update-title', {
+      title: remote.app.getName()
```
yuya-oc: Probably you can use `webview.getTitle()` by using index to get the element here.

src/browser/index.jsx
```diff
@@ -395,6 +401,12 @@ var MattermostView = React.createClass({
       }
     });
 
+    webview.addEventListener('page-title-updated', function(event) {
```
yuya-oc: Mattermost server changes the title by appending `*` when there are unread messages. So this event should be ignored when the tab is not activated. Probably `MattermostView` might need `active` prop.

Razzeee: Well I could just filter out any title starting with ```* ``` but that will fail if anybody has a channel that starts like that. (hopefully unlikely)

Razzeee:
One bug down one two go, working much better now already.
We just need to handle the ```*``` case and maybe also decide if we want to show the ````*``` in the title at all.

src/browser/index.jsx
yuya-oc: Sure, so we should not handle `*` here. I think the point is that this event is fired even if the tab is not activated. So the listener should check whether its webview is actually shown.

Or, you might be able to add/remove the listener in `handleSelect`, but probably the code would be complex.

Razzeee: But do we want it even for the tab thats currently selected?

yuya-oc: Yes. So the windows title should be updated only for the current tab. If the title became one of non-active tab, it would be confusing. As I wrote at first, the listener would be able to know the state by using a prop.

Razzeee: Yes, but how to change the data in the prop? I haven't seen any events on webview that would fit the bill.

yuya-oc: There are no events for this. So in my idea, you can add the prop to [`MattermostView`](https://github.com/mattermost/desktop/blob/v1.3.0/src/browser/index.jsx#L187-L188). It can be know by using `thisObj.state.key===index`.

Razzeee: Well ```index``` is only known outside of ```MattermostView```so I have no idea to get that working unfortunatly. And I can't set the prop from outside of ```MattermostView```

yuya-oc: I added at `Change-Title` branch. Please see https://github.com/mattermost/desktop/compare/0de211f...df4a1f4

Razzeee:
@yuya-oc 
I've cherrypicked your two commits.

Tested everything on windows 10 and can't find any problem anymore. If you ask me this is ready to be tested on every platform/merged.

yuya-oc:
Thanks. It's now ready to test and need one tester. Artifacts are available here. https://circleci.com/gh/mattermost/desktop/595#artifacts

jasonblais:
I have to say that this is quite nice! Just tested and love this improvement.

Great work @Razzeee and @yuya-oc! 

Could we have someone's help testing this on OS X? Perhaps @mmahkamov your help?

You can download [the artifact here](https://595-43964563-gh.circle-artifacts.com/0//tmp/circle-artifacts.6buRj5m/mattermost-desktop-osx.tar.gz)

yuya-oc:
Not necessary, your testing is enough. This doesn't depends on operating system.
Thanks!

PR #255, opened 2016-08-18
Bump prettify dependencies and apply prettification

Razzeee:
This is something we might want to wait with, until we got some PRs merged. As this will cause pretty much every PR to need a rebase.


Razzeee:
Btw you might want to use https://github.com/mattermost/desktop/pull/255/files?w=0 to review this

jasonblais:
Great! I've added a pending label for now -- as you said, we'll want to wait until more PRs have been merged to v1.4.

I love cleaning up code where possible, so thank you for submitting the PR :1

yuya-oc:
The indent size is defined as `2` in `.editorconfig` and `gulpfile.js`. Why it is changed in actual codes?

Razzeee:
Sorry, I thought those we're changes in the dependencies. Found the problem and will update the PR soon.

Razzeee:
Please have a look again, I hope that it now has the same functionallity

Two changes were needed:
- Use snake_case for options instead of camelCase https://github.com/tarunc/gulp-jsbeautifier/blob/master/CHANGELOG.md#200---2016-03-01
- Use gulp-diff instead of mode/showdiff https://github.com/tarunc/gulp-jsbeautifier#other

.gitattributes
```diff
@@ -0,0 +1,9 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Set our files to use lf, so our prettify will always be happy
```
Razzeee: Fixes the files switching line ending on windows after checkout and therefore showing up as changed after running prettify.

gulpfile.js
```diff
@@ -25,32 +26,38 @@ gulp.task('prettify:verify', ['prettify:sources:verify', 'prettify:jsx:verify'])
 
 var prettify_options = {
   html: {
-    eol: '\n',
-    indentSize: 2
+    indent_size: 2
   },
   css: {
-    eol: '\n',
```
Razzeee: We can let these eol go, as this is the default value already. So this doesn't change anything.

gulpfile.js
```diff
@@ -248,5 +255,5 @@ gulp.task('sync-meta', function() {
   appPackageJson.description = packageJson.description;
   appPackageJson.author = packageJson.author;
   appPackageJson.license = packageJson.license;
-  fs.writeFileSync('./src/package.json', JSON.stringify(appPackageJson, null, '  ') + '\n');
-});
+  fs.writeFileSync('./src/package.json', JSON.stringify(appPackageJson, null, '  '));
```
Razzeee: This fixes './src/package.json' changing each time you run prettify or build. As build will add this '\n' and prettify will remove it.

Razzeee:
I was able to address some more nice things that were bothering me, especially when building on windows.

Razzeee:
Btw, as the changelog is much smaller now, this should be able to get merged sooner and not wait for other PRs to go in.

yuya-oc:
Looks nice! Would you add also `"end_with_newline": true` for each option? It would make better.

Razzeee:
@yuya-oc 
changed.
Leads to some changes for html and css files, check the changelog for infos.

jasonblais:
:+1:  Removed `pending` label.

yuya-oc:
Thanks!

PR #257, opened 2016-08-20
Bump electron builder

Razzeee:
I've tested this on windows and the installer seems to perform as expected. And I couldn't find anything wrong with the program itself.

We might want to use this changed/new feature for macOS https://github.com/electron-userland/electron-builder/releases/tag/v5.6.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "package:windows": "gulp package:windows",
     "package:osx": "gulp package:osx",
     "package:linux": "gulp build && build --platform linux --arch all && npm run linux-additions",
-    "linux-additions": "cp resources/icon.png resources/linux/create_desktop_file.sh release/linux/ && cp resources/icon.png resources/linux/create_desktop_file.sh release/linux-ia32/",
+    "linux-additions": "mkdir -p release/linux/ && cp resources/icon.png resources/linux/create_desktop_file.sh release/linux/ && mkdir -p release/linux-ia32/ &&  cp resources/icon.png resources/linux/create_desktop_file.sh release/linux-ia32/",
     "package:all": "gulp package:all",
     "prettify": "gulp prettify",
     "installer": "node ./script/installer.js"
--- a/package.json
+++ b/package.json
@@ -12,6 +12,10 @@
   "engines": {
     "node": ">=4.2.0"
   },
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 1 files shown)


Razzeee:
Not sure why this is timing out on circleci. Already tried to push again, second time it hits the timeout.

Razzeee:
Changed to only bump electron builder. As ncu (https://www.npmjs.com/package/npm-check-updates) doesn't show the winstaller update for some reason and I've read this https://github.com/electron/windows-installer/issues/92#issuecomment-220921269

Razzeee:
Same thing happening, I'll lower the dependency and try to see which version "breaks" it. Might take a few builds.

Razzeee:
Updated and also includes a small linux fix

package.json
```diff
@@ -36,7 +40,7 @@
     "chai": "^3.5.0",
     "chai-as-promised": "^5.3.0",
     "devtron": "^1.3.0",
-    "electron-builder": "5.2.1",
+    "electron-builder": "7.9.0",
```
yuya-oc: v7.9.0 has the bug, generated app can't start. You should use [v7.10.2](https://github.com/electron-userland/electron-builder/releases/tag/v7.10.2).

yuya-oc: Sure. So we are glad if you test before.

Razzeee: I'm sorry, but from my understanding, that's a deb package problem only. Do you really expect me to test every platform? I thought the PRs are there to check the code and get tests from every platform deemed important?
The artifact provided in https://circleci.com/gh/Razzeee/desktop/49#artifacts/containers/0 runs fine in windows. Is my understanding right, that `electron-builder` is used to generate every platform, even windows?

Razzeee:
Bumped version also seems to run fine on windows 10

yuya-oc:
For now, `electron-builder` is used only for .deb package. So I expected you had already tested that. I have tested v7.10.2 in this PR, so it's acceptable after new push.

Razzeee:
Sorry, I think I really need to pick your brains about how everything is build and document it somewhere in the git/wiki

yuya-oc:
Sure.
But please understand it's very helpful for us to review if there is your analysis about what you are doing. Otherwise, I need to research everything from scrach whether I can merge the PR.

PR #251, opened 2016-08-16
Improved window state handling

Razzeee:
As stated in #249 the window handling was off under some circumstances.

This tries to improve it and is working fine with win10 x64.

There might be cases I missed, where this also needs to be done. Just let me know.

Issue #249, opened 2016-08-16
Maximized state is not restored when clicking on tray icon or notifications


jfperusse-bhvr:
I confirm (by marking "x" in the [ ] below): 

- [x] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [x] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [x] This is a product defect issue that can be reproduced using the steps listed below.
- [x] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

#### Summary

When clicking on a notification or on the tray icon, the maximized state of the window is lost if the window was minimized to the taskbar.

#### Steps to reproduce

This is for the Windows Desktop App v1.3.0.

- Open the Windows Desktop App.
- Maximize the window, then minimize it to the taskbar.
- Click on a notification or on the tray icon.

#### Expected behavior

The window should be restored to its maximized state.

#### Observed behavior

The window comes back to its original size, not maximized.

PR #261, opened 2016-08-23
Add access to settings through tray menu

St-Ex:
Hello,

Here is a small change to allow settings from tray menu. It's really usefull when your using frameless window (coming in another PR) to gain space.

I've tested on Linux Mint 17.

Regards,

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -78,7 +78,7 @@ You have to configure the application to interact with your teams.
    - Windows: Press `Alt` key to bring up the menu at the top of the window, then click `File -> Settings`.
    - OS X: Click `Mattermost` from the menu at the top of the screen, then click `Preferences...`.
    - Linux: Click `File -> Settings` on the menu.
-   - If tray menu : Right-click on tray icon and click `Settings`
+   - All : right-click on tray icon and click `Settings`
 
 2. Press `+` button next to the "Teams" label.
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -78,6 +78,7 @@ You have to configure the application to interact with your teams.
    - Windows: Press `Alt` key to bring up the menu at the top of the window, then click `File -> Settings`.
    - OS X: Click `Mattermost` from the menu at the top of the screen, then click `Preferences...`.
    - Linux: Click `File -> Settings` on the menu.
+   - If tray menu : Right-click on tray icon and click `Settings`
 
 2. Press `+` button next to the "Teams" label.
 
--- a/src/main/menus/tray.js
+++ b/src/main/menus/tray.js
@@ -29,6 +29,14 @@ function createTemplate(mainWindow, config) {
     }), {
       type: 'separator'
     }, {
+      label: 'Settings',
+      click: () => {
+        mainWindow.loadURL('file://' + __dirname + '/browser/settings.html');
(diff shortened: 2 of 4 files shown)


docs/setup.md
```diff
@@ -78,6 +78,7 @@ You have to configure the application to interact with your teams.
    - Windows: Press `Alt` key to bring up the menu at the top of the window, then click `File -> Settings`.
    - OS X: Click `Mattermost` from the menu at the top of the screen, then click `Preferences...`.
    - Linux: Click `File -> Settings` on the menu.
+   - If tray menu : Right-click on tray icon and click `Settings`
```
Razzeee: That `If` somehow doesn't make sense in my head, might be me?

Razzeee:
Thanks for your PR.

To everyone willing to test, the builds can be found here: 
https://circleci.com/gh/St-Ex/desktop/22#artifacts

Razzeee:
Working nicely on Windows 10 btw

docs/setup.md
St-Ex: As tray menu is an option and might not be present, i wanted to reflect it and not mislead people on an hidden menu.
Is is too much ?

Razzeee: Is it an option? I'm not aware of a platform that doesn't have it?

yuya-oc: It's an option in OS X and Linux, but I feel we don't have to mention about it here. If the user don't use the tray icon, the sentence would be ignored.

St-Ex: I've changed the phrasing.

src/main/menus/tray.js
```diff
@@ -29,6 +29,14 @@ function createTemplate(mainWindow, config) {
     }), {
       type: 'separator'
     }, {
+      label: 'Settings',
+      click: () => {
+        mainWindow.loadURL('file://' + __dirname + '/browser/settings.html');
+        mainWindow.show();
```
yuya-oc: This would not work properly when the window is minimized from the maximized state (#249). So you should handle the window like [other shortcuts](https://github.com/mattermost/desktop/blob/f85337db6598e0a8fa2cd94c7868ce1b440314cc/src/main/menus/tray.js#L15-L26).

Razzeee: @yuya-oc  'mainWindow.webContents.send('switch-tab', i);'? Why would that be needed for settings?

yuya-oc: Of course not necessary.

St-Ex: On Linux Mint 17, it is working. Even if you're not on the good workspace, it switches your ws and open up the window.
Is it something proper to OSX/Windows ?

yuya-oc: Yes, it's for Windows and OS X. The problem reported in #249 was not reproduced in Linux (https://github.com/mattermost/desktop/pull/251#issuecomment-241278012).

St-Ex: I've changed it. Btw, i just ran a search regarding show or restore. It is used in a lot of place on the code. Could it be good to call one method to do it or add the equivalent directly onto the mainWindow object ?

yuya-oc:
Tested on Windows 10. Thanks!

PR #263, opened 2016-08-26
Fix for window toggle on tray click

Razzeee:
Probably an oversight of my previous window handling refactor.

Especially the toogle window on tray icon click option was borked.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -258,7 +258,7 @@ app.on('ready', function() {
 
     trayIcon.setToolTip(app.getName());
     trayIcon.on('click', function() {
-      if (!mainWindow.isVisible()) {
+      if (!mainWindow.isVisible() || mainWindow.isMinimized()) {
         if (mainWindow.isMinimized()) {
           mainWindow.restore();
         }


yuya-oc:
Thanks! Would you indicate the commit of refactor here?

Razzeee:
I don't think so, as I would think it worked fine in 1.3.0 and it just broke in nighlies

yuya-oc:
The feature was broken by [`src/main.js`-L245](https://github.com/mattermost/desktop/blob/7429df05f5b845cc16bf4cb11f49a09953198565/src/main.js#L245) of 7429df0. As far as I read previous code, I feel the condition `mainWindow.isMinimized()` should be written there.

Razzeee:
Can change it back later today. In my tests it seemed like there is o one case which doesn't get hit, with the code as it zits now in this pr

Razzeee:
Updated

yuya-oc:
Tested following cases on Windows 10. Thanks!

- Minimize -> Tray click => Open window
- Maximize -> Minimize -> Tray click => Open maximized window
- Close -> Tray click => Open window
- Maximize -> Close -> Tray click => Open maximized window
- Tray click => Minimize window
- Maximize -> Tray click => Minimize window

PR #265, opened 2016-08-29
Jshint basic config and some cleanup

Razzeee:
I've tried to come up with a basic jshint setup. If this is not something we want to have/do, just let me know.

I took the liberty of doing some of the suggestions, let's see if it breaks something when building.


yuya-oc:
Thanks! But for this purpose, I have considered using eslint, which is used in /platform repo. It can support also JSX by using plugins. Any thoughts?

Razzeee:
If that's used in platform, we should absolutely use that.
I think JSHint also has an jsx extension, but doesn't really matter then. Should we keep the second commit of this PR anyway?

yuya-oc:
The second commit looks nice, so I think we should merge it. BTW to keep git-blame info, eslint should not be added in v1.4.0 development.

Razzeee:
@yuya-oc 
can you explain why it would screw up git-blame?

yuya-oc:
First, the rule of /platform doesn't match for /desktop repo, so it certainly screws up git-blame. On the other hand, we can decide a new rule for /desktop, but we have to find what is better because the current code is not using any guidelines.

Just in my thought, `eslint-config-airbnb` seems nice for the base rule, but it certainly affects code structure.

Razzeee:
I dropped the jshint stuff and only kept the code improvements. I will create an issue to talk about an linter.

yuya-oc:
Thanks!

yuya-oc:
Assuming my previous post, jshint is not bad idea until we prepare eslint because jshint didn't break current codes.
Sorry for many times, but I would be glad if you add a test command into `package.json`. It's better to execute jshint in CI.

Razzeee:
I think eslint shouldn't break anything either, we can set the rules to only be warnings and just have it in here as a long time goal. So that we can improve our warning score over time.

PR #272, opened 2016-09-04
Use electron-context-menu insted of bare contextmenu

jnugh:
This swaps the context menu implementation we used to have with `electron-context-menu`. This fixes multiple behavior issues with the default implementation.

Like:
 - Copy / Paste / Cut only work when the context matches
 - Copy links work flawless
 - Saving images 
 - (See https://github.com/sindresorhus/electron-context-menu)
---

  - Ubuntu Linux 16.04


Razzeee:
Hooray, finally :)
This is really awesome.

Windows 10 seems to work fine.

For everyone that wants or can test the other operating systems, you can find builds here: https://circleci.com/gh/mattermost/desktop/657#artifacts

jasonblais:
Yay, big thanks @jnugh!! This is great :)  

I tested it on Windows 10, but was having trouble copying links..I tried the following test case:

 1. Post a message `[Link to Docs](docs.mattermost.com)`
 2. Highlight the message and right-click
 3. Select `Copy Link`
 4. Attempt to paste in a message box

Observed: Nothing is pasted
Expected: Link `docs.mattermost.com` pasted to the message box.

Otherwise things look as expected, including saving images (which is really nice!)

jnugh:
I can't reproduce this behavior. Is this what you tried to do: Screencast?

yuya-oc:
Tested on Windows 10 and OS X 10.11.6.

On Windows, I could not select `Paste` in the @jasonblais's step 4. It seems copying works fine, and pasting is working after copying a text at the outside of the app.

src/package.json
```diff
@@ -21,6 +21,7 @@
     "react": "^15.3.0",
     "react-bootstrap": "~0.29.0",
     "react-dom": "^15.3.0",
-    "yargs": "^3.32.0"
+    "yargs": "^3.32.0",
+    "electron-context-menu": "^0.5.0"
```
yuya-oc: Should be placed with alphabetical order.

jasonblais:
Thanks @jnugh! That's correct, I tried the exact steps you showed in the gif screenshot (on Windows 10), but when attempting to paste into the message box, nothing is actually pasted.

It does work if I copy a link in the desktop app and paste it in the web app instead, as @yuya-oc mentions. So it's step 4 (pasting the link in the desktop app) that seems to fail.

Any thoughts from the community what might be causing it?

jasonblais:
@jnugh has helped investigate the issue and posted a question in the [electron-context-menu](https://github.com/sindresorhus/electron-context-menu/issues/18) repository.

Given the PR fixes several issues, propose we merge the pull request and open a separate issue to fix pasting links on Windows.

yuya-oc:
I agree to merge, so I have made an issue #279. Thanks all!

Issue #270, opened 2016-09-04
Changing find/search from CTRL+SHIFT+F to CTRL+SHIFT+S


jasonblais:
There was [discussion last week in our community channel](https://pre-release.mattermost.com/core/pl/do9y9xasb3bf9ykegotrgxtzcy) on what `CTRL+F` should do in the desktop app. 

Right now on `master`, `CTRL+F` let's you search the current channel deviating from browser behaviour and making the Mattermost experience a little inconsistent.

If we want to implement 'Find' for the desktop app that would be identical to `CTRL+F` in the browser, we'll need to change what `CTRL+F` does before 1.4 is released.

Hence, propose we have "Find" and "Search" as separate shortcuts on desktop:
 - `CTRL+F` = Find (to be implemented in the future)
 - `CTRL+S` = Search (focus is set to Mattermost search bar)
 - `CTRL+SHIFT+S` = Search in current channel (focus is set to Mattermost search bar with `in:channel` automatically added)

Feedback welcome,

jasonblais:
@MetalCar and @jnugh  suggested using `CTRL+SHIFT+F` as a shortcut, but this is considered to bring up a list of flagged posts (to be implemented by the Mattermost core team).

PR #262, opened 2016-08-25
Fix webview doesn't get focus correctly

yuya-oc:
For #231.

`webview.focus()` was used to focus on the `<webview>`, but it's a DOM element function. So only the tag was focused in the host window, and its `webContents` was not focused. As the result, the shortcuts related to Mattermost UI didn't work in some cases.
Electron is upgraded to 1.2.8 because of `webContents.isFocused()`.

## Tested environment
Windows 10 and OS X 10.11.6.

## Shortcuts
- Ctrl+F (provided by the app)
- Ctrl+K (provided by mattermost UI)

## Test conditions
- After clicking application window.
- After clicking a tab.

PR #273, opened 2016-09-04
Move current search in channel function to Shift+CTRL+S

Razzeee:
This addresses #270 

Tested and working on win10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ Release date: TBD
 - Disabled `eval()` function for security improvements.
 - Invalidate cache before load, to make server upgrades easy
 - Removed misleading shortcuts from tray menu, as they didn't work
-- Ctrl/Command+F puts cursor in search box to search in current channel.
+- Ctrl/Command+S puts cursor in search box to search in current team.
+- Shift+Ctrl/Command+S puts cursor in search box to search in current channel.
 - Add access to settings through tray menu
 - Removed unclear desktop notifications when failed to load tabs.
 - Reload automatically the failed tab when the computer becomes online.
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -152,7 +152,8 @@ Below lists menu options (shortcut keys are listed in brackets, `Ctrl` becomes `
   - **Copy** (Ctrl+C) - Copies selected text
   - **Paste** (Ctrl+V) - Pastes text from clipboard
   - **Select All** (Ctrl+A) - Select all text in input box
-  - **Search in Channel** (Ctrl+F) - Puts cursor in search box to search in current channel
+  - **Search in Team** (Ctrl+S) - Puts cursor in search box to search in the current team
+  - **Search in Channel** (Shift+Ctrl+S) - Puts cursor in search box to search in current channel
 - **View**
   - **Reload** (Ctrl+R) - Reload page from the server
   - **Clear Cache and Reload** (Ctrl+Shift+R) - Clear cached content in application and reload page
--- a/src/browser/index.jsx
+++ b/src/browser/index.jsx
@@ -81,6 +81,12 @@ var MainPage = React.createClass({
       webview.send('activate-search-box');
     });
 
+    // activate search box in current chunnel
+    ipcRenderer.on('activate-search-box-in-channel', (event) => {
+      let webview = document.getElementById('mattermostView' + thisObj.state.key);
+      webview.send('activate-search-box-in-channel');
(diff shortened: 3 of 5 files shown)


jasonblais:
Wow, what a quick turnaround! Great work! Tested on Windows 10 as well, works nicely

I did find one UX defect which is related to the PR (but not caused by it): when doing `CTRL+SHIFT+S` in a DM channel, the search bar is filled with two confusing strings.

It's caused by the fact that DM channel handles/URLs are automatically constructed to be `[userid_1]_[userid_2]` in Mattermost, and this is what `CTRL+SHIFT+S` is picking up.

image

Should `CTRL+SHIFT+S` only work in public and private channels but not in DMs for now? Auto-complete search options for `in` only work for public and private channels in Mattermost web app too.

jasonblais:
If anyone else would like to help with testing the PR, you can download a CircleCI artifact here: https://circleci.com/gh/Razzeee/desktop/11#artifacts

Test cases to consider:
 1. `CTRL+S` should set focus to the search bar
 2. `CTRL+SHIFT+S` should set focus to the search bar AND automatically add `in:[Channel URL]` to the search bar.
 3. `Menu Bar > Edit` should show two options: 
   - one for `Search in Team` which has equivalent behaviour to case 1
   - one for `Search in Channel` which has equivalent behaviour to case 2

Both of the above should allow you to perform search across public channels, private groups and direct messages.

Razzeee:
From what I've tested it seemed like the search also worked for DMs, will have to recheck. But I feel like it's a small price to pay for being able to search DMs

jasonblais:
Sorry @Razzeee, to clarify: DM search works, but the user experience for it is confusing.

I'm going to check with the platform team if there are any plans on handling DM search using the `in` modifier and respond back here

jasonblais:
Found an old ticket that was planning to address this case with usernames: https://mattermost.atlassian.net/browse/PLT-91

@Razzeee Propose making one small update to the PR, which is that when you use `CTRL+SHIFT+S` for a DM channel, the text that auto-populates the search bar is: `in:{username}`.

That would make the user experience less confusing I think, while still keeping the functionality for DMs

Razzeee:
Not easy to do in a sane way :/

jasonblais:
Makes sense, we would probably need the [Jira ticket](https://mattermost.atlassian.net/browse/PLT-91) fixed by the platform team first.

So I guess there's at least these options for v1.4:

1) have `CTRL+SHIFT+S` only work on public/private channels 
 - cons: loses ability to search in DMs, might make it feel like the shortcut is broken
2) have `CTRL+SHIFT+S` work on all channels, including DMs
 - cons: user experience for searching DMs is confusing and below Mattermost UX standards

Razzeee:
1)
Cons no idea how to track if we're in a dm channel, in a sane way

yangchen1:
Great work @Razzeee! Tests 1/2/3 passes for the general case. However if you switch tabs, and then try 1/2/3, focus is not set (the problem is being discussed/hopefully fixed in this PR https://github.com/mattermost/desktop/pull/262)

For all tests, searching within a channel for a direct message gives "in:ooze8mspijga3qufiap76i565r__x6sdh1ok1tyd9f4dgq4" or something similar. This should be changed to the username or display name.

jasonblais:
Hey @Razzeee, 

Given the functionality is already in, we can merge this PR and test how searching DMs feels before we release v1.4.

However, wondering if you'd be interested to work on this open issue in Jira: https://mattermost.atlassian.net/browse/PLT-91

It won't likely be completed by the platform team in the near future and would an improvement to the user experience of searching in DMs.

jasonblais:
@yuya-oc Tested in Windows and OS X, ready for code review and merge.

yuya-oc:
Yup, I've tested it on Ubuntu 14.04. It looks enough for now.

yuya-oc:
For DM issue, probably we can parse the username instead of URL when the active channel is DM. But the solution which is proposed in PLT-91 seems simple.

Razzeee:
@jasonblais 
not sure if there is an documentation for building platform on windows, last time I checked there wasn't. So I can't really do builds/test/improvements.

jasonblais:
Thanks @Razzeee, that makes sense

We have an install guide for Windows in progress, would love to have your help testing it once it's ready, if you'd like?

Razzeee:
I'm not sure, I think I will shift my focus to some private projects soonish :)
So I will have to cut down on time I use for ther stuff.

Issue #181, opened 2016-06-28
Show validation errors in the settings


Razzeee:
If you try to create a team and give it an invalid url, the app blocks you from doing it.
It might be nice to show the user what's going on by some kind of error handling.

Maybe like done in this example http://getbootstrap.com/components/#glyphicons-examples

jasonblais:
Would this be an easy update? It would be quite nice particularly with new user on-boarding as it's currently against [Mattermost UX guidelines > Error Messages](https://docs.mattermost.com/developer/fx-guidelines.html#input-hints)
.

Following the behaviour elsewhere on Mattermost, the user would be allowed to click on the `Add` button, but an error message would be shown below the input field:
 - if `Name` is not set: `Name is a required field.`
 - if only `URL` is not set: `URL is a required field.` 

![image](https://cloud.githubusercontent.com/assets/13119842/17835654/80c0dfaa-6744-11e6-82fc-23ea03320837.png)


We should probably also enforce `http://` and `https://` as required for the URL, but that's a separate issue from this.

yuya-oc:
I'm not sure. Probably we have to use `react-bootstrap` now, but we need to migrate to latest version in order to refer to documents because it looks still unstable.
If I could, I would like to consider this, including refactoring. It might be better not to use `react-bootstrap` now.

jasonblais:
@yuya-oc By next version, do you mean after v1.4?

If it involves significant refactoring, I would agree. It would be better to work on it early for v1.5 in that case.

yuya-oc:
Ah, sure. That should be done after v1.4. Then I tried adding a simple one at 255ff9a.

<img width="551" alt="blank_name" src="https://cloud.githubusercontent.com/assets/1412057/17855363/5c217a54-68b3-11e6-85ff-58004e8ca697.png">
<img width="549" alt="blank_url" src="https://cloud.githubusercontent.com/assets/1412057/17855366/5c29aa08-68b3-11e6-834c-79341ed70b31.png">
<img width="550" alt="invalid_url" src="https://cloud.githubusercontent.com/assets/1412057/17855364/5c281472-68b3-11e6-93d6-534ce5274387.png">

jasonblais:
Awesome, many thanks @yuya-oc! 

I've queued it for a meeting to get the texts UX approved. Starting to get so excited for v1.4 with all the improvements :)

jasonblais:
@yuya-oc Here's some feedback:

[...]

PR #274, opened 2016-09-06
Add simple validation for Name and URL in the settings page

yuya-oc:
For #181.

Tested on Windows 10. The appearance is below.

It's only for the appearance of html, so I think it's enough if we have one tester. Artifacts are available at https://circleci.com/gh/yuya-oc/desktop/7#artifacts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Release date: TBD
 - Added back/forward features for the current tab.
   - Windows and Linux: Alt+Left, Alt+Right
   - OS X: Command+[, Command+]
+- Added simple validation for Name and URL in the settings page.
 
 #### Windows
 - Added an option to toogle the red dot icon for unread messages (default is on).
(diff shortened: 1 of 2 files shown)


jasonblais:
Thanks @yuya-oc!

During testing, I expected to be able to click on the 'Add' button, *after which* the error would be shown.

I actually didn't see the `Name is required` on my first read.

Not sure if others have thoughts?

image

yuya-oc:
Ah, surely we didn't define the timing to show validation. I expected the user can't click 'Add' until all input fields are validated. The user can know what is wrong before clicking the button.

jasonblais:
Thanks @yuya-oc, makes sense.

Showing the error message after a user attempts to add the team would be more consistent with the rest of the Mattermost experience.

Would it be a lot of work adding this change?

yuya-oc:
@jasonblais Thanks for feedback. Probably it's not hard, so I'll try later.

yuya-oc:
Updated. The artifacts are here. https://circleci.com/gh/yuya-oc/desktop/10#artifacts

jasonblais:
Tested on Windows 10, looks great. Thank you @yuya-oc !

PR #283, opened 2016-09-13
Save autostart settings always when saving options

yuya-oc:
For [the issue reported in #230](https://github.com/mattermost/desktop/pull/230#issuecomment-246700060).
When upgrading node-auto-launch, registry info was not updated because the logic updated the info when the option was changed. So now the settings UI saves the autostart info when pressing 'Save' button.


Razzeee:
Needs to be prettified :)

jasonblais:
I think this PR might not be necessary with the fix planned here https://github.com/mattermost/desktop/issues/291

yuya-oc:
Sure, so I close this.

Issue #287, opened 2016-09-15
Windows 10: Confusing text appears at the bottom of desktop notifications


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

#### Summary

Windows 10: Confusing text appears at the bottom of desktop notifications

#### Steps to reproduce

- Operating System: Windows
- Mattermost Desktop App version: v3.4 (RC1)
- Mattermost Server version: v3.4
- Steps to reproduce: Have another user send a direct message to you and observe the desktop notification.

#### Expected behavior

No text at bottom of notifications.

#### Observed behavior

![image](https://cloud.githubusercontent.com/assets/13119842/18552519/bb25d6d8-7b2a-11e6-93fd-be95e5ae0adb.png)

#### Possible fixes

N/A

jasonblais:
Might reproduce on other platforms too.

We should consider fixing this for 1.4

yuya-oc:
This reproduces on Windows only. The reason is that the application is not using ApplicationUserModelID which is used in the start menu shortcut. It's `com.squirrel.mattermost.Mattermost`.

yuya-oc:
In the setup process of `mattermost-setup-*.exe`, it should create the shortcut which has ApplicationUserModelID. But it seems the ID is not attached to the shortcut. So the application name is not applied there.

jasonblais:
I see, thanks @yuya-oc. Would that be feasible to fix for 3.4?

jasonblais:
Fixed by #290

PR #290, opened 2016-09-16
Fix ApplicationUserModelID in the start menu shortcut

yuya-oc:
For #287

To correctly show the application name in toast notifications on latest Windows 10, the shortcut which has "ApplicationUserModelID" is necessary. The application expected `com.squirrel.mattermost.Mattermost` due to electron-winstaller and Squirrel. But probably the package name of `package.json`, `mattermost-desktop` was used when creating the shortcut. Unfortunately I was not sure what was actually used, so I explicitly fixed the ID.

This might affect the internal name of .deb packages. It's not critical for now, but if the server provided .deb packages, they might conflict.

When testing this PR, please use installers. If the problem is not solved, please try uninstalling the existing app.

Tested on Windows 10 RS1

https://circleci.com/gh/yuya-oc/desktop/26#artifacts

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "mattermost-desktop",
+  "name": "mattermost",
   "productName": "Mattermost",
   "version": "3.4.0",
   "description": "Mattermost Desktop application for Windows, Mac and Linux",
--- a/src/main.js
+++ b/src/main.js
@@ -30,6 +30,7 @@ if (process.platform === 'win32') {
   }
 }
 
+app.setAppUserModelId('com.squirrel.mattermost.Mattermost'); // Use explicit AppUserModelID
 require('electron-squirrel-startup');
 
 const fs = require('fs');
--- a/src/package.json
+++ b/src/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "mattermost-desktop",
+  "name": "mattermost",
   "productName": "Mattermost",
   "desktopName": "Mattermost.desktop",
   "version": "3.4.0",


yuya-oc:
The tests excluding prettifying was succeeded in the previous CircleCI. https://circleci.com/gh/yuya-oc/desktop/25#tests

jasonblais:
@yuya-oc It appears the builds failed for the above artifacts too.. Can you help taking a look?

yuya-oc:
Actually no problem. Please use the artifacts.

There are two parts in the tests(app functionality, code format), both of them should succeed when merging.

- 1st: Failed by code format. So actually test succeeded.
- 2nd: Failed by test's unstability. I have changed only code format.

jasonblais:
I tried testing on the artifacts here https://circleci.com/gh/yuya-oc/desktop/26#artifacts but I receive the same text at the bottom of the desktop notification box.

image

yuya-oc:
@jasonblais Hmm... I have tried again, but it seems the problem is solved.

Steps to test:

1. Uninstall existing app.
2. Install v1.3.0.
3. Toast shows `com.squirrel.mattermost.Mattermost` (the issue).
4. Quit the app, then install v3.4.0.
5. Toast shows `Mattermost`.

toast

jasonblais:
@yuya-oc I first tried the above steps without uninstalling the app, the again after uninstalling it first as you advised above.

The PR works if the app is uninstalled and the problem is resolved.

Can you help me understand why the uninstallation step would be expected? Will the PR work for those who are currently using v1.3 or an RC for v3.4, then later download the final release for v3.4?

yuya-oc:
The purpose of steps from 1 to 3 is to reproduce the original issue and to prepare the actual condition for upgrading from v1.3.0.

Then I tried two more cases. In both cases, the artifact properly overwrote the shortcut. As the result, the PR should work for both cases, currently using v1.3.0 or v3.4.0-rc.

1. At step 4, don't quit the existing app. (Test upgrading the app which is still running)
2. At step 2, install v3.4.0-rc1 instead of v1.3.0. (Test upgrading from RC)

jasonblais:
Sounds good, thanks @yuya-oc for testing the above two additional cases. Much appreciated!

Should be ready to merge then!

Issue #293, opened 2016-09-19
Disable search in current team `CTRL+SHIFT+S` until DM search works properly


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

#### Summary

Disable search in current team `CTRL+SHIFT+S` until DM search works properly

#### Steps to reproduce

1. [Download v3.4 RC1](https://github.com/mattermost/desktop/issues/285) for any platform
2. Use `CTRL+SHIFT+S` in a DM channel

#### Expected behavior

App sets focus on the search bar, then adds `in:[{username1}_{username2}]`

where {username1} is the display name of the user.

#### Observed behavior

App sets focus on the search bar, then adds `in:[userid1_userid2]`

![image](https://cloud.githubusercontent.com/assets/13119842/18636712/cb0f59cc-7e57-11e6-9df9-7afc0464083c.png)

#### Possible fixes

Propose disabling `CTRL+SHIFT+S` until DM URLs use the usernames instead of user IDs, [ticket tracked here](https://mattermost.atlassian.net/browse/PLT-91).

This feature is not critical for the desktop app (although would be very nice to have), and appears as a bug for end users.

yuya-oc:
Closed by #295.

PR #295, opened 2016-09-19
Disable search in current team `CTRL+SHIFT+S`

yuya-oc:
Due to #293
Need to revert when DM search works properly.

This just removed the menu, so just confirmation is enough.


jasonblais:
Thanks! Will test it once the build has completed

jasonblais:
@yuya-oc verified on Windows 10:
 - `CTRL+SHIFT+S` removed from the menu bar and does no action on the app
 - `CTRL+S` remains on the menu bar and sets focus to the Mattermost search box

Issue #291, opened 2016-09-17
If auto-starting the desktop app is turned on, an old app will continue to auto-start unless the auto-start option is reset from the Settings page.


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

#### Summary

Registry key not updated if using auto-start and downloading a new version

#### Steps to reproduce

Operating System: Windows 10
MM Desktop App version: 3.4 RC1
MM Server version: 3.4

1. Download Desktop v1.3 from the [downloads page](https://about.mattermost.com/downloads/)
2. Go to the Settings page via **File > Settings**
3. Turn on **Start app on login.** and hit **Save**
4. Download the RC for v1.4/3.4: https://github.com/mattermost/desktop/issues/285
 - [64-bit installer here](https://releases.mattermost.com/desktop/3.4.0-rc1/mattermost-setup-3.4.0-rc1-win64.exe)
5. Verify download was successful by checking the version number from **Help > Version**
6. Exit the application from **File > Exit**
7. Restart your machine

#### Expected behavior

The latest downloaded version of Mattermost starts.

#### Observed behavior

Mattermost v1.3 auto-starts.

#### Possible fixes

Registry key doesn't update when you download the desktop app. Thus, the old version will start unless the **Start app on login.** setting is reset (which only happens if you first turn the option off, save, then turn the setting back on)

@Razzeee has started progress with an [initial artifact here](https://circleci.com/gh/Razzeee/desktop/18#artifacts).

jasonblais:
The setting was new for v1.3, so releasing v1.4/3.4 will be the first time users will hit this issue, and they most likely will hit it.

This is a really important bug to fix before the release - could we get the community's help here to chime in?

jasonblais:
@Razzeee @yuya-oc, any potential thoughts here?

yuya-oc:
I feel [Razzeee's branch](https://github.com/Razzeee/desktop/commit/66757ca406e9b3796f7d4c0c95fb43fdaec4dea9) would potentially help this.

[...]

PR #296, opened 2016-09-20
Make sure our auto launch link is up do date

Razzeee:
So like we talked about, this makes sure the windows registry is kept up to date when upgrading versions.
Fixes/closes: https://github.com/mattermost/desktop/issues/291

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -17,15 +17,26 @@ process.on('uncaughtException', (error) => {
 
 if (process.platform === 'win32') {
   var cmd = process.argv[1];
+  var AutoLaunch = require('auto-launch');
+  var appLauncher = new AutoLaunch({
+    name: 'Mattermost',
+    isHidden: true
+  });
   if (cmd === '--squirrel-uninstall') {
-    var AutoLaunch = require('auto-launch');
-    var appLauncher = new AutoLaunch({
-      name: 'Mattermost',
-      isHidden: true
-    });
+    // If we're uninstalling, make sure we also delete our auto launch registry key
     appLauncher.isEnabled().then(function(enabled) {
       if (enabled)
         appLauncher.disable();
+    });
+  }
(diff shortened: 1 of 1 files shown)


jasonblais:
Tested on Windows 10 with following steps:

1 - [Download v1.3 from downloads page](https://about.mattermost.com/downloads)
2 - Confirm registry key in `C:\...\AppData\Local\mattermost` is v1.3

image
3 - [Download artifact](https://circleci.com/gh/Razzeee/desktop/17#artifacts), in this case 64-bit Windows installer

4 - Confirm registry key in `C:\...\AppData\Local\mattermost` has now changed to v1.4

image
So this appears to work for the installer. I also tried a couple other configurations (including when the auto-start option is on vs. off) and it's working smoothly.

-------

However, if I [download the 64-bit .zip artifact](https://17-58761399-gh.circle-artifacts.com/0//tmp/circle-artifacts.xXpeVy0/mattermost-desktop-win64.zip), and carry out the above steps, the registry key remains v1.3.

Is this expected? 

If so, I think that's okay as the majority of the Windows Desktop app users download the installer. We should just need to communicate it in our changelog if that 's the case (which I can update).

Razzeee:
It's expected, as the installer runs the upgrade script and the zip file has no installation.

jasonblais:
Sounds good.

Should I try and test other configurations as well?

Razzeee:
No clue what else would need to be tested.

jasonblais:
@Razzeee Sorry, just realized. Can you help submit the PR against the `release-3.4` branch?

Otherwise it won't be included in the 3.4 builds

yuya-oc:
To make sure, I also confirmed with Windows 7 32-bit, it works well.

jasonblais:
Perfect, thanks @yuya-oc!

yuya-oc:
Thanks!

Issue #150, opened 2016-05-21
Own installation animation 


Razzeee:
It would be nice, to get someone to do a mattermost is now installing animation as a gif. So that we can use it when the setup runs.

jasonblais:
FYI, there's some progress on this front, we should have a loading animation gif for 1.4!

jasonblais:
In light of the progress from the Mattermost team, I have tentatively added this for the 1.4 milestone

Razzeee:
Any news here @jasonblais ?

jasonblais:
It didn't make it for Mattermost v3.3, but I'll push for it to be added for v3.4.

That will let us use it for the desktop app too.

jasonblais:
Update: The GIF is *almost* ready now, @asaadmahmood will take a look at appropriate colouring for it as black might be too dark.

jasonblais:
Unfortunately it didn't make it to Mattermost v3.4, so removing from v1.4 milestone.

jasonblais:
The GIF loading icon is ready, attached is the 50x50 version below.

We should reduce opacity of the GIF by 0.5 so that the icon is grey rather than black - black is too heavy for the desktop app.

![2016-07-21 20_07_48](https://cloud.githubusercontent.com/assets/13119842/18682296/af053a5c-7f39-11e6-9dfc-d9c5a16a498d.gif)

jasonblais:
We might want to build an installation wizard instead of using an installation icon. That's much more common for Windows

PR #298, opened 2016-09-20
Use mattermost branded loading animation for installation

Razzeee:
This replaces the generic installation animation used by the windows installer, with the mattermost loading gif.

I checked it on a local build of the installer and it works fine on windows. We should test the circleci build, but other platforms must not be checked.


Razzeee:
Adresses https://github.com/mattermost/desktop/issues/150#issuecomment-234442469

jasonblais:
I think we should use the generic installation animation GIF for now. It looks super-weird as an installation animation at the moment.

image

The one we have here works better as a loading icon, that replaces the "Loading..." text, something like attached.

image

Issue #279, opened 2016-09-12
Context menu can't paste a URL which is copied from a link


yuya-oc:
#### Summary

Reported in #272. Context menu can't paste a URL which is copied from a link on Windows.

#### Steps to reproduce

- Operating System: Windows 10
- Mattermost Desktop App version: v1.4 (#272 artifact)
- Mattermost Server version: N/A
- Steps to reproduce the issue:

1. Post a message `[Link to Docs](docs.mattermost.com)` (any url)
2. Highlight the message and right-click
3. Select `Copy Link`
4. Attempt to paste in a message box, but can't select `Paste`.

#### Expected behavior

Select `paste` can be selected and the URL is pasted.

#### Observed behavior

See [@jasonblais's comment in #272](https://github.com/mattermost/desktop/pull/272#issuecomment-244635083) and further comments.

On OS X and Linux, this doesn't happen.

#### Possible fixes

sindresorhus/electron-context-menu#18 or fixes on Electron.

jasonblais:
Fixed by #306 :tada:

Big thanks to @Razzeee, and again to @jnugh who originally added the [`electron-context-menu`](https://github.com/mattermost/desktop/pull/272) for v3.4

PR #306, opened 2016-09-21
Bump dependency to get rid of windows copy bug

Razzeee:
This should fix the windows copy bug.
I haven't tested it yet, but if upstream/@jnugh did it's job right it should be fine ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Release date: TBD
 ### Improvements
 
 #### Windows
+ - Copying a links address and pasting it inside the app now works
 
 #### macOS
 
--- a/src/package.json
+++ b/src/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "auto-launch": "^4.0.1",
     "bootstrap": "^3.3.7",
-    "electron-context-menu": "^0.5.0",
+    "electron-context-menu": "^0.6.0",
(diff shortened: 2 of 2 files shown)


Razzeee:
Should close https://github.com/mattermost/desktop/issues/279

yuya-oc:
Tested on Windows 10, it works well.

ref: sindresorhus/electron-context-menu#19

CHANGELOG.md
```diff
@@ -88,7 +88,6 @@ Version number updated to 3.4 to make numbering consistent with Mattermost serve
 ### Known Issues
 
 #### Windows
- - Copying a link address and pasting it inside the app doesn't work
```
yuya-oc: Please keep this line. Instead, please write "Bug Fixes" section of the latest changelog.

Razzeee: Why should we keep this line if it's fixed and working?

Razzeee:
Updated

CHANGELOG.md
jasonblais: I think we should still keep it in the v3.4 changelog as it'll continue to be a bug in that version.

We should just ensure it's removed in the next version's changelog

Razzeee: Your obviously right. When I drafted this, I hoped it would still make it to release ;)
Will change.

yuya-oc:
Thanks!

PR #307, opened 2016-09-21
Bump some dev dependencies

Razzeee:
Seems to work fine with Mattermost platform pre release and 3.2

You can find builds here: https://circleci.com/gh/Razzeee/desktop/27#artifacts/containers/0


package.json
```diff
@@ -37,9 +37,9 @@
     "chai-as-promised": "^5.3.0",
     "devtron": "^1.3.0",
     "electron-builder": "5.2.1",
-    "electron-connect": "~0.3.9",
-    "electron-packager": "^7.0.1",
-    "electron-prebuilt": "1.2.8",
+    "electron-connect": "~0.6.0",
+    "electron-packager": "^8.0.0",
```
yuya-oc: You don't have to bump `electron-packager`. I'm working on updating build commands.
(In addition, `gulpfile.js` uses the deprecated parameter, `version-string`.)

package.json
```diff
@@ -37,9 +37,9 @@
...
+    "electron-prebuilt": "1.4.0",
```
yuya-oc: I have checked [a known issue](https://github.com/mattermost/desktop/pull/214#issuecomment-237263102) with servers, pre-release and v3.2.0. The issue doesn't happen. Did you check?
Then, to make sure, we might have to write its compatibility and workaround in the changelog. Any thoughts?

yuya-oc:
If you write the detail, for example, the motivation to bump, the part which you checked, etc., it's helpful to review.

yuya-oc:
As you wrote, electron 1.4.0 seems to work with the pre-release server and v3.2.0 server. So I think we can bump electron.

Razzeee:
Updated

jasonblais:
welcome back @Razzeee :)

yuya-oc:
@Razzeee Just curious, post's background blinks when moving the mouse cursor across the posts. Have you experienced?

Razzeee:
@yuya-oc 
They just highlight, as normal for me on windows 10

yuya-oc:
Tested again on Window 10 and macOS.

On Windows 10, I saw the blinking like the gif (actually, it's very fast blinking). @jasonblais and others, would you check this?

background_blinking

On macOS, the app seems to work well.

jasonblais:
@yuya-oc  

Hmm, haven't been able to reproduce the blinking. I tried it on the very same messages you shared above. Does it reproduce consistently across channels and messages?

yuya-oc:
@jasonblais Yes, it happens also on channel list, but messages cause that more often. It seems that the vertical position of mouse cursor is randomly wrong in the window, so I see the blinking.

Then, I tried the app on some DPI. On 100% and 125%, the problem seems not to happen. On 150%(I'm using), it happens. I think almost of users use 100% or 125%, so I feel we might as well merge this for now. Any thoughts?

jasonblais:
Yeah, I think merging makes sense. 

Should we add a known issue for the next changelog that using DPI of 150% might cause the vertical position of mouse cursor to be incorrect? It's okay not to fix it, but let's others know about it.

yuya-oc:
@jasonblais Sure, that should be written. I add it later.

PR #310, opened 2016-09-22
Stop using Docker and using Ubuntu 14.04 for this in CircleCI

yuya-oc:
This makes the test faster than current. The purpose of PR is just for notice.

Before merging, we have to change the setting in CircleCI.
After merging, further commits for existing PRs must rebase in order to pass the test.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,9 +1,4 @@
 machine:
-  pre:
-    - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.6 10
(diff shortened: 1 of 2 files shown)


Razzeee:
Nice, when this is in, we should check if the  electron builder pr stops with the timeouts

yuya-oc:
The problem should be gone. The point was wine's configuration.
`wineboot --init` is executed when using electron-builder to setup wine environment, but it seems to block the process if there is X11 display (probably due to GUI). So I used `unset DISPLAY && wineboot --init` before using electron-builder.

Razzeee:
Sounds promising, thank you.

PR #339, opened 2016-10-08
Use only electron-builder to package the app

yuya-oc:
Tested on Windows 10 64-bit, macOS 10.12 and Ubuntu 16.04 64-bit.

Only `electron-builder` will be used to packaging the app. It improves complicated packaging-commands and supports code-signing for our public release.

In this PR, please let me know whether the artifacts correctly start.

https://circleci.com/gh/yuya-oc/desktop/68#artifacts


Razzeee:
Both .exe files (32 and 64 bit) work fine for me on windows 10

jasonblais:
Works on the Windows .zip files too

docs/development.md
```diff
@@ -24,6 +24,26 @@ $ npm run build
...
+### Packaging
+Package `dist/` directory as distributable formats with [`electron-builder`](https://github.com/electron-userland/electron-builder).
+Packages will be generated into `release/` directory.
+
+```
+$ npm run package:<all | windows | mac | linux>
+```
+
+#### Code signing
+To build trusted packages, need to set some environment variables.
```
jasonblais: Maybe: "Set environment variables to build trusted packages" 

Just for consistency with other sections (Building, Packaging, ...) which start with an action word.

yuya-oc:
Updated `development.md`.

jasonblais:
@jungh, @yangchen1 and @magicmonty, would you like to help test the pull request?

Builds available to download here: https://circleci.com/gh/yuya-oc/desktop/68#artifacts

Main thing to test is to check if the app opens properly. It's good also to test that the basic functionality works fine (e.g. posting a message).

jnugh:
I'm not able to upgrade the deb package, unless I use `--force-overwrite` or remove the old package before installation. This is because the package is named mattermost instead of mattermost-desktop.

When successfully installed, things work as expected.

yuya-oc:
I found an issue for package name in the latest push. The name of deb package depends on `name` of `src/package.json`. If it's set as `mattermost-desktop`, it causes invalid application user model id for Windows.

Windows app: `com.squirrel.mattermost.Mattermost` by `src/main.js`
Windows installer: `com.squirrel.mattermost-desktop.Mattermost-desktop` by `name` of `src/package.json`
This mismatch causes #287, so should be avoided.

To resolve this, we have two choices. I feel 1 is better due to Any thoughts?

|| `name` | AppUserModelID | deb package | side effects |
|---|---|---|---|---|
| 1 | `mattermost` | Keep `com.squirrel.mattermost.Mattermost` | Become `mattermost` | * Need `--force-overwrite` or reinstalling in debian/ubuntu. <br>* If platform provided deb package in future, need to consider its name (`mattermost-server`?).
| 2 | `mattermost-desktop` | Become `com.squirrel.mattermost-desktop.Mattermost-desktop` | Keep `mattermost-desktop` | If pinned shortcuts exist, the taskbar icon might be duplicated on Windows. Need to unpin once. And the user might lose notification settings on Windows |

jnugh:
We cannot have both, can we? Some special build flags or something?
All Linux (deb) users need to be instructed to remove the old version before install.
I tried to install the deb with Software Center at first and didn't get a good error message. When I was a user I might not see the info about reinstalling depending on where it is.

yuya-oc:
Unfortunately no in `electron-builder`. So we have to add a script to control `name` by depending on the target.

yuya-oc:
Updated. Somehow I had thought `--em` option of `electron-builder` doesn't work. But now it's working, so both of `com.squirrel.mattermost.Mattermost` and `mattermost-desktop` are available.
https://circleci.com/gh/yuya-oc/desktop/74#artifacts/

jasonblais:
Option 1: 

Between the two options, the first seems to have fewer side effects, agree.

Since the .deb packages are unofficial, we'll only provide them in the release tag and we can add a note there that installing the app requires a re-install.

I'm a little worried about forward-compatibility -- if the platform team provided .deb packages, the `mattermost` tag would be more obvious for the server than for the desktop app.

Option 2:

If we went with `mattermost-desktop`, would we need help from the Debian community to fix these two issues:

a) taskbar icon might be pinned more than once
b) the user might lose notification settings on Windows

Also, would 2b) only happen for the .deb packages?

yuya-oc:
@jasonblais a) and b) are only for Windows when choosing the case 2.But now the latest artifacts should solve this problem, so they have no changes in their package name.

jasonblais:
Thanks @yuya-oc! Can confirm #287 does not reproduce on 64-bit .exe or .zip files

jnugh:
Looks good to me :) Instally cleanly and works as expected.

jnugh:
I guess he fixed it with a custom build flag. No decision needed anymore,
we can have both ;) I didn't check the new deb yet however.

Jason Blais <notifications@github.com> schrieb am Sa., 22. Okt. 2016, 20:41:

yangchen1:
+1 @yuya-oc, works on MacOS and #287 doesn't reproduce. Great work!

yuya-oc:
Thanks all! 🎉

PR #345, opened 2016-10-19
Purge cache on app update

jnugh:
First of all, please read [`CONTRIBUTING.md`](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)

---

- [x] Update `CHANGELOG.md` and/or `doc/*.md` if it's necessary.
  - Ubuntu 16.10

### DIFF
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -45,8 +45,10 @@ var upgradeV0toV1 = function(config_v0) {
   return config;
 };
 
-var upgrade = function(config) {
+var upgrade = function(config, newAppVersion) {
   var config_version = config.version ? config.version : 0;
+  if (newAppVersion)
+    config.lastMattermostVersion = newAppVersion;
   switch (config_version) {
     case 0:
       return upgrade(upgradeV0toV1(config));
--- a/src/main.js
+++ b/src/main.js
@@ -82,14 +82,15 @@ var config = {};
 try {
   var configFile = global['config-file'];
   config = settings.readFileSync(configFile);
-  if (config.version != settings.version) {
-    config = settings.upgrade(config);
+  if (config.version != settings.version || wasUpdated()) {
+    clearAppCache();
+    config = settings.upgrade(config, app.getVersion());
     settings.writeFileSync(configFile, config);
   }
 }
 catch (e) {
   config = settings.loadDefault();
-  console.log('Failed to read or upgrade config.json');
+  console.log('Failed to read or upgrade config.json', e);
 }
 ipcMain.on('update-config', () => {
   config = settings.readFileSync(configFile);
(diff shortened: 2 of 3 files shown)


src/common/settings.js
```diff
@@ -2,6 +2,7 @@
 
 const fs = require('fs');
 const settingsVersion = 1;
+const app = require('electron').app;
```
yuya-oc: The js is tested in `test/specs/settings_test.js` by mocha. Please consider not using `require('electron')` here. For example, add an arg for `upgrade()`.

src/main.js
```diff
@@ -254,6 +273,7 @@ allowProtocolDialog.init(mainWindow);
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 app.on('ready', function() {
+
```
yuya-oc: Unexpected blank line.

jnugh:
i rebased my PR to address your the changes mentioned by your code review.

jasonblais:
oThanks @jnugh, @yuya-oc! 

Is there a way for us to test this PR? Or is it something where we'd need to monitor for caching issues once the new desktop app is released?

yuya-oc:
By editting `lastMattermostVersion` of `config.json`, the app would assume that the app version has been changed. So you will see the app starts reloading when launching.

Wndows: `%HOME%\AppData\Roaming\Mattermost\config.json`

jnugh:
You could check if the Cache folder in the same directory gets smaller after a version change...
But it might only be smaller for a few seconds, because when loading the webview cache should build up again.

jasonblais:
Hmm, so I tried with the following steps:

1) Download artifact from: https://circleci.com/gh/mattermost/desktop/747#artifacts
2) Install Mattermost (using the 64-bit Windows installer)
3) Go to `%HOME%\AppData\Roaming\Mattermost\config.json` and modify `lastMattermostVersion` from `3.4.1` to `3.5.0`

It seems the app doesn't detect the change, but keeps the version at `3.4.1`. Is there a step I missed?

I also tried closing and re-opening the app.

yuya-oc:
Tested on Windows 10, and the feature works well.

@jasonblais The version is rewritten when launching the app, so keeping `3.4.1` is correct. The clearing caches works very quickly, so probably you can't see reloading. You can see it by watching `%HOME%\AppData\Roaming\Mattermost\Cache`.

jnugh:
The version number in config.json should change back to 3.4.1 when you
start the application next time. The cache will perform purged before the
version string is written to config.json.

Jason Blais <notifications@github.com> schrieb am Sa., 22. Okt. 2016, 21:02:

yangchen1:
+1 @jnugh! Good work.

Issue #353, opened 2016-10-28
Change target screen resolution from 800x600 to 1280x800


jasonblais:
I confirm (by marking "x" in the [ ] below): 
- [X] This is not a feature idea. Feature ideas go here: http://www.mattermost.org/feature-ideas/
- [X] This is not a troubleshooting question. Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/
- [X] This is a product defect issue that can be reproduced using the steps listed below.
- [X] I have read [CONTRIBUTING.md](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---
#### Summary

[An estimated 94% of browsers are 1280x800 or larger](http://www.w3schools.com/browsers/browsers_display.asp), and using the current 800x600 resolution can cause issues such as hiding the channel header on initial launch.

Thus, propose changing our target screen resolution to be 1280x800.

(Note that the default desktop app window size should be a little smaller vertically than the target 1280x800 resolution, so that it's not hidden by the bottom taskbar)

jasonblais:
(Note that the default desktop app window size should be a little smaller vertically than the target 1280x800 resolution, so that it's not hidden by the bottom taskbar)

yuya-oc:
The default window dimensions should be smaller than target screen to avoid conflicting with other UI, e.g. task bar. So the dimensions are set to 1000x700.

jasonblais:
We're also targeting 1024x768 screen resolution instead, update title and description.

This is because 1366x768 resolution is common, in which case targeting 1200x800 isn't recommended

PR #354, opened 2016-10-28
Increase default window size

danrabinowitz:
Per https://github.com/mattermost/desktop/issues/353

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -386,8 +386,8 @@ app.on('ready', function() {
     window_options = JSON.parse(fs.readFileSync(bounds_info_path, 'utf-8'));
   }
   catch (e) {
-    // follow Electron's defaults
-    window_options = {};
+    // follow Electron's defaults, except for window dimensions
+    window_options = {width: 1100, height: 800};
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -386,8 +386,8 @@ app.on('ready', function() {
     window_options = JSON.parse(fs.readFileSync(bounds_info_path, 'utf-8'));
   }
   catch (e) {
-    // follow Electron's defaults
-    window_options = {};
+    // follow Electron's defaults, except for window dimensions
+    window_options = {width: 1100, height: 800};
```
yuya-oc: As far as reading #354, the height should be smaller than 800 in order to avoid that the window is hidden by the taskbar.

jasonblais:
Many thanks @danrabinowitz for the pull request!

It looks like the Circle CI build is failing and I think they would be resolved by running `npm run prettify`.

If you're able to address @yuya-oc's notes above, we'll be able to start the review process and approve the pull request swiftly.

Thanks again @danrabinowitz!

PR #370, opened 2016-11-23
Proposed release process for desktop app

jasonblais:
Proposed release process for desktop app, a subset of what's used by the Mattermost core team for server releases.

This will help smooth out the release process, and help everyone monitor how the release is progressing (each checklist will be posted in the Desktop App channel on their respective dates)

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -1,0 +1,134 @@
+# Release Process
+
+This document outlines the release process for the Mattermost Desktop App. It is drawn from the development process used by the [Mattermost core team](https://docs.mattermost.com/process/release-process.html).
+
+## Release Timeline
+
+Notes: 
+- All cut-off dates are based on 15:00 GMT (UTC-00) on the day stated. 
+- T-minus counts are measured in "working days" (weekdays other than major holidays concurrent in US and Canada) prior to release day.
+
+### A. (Code complete date of previous release) Beginning of release
+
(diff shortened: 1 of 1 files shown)


yuya-oc:
It might not be written in this document though, how would the code completion date be announced?

jasonblais:
I was thinking that each one would be posted in Desktop App channel as we work through the release.

Once we've locked on a release date, anyone can then check from this release process when each step would occur, as we give the generic dates for each (e.g. T-minus 10).

That being said, another option is having it in the mattermost/docs repo.

yuya-oc:
I see. Just I wanted to know that because the doc doesn't say about the code completion date. Thanks!

jasonblais:
> I see. Just I wanted to know that because the doc doesn't say about the code completion date. Thanks!

Can you help me understand more? Want to make sure your concern is addressed

yuya-oc:
I felt that if the code completion date is previously announced as a part of release process, contributors can consider their schedule to submit codes.

jasonblais:
Ah interesting. Maybe we can open a GitHub issue earlier for the release, say 2 weeks before code complete?

The issue could include a shortlist of issues left for the release + planned code complete and release dates.

I actually did so for last couple of releases, and it worked pretty well.

yuya-oc:
Yeah, surely it seems enough. It was forgotten probably because it was not done in upcoming release.

Issue #378, opened 2016-12-04
Mac: Mattermost app opens in an "invisible" screen


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

**Summary**
Mac: Mattermost app opens in an "invisible" screen

**Steps to reproduce**

Mac OS X; Mattermost Server v3.4.0; Mattermost Desktop v3.4.1

1) Connect Mac to two screens
2) Have Mattermost open on Screen 2
3) Disconnect display from Screen 2
4) Click on Mattermost icon on the dock menu

**Expected behavior**

Mattermost opens in Screen 1 (which is still open)

**Observed behavior**

Mattermost opens in Screen 2 (which is invisible)

jasonblais:
An issue with the above description was reported -- wondering if others in the community have noticed this or are able to reproduce?

If you're able to reproduce, can you also help:
1) Launch "System Preferences" from "Launchpad"(application list).
2) Select "Displays".
3) Select "Show mirroring options in the menu bar when available"

and see if it also reproduces when the mirroring option is selected in Mac display settings.

![image](https://cloud.githubusercontent.com/assets/13119842/20868145/3443cdfa-ba22-11e6-9ff3-3bf73331a8ed.png)

jnugh:
I was able to reproduce this on Linux (Unity) and wrote a PR that fixes it for me :) https://github.com/mattermost/desktop/pull/386

yuya-oc:
I confirmed the test case of #386 with v3.5.0 on macOS again, but the original issue didn't reproduced (I unplugged the cable instead of disabling because "disable" doesn't exist).

As far as I observed the behavior, the original behavior is a little complicated on macOS.

- Launched from Dock or Launchpad: The window opens in the monitor where Dock exists.
- Launched from Finder: The window opens at the previous position (expected behavior of source code).
  - If the monitor doesn't exist, the window position is fitted to the current monitor.

jasonblais:
[...]

PR #386, opened 2016-12-06
Do not place windows off screen

jnugh:

This fixes an issue where a window could be placed off screen which happend when the display it was previously on has been disconnected before launching the application.

> The application will always be placed on an existing screen.

https://github.com/mattermost/desktop/issues/378

**Test Cases**
* Have 2 monitors
* Place the app on one of them
* Close the app
* Start the app -> It should be exactly were it was when you closed it
* Close the app
* Disable the monitor
* Start the app -> It should be on your main screen

**Additional Notes**
I do not check if the window fits onto the screen , so if you plug in another monitor with a smaller resolution the window might not fit the screen (which I actually didn't test, might as well be covered by electron).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -219,6 +219,35 @@ function clearAppCache() {
   }
 }
 
+function getValidWindowPosition(state) {
+  // Screen cannot be required before app is ready
+  const {screen} = require('electron');
+
+  // Check if the previous position is out of the viewable area
+  // (e.g. because the screen has been plugged off)
+  const displays = screen.getAllDisplays();
+  let minX = 0;
+  let maxX = 0;
+  let minY = 0;
+  let maxY = 0;
+  for (let i = 0; i < displays.length; i++) {
+    const display = displays[i];
+    maxX = Math.max(maxX, display.bounds.x + display.bounds.width);
+    maxY = Math.max(maxY, display.bounds.y + display.bounds.height);
+    minX = Math.min(minX, display.bounds.x);
+    minY = Math.min(minY, display.bounds.y);
+  }
+
+  if (state.x > maxX || state.y > maxY || state.x < minX || state.y < minY) {
+    Reflect.deleteProperty(state, 'x');
+    Reflect.deleteProperty(state, 'y');
+    Reflect.deleteProperty(state, 'width');
+    Reflect.deleteProperty(state, 'height');
+  }
+
+  return state;
+}
+
 // Quit when all windows are closed.
 app.on('window-all-closed', () => {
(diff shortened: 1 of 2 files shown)


jasonblais:
Thanks @jnugh!! 

I don't have two monitors to test..so wondering if someone in the community would like to help with testing?

[You can find the test builds here](https://circleci.com/gh/mattermost/desktop/809#artifacts), and test cases have been included above.

If you have two monitors with different sizes, please help test the corner case provided by @jnugh above.

yuya-oc:
I could not reproduce the issue on Ubuntu 16.04 hosted on VirtualBox. However this feature looks like working well.

jasonblais:
Let me see if I can find someone who can help test  this PR on Mac/Windows.

src/main.js
```diff
@@ -201,6 +201,31 @@ function clearAppCache() {
   }
 }
 
+function restoreWindowState(state) {
```
jnugh: validateWindowPosition sounds as if it would return a bool value, I will go for getValidWindowPosition

jasonblais:
@yuya-oc would you be able to test this on Windows, just to confirm it's working as expected?

I have a couple of people I can ask to try as well, but it would be good to try ourselves first (I would, but again only have one monitor).

yuya-oc:
As I had written in https://github.com/mattermost/desktop/issues/378#issuecomment-267035909, the original issue didn't reproduced on my MacBook. But this feature works well if I launched the app from CLI.

For Windows, I need to bring another monitor to test this, so I will check this in weekend.

yuya-oc:
Tested on Windows 10. In the case of that the secondary monitor is located at left of main monitor, `x` has a negative value. So the feature doesn't work.

secondary_is_located_at_left

src/main.js
```diff
@@ -201,6 +201,31 @@ function clearAppCache() {
...
+  const displays = screen.getAllDisplays();
+  let maxX = 0;
+  let maxY = 0;
+  for (let i = 0; i < displays.length; i++) {
+    const display = displays[i];
+    maxX = Math.max(maxX, display.bounds.x + display.bounds.width);
+    maxY = Math.max(maxY, display.bounds.y + display.bounds.height);
+  }
+
+  if (state.x > maxX || state.y > maxY) {
```
yuya-oc: `(state.x < minX || state.y < minY)` should also be handled.

jnugh:
You are absolutely right, could you please check the updated version.

yuya-oc:
Tested on Windows 10 by editing `<userData>/bound-info.json`. The feature works well.

Issue #401, opened 2016-12-18
Desktop UX Improvements: Managing servers


jasonblais:
This ticket is one of three improvement tickets planned for Desktop user experience.

- [x] Navigation bar -- PR #415
- [x] Adding a new server from the “+” icon -- PR #415
- [x] Opening an app with no servers in Settings page -- PR #435

----

### PART 1 - Navigation bar: 

The navigation bar will display on top of the app itself similar to browser tabs and will have the server display name as specified on the Settings page.

This bar will be hidden if there is just one server specified on the Settings page.

The UI screen below shows:
 - two servers, the first and active server is “Pre-release” and the second is “Bladekick”
 - a “+” button beside the second tab which can be used to add a new server

![image](https://cloud.githubusercontent.com/assets/13119842/21290354/c4869516-c484-11e6-9314-64a488708a40.png)

### PART 2 - Adding a new server from the “+” icon

If a user clicks on the “+” icon, they can enter the name and URL similar to the Settings page. The dialog should be the same from #400, PART 3

![image](https://cloud.githubusercontent.com/assets/13119842/21290364/52baa08e-c485-11e6-8770-bc96c631111d.png)

After clicking “Add”, the new server will appear on the right-most tab of the navigation bar, and the sign in page will be displayed to the user.

### PART 3 - Opening an app with no servers in Settings page

If a user has no servers on the Settings page, a dialog will automatically appear asking them to add a server. The dialog is the same from PART 2 in this ticket (and from #400, PART 3)

Once done, the server is added to the Server Management section.

Note: This also covers the case when a user starts the app for the first time.

---- 

Note that "team management" in the desktop app will change to "server management" once [team sidebar PR on the platform repo is merged](https://github.com/mattermost/platform/pull/4569
), which adds an additional layer on the left hand sidebar for team management

![image](https://cloud.githubusercontent.com/assets/13119842/21290644/97b60448-c490-11e6-92a9-3ea97857d059.png)

----

Future improvements:

Allow users to reorder servers directly on the Server Management page and on the navigation bar. https://mattermost.uservoice.com/forums/306457-general/suggestions/15771343-allow-reordering-teams-in-desktop-application

jasonblais:
Closed via #415 and #435, thanks all!

Issue #400, opened 2016-12-18
Desktop UX Improvements: Settings Page


jasonblais:
This ticket is one of three improvement tickets planned for Desktop user experience.

- [x] 1. Basics -- PR #423
- [x] 2. Clicking a server -- PR #425 
- [x] 3. Clicking on “Add new server” opens a dialog -- PR #415
- [x] 4. Clicking on “Edit” opens the same dialog from part 3 -- PR #415
- [x] 5. Clicking on “Remove” opens a dialog to confirm that the user wants to remove the server -- PR #421

----

### PART 1 - Basics

 Core aspects of the current settings page should remain unchanged, apart from minor string changes:
 - Title: Settings
 - **Section 1**: Server Management
   - Displays the Name and URL
   - "Add new server" option
   - "Edit" and "Remove" options
 - **Section 2**: App Options
   - “Cancel” and “Save” buttons hovering over the page
   - “X” on top right corner, which closes the settings page and directs the user to the first server tab, if applicable. If the user has not added a server, the settings page should remain open

![image](https://cloud.githubusercontent.com/assets/13119842/21290256/63be111c-c481-11e6-8f2a-3dae75719216.png)

The following includes improvements to the Settings page:

### PART 2 - Clicking a server

When you click one of the servers under “Server Management”, the Settings page closes and the tab corresponding to that server opens.

### PART 3 - Clicking on “Add new server” opens a dialog

**DIALOG**:
- Title: Add Server
- Input Field 1:
  - Title: Server Display Name
  - Placeholder Text: Server Name
  - Help Text: The name of the server displayed on your desktop app tab bar. 
- Input Field 2:
  - Title: Server URL
  - Placeholder Text: https://example.com
  - Help Text: The URL of your Mattermost server. Must start with http:// or https://.
- Dialog options: 
  - Cancel → Closes the dialog without saving changes
  - Save → Adds server to the end of the Server Management list

**ERROR MESSAGES**:
- If no name is provided, show text “Name is required.”
- If no URL is provided, show text “URL is required.”
- If URL is provided but is missing http:// or https://, show text “URL should start with http:// or https://.”

![image](https://cloud.githubusercontent.com/assets/13119842/21290323/e9753e3c-c483-11e6-8786-c974cc5faa95.png)

**LOCATION:**

The dialog should ideally be located somewhere in the top middle part of the currently active window. Currently, the dialog would always open in the top left corner of the screen, which might be very different from the actual location of the window and therefore not ideal.

[...]

PR #415, opened 2017-01-16
Ux manage servers

jnugh:

**Summary**
Adds a button to the Tab Header which allows the user to add a new team without opening the settings page. Use the new Modal in the settings view as well.

**Issue link**
#400, #401 

**Test Cases**
Add a new server from the tab header and the settings page.

I added multiple specs that test if newly added teams are being persisted and if the modal behaves as expected in general.

I'm not sure if the way I do error handling is a good idea, I just saw that #400 defines an area where the error message should be displayed - I think that it's best to have the message next to the input that is incorrect but I could change it (might be a good idea to discuss).

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -10,6 +10,8 @@ const MattermostView = require('./MattermostView.jsx');
 const TabBar = require('./TabBar.jsx');
 const HoveringURL = require('./HoveringURL.jsx');
 
+const NewTeamModal = require('./NewTeamModal.jsx');
+
 // Todo: Need to consider better way to apply styles
 const styles = {
   hoveringURL: {
@@ -237,6 +239,11 @@ const MainPage = React.createClass({
       this.setState({targetURL});
     }
   },
+  addTeam() {
+    this.setState({
+      showNewTeamModal: true
+    });
+  },
   render() {
     var self = this;
 
@@ -253,6 +260,7 @@ const MainPage = React.createClass({
             mentionAtActiveCounts={this.state.mentionAtActiveCounts}
             activeKey={this.state.key}
             onSelect={this.handleSelect}
+            onAddTeam={this.addTeam}
           />
         </Row>
       );
@@ -296,6 +304,25 @@ const MainPage = React.createClass({
       authServerURL = `${tmpURL.protocol}//${tmpURL.host}`;
       authInfo = this.state.loginQueue[0].authInfo;
     }
+    var modal;
+    if (this.state.showNewTeamModal) {
+      modal = (
+        <NewTeamModal
+          onClose={() => {
+            this.setState({
+              showNewTeamModal: false
+            });
+          }}
+          onSave={(newTeam) => {
+            this.setState({
+              showNewTeamModal: false
+            });
+            this.props.teams.push(newTeam);
+            this.render();
+          }}
+        />
+      );
+    }
     return (
       <div>
         <LoginModal
@@ -323,6 +350,9 @@ const MainPage = React.createClass({
               targetURL={this.state.targetURL}
             /> }
         </ReactCSSTransitionGroup>
+        <div>
+          { modal }
+        </div>
       </div>
     );
   }
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -130,6 +130,10 @@ const MainPage = React.createClass({
         mattermost.goForward();
       }
     });
+
+    ipcRenderer.on('add-server', () => {
+      this.addServer();
+    });
   },
   componentDidUpdate(prevProps, prevState) {
     if (prevState.key !== this.state.key) { // i.e. When tab has been changed
@@ -240,7 +244,7 @@ const MainPage = React.createClass({
       this.setState({targetURL});
     }
   },
-  addTeam() {
+  addServer() {
     this.setState({
       showNewTeamModal: true
     });
@@ -261,7 +265,7 @@ const MainPage = React.createClass({
             mentionAtActiveCounts={this.state.mentionAtActiveCounts}
             activeKey={this.state.key}
             onSelect={this.handleSelect}
-            onAddTeam={this.addTeam}
+            onAddServer={this.addServer}
           />
         </Row>
       );
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -315,10 +315,11 @@ const MainPage = React.createClass({
             });
           }}
           onSave={(newTeam) => {
-            this.setState({
-              showNewTeamModal: false
-            });
             this.props.teams.push(newTeam);
+            this.setState({
+              showNewTeamModal: false,
+              key: this.props.teams.length - 1
+            });
             this.render();
             this.props.onTeamConfigChange(this.props.teams);
           }}
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -38,7 +38,8 @@ const MainPage = React.createClass({
   propTypes: {
     disablewebsecurity: React.PropTypes.bool.isRequired,
     onUnreadCountChange: React.PropTypes.func.isRequired,
-    teams: React.PropTypes.array.isRequired
+    teams: React.PropTypes.array.isRequired,
+    onTeamConfigChange: React.PropTypes.func.isRequired
   },
 
   getInitialState() {
@@ -319,6 +320,7 @@ const MainPage = React.createClass({
             });
             this.props.teams.push(newTeam);
             this.render();
+            this.props.onTeamConfigChange(this.props.teams);
           }}
         />
       );
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -309,27 +309,25 @@ const MainPage = React.createClass({
       authServerURL = `${tmpURL.protocol}//${tmpURL.host}`;
       authInfo = this.state.loginQueue[0].authInfo;
     }
-    var modal;
-    if (this.state.showNewTeamModal) {
-      modal = (
-        <NewTeamModal
-          onClose={() => {
-            this.setState({
-              showNewTeamModal: false
-            });
-          }}
-          onSave={(newTeam) => {
-            this.props.teams.push(newTeam);
-            this.setState({
-              showNewTeamModal: false,
-              key: this.props.teams.length - 1
-            });
-            this.render();
-            this.props.onTeamConfigChange(this.props.teams);
-          }}
-        />
-      );
-    }
+    var modal = (
+      <NewTeamModal
+        show={this.state.showNewTeamModal}
+        onClose={() => {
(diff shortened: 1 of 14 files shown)


yuya-oc:
Somehow CircleCI didn't build the PR (probably due to preferences). So I made the artifacts in my fork. https://circleci.com/gh/yuya-oc/desktop/132#artifacts

jnugh:
What's going on with the tests? It worked for me locally...

jnugh:
That was pretty obvious... :+1:

yuya-oc:
Testable artifacts are available. https://circleci.com/gh/mattermost/desktop/839#artifacts

jasonblais:
Yay, this is great @jnugh! Will test it out today! 🎉

jasonblais:
Thanks @jnugh! This is really well done! I've been testing it out yesterday and today.

There are a few notes / pieces of feedback to consider when you have the chance - even though they might seem like a lot, it's just mostly small UX notes. Functionally everything is working really well! Love it :) 

#### 1) "Add Server" dialog

image

Proposals (based on #400, part 3):

VISUAL: 

1a) Change title from `Add a new Team` to `Add Server` --> a new team sidebar was added to Mattermost Server 3.6, which allows you to display all your teams on the single interface. So you no longer need a separate tab for your teams, only for your servers
1b) I feel `Please specify a server name and a valid Mattermost URL` is redundant since we give help text below the input boxes - we could consider removing that sentence from the modal, what do you think?
1c) Update `https://example.org` to `https://example.com` (it'll be consistent with other similar examples in Mattermost)

FUNCTIONALITY:

Propose that when a user presses
1d) Enter, the "Add" button action is taken
1e) Esc, the "Cancel" button action is taken
1f) Clicking outside of the modal closes the dialog (carries out the "Cancel" button action) --> I'm not sure how hard this would be to implement, so definitely okay making a separate improvement ticket for it
1g) After hitting "Add" on the dialog (when pressing the "+" icon, not on the Settings page), and the new server appears on the right-most tab of the navigation bar, propose automatically activating that tab for the user.

ERROR HANDLING:

1h) Propose that errors aren't displayed until the user hits the "Add" button or hits Enter. This is consistent with other Mattermost dialogs, and is more forgiving to the user (we want to avoid telling the user they made a mistake, unless they actually made one).

When the user hits "Add" and if there's an error that prevents adding the server, then 
   - highlight the input field in red only (no highlights for the title or help text)
   - show the error message below the input field, but above the help text, similar to when creating a new channel (see below)

image

1i) Error messages (from #400)
 - If no name is provided, show text “Name is required."
 - If no URL is provided, show text “URL is required.”
 - If URL is provided but is missing http:// or https://, show text “URL should start with http:// or https://.”

#### 2) "Edit Server" dialog

It looks like clicking "Edit" on the Settings page opens the same "Add Server" dialog without pre-populated information.

Propose that 
  - fields are pre-populated with the previously specified server display name and server URL
  - title is "Edit Server" instead of "Add Server"
  - pressing "Cancel" undoes changes and closes the modal
  - pressing "Save" closes the modal and saves the changes

#### 3) "+" icon

3a) Propose removing the border around the "+" icon in server navigation tab bar -- it makes it look a bit prominent (I noticed that my eyes keep getting drawn to it). (Propose we keep the hover effect though)
3b) To follow the ["obvious" design principle](https://docs.mattermost.com/developer/fx-guidelines.html#obvious), propose we add a tooltip when hovering over the "+" icon, similar to Google Chrome (see below) with text `New server`. 

image

#### 4) Miscellaneous

This is just something that came to mind during testing and would probably be a separate improvement ticket:

If there is only one server tab, then the "+" icon isn't visible since the server navigation bar is hidden automatically. Hence, it's not obvious how to add a new server in that case.

I'm not a huge fan for showing the tab bar when there is only one server, and I think we should continue to hide it. Otherwise, it would just be wasted space on the app window for those who only have one server.

However, we could consider adding a new option in the menu bar such as **Sign in to Another Server** or **Add a New Server**, perhaps under **File** or **Window**, which opens the "Add Server" dialog when clicked. That way it'll at least be more discoverable.

Other ideas would be welcome.

src/browser/components/TeamList.jsx
```diff
@@ -84,12 +85,17 @@ const TeamList = React.createClass({
...
-          teamIndex={this.state.team.index}
-          teamName={this.state.team.name}
-          teamUrl={this.state.team.url}
+        <NewTeamModal
+          onClose={this.props.toggleAddTeamForm}
+          onSave={(newTeam) => {
+            this.setState({
+              showNewTeamModal: false
+            });
+            this.props.teams.push(newTeam);
```
yuya-oc: A new array should be created.

jnugh:
Alright, most of these things should be pretty easy to change.
Regarding error handling: There is still some stuff to do but I didn't want to focus on the actual behavior before discussing where to place the message (there is this mockup where it goes to the bottom left): 

So when I change the style for invalid inputs, that would would be a general change for all inputs in the desktop app (not inside the web view) - To be honest I'm not sure if we highlight invalid input somewhere else yet - if we do these places would change as well. But as this would be consistent with the Mattermost platform I feel like this should actually be the best way to go.

1a-g) :+1: 
1h-i) See above :+1: 
2) Nice catch, didn't even notice that I changed the behavior of the server edit function. We could still keep the old edit behavior - might make sense when we add drag/drop reordering...
3) I change the + button into a link - let's see.
4) Would be very easy to add :+1:

jasonblais:
Thanks for your response! All sounds good.

1h-i): Ah, you're right. Yes, let's add it at the bottom left instead. User's eyes are drawn towards the "Add" button and hence you notice the error message more easily if it's close to it.

That way you also don't have to move the help text down if there's an error message present.

2) Yeah, I think using the "Edit Server" dialog would be better for a consistent experience. Also, right now the 'edit server' input fields appear at the bottom of the server list, which isn't necessarily clear to users (it looks like you're adding a new server).

This would also cover part 4 of #400 ticket.

3): Let me know if that's tricky to change, happy to discuss it
4): Awesome, can you help add it in that case? We can add it under **File > Sign in to Another Server**, which just performs the same exact action as the "+" icon.

Propose adding it as the first item on the File drop-down list.

jasonblais:
By the way, @yuya-oc, if you have any feedback about my proposals, please let me know.

Same for you @jnugh, please don't hesitate to suggest other options

src/browser/components/MainPage.jsx
```diff
@@ -296,6 +305,26 @@ const MainPage = React.createClass({
...
+          onClose={() => {
+            this.setState({
+              showNewTeamModal: false
+            });
+          }}
+          onSave={(newTeam) => {
+            this.setState({
+              showNewTeamModal: false
+            });
+            this.props.teams.push(newTeam);
```
yuya-oc: prop is changed.

src/browser/components/TeamList.jsx
yuya-oc: prop is changed.

yuya-oc:
Almost was done in above discussion 👍 
So I added a review in code.

src/browser/components/MainPage.jsx
```diff
@@ -296,6 +305,26 @@ const MainPage = React.createClass({
       authServerURL = `${tmpURL.protocol}//${tmpURL.host}`;
       authInfo = this.state.loginQueue[0].authInfo;
     }
+    var modal;
+    if (this.state.showNewTeamModal) {
+      modal = (
+        <NewTeamModal
```
yuya-oc: Possibly you can use `show` parameter of ReactBootstrap.Modal. In my memory, it would make better animation. Please see `LoginModal`.

jasonblais:
Thanks for the continued progress @jnugh!

jasonblais:
Hey @jnugh! Just checking how things are going? :)

jnugh:
Here comes another bunch of changes :) From what I see, the only thing missing is to change the plus sign... I tried a usual link but that does not match the layout at all. So some more Styling is needed here.

jasonblais:
Awesome! :tada: 

Let me know if you need any help with CSS on the plus sign styling, I can ask our UI designer to help take a look too.

yuya-oc:
Please let me confirm. Is the part 4 of #400 intended in this PR? I could not close the modal which is opened by "Edit". And if I click the outside of modal or "Cancel", the server is unexpectedly duplicated.

Repro steps:
1. Click "Edit" in settings page. ("Add Server" modal opens. This should be "Edit Server".)
2. Click "Cancel" or outside of the modal.
3. Click "Add". (should be "Save")
4. The server is duplicated as a new server.

If the step 2 is skipped, the modal correctly works.

However the key functionality (especially modal) would be done in this PR. So I feel that I can take over the rest though, how do you feel?

jnugh:
That should be fixed now :)
I guess it can be merged if you have no further additions or changes. The Plus button can be changed afterwards...

jasonblais:
Thanks @jnugh! Let me test it out as well, thanks for all your help here!

yuya-oc:
The feature looks working 👍

jasonblais:
Thanks @jnugh! Looking great!

Just minor notes:

1) Team URL should actually be "Server URL". I missed this on the first review

image

2) Can we make the whitespace between
  a. top line and `Server Display Name`
  b. bottom line and `The URL of ...` help text

equal?

image

3) Animation when closing the dialog

We have a really nice animation for the "Add Server" dialog when it appears (yay!).

However, the same animation doesn't happen when the dialog closes. It's more abrupt.

Wondering if this is easy to update so the experience is consistent?

4) I noticed the error handling for `http://` isn't correct: it actually accepts a URL with just `http:`

I couldn't reproduce it with our 3.5 release, so seems it might be caused by this PR.

image

5) + icon: How likely do you think you'll have time to update the styling of the "+" icon? We're considering to cut the release candidate for 3.6 on Feb 7th.

I could ask our UI guru (asaad) to take a look. It might be quick for him and we' be able to apply it already on this PR.

jnugh:
4. : I use https://www.npmjs.com/package/valid-url, I might switch to something different - any other ideas?
5. : Yeah if he has time he can definitely do this a lot faster then I could. Would be awesome!

jasonblais:
4 - What did we use before? @yuya-oc any suggestions?
5 - I'll touch base with him 👍 

Thanks @jnugh for your continued help!

jasonblais:
Update: I tested 1) and 2), looks good! 

The commit for 3) seems to break the build, so I couldn't test it yet: https://github.com/mattermost/desktop/pull/415/commits/1343dce0edaaba110334f3a52f327130a91b94fc

Also asked Asaad to take a look at 5)

jnugh:
Tests were not waiting for animation - fixed that one.
4) I will use the regex which was used before, drop `valid-url`.

jasonblais:
@jnugh I tested 3) and 4), works great.

@asaadmahmood will take a look at the styling of the "+" sign once this PR has merged.

@yuya-oc please feel free to merge if this looks good to you

yuya-oc:
Looks good. Thanks @jnugh many times!

Issue #461, opened 2017-03-07
Use power-monitor API to re-load the app after computer wakes from sleep


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

**Summary**
Use [power-monitor](https://github.com/electron/electron/blob/master/docs/api/power-monitor.md) API to re-load the app after computer wakes from sleep.

This should help with missing messages reported by people when the wake their computer from sleep.

yuya-oc:
@jasonblais To make sure, would you check webapp's behavior in this case? I'm afraid there are missing messages in web browsers.

jasonblais:
This isn't an easy-to-solve solution on the webapp because we are limited by what the browser tells us and have no access to operating system events that would normally make this trivial. 

However, if the app reloads after the computer wakes from sleep, we can be more certain that messages shouldn't be missing on the desktop app.

As a note, the people who reproduced the issue could do it on v3.6 but not v3.4.1 (of the desktop app), and we're running on the same Mattermost server (v3.6.2 at the time).

yuya-oc:
Ah, thanks. Probably I confused with online/offline-events. But the computer might be offline when it wakes from sleep, so I will take care about this case.

> on v3.6 but not v3.4.1

Separately, it's a little weird :(

jasonblais:
Since Mattermost Server v3.8, there haven't been issues with missing messages after this PR on server side: https://github.com/mattermost/platform/pull/5907

Closing the issue

PR #495, opened 2017-03-27
Reload tabs when the computer resumes

yuya-oc:

**Summary**
Reload tabs when the computer resumes.

If the computer is offline when resuming from sleep, the app doesn't reload any tabs to keep their contents.

**Issue link**
#461 

**Test Cases**
Tested on macOS 10.12.3

Case 1:

1. Let the computer sleep.
2. Resume from sleep.
3. The app should reload all tabs immediately.

Case 2:

1. Turn off Wi-Fi or unplug LAN cable.
2. Let the computer sleep.
3. Resume from sleep.
4. The app should NOT reload any tabs immediately.
5. Connect the computer to a network.
6. The app should reload all tabs.

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/200#artifacts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Release date: TBD
 
 ### Improvements
 - Removed "Display secure content only" option it's no longer necessary.
+- Reload tabs to fetch latest messages when the computer resumes.
 
 #### Windows
 - Removed Japanese fonts support it's no longer necessary.
--- a/src/browser/components/ErrorView.jsx
+++ b/src/browser/components/ErrorView.jsx
@@ -31,10 +31,17 @@ const errorPage = {
 };
 
 function ErrorView(props) {
+  const classNames = ['errorView'];
+  if (!props.active) {
+    classNames.push('errorView-hidden');
+  }
+  if (props.withTab) {
+    classNames.push('errorView-with-tab');
+  }
   return (
     <Grid
       id={props.id}
-      style={props.style}
+      bsClass={classNames.join(' ')}
     >
       <div style={errorPage.tableStyle}>
         <div style={errorPage.cellStyle}>
@@ -87,7 +94,8 @@ function ErrorView(props) {
 ErrorView.propTypes = {
   errorInfo: React.PropTypes.object,
   id: React.PropTypes.number,
-  style: React.PropTypes.object
+  active: React.PropTypes.bool,
+  withTab: React.PropTypes.bool
 };
 
 module.exports = ErrorView;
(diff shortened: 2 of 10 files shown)


jasonblais:
@yuya-oc tested on Windows 10

After waking computer from sleep, each tab begins to reload

image

However, it seems the reloading process takes a really long time -- the page doesn't actually appear until I refresh the tab. Reproduced on the pre-release server and one of our test servers. I attached some console errors that might be relevant

image

yuya-oc:
Thanks for catching! I missed that because badges was correctly updated.

I'm not sure the reason why this problem happens, but it's fixed by applying [Electron's notes for webview](https://electron.atom.io/docs/api/webview-tag/#css-styling-notes) on my Mac. Originally I worked on this at another branch though, I added it to this PR.

https://circleci.com/gh/yuya-oc/desktop/207#artifacts

jasonblais:
I seem to be hitting the same issue - the tabs do start to reload, but it seems to take a long time

image

Interestingly, the tab loads properly for test sites like example.com

image

and test.com

image

yuya-oc:
Hmm...interesting. But I'm not sure. Perhaps it's too early to reload even when the computer is online in resuming.

And unfortunately I found another problem on my Windows 10 desktop. The reloading fails due to `ERR_NETWORK_CHANGED` error. Long (more than 5 seconds) delay was necessary to avoid it, so the delay might not comfortable.

Just curious, did https://github.com/mattermost/platform/pull/5120 make sense for the original problem?

jasonblais:
Yeah, https://github.com/mattermost/platform/pull/5120 is related -- however the PR was included in Mattermost 3.6, and some people reported missing messages on that server.

However, we just had another PR merged (https://github.com/mattermost/platform/pull/5907), which adds a sequence number to WebSocket connections and events -- that sequence is used to determine if any events were missed and run reconnect logic to get that missed data.

So I can check on Monday whether that PR would be sufficient to cover #461

jasonblais:
@yuya-oc chatted with Joram - although https://github.com/mattermost/platform/pull/5907 won't be the perfect solution for resolving missing messages, it might be enough for this purpose. So propose keeping this PR as `pending` until Mattermost v3.8 is released and asking the original issuer to report back whether missing messages continue

yuya-oc:
@jasonblais Okay. But [the patch for webview](https://github.com/mattermost/desktop/pull/495#issuecomment-290460373) seems necessary, so I will apply only it later.

jasonblais:
Thanks @yuya-oc, sounds good

jasonblais:
@yuya-oc Since Mattermost Server v3.8, there haven't been issues with missing messages. Hence, propose closing this PR as well as https://github.com/mattermost/desktop/issues/461

yuya-oc:
@jasonblais I agree. Thanks for feedback.

PR #496, opened 2017-03-27
Update current year for license file

jasonblais:
Propose using `2015-present` format so we don't have to worry about updating the year in the future :)


yuya-oc:
Sounds good! But I'm not sure whether this format is applicable for license file. Is there any project which uses the format?

jasonblais:
Yeah, for instance Facebook uses it: https://github.com/facebook/facebook-ios-sdk/blob/master/LICENSE

yuya-oc:
Got it! Thanks!

PR #509, opened 2017-04-09
Add display-metrics-changed handler to place window on screen

### LINKED ISSUE #471: Windows spawn off-screen after remote desktop
I confirm (by marking "x" in the [ ] below): 

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
---

**Summary**
Mattermost client spawns off-screen after using the remote desktop feature.

- Operating System: Windows 10 Enterprise
- Mattermost Desktop App version: 3.6.0
- Mattermost Server version: 3.6.0 (3.6.1)

**Steps to reproduce**

1-Open mattermost.
2-Move the window to a secondary screen.
3-Access the computer using the remote desktop feature as the same user.
4-Disconnect from the remote desktop session.
5-After logging in the computer directly, the mattermost window is still running but it spawns to an off-screen location.

**Expected behavior**
The mattermost window should be visible on the secondary screen like before the remote desktop session.

**Observed behavior**
The mattermost window spawns off-screen.

**Possible fixes**
As depicted in a previous issue marked as [solved](https://github.com/mattermost/desktop/issues/44), deleting
`%USERPROFILE%\AppData\Roaming\Mattermost\bounds-info.json`
after completely closing the client works.

jnugh:
Add display-metrics-changed handler to place window on screen when metrics change while the application is running.

**Summary**
I added a listener for `display-metrics-changed` to handle changes while the app is running.
However the event doe snot fire on my OS so I'm not able to test this (+ I can't reproduce the bug on Linux so there is actually no bug to fix in my case).
However this might fix #471 if it works correctly...

**Test Cases**
See #471, place window on a dedicated monitor, remove that monitor. Hope the event fires and the window is moved to an existing screen.

**Additional Notes**
Not really tested, event does not fire on my OS. I hope it does on Window...

### DIFF
--- a/src/main/mainWindow.js
+++ b/src/main/mainWindow.js
@@ -14,35 +14,6 @@ function saveWindowState(file, window) {
   }
 }
 
-function getValidWindowPosition(state) {
-  // Screen cannot be required before app is ready
-  const {screen} = require('electron'); // eslint-disable-line global-require
-
-  // Check if the previous position is out of the viewable area
-  // (e.g. because the screen has been plugged off)
-  const displays = screen.getAllDisplays();
-  let minX = 0;
-  let maxX = 0;
-  let minY = 0;
-  let maxY = 0;
-  for (let i = 0; i < displays.length; i++) {
-    const display = displays[i];
-    maxX = Math.max(maxX, display.bounds.x + display.bounds.width);
-    maxY = Math.max(maxY, display.bounds.y + display.bounds.height);
-    minX = Math.min(minX, display.bounds.x);
-    minY = Math.min(minY, display.bounds.y);
-  }
-
-  if (state.x > maxX || state.y > maxY || state.x < minX || state.y < minY) {
-    Reflect.deleteProperty(state, 'x');
-    Reflect.deleteProperty(state, 'y');
-    Reflect.deleteProperty(state, 'width');
-    Reflect.deleteProperty(state, 'height');
-  }
-
-  return state;
-}
-
 function createMainWindow(config, options) {
   const defaultWindowWidth = 1000;
   const defaultWindowHeight = 700;
@@ -53,11 +24,12 @@ function createMainWindow(config, options) {
   const boundsInfoPath = path.join(app.getPath('userData'), 'bounds-info.json');
   var windowOptions;
   try {
-    windowOptions = getValidWindowPosition(JSON.parse(fs.readFileSync(boundsInfoPath, 'utf-8')));
+    windowOptions = JSON.parse(fs.readFileSync(boundsInfoPath, 'utf-8'));
   } catch (e) {
     // Follow Electron's defaults, except for window dimensions which targets 1024x768 screen resolution.
     windowOptions = {width: defaultWindowWidth, height: defaultWindowHeight};
   }
+
   if (process.platform === 'linux') {
     windowOptions.icon = options.linuxAppIcon;
   }
--- a/src/main.js
+++ b/src/main.js
@@ -213,6 +213,55 @@ app.on('window-all-closed', () => {
   }
 });
 
+function getValidWindowPosition(state, screen) {
+  // Check if the previous position is out of the viewable area
+  // (e.g. because the screen has been plugged off)
+  const displays = screen.getAllDisplays();
+  let minX = 0;
+  let maxX = 0;
+  let minY = 0;
+  let maxY = 0;
+  for (let i = 0; i < displays.length; i++) {
+    const display = displays[i];
+    maxX = Math.max(maxX, display.bounds.x + display.bounds.width);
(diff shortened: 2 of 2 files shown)


jasonblais:
@jnugh Big thanks for this PR!

@yuya-oc can you help review/test this PR?

yuya-oc:
Sure. But it's hard today, so I will test tomorrow if possible.

jasonblais:
Yeah for sure, not urgent 👍

src/main/mainWindow.js
```diff
@@ -70,6 +84,7 @@ function createMainWindow(config, options) {
   });
 
   const mainWindow = new BrowserWindow(windowOptions);
+  handleScreenResize(screen, mainWindow);
```
yuya-oc: How about using [`browser-window-created`](https://electron.atom.io/docs/api/app/#event-browser-window-created) of app module? All windows can get this feature.

yuya-oc:
I tested the artifact of https://circleci.com/gh/mattermost/desktop/940#artifacts . It looks the problem does not exist even when the event is not fired.

Possibly this might be needed to be solved by upstream Electron in future.

src/main/mainWindow.js
```diff
@@ -43,7 +40,23 @@ function getValidWindowPosition(state) {
   return state;
 }
 
+function handleScreenResize(screen, browserWindow) {
+  screen.on('display-metrics-changed', () => {
```
jnugh: I thought display removed would lead to a display metrics change. However I'm not sure about that ;)
Adding more handlers does not hurt in this case I guess...

yuya-oc: Yeah, I had thought that, too. But actually the event was not fired on macOS.
Would you wait the result of Windows? (Sorry, I haven't tested yet)

yuya-oc: You're right. On Windows, `display-metrics-changed` is emitted after `display-removed`.

yuya-oc:
I have tested both of v3.6.0 and the artifact on Windows 10 by following [the walk-through](https://github.com/mattermost/desktop/issues/471#issuecomment-294154572). It seems the problem happens when the window is minimized.

In that case, the problem was not fixed by this PR (in other words, this PR's approach might not be appropriate). Possibly a minimized window might have a special characteristic.

jasonblais:
Thanks @yuya-oc, great work investigating this!

@jnugh would you have any thoughts on the most recent findings?

jnugh:
Running the check on window restore might b an option I guess...

Jason Blais <notifications@github.com> schrieb am Mi., 19. Apr. 2017, 02:21:

jnugh:
Updated the commit. Added a restore handler (and use this globally across all browser windows).

jasonblais:
Thanks Jonas! @yuya-oc can you help take a look?

jnugh:
As stated initially - this has not been verified to work correctly as I have no possibility to test on windows at the moment.
My test was to place the window off-screen (by closing the app and changing the restore coordinates) and check if it will be restored on-screen, which it did in my case. However that worked with prior versions of this PR as well and didn't fix the issue on windows.

So if this attempt does not work, someone else might be able to debug this further, by actually using windows for that ;)

yuya-oc:
Thanks @jnugh!

Here is the result.

1. Move the window to secondary monitor.
2. Minimize the window.
3. Start Remote Desktop session.
4. Restore the window.
5. The window should move to (0, 0), but it's restored at other coordinates.
6. Minimize and restore the window again. It's located at (0 ,0).

In my logging, the position was valid in `getValidWindowPosition` on `display-metrics-changed` and `restore` of step 3 and 4. But the position becomes invalid (too large x for me) on step 6. To solve this, I needed to remove `display-metrics-changed`. So only `restore` event was sufficient for me.

Does using only `restore` event make sense also on Linux?

jnugh:
I guess `display-metrics-changed` is called before the new screen size is actually calculated. A timeout might work also but when waiting for `restore` is enough I guess we should go with it.

So the Desktop I use does not even place windows off screen when removing a display but only when the window is reopened after removing the display it has been on previously.
So yeah as long as it works on windows this should be okay. As the position is still checked on window create.
I will update the commit.

yuya-oc:
Looks good for me on both of Windows and Mac. On Mac, originally the problem doesn't reproduce and I could not find any regression.

yuya-oc:
@jasonblais Is there anything to test in your side?

jasonblais:
I don't have multiple monitors to try..but the work here looks good!

Thanks @jnugh and @yuya-oc! 👍

yuya-oc:
Thanks all!

Issue #520, opened 2017-04-28
Client freezes in Windows 10 (when app is minimized)


mdementev:
I confirm (by marking "x" in the [ ] below): 

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

**Summary**
Client freezes in Windows 10 on getting messages

**Steps to reproduce**

1. Minimize app
2. Wait for mesasges
3. Try to open app

**Expected behavior**
App should be opened and interactible

**Observed behavior**
Main window is opened, but doesn't correspond to any actions. Will work again after few minutes.
Console output is clear.

**Possible fixes**
Problem has been solved for my team, by disabling Windows notifications for Mattermost. So it could be related.

**Notes**
Added comment in closed ticket: https://github.com/mattermost/desktop/issues/347

jasonblais:
@mdementev Really appreciate the report,

When you say "minimized", do you mean 
 - minimized to the taskbar, or 
 - closed (which minimizes to the system tray / notifications area)

jasonblais:
Related to https://github.com/mattermost/desktop/issues/494

mdementev:
@jasonblais to taskbar

yuya-oc:
Thanks @mdementev

Actually the code is injected for desktop notifications in the application. So it's most significant difference from normal web browsers.

jasonblais:
Closed via #563

Boababa:
I'm running 3.7.1 on Win10, bug still happens in both 32 and 64 bit client versions. Is the fix in 3.7.1 or slated for 3.8.0?

jasonblais:
Hm, that's unfortunate to hear @Boababa. It was indeed included in the v3.7.1 release. Are the reproduction steps the same as in the issue description, and turning off desktop notifications no longer causes the app to freeze?

@mdementev Are you still experiencing the issue on v3.7.1 as well?

Boababa:
[...]

Issue #494, opened 2017-03-27
Client constantly hangs in random moments


jpVm5jYYRE1VIKL:
It is not possible to predict when it will hang next time. But look like mattermost client have dead thread lock somewhere . Have same behavior at least on 3 Different computers (1 PC Windows 10 PRO and 2 PC Windows 10 home). Possible fix only over kill mattermost client processes in Task Mananger and restart. In most cases it happens in moment click trey icon during conversation.

jasonblais:
Thanks @jpVm5jYYRE1VIKL for the report, few questions that come to mind,

1. Which Mattermost server version are you using?
2. Which Mattermost desktop version are you using?
3. Does it reproduce after accessing a particular channel/team?
4. Do you see a CPU spike during this time?

jpVm5jYYRE1VIKL:
Sorry that forced to ask extra questions. 
1. Server was 3.6.2
2. Client was 3.5.0 
3. Nope completely random. Sometimes i just click on window and understand that it hanged. 
4. It is hard to detect because all issues happens on powerfull computers (core i7) i not feel cpu spike. 
Now i upgraded server to version 3.7.3 and client to version 3.6.0 to see that happens.

jasonblais:
Awesome, thanks! Let us know if the issue reproduces on the latest server/client version

jpVm5jYYRE1VIKL:
Small update to issue. Just now happened again. Issue is not real hanging. It is freeze for 1 - 2 minutes. During this time application GUI completely not respond. Cpu not have any spikes.  In task manager application not marked as not responding and not eat CPU. 
After 1 - 2 minutes application unfreeze itself and continue to work. Detected such behavior because started to search debugger to attach and suddenly found that application not hang anymore.  Issue happened with 3.7.3 server and 3.6.0 client.

jasonblais:
Thanks @jpVm5jYYRE1VIKL - how did you set up your Mattermost environment? 

Are you using any proxies?

jpVm5jYYRE1VIKL:
Yes using. https://nginx -> http://mattermost:8065

jasonblais:
@yuya-oc is there any debugging available for the desktop app?

Someone else has the client freezing in Windows 10: https://github.com/mattermost/desktop/issues/347#issuecomment-291153639

yuya-oc:
> It is freeze for 1 - 2 minutes.

[...]

PR #563, opened 2017-07-21
Use setTimeout instead of setInterval to get unread count

yuya-oc:

**Summary**
Change the way to take intervals when counting unread channels/mention.

Possible fix for related issues. Unread counter is updated by the function invoked by `setInterval()`. But this function doesn't take care whether the invoked function finished. When the app is minimized, internal Chrome reduces performance by itself. So I think possibly the function might not be able to finish by the next interval.

Eventually we should also improve the function performance though, I use `setTimeout()` instead in order to take proper intervals first.

@jasonblais I haven't met the freezing issue. Would you try the artifact for a few days?

**Issue link**
#494 #520

**Test Cases**
1. Open desired servers.
2. Get messages in channels/DM.
3. Badge should be updated.

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/295#artifacts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Release date: TBD
 
 ### Improvements
 
+#### All Platforms
+ - Possible improvement when the app is in background state.
+ [#563](https://github.com/mattermost/desktop/issues/563)
+
 #### Windows
  - [Windows 7/8] Added support to open the message when clicking desktop notification.
  [#67](https://github.com/mattermost/desktop/issues/67)
--- a/src/browser/webview/mattermost.js
+++ b/src/browser/webview/mattermost.js
@@ -5,6 +5,8 @@ const ipc = electron.ipcRenderer;
 const webFrame = electron.webFrame;
 const EnhancedNotification = require('../js/notification');
 
+const UNREAD_COUNT_INTERVAL = 1000;
+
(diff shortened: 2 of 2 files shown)


jasonblais:
My freezing issue was something else..not related to the desktop app.

I'll share the build with one of our customers though, and ask if they'd like to try it.

It would be fantastic if this fixes the issue, 🤞

jasonblais:
@yuya-oc Are there any potential side effects we should try? I can test for those.

yuya-oc:
@jasonblais We should test whether the interval loop never stop. In other words, the app should properly update badges as before. So "try the app for a few days" is a good test for us.

jasonblais:
Got it 👍 

I've downloaded the artifact and will use it for the next few days

jasonblais:
By the way, I've been using the app and haven't noticed any issues.

We did send the build to one of our customers who is going to try it out. So we can wait to hear from them, and then merge if it indeed fixes the freezing issue.

jasonblais:
We haven't heard back yet, but this should be fine to merge regardless.

Let's not do 3.7.1 release until we've heard confirmation from at least one other user. If we don't, we can include this in the 3.8.0 release.

yuya-oc:
Thanks!

Issue #483, opened 2017-03-22
Allow admins to pre-configure desktop apps


jasonblais:
Allow admins to pre-configure desktop apps so end users don't have to input server URLs and app settings.

For auto-start option, there are 2 ways to sync this kind of setting:
 - Query the OS on app start and set config.json to whatever is set in the OS -> Changing config.json would have no effect
 - Set whatever is found in config.json and write it to OS (create or remove autostart file). -> Removing the autostart file using the settings provided by windows would have no effect

Neither is ideal so we'll want to carefully evaluate the option. See here for more details: https://pre-release.mattermost.com/core/pl/136f8r3w8f835m3f4mpho7bhua

yuya-oc:
In current implementation, [we use registry or hidden files](https://github.com/Teamwork/node-auto-launch#how-does-it-work). Probably users don't modify them, so I feel the second option looks good.

dmeza:
@yuya-oc @jasonblais the actual fix for this is `BASH-14 Set default server/team in settings`: https://github.com/mattermost/desktop/pull/594 that is built on top of `BASH-20`

jasonblais:
@yuya-oc With https://github.com/mattermost/desktop/pull/594, https://github.com/mattermost/desktop/pull/586 and https://github.com/mattermost/desktop/issues/618, is there additional work required for this issue?

yuya-oc:
@jasonblais One addition, the core concept of this issue (admins can set pre-config without building app from source) would be needed.

yuya-oc:
The possible location to put the pre-configuration file:

- Windows: `C:\ProgramData\Mattermost\`
- Mac: `/Library/Application Support/Mattermost/`
- Linux: `/etc/mattermost/`

Are there any better choices?

yuya-oc:
Probably we haven't decided how to prioritize pre-configuration against to user preferences.

I think there are two choices for that:
  1. defaultPreferences are overridden. So users can change overridden parameters later.
  2. config.json is overridden. So users can't change overridden parameters.

@jasonblais I feel 2 is suitable for admins to control their clients. Any thoughts?

jasonblais:
@yuya-oc 1. is more consistent with the server.

For instance, if the [admin sets a default theme](https://docs.mattermost.com/administration/config-settings.html#default-theme), the user can change it unless they modify some of the other settings such as [theme selection](https://docs.mattermost.com/administration/config-settings.html#theme-settings-experimental).

[...]

PR #586, opened 2017-09-08
(#1) BASH-20 Add base and override config

csduarte:
**Description**
This PR moves configuration values to a base.json file (src/common/config/base.json) and allow to override those config values on file override.json (src/common/config/override.json). That way the override values can be stored on a separate repository when building the clientes.

**Test Cases**
Add override values like this:
```json 
// Example override.js
{
  "1": {
    "showTrayIcon": true,
    "trayIconTheme": "light"
  }
}
```


yuya-oc:
Thanks for a lot of PRs!

This mechanism doesn't require Electron, so would you write tests in `test/specs/settings_test.js`? At least, upgrading from v0 to v1 is breaking. You can use `describe.only()` or `it.only()` to execute only your tests.

I thinks this PR forces only default settings. So users can change all settings. Is this correct?

yuya-oc:
Just curious, why does `override.json` require version number? I feel it's enough there is one config in the file.

src/common/settings.js
```diff
@@ -1,39 +1,38 @@
...
 
 function merge(base, target) {
   return Object.assign({}, base, target);
 }
 
+function deepMergeArray(dest) {
+  return dest;
+}
+
 function loadDefault(version, spellCheckerLocale) {
```
yuya-oc: Please test how `loadDefault()` works. Possibly `require('override.json')` doesn't match for writing tests because its static file.

yuya-oc:
@jasonblais This would be one of solution for #483, but admins have to build apps from source on their own. Of course code-signing would be gone when they don't. Is this beneficial?

yuya-oc:
Please let me confirm. What is the purpose of this feature? At first, I thought that this PR let users use pre-configured preferences.
However users can change their preferences. And in other PRs, this feature is also used to change "non user-preferences" behavior like `./configure.sh` of C/C++ projects. I feel this will be confusing later, so this should be separated to two modules.

csduarte:
@yuya-oc We were trying to make the smallest possible changes, because we know that getting bigger changes back is quite hard. So we used the only config there was to allow what we think of 'Feature Flags'. 

Being able to flip something on or off, sometimes dynamically. -- If you could imagine, in the future, we could be developing a brand new feature in the master branch, but have it disabled because of a flag in this config, and that could be flipped. So it certainly crosses the bounds between system config and user config, but eventually we decided a division doesn't make a lot of sense. Just some configs are visible to the user and some are not, and maybe an admin could have access to all of them to flip on and off.

src/package.json
```diff
@@ -12,6 +12,7 @@
     "auto-launch": "^5.0.1",
     "bootstrap": "^3.3.7",
     "create-react-class": "^15.5.3",
+    "deepmerge": "^1.5.1",
```
jasonblais: Is assume this is for https://github.com/KyleAMathews/deepmerge?

jasonblais:
@csduarte In terms of the use case, when developing a new feature in the master branch, we'd have it turned on by default, but have that setting disabled via config/override.json until the feature is finished? Is that correct?

csduarte:
@jasonblais The exact use case is quote flexible. In products that I know better, there is actually another service that will give the config option to the user on boot and show/hide certain features. Basically, An easy way to to A/B testing and test features, all while maintaining continuous integration and deployment.

src/package.json
csduarte: correct, also seen here https://www.npmjs.com/package/deepmerge

src/common/settings.js
csduarte: @yuya-oc do you want an automated test?

We followed the exact same style as the mobile app in terms of config and override. Possibly would like to match that to start with and then diverge?

yuya-oc: If possible. At least, I want that existing tests should be passed.
Unfortunately I'm not sure about mobile app source code. Would you tell me the part?

jasonblais:
Thanks! The context is very helpful.

Would you want to test on a live deployment however? We're open to taking the PR, just wondering if it's the optimal approach for the use case.

jasonblais:
Hey @dmeza, @csduarte 

Not urgent, but wondering for your thoughts on the following:

1 - Having both `default` and `1` lists in the `/config/base.json` file might be confusing to admins/developers building the apps from the source. It's not immediately obvious why there are two sets of default settings present and what their purpose is.

Would it make sense to split this file into `base.json` (containing `"default"`) and `pastDefault.json` (containing `"1"`, and other versions in the future)?

2 - Wondering your thoughts about keeping the core PR the same, but splitting the `base.json` file into `defaultPreferences.json` (user config) and `buildConfig.json` (system config)?

^Open for feedback, as we want to keep further changes from you minimal.

dmeza:
Hey @jason, @yuya-oc,
Sorry for the delay to reply. We've been focused on mobile and platform.
Our proposed changes/solution is based on the code that's already there, we're open to getting to the same result with different changes. I have a question as if `default` and `1` are really needed or there could be a different solution. 
Our goal is having configuration values that can enable/disable features and for white-labeling, so both of the proposed changes work.

dmeza:
@yuya-oc: settings tests are passing now. Fixed the `upgrade` test and added a test for `loadDefault`.

test/modules/environment.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+process.env.TEST = 'test';
```
yuya-oc: Was this necessary for you?

test/specs/settings_test.js
```diff
@@ -10,4 +11,14 @@ describe('common/settings.js', () => {
     config.teams[0].url.should.equal(v0Config.url);
     config.version.should.equal(settings.version);
   });
+
+  it('should loadDefault config for version 1', () => {
+    const baseConfig = require('../../src/common/config/base.json');
+    const overrideConfig = require('../../src/common/config/override.json');
+    const expectedDefaults = deepmerge(
+      baseConfig[1], overrideConfig[1] || {}, {clone: true, arrayMerge: settings.deepMergeArray}
```
yuya-oc: This part is actually same to `loadDefault()` implementation. So this test always passes. Is this intended?

package.json
```diff
@@ -60,5 +61,6 @@
     "webpack": "^2.5.1",
     "webpack-dev-server": "^2.4.5",
     "webpack-merge": "^4.1.0"
-  }
+  },
+  "dependencies": {}
```
yuya-oc: This seems accidentally changed.

test/modules/environment.js
dmeza: @yuya-oc this the packages context for build and test seems to be different and needed a different `deepmerge` require:
`
let deepmerge = require('deepmerge').default;
if (process.env.TEST) {
  deepmerge = require('deepmerge'); // eslint-disable-line
}
`
https://github.com/mattermost/desktop/pull/586/files#diff-7cdc7778ed182028c72b7c53ac2af20bR4

test/specs/settings_test.js
dmeza: @yuya-oc I think that it will provide the context to be able to test specific base config and override changes. What other test do you think we should add here?

dmeza:
Hey @jason, @yuya-oc, one way to simplify is to have a single version on the `override.json` file.
Not this:
`
// Example override.js
{
  "1": {
    "showTrayIcon": true,
    "trayIconTheme": "light"
  }
}
`
But this:
`
// Example override.js
{
  "showTrayIcon": true,
  "trayIconTheme": "light"
}
`

test/modules/environment.js
yuya-oc: @dmeza Got it. Assuming that, this sentence is not needed. When `src/common/settings.js` requires deepmerge, it correctly uses `src/node_modules/deepmerge`. Probably you met the error of that `test/specs/settings_test.js` requires deepmerge at outside of `src/`.

test/specs/settings_test.js
yuya-oc: @dmeza I had imagined the test simply comparing to "expected" default config. So in this case, something like `expectedDefaults = { teams: [], showTrayIcon: false, ... }` instead of calling deepmerge() because a plain object is obvious as an expected value.
However, surely this might be enough for now. I can accept.

src/common/settings.js
```diff
@@ -1,35 +1,37 @@
 'use strict';
 
 const fs = require('fs');
+let deepmerge = require('deepmerge').default;
```
yuya-oc: As mentioned at https://github.com/mattermost/desktop/pull/586#discussion_r142157918, `const deepmerge = require('deepmerge')` is enough.

dmeza: @yuya-oc this form `const deepmerge = require('deepmerge')` failed when building. Did you try it and worked for you for both cases building and running tests?

yuya-oc: @dmeza Thanks for feedback. Somehow deepmerge becomes `object` from `function` in webpack. So error occurs at run time.
I'm not sure whether it's bug or spec of webpack though, let's keep this for now.

dmeza:
@yuya-oc from your comments I gather that this PR is ready, right?

yuya-oc:
@dmeza Ah, yeah. I had not been sure whether you are preparing more updates about simplified override.json or splitting to two config files.
If there are no more updates here, I think it’s ready to merge.

cc: @jasonblais

jasonblais:
Thanks all! Let's go ahead and merge the PR.

yuya-oc:
@csduarte @dmeza Huge thanks for many iteration!

PR #587, opened 2017-09-08
BASH-1 Config value to display or not the URL on hover at bottom of screen

csduarte:
**Description**
This PR allows configurable url preview (the small gray bar in bottom corner) on mouse hovers over a link.

- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
 - [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
 - [x] executed `npm run lint:js` for proper code formatting

**Test Cases**
Load application and hover over any link tag (<a>). See gray bar with url in lower corner.
Disable Config and hover any tag, no gray bar rendered
```json 
//override.js
{
  “1”: {
    "displayUrlOnHover": false
  }
}
```
**Notes**
Includes bash-20

screen shot 2017-09-08 at 10 54 49 am

PR #594, opened 2017-09-08
BASH-14 Set default server/team in settings

csduarte:
**Description**
This PR allows for a defaultTeam to be defined in the override.json file (src/common/config/override.json). When this value is defined and the config file does not contain any teams in the team’s array, the team defined in the defaultTeam property will be added to the team's array in the config.

**Test Cases**
Add override values like this:
```json 
// Exampleoverride.js
{
  "1": {
    "defaultTeam": {
       “name”: “Test Team”,
       “url”: “https://test.team.com” 
    }
  }
}
```
Remove the application support folder by running rm -rf /Users/{yourusername}/Library/Application Support/Mattermost

When the application starts the login for the defaultTeam will be shown.

**notes**
includes bash-20

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -71,6 +71,12 @@ try {
   const spellCheckerLocale = SpellChecker.getSpellCheckerLocale(app.getLocale());
   config = settings.loadDefault(null, spellCheckerLocale);
   console.log('Failed to read or upgrade config.json', e);
+  if (!config.teams.length && config.defaultTeam) {
+    config.teams.push(config.defaultTeam);
+
+    const configFile = app.getPath('userData') + '/config.json';
+    settings.writeFileSync(configFile, config);
+  }
 }
 
 ipcMain.on('update-config', () => {
--- a/src/common/settings.js
+++ b/src/common/settings.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const fs = require('fs');
+
+const path = require('path');
 let deepmerge = require('deepmerge').default;
 if (process.env.TEST) {
   deepmerge = require('deepmerge'); // eslint-disable-line
@@ -14,7 +16,7 @@ function merge(base, target) {
   return Object.assign({}, base, target);
 }
 
-function deepMergeArray(dest) {
+function deepMergeArray(source, dest) {
   return dest;
 }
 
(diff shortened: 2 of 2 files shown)


jasonblais:
@yuya-oc Same as in #587, can you help share a link for the test build from circle-ci? I'll test this out.

yuya-oc:
https://circleci.com/gh/mattermost/desktop/1023#artifacts However at this time, you would not see any changes if the app is correctly implemented.

jasonblais:
Hm, that's correct. The settings would have to be specified before building the app. You enable/disable the display of URL hover before building and dist.

@csduarte @dmeza Wondering your thoughts if this should be a setting controlled by the Mattermost server? I would see a benefit of having a default team at server-level.

dmeza:
@jasonblais @csduarte the idea of this setting is to already have a default server and not even have to enter it like in this ticket `BASH-18 Add configs to show/hide server management and multiteam settings`: https://github.com/mattermost/desktop/pull/600
I think you need the desktop specific changes because it sends you to add a server if one is not configured. The process to check anything server side would mean that you're already connected to a specific server.

jasonblais:
@dmeza aah, got it. I was thinking of teams as "Mattermost teams" not "servers".

Can you pre-configure more than one server?

dmeza:
@jasonblais @csduarte: Yes, if you add more teams directly via the `config.json` file you would see multiple tabs. If no teams are configured and if `defaultTeam` is properly configured then you would connect to that specific server directly by default.

jasonblais:
@yuya-oc Once rebased, would you be able to help with a test build where `defaultTeam` is set?

dmeza:
@yuya-oc rebased and tested.

yuya-oc:
@jasonblais Added https://example.com as the defaultTeam. https://circleci.com/gh/yuya-oc/desktop/420#artifacts

As described in the first comment, please remove `AppData\Roadming\Mattermost\config.json` before starting the app.

src/main.js
```diff
@@ -71,6 +71,12 @@ try {
   const spellCheckerLocale = SpellChecker.getSpellCheckerLocale(app.getLocale());
   config = settings.loadDefault(null, spellCheckerLocale);
   console.log('Failed to read or upgrade config.json', e);
+  if (!config.teams.length && config.defaultTeam) {
```
jasonblais: @yuya-oc Should we consider renaming the term "team" to a "server" to reduce confusion? I searched through the GitHub repo and realized we refer to servers as teams in many places. This is probably because of how multi-team support worked before the team sidebar introduced a few months ago.

I could do a search and replace of the terms across our files after this PR and https://github.com/mattermost/desktop/pull/600 are merged.

yuya-oc: @jasonblais Yeah, we should do that. However, "team" is already used in `config.json`. So to avoid confusing, I don't think we need to do in v3.8.

yuya-oc:
@jasonblais Added a issue to rename variables. #623

PR #595, opened 2017-09-08
BASH-16 Add to eslint config

csduarte:
Add more rules to eslint config and packaging rules.


.eslintrc.json
```diff
@@ -1,17 +1,169 @@
 {
-    "extends": "./.eslintrc-platform.json",
```
yuya-oc: The purpose is to use same base config with mattermost-webapp. You can write only additional rules here. What was actually needed for you?

package.json
```diff
@@ -31,7 +31,8 @@
     "package:mac": "build --mac --publish=never",
     "package:linux": "build --linux --x64 --ia32 --em.name=mattermost-desktop --publish=never",
     "manipulate-windows-zip": "node scripts/manipulate_windows_zip.js",
-    "lint:js": "eslint --ext .js --ext .jsx ."
+    "lint:js": "eslint --ext .js --ext .jsx .",
+    "eslint": "eslint --ext .js,.jsx src/browser src/common src/main"
```
yuya-oc: Why is this script added?

csduarte:
@dmeza I couldn't say for sure, I'm not the one who added it. But if there is a platform specific eslint, will the editor pick up on it?

jasonblais:
> But if there is a platform specific eslint, will the editor pick up on it?

@yuya-oc thoughts on this question?

yuya-oc:
Honestly I'm not sure what I should say now though, `"extends"` is a feature of eslint. https://eslint.org/docs/user-guide/configuring#extending-configuration-files So any editors should use both of `.eslintrc-platform.json`(base config) and `.eslintrc.json`(override).

In `.eslintrc.json`, we can write only desktop app specific rules. However this PR would make difficult to know the rules.

csduarte:
Okay closing, will reopen if we discover any editor issues (which is why I think we created in the first place)

PR #597, opened 2017-09-08
BASH-23 Add npm clean script

csduarte:
**Description**
Adds a simple rm command to npm to remove all dynamically generated files to make it easy to reset between builds/branches

**Test Cases**
Run `npm run clean`

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "build:main": "cross-env NODE_ENV=production webpack --bail --config webpack.config.main.js",
     "build:renderer": "cross-env NODE_ENV=production webpack --bail --config webpack.config.renderer.js",
     "start": "electron src --disable-dev-mode",
+    "clean": "rm -rf release/ node_modules/ src/node_modules/ && find src -name '*_bundle.js' | xargs rm",
     "watch": "run-p watch:*",
     "watch:main": "node scripts/watch_main_and_preload.js",
     "watch:renderer": "webpack-dev-server --config webpack.config.renderer.js",


package.json
```diff
@@ -20,6 +20,7 @@
     "build:main": "cross-env NODE_ENV=production webpack --bail --config webpack.config.main.js",
     "build:renderer": "cross-env NODE_ENV=production webpack --bail --config webpack.config.renderer.js",
     "start": "electron src --disable-dev-mode",
+    "clean":"rm -rf dist/ release/ node_modules/ src/node_modules/ build/ browser/**/*_bundle.js}",
```
yuya-oc: You should remove `src/**/*_bundle.js` instead of `browser/**/*_bundle.js` to remove `src/main_bundle.js`.

yuya-oc: - missing a white space after `clean:`.
- there is unexpected trailing token `}`.

dmeza:
@yuya-oc rebased to latest. Fixed all your comments built, packaged and run linting. Let me know if that's the correct change.

yuya-oc:
Thanks @dmeza! But sorry, I missed one thing. `src/**/*_bundle.js` might not enough at least on git-bash. Would you confirm whether `src/browser/webview/mattermost_bundle.js` is deleted?

dmeza:
@yuya-oc yeah, it did not work so I added these two: `src/*_bundle.js` and `src/**/*_bundle.js` and it cleans both directories `src` and `src/browser`.

package.json
```diff
@@ -20,6 +20,7 @@
     "build:main": "cross-env NODE_ENV=production webpack --bail --config webpack.config.main.js",
     "build:renderer": "cross-env NODE_ENV=production webpack --bail --config webpack.config.renderer.js",
     "start": "electron src --disable-dev-mode",
+    "clean": "rm -rf dist/ release/ node_modules/ src/node_modules/ build/ src/*_bundle.js src/**/*_bundle.js",
```
yuya-oc: Just curious, `dist` and `build` are not used in this project. Is it necessary removing them explicitly?

yuya-oc: Is `rm -rf dist/ release/ node_modules/ src/node_modules/ build/ && find src -name '*_bundle.js' | xargs rm` appropriate for you? This can remove `src/browser/webview/mattermost_bundle.js`

yuya-oc: `dist` and `build` are not used in this project. Is it necessary explicitly removing them?

dmeza: @yuya-oc: I thought I saw that some updates didn't get picked up because those folders already existed. If this is wrong, we should remove those from the clean task.

yuya-oc: @dmeza Yeah, please remove. Those folders were actually used in past versions, so I think that you possibly saw them.

yuya-oc: @dmeza What about this?

dmeza: @yuya-oc changed and tested.

PR #616, opened 2017-10-10
Protocol deep linking

csduarte:
**Description**
Adds protocol mattermost to fire up the desktop application from links. Also adds deep linking to be able to go to the specific channel on the link, this was done without the need for a hard refresh if the application is already running.
A use case are links on emails that would allow the user to access the desktop application directly. This was added for mac and windows.
Based on this feature proposal:
- https://mattermost.uservoice.com/forums/306457-general/suggestions/18684067-protocol-handler-for-deep-linking-to-channel-in-de
- https://mattermost.atlassian.net/browse/PLT-5853

 read and understood our Contributing Guidelines
 completed Mattermost Contributor Agreement
 executed npm run lint:js for proper code formatting

**Test Cases**
Tested with OSX El Capitan 10.11 and Windows 8.1. 
Package and run the application and run example links in a browser:
- mattermost://pre-release.mattermost.com/core/channels/ps-uber
- mattermost://pre-release.mattermost.com/core/channels/apiv4
- mattermost://pre-release.mattermost.com/core/channels/bugs

There are two cases, when the application is already running. Or after the protocol has registered in the OS, then close the app and try the protocol from a restarted browser and it should start the application correctly and go to the right channel. (edited)


yuya-oc:
@jasonblais For now, please restart the app. It seems that the problem happens while installing the app via Squirrel installer. Because it adds command line options, so the deeplink feature is confused at the first time.

yuya-oc:
Updated the first comment for our readability.

src/main.js
```diff
@@ -325,10 +349,20 @@ app.on('ready', () => {
       catch((err) => console.log('An error occurred: ', err));
   }
 
+  // Protocol handler for win32
+  if (process.platform === 'win32') {
+    // Keep only command line / deep linked arguments
+    if (Array.isArray(process.argv.slice(1)) && process.argv.slice(1).length > 0) {
```
yuya-oc: You should also handle the case, command line includes [`--squirrel-*` options](https://github.com/electron-archive/grunt-electron-installer#handling-squirrel-events). In that case, the app should run as an installer/uninstaller. So please do not anything.

src/browser/components/MattermostView.jsx
```diff
@@ -44,8 +43,18 @@ const MattermostView = createReactClass({
     var self = this;
     var webview = findDOMNode(this.refs.webview);
 
+    ipcRenderer.on('protocol-deeplink', (event, lastUrl) => {
```
yuya-oc: You should handle which tab should load the deeplink. Otherwise, non-appropriate tabs may also open the link.

src/browser/components/MattermostView.jsx
```diff
@@ -223,14 +232,18 @@ const MattermostView = createReactClass({
     if (!this.props.active) {
       classNames.push('mattermostView-hidden');
     }
+
+    const deeplinkingUrl = remote.getCurrentWindow().deeplinkingUrl;
```
yuya-oc: You should handle which tab should load the deeplink. Otherwise, non-appropriate tabs may also open the link.

jasonblais:
@yuya-oc Do you know if it's a lot of work to support this for Linux as well? Wondering if we could add it ourselves after this PR is merged, or if I should create a separate issue for us to look at later.

@csduarte @dmeza Is server work required to support deep linking from email notifications?

jasonblais:
Also, any security concerns we should be mindful of?

dmeza:
@jasonblais @yuya-oc no server work is required just to use the protocol and deep linking, but it's good practice to have a server side landing page that can show download instructions in case the native app is not installed. I have a prototype version of it ready and will create a PR for it once we have it.
It's like the last part of this tutorial:
https://developers.tune.com/sdk/deep-linking-to-your-mobile-app-from-your-website/
No security concerns that I know of. cc @csduarte

dmeza:
@yuya-oc do you know of an easy way of reading the `protocol` value from `electron-builder.json` so we don't have to have it in `src/main.js`. The other option is to add it as a config and be able to override it. What do you think?
cc @csduarte

jasonblais:
^(I'd prefer to avoid adding a config setting whenever possible)

yuya-oc:
@dmeza Please use `require('../electron-builder.json')`. The json is imported into `src/main_bundle.js` by webpack. Unfortunately I'm not sure how electron-builder uses the `protocol` (and not documented on https://www.electron.build/configuration/configuration).

Possibly this is not security, but we should take care that users might take mistake if the tab opens incorrect server by deeplinking. As I wrote in review comments, the link should be opened in the appropriate tab.

yuya-oc:
@jasonblais Possibly we can support Linux by adding `MimeType` entry into `Mattermost.desktop` file. http://archive.is/8C3zb First, we should know whether electron-builder adds it for .deb packages.

dmeza:
@yuya-oc @jasonblais added logic for:
- deep linking to go to the right tab and channel based on domain
- take value of protocol from electron-builder.json to be able to white-label easily.

protocol_deeplink

Not really sure on how to solve and test the squirrel comment.

yuya-oc:
https://circleci.com/gh/mattermost/desktop/1075#artifacts

src/main.js
```diff
@@ -313,6 +326,27 @@ ipcMain.on('download-url', (event, URL) => {
   });
 });
 
+let scheme;
+if (protocols && protocols[0] &&
```
yuya-oc: Should we consider the case where multiple protocols are listed?

src/utils/util.js
```diff
@@ -0,0 +1,6 @@
+const REGEXP_DOMAIN = /(?:[^/]*\/){3}/;
+
+export function getDomain(url) {
+  const matched = url.match(REGEXP_DOMAIN);
```
yuya-oc: Just curious, why not using Node.js URL module?

And I expected the return value of this function is `example.com` when url is `https://example.com/test`, but it seems that the actual return value is `https://example.com/`. Would you declare more right name? (It's called `origin` in Node.js URL module https://nodejs.org/dist/latest/docs/api/url.html)

src/main.js
```diff
@@ -313,6 +326,27 @@ ipcMain.on('download-url', (event, URL) => {
...
+  if (scheme) {
+    deeplinkingUrl = url.replace(scheme, 'https');
+  }
+}
+
+// Protocol handler for osx
+app.on('open-url', (event, url) => {
+  event.preventDefault();
+  setDeeplinkingUrl(url);
+  mainWindow.webContents.send('protocol-deeplink', deeplinkingUrl);
```
yuya-oc: You should also call `mainWindow.show()` to bring up hidden/minimized window.

src/browser/components/MainPage.jsx
```diff
@@ -108,6 +121,20 @@ const MainPage = createReactClass({
     ipcRenderer.on('focus-on-webview', () => {
       this.focusOnWebView();
     });
+
+    ipcRenderer.on('protocol-deeplink', (event, lastUrl) => {
+      const mattermostViews = document.getElementsByClassName('mattermostView mattermostView-with-tab');
```
yuya-oc: This is equal to `props.teams.length`. And you can access to the view via `refs[mattermostView${i}]`. So you don't have to search around DOM.

By adding `getSrc()` member function to `MattermostView`, `MainPage` doesn't have to know about `<webview>`. It should be the role of `MattermostView`.

jasonblais:
Works on Windows as well.

@dmeza Curious why we didn't use a protocol handler that was more consistent with other apps, in the form:

```
mattermost://channel?id=<CHANNEL-ID>&team=<TEAM-ID>
```

Is this something we should consider?

----

Created two follow-up tickets:
 - Support protocol deep linking on Linux (@yuya-oc): https://github.com/mattermost/desktop/issues/627
 - Add a server landing page when the app isn't installed (@dmeza): https://github.com/mattermost/desktop/issues/628

dmeza:
@jasonblais @yuya-oc the protocol handler is this part `mattermost://` the other part are the parameters used for deep linking.
In the case of the desktop app we need the domain to know what tab to use and the rest of the url to go to that specific path inside of that server and not have to refresh the all the webview.

For the IOs and Android apps we might need to add the other parameters at the end. For example: 
`mattermost://pre-release.mattermost.com/core/channels/bugs?id=<CHANNEL-ID>&team=<TEAM-ID>`

jasonblais:
@dmeza Why would we need the additional parameters for mobile?

Makes sense that we need the server URL to know which tab to switch to (I realize now it wouldn't work without it). The existing implementation sounds good.

@dmeza Please help review Yuya's earlier comments from code review, and this should be good to merge.

Also adding @MusikPolice as a second code reviewer.

dmeza:
@jasonblais the additional parameters are needed for mobile because they don't use urls like platform or desktop. We discussed with @jarredwitt that they make APIv4 calls directly.

I have the changes for @yuya-oc's comment almost ready. I'll push later today.

jasonblais:
@dmeza Got it, thanks!

And that's great, thanks for your continued work on this PR

src/main.js
dmeza: @yuya-oc I tried adding multiple protocols but after a while I realized that it's not possible because: `app.setAsDefaultProtocolClient(protocol)` just registers the last one.

dmeza:
@yuya-oc @jasonblais except for the multi protocol that's not possible, all other comments are done. Tested on mac and windows that tabs display properly based on domain and that it goes to the right channel. Also tested that it works with the app closed.
Let me know if you see any other issues.

yuya-oc:
Tested https://circleci.com/gh/mattermost/desktop/1080#artifacts

Almost okay, but when the application was launched by deeplinking, it keep the link even when after using the settings page. https://github.com/mattermost/desktop/pull/616#discussion_r146103896

And a permalink as deeplinking also works.
`mattermost://pre-release.mattermost.com/core/pl/bc3he4u37fdwzqwfdtychcc1go` (You would see "v3.7.1 has been published" of Desktop App channel)

src/browser/components/MainPage.jsx
```diff
@@ -25,14 +27,25 @@ const MainPage = createReactClass({
   },
 
   getInitialState() {
+    const deeplinkingUrl = remote.getCurrentWindow().deeplinkingUrl;
+    let key = this.props.initialIndex;
+    if (deeplinkingUrl !== null) {
+      for (var i = 0; i < this.props.teams.length; i++) {
+        if (deeplinkingUrl.includes(this.props.teams[i].url)) {
+          key = i;
```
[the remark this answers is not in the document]
dmeza: This case was weird and hard to handle, but it's done and tested.

dmeza:
@yuya-oc fixed this case: https://github.com/mattermost/desktop/pull/616#discussion_r146103896

yuya-oc:
https://circleci.com/gh/mattermost/desktop/1084#artifacts

The test case has been resolved 👍

src/main.js
```diff
@@ -46,7 +46,7 @@ const assetsDir = path.resolve(app.getAppPath(), 'assets');
 // be closed automatically when the JavaScript object is garbage collected.
 var mainWindow = null;
 let spellChecker = null;
-let deeplinkingUrl = null;
```
dmeza: @yuya-oc @jasonblais removing the global variable `deeplinkingUrl` breaks the logic to be able to deep link when the application is closed. Test case would be to have multiple tabs, close the app completely and deep link to a channel in the second tab. It stays on the first tab.

dmeza:
@yuya-oc a question just to educate me, why is it a good idea to disable deep linking in dev mode?

src/main.js
yuya-oc: @dmeza I believed there are no cases which must refer the global variable though, I’ll check again. Thanks.

yuya-oc:
@dmeza It seems that electron.exe is registered as the protocol client in dev mode. In such case, the application couldn’t launch correctly because the second arg to specify runtime code is missing.

yuya-oc:
As mentioned by @dmeza https://github.com/mattermost/desktop/pull/616#discussion_r146602059, the global variable is needed for mac at `app.on('open-url')` event. So I reverted the commit ea746bad2915b604eb816b34027d4be889eccb9a as d2c4aa0546361ad6ae4722e8eea9ac6bd9fcbe85. Thanks!

Issue #636, opened 2017-10-29
OSX: tri-tone sound playing instead of Mattermost "ding" sound


philip-peterson:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Mac OSX "installation finished" (aka tri-tone) sound plays every time there is a Mattermost notification.

**Steps to reproduce**

Using Mattermost app on any OSX system, have a friend send you a message when the app is not focused while the volume is on.

**Expected behavior**
Mattermost ding sound will play.

**Observed behavior**
OSX Tri-tone sound will play.

**Possible fixes**
I have a PR that fixes this issue. Just wanted to create a ticket first in case someone has an objection.

amyblais:
Closing via feature request: https://mattermost.uservoice.com/forums/306457-general/suggestions/8686084-change-notification-sound.

PR #637, opened 2017-10-29
Play ding.mp3 instead of tri-tone sound for notifications on OSX. #636

philip-peterson:

**Summary**
Play ding.mp3 instead of tri-tone sound for notifications on OSX. 

**Issue link**
https://github.com/mattermost/desktop/issues/636

**Test Cases**
* OSX 10.12.6

**Additional Notes**
None.


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Release date: TBD
  [#512](https://github.com/mattermost/desktop/issues/512)
  - Added the feature to open the application via `mattermost://` link.
  [#616](https://github.com/mattermost/desktop/pull/616)
+ - Change to use Mattermost's own notification sound instead of
+ the built-in tri-tone OSX sound. [#636](https://github.com/mattermost/desktop/issues/636)
```
philip-peterson: Not super clear here on whether we want issue numbers or pull request numbers here...

yuya-oc:
Thank you for your feedback.

The app would always play the sound even when the user disabled notification sounds in OS preferences (On macOS, Preferences.app -> Notifications). At least the app should follow the preference. It's a reason why we use built-in sound.

Unfortunately I'm not familiar at macOS native part, so I'm not sure which Electron API can get the appropriate value. (Possibly [`systemPreferences.getUserDefault()`](https://electron.atom.io/docs/api/system-preferences/#systempreferencesgetuserdefaultkey-type-macos)?)

JuniorGenius:
Actually, I already use the code from this PR and that's not the case for me. The notification sound setting still works. If there's a better way to solve it, then by all means, but this solution does work.

yuya-oc:
It would be helpful if the case is concretely written in detail.

I'm sure that's confusing point though, the sound for desktop app's notifications depend on two kinds of preferences.

- A: Mattermost webapp -> Main Menu [≡] -> Account Settings -> Notifications -> Desktop notifications -> Notification sound
- B: OS preferences (On macOS, Preferences.app -> Notifications or "Do Not Disturb" of Notification Center)

When B is off, the app should not play the sound even if A is on. i.e. B has higher priority to disable the sound. But this PR plays the sound in that case.

JuniorGenius:
That's just a by-product of using a custom alert sound. As far as I'm aware, there's no way around it, and most major apps (i.e. Discord, Slack) that use alerts just provide a do not disturb setting from within the app (or some other way to turn off notification sounds, which we already have)

yuya-oc:
A little researched. Surely it seems there is no way to get Do Not Disturb settings. However I couldn't hear sound of notifications from Slack in No Not Disturb. But the sound is different, so I think there is a way to change it. (In my memory, `sound` parameter of `new Notification()` is not supported yet...)

Oppositely, I'm not sure why ding.mp3 should be used instead of built-in sound. Would you explain about your case?

JuniorGenius:
To clarify; for me, the slack sound _is_ muted when Slack's built-in do not disturb status is enabled, but the sound always plays when only OSX's do not disturb mode is on. That's how it's always been for me, presumably because there's no way to check the setting in the API

The custom sound is just a lot better-suited. I'd be fine with another sound, but in my opinion, the default tri-tone sound is very loud, jarring, dated, and out of place for a messaging alert.

(And I'm not the only one that feels this way):
https://mattermost.uservoice.com/forums/306457-general/suggestions/17213822-use-custom-notification-sound
https://mattermost.uservoice.com/forums/306457-general/suggestions/8686084-change-notification-sound

yuya-oc:
I see about the situation. By the way, I missed a changelog in Electron. Since v1.7.7, `sound` is supported on macOS. I'm wondering if you would like to investigate it. I think that it properly uses custom sounds and 'Do Not Disturb'.

As described at https://github.com/electron/electron/pull/10293, you would need to specify the local file.

JuniorGenius:
Ah, very cool! Will do.

philip-peterson:
@yuya-oc Thinking about how to tackle this. While the new electron notification class does support sounds, unfortunately it does not seem to be do-not-disturb-aware. So the only reason to use the `sound` property would be idealism. There's actually a reason _not_ to use it as well, which is the throttling feature that's used for Windows. Let me know if you'd rather I use that property (my vote would be not).

Luckily, there is a way to check whether notifications are muted via dnd. Will submit a PR for that soon.

philip-peterson:
We actually might want to use `sound` if it works on Windows as it'll let us kill the `throttle` code. I don't have a Win machine so can't test unfortunately

jasonblais:
Hey @philip-peterson @JuniorGenius Any updates here?

philip-peterson:
@jasonblais Thanks for checking. I have been attempting to use the new `sound` property, but there seems to be no documentation on what kind of path `sound` should be. I tried a Data URI and also an absolute path to the Assets folder. Neither worked -- starting to think it might need to be some kind of relative path since that's the only remaining option, but it's unclear what the URL should be relative to. Any ideas?

jasonblais:
Thoughts on this @yuya-oc?

yuya-oc:
`sound` is described at https://electronjs.org/docs/api/notification#playing-sounds . It's internally used by [NSUserNotification](https://developer.apple.com/documentation/foundation/nsusernotification?language=objc) in electron's native part. In my guess, it's finally used by [NSSound](https://developer.apple.com/documentation/appkit/nssound/1477318-soundnamed?language=objc). So probably we need to set "sound name".

When investigating, I suggest to create a simple electron app before editing our code. If you are not familiar at electron stack, you can use my repository, [yuya-oc/electron-scaffold](https://github.com/yuya-oc/electron-scaffold). It creates a simple electron app.

jasonblais:
@philip-peterson Let us know if we can help, or if you have any questions about Yuya's comment above?

philip-peterson:
@jasonblais @yuya-oc I've been working to get the Electron API to find the sound. The [docs](https://electronjs.org/docs/api/notification#playing-sounds) suggest that the file needs to be present in the Resources folder, so I tweaked the `electron-builder.json` to include that file. Unfortunately even after doing so, Electron seems to be unable to find the file. I've tried relative paths, absolute paths, and also referring to the file by name with no extension. Please let me know if you have any other suggestions.

yuya-oc:
@philip-peterson I have looked up docs and codes. I could get the custom sound in the main process when I converted the mp3 to aiff, but I'm not sure why.

```javascript
const notification = new Notification({
  title: 'Main Process',
  sound: 'ding' // Locate ding.aiff to ~/Library/Sounds/
})
```

[The doc](https://electronjs.org/docs/api/notification) is writing about the main process. But [its implementation](https://github.com/electron/electron/blob/v1.8.4/brightray/browser/mac/cocoa_notification.mm#L44-L50) is also used by renderer processes. So I had thought `sound` should work in webview. But when using debugger, `sound` was not passed to the native part. Probably `sound` is not implemented yet in Chromium because [`sound` doesn't exist in content::PlatformNotificationData struct](https://github.com/electron/electron/blob/v1.8.4/brightray/browser/platform_notification_service.cc#L27).

A way to avoid the problem is to use the main process and IPC. But that would need more complicated codes because IPC can't pass callback functions.

jasonblais:
Hey @philip-peterson - any questions on Yuya's note above?

esethna:
Closing this PR due to inactivity, let us know if you'd like it reopened :)

Issue #643, opened 2017-11-06
Lag when clicking "Allow" or "Block" after allowing app to show notifications


jasonblais:
Follow-up for https://github.com/mattermost/desktop/pull/609 where clicking "Allow" or "Block" results in a 1-2 second lag until the action is accepted.

Possible related to React's state management although more work is needed to find the actual problem.

jasonblais:
Assigning to v3.8.0 for now to track. We may want to investigate it for the release.

PR #655, opened 2017-11-16
Fix permission dialog lag

yuya-oc:

**Summary**
Fix lag to show/hide permission dialog.

If the LHS has unread channels, the dialog doesn't correctly appear/hide due to complicated rendering function of TabBar.jsx. Now some `if` statements are removed, so the TabBar correctly render the dialog in any cases.

By the way, I feel `notifications` should be allowed as default. Users might not notice the dialog until they open the tab.

**Issue link**
#643

**Test Cases**
1. Remove `AppData\Roaming\Mattermost\permisson.json`
2. Start the app.
3. Receive new messages or WebRTC concall

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/499#artifacts

### DIFF
--- a/src/browser/components/PermissionRequestDialog.jsx
+++ b/src/browser/components/PermissionRequestDialog.jsx
@@ -79,8 +79,8 @@ function PermissionRequestDialog(props) {
 }
 
 PermissionRequestDialog.propTypes = {
-  origin: PropTypes.string.isRequired,
-  permission: PropTypes.oneOf(['media', 'geolocation', 'notifications', 'midiSysex', 'pointerLock', 'fullscreen', 'openExternal']).isRequired,
+  origin: PropTypes.string,
+  permission: PropTypes.oneOf(['media', 'geolocation', 'notifications', 'midiSysex', 'pointerLock', 'fullscreen', 'openExternal']),
   onClickAllow: PropTypes.func,
   onClickBlock: PropTypes.func,
   onClickClose: PropTypes.func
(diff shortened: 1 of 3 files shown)


jasonblais:
@yuya-oc This works great to me. 👍 

> By the way, I feel notifications should be allowed as default. Users might not notice the dialog until they open the tab.

That might be correct - but I think Chrome requests permissions from the user first before sending desktop notifications on Windows. Have you noticed the same?

yuya-oc:
@jasonblais Yeah, Chrome's behavior is appropriate as a web browser. But in our case, notification is a one of core intended features. In almost case, notification would be allowed. Of course, the dialog should appear for untrusted (not added in the settings page) origins.

jasonblais:
@yuya-oc That makes sense. I'm actually reading the Doyensec's report, and it requires one to use setPermissionRequestHandler for untrusted origins. But then for trusted origins this wouldn't be required and we could skip this step.

What do you think? If so, we probably wouldn't need an advanced settings modal either to manage your permissions, given the user would automatically give permissions for notifications and microphone/video access within Mattermost.

jasonblais:
> given the user would automatically give permissions for notifications and microphone/video access within Mattermost

And they can block desktop notifications anytime from Account Settings, and turn off camera and video during video calls, if they're in a Mattermost server.

yuya-oc:
@jasonblais I agree. I was a little worried about microphone/video access without permissions, but in almost case it's also intended. By removing `permission.json`, users can always reset allowed/denied permissions, so the dialog would not be needed.

jasonblais:
@yuya-oc Would this be a difficult change? Should I create a separate ticket for it?

yuya-oc:
I think not so much. But I feel we should have another ticket.

jasonblais:
Sounds good, created it here: https://github.com/mattermost/desktop/issues/659

Thanks @yuya-oc!

PR #656, opened 2017-11-17
Replace hard-coded doc link with a redirect

jasonblais:
Instead of hard-coding the help link to https://docs.mattermost.com/help/apps/desktop-guide.html, added a redirect from https://about.mattermost.com/default-desktop-app-documentation/.

This is so that if the help documentation moves, we can just update the redirect at .com page, instead of making a change in the product.

Note: https://about.mattermost.com/default-desktop-app-documentation/ gives a 404 error until the marketing team pushes the change. Should be in the next couple of days.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Packages are available on the [releases page](http://github.com/mattermost/deskt
 ## Usage
 
 ### Installation
-Detailed guides are available at [docs.mattermost.com](https://docs.mattermost.com/help/apps/desktop-guide.html).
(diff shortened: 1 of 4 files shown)


yuya-oc:
@jasonblais Should we also consider to update these two lines? https://github.com/mattermost/desktop/blame/2aea3150f81a6ed3ab41d6c4ee91a547b5985531/TESTING.md#L16-L17

jasonblais:
Good catch! Updated those two links

PR #665, opened 2017-11-23
Remove dev dependancies from NOTICE.txt

jasonblais:
These are dev dependancies in https://github.com/mattermost/desktop/blob/master/package.json and not bundled with the release build like regular dependancies.

@yuya-oc please confirm if any of these should be considered dependancies instead of dev dependancies 

https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

PS: Not urgent to review, any time in the next week or two before the v4.0 release.


yuya-oc:
At least `dependencies` of `src/package.json` should be listed. Unless we have mistake at two `package.json`, they and their [nested dependencies](https://maxogden.com/nested-dependencies.html) are bundled into compiled JavaScript by webpack.

jasonblais:
@yuya-oc Yes, I kept those in NOTICE.txt and only removed the `devDependancies` of `package.json`.

yuya-oc:
@jasonblais Okay, looks good to me.

Issue #678, opened 2018-01-12
mattermost:// protocol doesn't launch the desktop app if it's not already open


jasonblais:
I confirm (by marking "x" in the [ ] below: [x]):

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
mattermost:// protocol doesn't launch the desktop app if it's not already open

**Steps to reproduce**

Mac; Mattermost server v4.6.0-RC1; Mattermost Desktop App v4.0.0-beta2

1. Sign in to https://pre-release.mattermost.com on your desktop app
2. Post http://pre-release.mattermost.com/core/channels/developers into a channel
3. Quit the desktop app
4. Sign in to https://pre-release.mattermost.com on your browser
5. Click on the link you posted in step 2.

**Expected behavior**
Desktop app launches, and opens the Developers channel.

**Observed behavior**

Opens in browser, doesn't launch desktop app.

NOTE: If the desktop app is already open, the app gets focus and redirects to the Developers channel.

yuya-oc:
Reproduced also on my Mac. But not reproduced on Windows 10.

PR #588, opened 2017-09-08
BASH-2 Bundle Mac Desktop App as DMG

csduarte:
**Description**
Changed build config for mac to generate DMG. User experience is much better when installing via DMG.

- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
 - [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
 - [x] executed `npm run lint:js` for proper code formatting

**Test Cases**
run : `npm run package:mac`
Execute the resulting `Mattermost-xxx.dmg` file.

screen shot 2017-09-07 at 12 51 04 pm

PR #680, opened 2018-01-16
Fix deep linking not working when the app is not already opened

yuya-oc:

**Summary**
Fix deep linking not working when the app is not already opened.

On macOS, deep linking is implemented with `open-url` event. When the app is launched by URL scheme, the event is emitted before `ready` event. In this case, the main windows is not created yet. So there was null reference error and the app was not able to start correctly.

**Issue link**
#678 

**Test Cases**
1. Register 2 servers to the app. https://pre-release.mattermost.com must be the second server.
2. Quit the app.
3. Open mattermost://pre-release.mattermost.com/core/channels/developers in your web browser.
4. App should be launched then the Developers channel of pre-release should be opened.

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/534#artifacts

### DIFF
--- a/src/browser/components/MattermostView.jsx
+++ b/src/browser/components/MattermostView.jsx
@@ -8,6 +8,8 @@ const {findDOMNode} = require('react-dom');
 const {ipcRenderer, remote, shell} = require('electron');
 const url = require('url');
 const contextMenu = require('../js/contextMenu');
+const {protocols} = require('../../../electron-builder.json');
+const scheme = protocols[0].schemes[0];
 
 const ErrorView = require('./ErrorView.jsx');
 
@@ -72,7 +74,7 @@ const MattermostView = createReactClass({
     webview.addEventListener('new-window', (e) => {
       var currentURL = url.parse(webview.getURL());
       var destURL = url.parse(e.url);
-      if (destURL.protocol !== 'http:' && destURL.protocol !== 'https:') {
+      if (destURL.protocol !== 'http:' && destURL.protocol !== 'https:' && destURL.protocol !== `${scheme}:`) {
         ipcRenderer.send('confirm-protocol', destURL.protocol, e.url);
         return;
       }
--- a/src/main.js
+++ b/src/main.js
@@ -361,9 +361,16 @@ app.on('will-finish-launching', () => {
   app.on('open-url', (event, url) => {
     event.preventDefault();
     setDeeplinkingUrl(url);
-    if (mainWindow) { // 'open-url' is emitted before 'ready' when the app is launched by URL scheme.
-      mainWindow.webContents.send('protocol-deeplink', deeplinkingUrl);
-      mainWindow.show();
+    if (app.isReady()) {
+      function openDeepLink() {
+        try {
+          mainWindow.webContents.send('protocol-deeplink', deeplinkingUrl);
+          mainWindow.show();
+        } catch (err) {
+          setTimeout(openDeepLink, 1000);
+        }
+      }
+      openDeepLink();
     }
   });
 });
--- a/src/main.js
+++ b/src/main.js
@@ -356,14 +356,16 @@ function setDeeplinkingUrl(url) {
   }
 }
 
-// Protocol handler for osx
-app.on('open-url', (event, url) => {
-  event.preventDefault();
-  setDeeplinkingUrl(url);
-  if (mainWindow) { // 'open-url' is emitted before 'ready' when the app is launched by URL scheme.
-    mainWindow.webContents.send('protocol-deeplink', deeplinkingUrl);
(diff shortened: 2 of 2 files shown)


jasonblais:
@yuya-oc Sorry, I didn't notice your PR right away.

Step 4 doesn't occur for me. I see the Mattermost icon appear for a moment in my dock bar, but then it disappears and the app doesn't launch.

Second, if I click on mattermost://pre-release.mattermost.com/core/channels/developers in the desktop app, I get the following dialog. Is that expected?

image

yuya-oc:
The step 4 was reproduced also for me. But it's gone after a moment. I'm not sure why it happens though, I followed [Electron's doc](https://electronjs.org/docs/api/app#event-will-finish-launching). Would you try this again?
https://circleci.com/gh/yuya-oc/desktop/541#artifacts

Then the dialog is expected because `mattermost` is not `http/https`. Probably should we allow the protocol implicitly?

jasonblais:
@yuya-oc Yes, let's allow the `mattermost` protocol implicitly, similar to `http` and `https` protocols.

jasonblais:
Sorry @yuya-oc, it still doesn't work for me. Any further information I can provide to help troubleshoot?

yuya-oc:
Updated. https://circleci.com/gh/yuya-oc/desktop/545#artifacts

@jasonblais I prepared a test version for Mac. This creates a log as `/tmp/mattermost-desktop.log`. Would you try this one?
https://circleci.com/gh/yuya-oc/desktop/544#artifacts

jasonblais:
@yuya-oc Sorry..where is the `tmp` folder located?

yuya-oc:
@jasonblais Ah, possibly Finder doesn’t show it. So the folder can be opened with `open /tmp` command. And please use 544 artifact. The difference is whether there is the log.

jasonblais:
@yuya-oc With 544 artifact, I got this in the log

```
will-finish-launching
ready
open-url: mattermost://ci-linux-postgres.mattermost.com/jira-demo/channels/off-topic
open-url: mainWindow: true

```

yuya-oc:
That's weird. On my macOS Sierra, the log is below.

```
will-finish-launching
open-url: mattermost://pre-release.mattermost.com/core/channels/developers
open-url: mainWindow: false
ready
```

In this case, the feature should work.

For now I'm not sure why 'ready' is emitted before 'open-url' though, I'll take a look.

yuya-oc:
@jasonblais Unfortunately I'm not sure whether this can fix the problem though, would you try this?

- https://circleci.com/gh/yuya-oc/desktop/551#artifacts (normal)
- https://circleci.com/gh/yuya-oc/desktop/552#artifacts (with log)

jasonblais:
@yuya-oc Sure thing, testing now.

jasonblais:
@yuya-oc Unfortunately it didn't. But I just realized it's asking to open the "Mattermost 10" desktop app, which is the one I downloaded last Friday and doesn't have your most recent bug fixes.

So it seems the link is trying to open an old version of the app.

image

jasonblais:
@yuya-oc I uninstalled the app, re-installed artifact 552 and it's working now.

Wondering if this is something we should investigate, or do you think it was an issue on my system?

jasonblais:
I installed artifact 552. When I carry out the test steps, the version I downloaded 20 minutes ago opens.

Log
```
will-finish-launching
open-url: mattermost://ci-linux-postgres.mattermost.com/jira-demo/channels/off-topic
open-url: mainWindow: false
app is NOT ready
ready
```

MacOS system info:
image

yuya-oc:
Now two events were emitted as I expected. So there is a case where the order was changed due to some reason. So this PR would not solve the problem completely.

Probably we need to investigate Electron in detail. So I think this PR is enough for now though . Any thoughts?

jasonblais:
@yuya-oc One thing I worry is that if a user clicks a link with the mattermost protocol, will an old version of the app open (or try to open)?

yuya-oc:
@jasonblais On Mac, usually the apps should be installed into `/Application`. So probably it's used first. As long as users use the folder, I think your worry would not happen.

Our current .tar.gz is a simple archive file though, so users might not use `/Application`. But .dmg of #588 helps to install the app into `/Application`.

jasonblais:
Fair enough. @yuya-oc you can merge. Can you also create a follow-up ticket for further investigation.

jasonblais:
@yuya-oc If you merge the other two PRs in https://github.com/mattermost/desktop/milestone/7 milestone, we can go ahead and cut 4.0.0-RC1 as well.

yuya-oc:
Thanks. Created #688.

PR #745, opened 2018-03-26
MM-9922 Don't show hovered URL for internal links

hmhealey:
**Summary**
We keep getting tickets to hide different links throughout the app, but the root problem seems to be that these are links within the app, so they're not considered important. We still want the full URL to appear on hover for external links so the user can avoid malicious ones, but this should solve it without us making additional changes to the web app.

**Issue link**
https://mattermost.atlassian.net/browse/MM-9922

**Test Cases**
Tested locally using `yarn run watch`. Let me know if there's anywhere else that I should be adding tests

### DIFF
--- a/src/browser/components/HoveringURL.jsx
+++ b/src/browser/components/HoveringURL.jsx
@@ -2,10 +2,6 @@ const React = require('react');
 const PropTypes = require('prop-types');
 
 function HoveringURL(props) {
-  if (props.targetURL.startsWith(props.currentTeamURL)) {
-    return <div/>;
-  }
-
   return (
     <div className='HoveringURL HoveringURL-left'>
       {props.targetURL}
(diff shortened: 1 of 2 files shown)


jasonblais:
Tested on Mac, works great. Tested on both a new (v4.8) and old (v3.8) server versions.

@amyblais Can you help test on Windows? https://clanhq.battlehouse.com/ is running an old (v3.8) public version to try in.

src/browser/components/HoveringURL.jsx
```diff
@@ -2,6 +2,10 @@ const React = require('react');
 const PropTypes = require('prop-types');
 
 function HoveringURL(props) {
+  if (props.targetURL.startsWith(props.currentTeamURL)) {
```
yuya-oc: When mouse cursor moves between internal links and external links, hovering animation is not played because `<CSSTransition>` still exists even if the URL is hidden. So the condition should be examined at MainPage.jsx.

https://github.com/mattermost/desktop/blob/d1b70863b53a2485bdc5d55f9bc2092782bed5bc/src/browser/components/MainPage.jsx#L367-L368

Issue #518, opened 2017-04-26
Apply a maximum server tab width of 17px


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

**Summary**
Add a maximum server tab width of 17px

**Steps to reproduce**

Create a very long server tab name, 40-50 characters long

**Expected behavior**

Server name is truncated (common in web browsers such as Firefox)

![image](https://cloud.githubusercontent.com/assets/13119842/25448428/723ada48-2a86-11e7-98d4-6ee934b3f2fc.png)

**Observed behavior**

The entire server name appears in the tab bar. This can cause issues where if you have two or three servers each with a long server name, the last one isn't visible on the desktop app UI
![image](https://cloud.githubusercontent.com/assets/13119842/25448364/2ff53f8e-2a86-11e7-94b2-bdff80ab67de.png)

**Possible fixes**

Add a maximum server tab width of 17px. This should fit up to 20 character server names.

If the name doesn't fit within 17px, the server name should be truncated with `...` at the end.

jasonblais:
^@yuya-oc feel free to comment if you feel the 17px limit is too restrictive, but I estimated it based on the Firefox browser tab widths.

(in Chrome, the browser tab width auto-adjusts based on how many tabs you have open, but I don't think we need that since it's not likely for a user to belong to a lot of Mattermost instances at once)

PR #760, opened 2018-04-13
Truncate long server names in server tabs

kethinov:
- Should resolve https://github.com/mattermost/desktop/issues/518.
- Shows server name in a `title` attribute so that you can still see the full name on hover.
- Unusual placement of CSS changes is to prevent a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content).

Here is a screenshot of how it looks.

screen shot 2018-04-13 at 11 39 36 am

If you'd like it to look different, let me know, or if you think we should solve this problem a different way, let me know.

### DIFF
--- a/src/browser/css/components/TabBar.css
+++ b/src/browser/css/components/TabBar.css
@@ -1,3 +1,11 @@
+.TabBar .teamTabItem span {
+  display: inline-block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  max-width: 170px;
+}
+
 .TabBar>li>a {
   background: rgba(0, 0, 0, 0.05);
   border-radius: 2px 2px 0 0;
@@ -38,6 +46,6 @@
   border-radius: 50%;
 }
 
-.TabBar .teamTabItem-label {
+.TabBar .teamTabItem-unread {
   font-weight: bold;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Release date: TBD
 [#746](https://github.com/mattermost/desktop/pull/746)
 - Hide hovering URL bar for internal links.
 [#745](https://github.com/mattermost/desktop/pull/745)
+- Truncate long server names in server tabs.
+[#518](https://github.com/mattermost/desktop/issues/518)
 
 #### Windows
 - [Windows 7/8] Desktop notifications now respect the duration setting of Control Panel.
--- a/src/browser/components/TabBar.jsx
+++ b/src/browser/components/TabBar.jsx
@@ -61,7 +61,7 @@ export default class TabBar extends React.Component { // need "this"
           ref={id}
           draggable={false}
         >
-          <span className={unreadCount === 0 ? '' : 'teamTabItem-label'}>{team.name}</span>
+          <span title={team.name} className={unreadCount === 0 ? '' : 'teamTabItem-unread'}>{team.name}</span>
           { ' ' }
           { badgeDiv }
           {permissionOverlay}


yuya-oc:
@jasonblais #518 says 17px, but probably it's mistake because it's too small. Is 117px appropriate? And this PR makes static width instead of maximum width. Fixed width might make sense in appearance.

src/browser/index.html
```diff
@@ -4,6 +4,26 @@
...
+     *
+     * Unusual placement of these styles is so that they activate while the loading
+     * screen is up rather than applying after React initializes to prevent a flash
+     * of unstyled content.
+     *
+     */
+
+    .TabBar .teamTabItem span {
+      display: inline-block;
+      font-weight: bold;
```
yuya-oc: `font-weight: bold` must be used only in the case where the guest page has unread channels.

src/browser/index.html
```diff
@@ -4,6 +4,26 @@
...
   <meta charset="UTF-8">
   <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.min.css">
+  <style>
+
+    /*
+     * Note:
+     *
+     * Unusual placement of these styles is so that they activate while the loading
+     * screen is up rather than applying after React initializes to prevent a flash
+     * of unstyled content.
```
yuya-oc: Probably I have never noticed FOUC in the desktop app. And I think CSS is loaded [here](https://github.com/mattermost/desktop/blob/3fdcb64358909fa530f0ca9e1a69a9e784cb99f3/src/browser/index.jsx#L3) before rendering React compoentns.
Would you provide some gif or picture to know what you saw?

src/browser/css/components/TabBar.css
```diff
@@ -37,7 +37,3 @@
   margin-top: 5px;
   border-radius: 50%;
 }
-
-.TabBar .teamTabItem-label {
-  font-weight: bold;
```
yuya-oc: `font-weight: bold` must be used only in the case where the guest page has unread channels. In the meaning of that, the original `.teamTabItem-label` is not appropriate class name though. So feel free to edit class name if you need.

src/browser/index.html
kethinov: This gif shows how it behaves with all my CSS changes removed from index.html and moved into TabBar.css like this:

```css
.TabBar .teamTabItem-label {
  display: inline-block;
  font-weight: bold;
  overflow: hidden;
  text-overflow: ellipsis;
  white-space: nowrap;
  width: 117px;
}
```

screencast 2018-04-15 10-28-04

At the start of the gif, the app is closed and I open it fresh.

After it is opened, you can see a succession of different styles applied as different CSS files load.

I poked around for a while and the best spot I could find to place the CSS was right in index.html. That was the only way I could suppress the FOUC.

src/browser/css/components/TabBar.css
kethinov: Pushed changes to address that.

src/browser/index.html
kethinov: Ah, I get what you're going for now. The fact that original `teamTabItem-label` class wasn't being applied until later (when unread messages were detected) created this apparent FOUC in my test case.

When I applied my new CSS (sans the bold rule) using the new selector I wrote to TabBar.css, the FOUC went away. My latest commit should resolve that.

yuya-oc:
Looks good to me.

@jasonblais The remaining issue is following; Which kind of width should we use, fixed width or variable width?

kethinov:
Pushed a change to convert it to `max-width`.

kethinov:
Done.

yuya-oc:
Thanks!

Tested on Windows and Ubuntu 16.04. Both okay.

PR #766, opened 2018-04-23
Update CHANGELOG.md

amyblais:
Lines 27 and 30 need Yuya's attention.

I'll update the date after.


yuya-oc:
@amyblais Can we have new PR against to release-4.1 branch then close this? It has an updated changelog because several PRs are newly merged there. And we can get complete changelog when creating a new tag. It would make sense when checking out the tag later.

Issue #707, opened 2018-02-09
Right-clicking an image, then choosing "Save Image" does nothing


jasonblais:
I confirm (by marking "x" in the [ ] below: [x]):

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Right-clicking an image, then choosing "Save Image" does nothing

**Steps to reproduce**

Mac and Windows; Mattermost server v4.6 and v4.7-RC1. Mattermost Desktop App v4.0.0

1) Upload an image and post it to a Mattermost channel
2) Right-click the image, then choose "Save Image"

**Expected behavior**
File explorer (on Windows) / Finder (on Mac) opens and lets you choose where to save the file.

**Observed behavior**
Nothing happens

yuya-oc:
Probably related. https://github.com/sindresorhus/electron-context-menu/issues/13
However I'm not sure why this makes issue now because it was reported one and a half year ago.

jasonblais:
@yuya-oc Yeah, I tested with Desktop App v3.7.1 and it was working as expected. So I'd be surprised if https://github.com/sindresorhus/electron-context-menu/issues/13 was the issue.

(The maintainer also makes a note that the proper text would be `Save Image As...`.

jasonblais:
You don't think it's anything related to Redux changes in the webapp?

yuya-oc:
Closed via #774.

PR #774, opened 2018-04-25
Pin electron-context-menu to version 0.9.0 (Fix for save image context item)

Autre31415:
This is a resolution to #707 

The root cause to this loss of functionality was this pull request merged into electron-context-menu about 1 year ago: https://github.com/sindresorhus/electron-context-menu/pull/40/files#diff-168726dbe96b3ce427e7fedce31bb0bcL48

By eliminating the arguments in the click event important context was lost in its downstream dependency [electron-dl](https://github.com/sindresorhus/electron-dl/blob/master/index.js#L129) which handled the target for the image download.

For now this issue can be solved by pinning version 0.9.0 which doesn't include that change, and in the mean time I've also opened a pull request to electron-context-menu itself that restores these parameters: https://github.com/sindresorhus/electron-context-menu/pull/54

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -12,7 +12,7 @@
     "auto-launch": "^5.0.5",
     "bootstrap": "^3.3.7",
     "create-react-class": "^15.6.3",
-    "electron-context-menu": "^0.9.0",
+    "electron-context-menu": "0.9.0",
     "electron-devtools-installer": "^2.2.3",
     "electron-is-dev": "^0.3.0",
     "electron-squirrel-startup": "^1.0.0",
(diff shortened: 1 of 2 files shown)


amyblais:
Thank you for submitting this fix! Approving for Windows.

jasonblais:
By the way, just wanted to say 'thanks' one more time for this fix @Autre31415 - we didn't figure it out when we looked into it earlier and have a few people who reported it.

Autre31415:
Happy to help!

Hopefully the electron-context-menu people have a chance to look at my PR on their end, as I imagine any app using the functionality in a similar way would also have the issue. Unfortunately the repo hasn't seemed very active lately.

PR #775, opened 2018-04-25
Fix linting in TabBar.jsx

Autre31415:
While working on #774 I noticed a small unrelated linting error. I thought it would be better to separate that fix into another branch rather than conflating it with the dependency pin.


jasonblais:
@amyblais Since this is a lint change, probably no testing needed here. Should have no functional changes. 

//cc @yuya-oc

PR #777, opened 2018-04-26
Revert "Merge pull request #745 from mattermost/mm9922"

hmhealey:
**Summary**
Unfortunately, the actual behaviour that we want where hovered links in posts always show the URL isn't possible with this behaviour since I haven't found a way to identify what in the WebView has been hovered from these events, so this will have to be handled from the web app.

**Issue link**
https://mattermost.atlassian.net/browse/MM-9922

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -317,7 +317,6 @@ const MainPage = createReactClass({
       authServerURL = `${tmpURL.protocol}//${tmpURL.host}`;
       authInfo = this.state.loginQueue[0].authInfo;
     }
-    var currentTeamURL = this.props.teams[this.state.key].url;
     var modal = (
       <NewTeamModal
         show={this.state.showNewTeamModal}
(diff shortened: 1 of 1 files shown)


amyblais:
@yuya-oc I believe this is in release-4.1 branch?

yuya-oc:
Thanks!

@hmhealey Just curious, is there something like class which we can know whether we should show URL? Possibly we might be able to try again by combining other APIs.

hmhealey:
@yuya-oc I couldn't find anything on the [update-target-url](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#event-update-target-url) event that allows us to determine what element on the page triggered the event. The event's target field just points to the webview.

To get the functionality we want, I'm looking at reusing the component we have in the channel sidebar that renders an `<a>` tag in the web app and a `<button>` with a custom context menu in the desktop app since that works how we want. That just means we'll need to apply it manually when we want to get rid of the hovered URL

yuya-oc:
@hmhealey Yeah, the event only emits URL. So I had thought possibly we could get the target element by injecting JavaScript into webview. But your approach is already working in the LHS, so it would be better result.

hmhealey:
Actually, if that's possible by injecting JavaScript to look where the link is, that would be preferable, so I'll take a look at doing that. The approach in the LHS requires some more complicated logic and an extra library that adds custom right click menus that I'd like to get rid of if possible

PR #837, opened 2018-07-11
Update README.md

scherno2:
right -> Mattermost.exe --args --data-dir C:\my-mattermost-data
wrong -> Mattermost.exe -- --data-dir C:\my-mattermost-data

missing 'args'

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Configuration will be saved into Electron's userData directory:
 
 A custom data directory location can be specified with:
 
-* `Mattermost.exe -- --data-dir C:\my-mattermost-data` on Windows
+* `Mattermost.exe --args --data-dir C:\my-mattermost-data` on Windows
 * `open /Applications/Mattermost.app/ --args --data-dir ~/my-mattermost-data/` on macOS 
 * `./mattermost-desktop --args --data-dir ~/my-mattermost-data/` on Linux
 


yuya-oc:
Sorry for late reply, @scherno2. I think `--` is enough for working with arguments. Although should we use `--args` here?

scherno2:
well, the point is, its missleading:

**working**
`Mattermost.exe --args --data-dir C:\my-mattermost-data`
`Mattermost.exe --data-dir C:\my-mattermost-data`

**not working**
`Mattermost.exe -- --data-dir C:\my-mattermost-data`

i am no coder, but deployed mattermost-desktop at our company and failed at this part.
so i think other might have the same problem to understand why 
`Mattermost.exe -- --data-dir C:\my-mattermost-data`
is not working as expected,it will create the folder: 

`%AppData%\Mattermost`
`AppData\Roaming\Mattermost`

i am no further help but if you need the console output

[cmd_output.txt](https://github.com/mattermost/desktop/files/2224764/cmd_output.txt)

yuya-oc:
Much appreciate, @scherno2! Perhaps `--data-dir` might be a little different from other options I know.

For the log folder, probably we need to implement a new option to change it. (via [`app.setPath('logs', path)`](https://electronjs.org/docs/api/app#appsetpathname-path))

PR #864, opened 2018-09-22
Add flag for AppImage packaging

svelle:
**Summary**
Sets a flag in the electron builder that enables AppImage packaging.

**Additional Notes**
This would enable most linux users to simply run the application from the downloaded image and if wanted sets the necessary path variables to launch it from the preferred desktop environment. 
IMO it is better than distributing with a .deb or .rpm package, because it's distribution independent and comes with all needed dependencies, if any. Also it is just a minor change in packaging that enables a lot more users to comfortably install and use the app on their system.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -26,7 +26,8 @@
     "category": "Network;InstantMessaging",
     "target": [
       "deb",
-      "tar.gz"
+      "tar.gz",
+      "appimage"
     ],
     "extraFiles": [
       {


svelle:
If anyone wants to test this I can upload the AppImages i built or just clone the repository from my account.
Tested it successfully on Fedora 28 will test later today on Ubuntu 18.04.

ExNG:
Build and tested on Debian 9.0, works

hansendx:
Build and tested on Ubuntu 16.04, works

svelle:
Build and tested on Ubuntu 18.04. Works, too!

yuya-oc:
Thanks @svelle. Just curious, how popular are AppImage packages among Linux people? Unfortunately I have never used them in my usual desktop environment.

@jasonblais @esethna This would make easier to install the app on recent Linux. Should we officially support this?

svelle:
> Thanks @svelle. Just curious, how popular are AppImage packages among Linux people? Unfortunately I have never used them in my usual desktop environment.  

So they are pretty common. Some bigger projects use them like Krita, or LibreOffice and iirc Linus Torvalds also has some software that he packages as an AppImage. The consensus between bigger Linux Projects seems that this is a very much viable option to package and ship your product.   
 
I personally think it's a good option for projects like Mattermost because it's self contained and runs on pretty much every Linux distro because it is distro-agnostic and doesn't need any depencies. You can run it straight out of the box.   
Also it doesn't need any root permissions so every user can run it which can be especially useful for users who get a machine from their company on which they have no admin rights. Which for example would be the case at my workplace for ~90% of the users.  

From all the options out there, there are only two viable options for Mattermost, imo. One being AppImage and the other one being FlatPak. FlatPak would have the additional advantage of having the possibility to update directly from FlatHub. But for FlatPak to work the user needs to install the FlatPak depencies on their machine.   

I'm completely open for discussion on this subject and put this PR here mainly because it's such a simple change which can possibly make a lot of users happy without being much additional work for the maintainers. :)

yuya-oc:
Much appreciate! I haven't known about Linus's project, that's great news.

@esethna @jasonblais I think it's good to merge this because there is no change in program code and AppImage is not limited to specific Linux distribution. What we need to do is merely to release .appimage file.

esethna:
Thanks @svelle and @yuya-oc, I'm reaching out to the team to get some developer opinions on supporting this. A couple questions:

1. Compared to deb and tar.gz what would we expect the usage to be on the appimage package in terms of downloads?

2. If we did add the appimage package I would propose we offer it unofficially (similar to how we offer the deb packages currently), since it would be mostly community driven. Is there a large community around appimage to help support and maintain this package?

ExNG:
1. AppImages are portable they don't need to be installed and will run without any modification on any major distro, this would make Mattermost's Desktop App more accessible for a wider range of Linux distros.
2. Actually yes just search any major software that is available for Linux and add AppImage o the search, someone probably wrote a script to package it as AppImage, but for Mattermost there isn't much that need to be maintained, its just an entry in the `electron-builder.json`.

I would really like to see Mattermost to publish its desktop as AppImage, even if its unofficial. :smile:

svelle:
@ExNG summed it all up pretty good. For every bigger linux project there are often community built appimages and sometimes even official.
So adding this with an unofficial tag would probably be the best way to go. :)

esethna:
Great, thanks @svelle @ExNG, good info. 

@yuya-oc how do you feel about merging this and offering the appimage package as unofficial?

yuya-oc:
@esethna Sounds good to me.

SemaiCZE:
AppImage is a great start, but I'd really love to have a flatpak version which I could update with my other apps. Slack app is also on FlatHub, so packaging of Mattermost should not be that different and for many users it could simplify transition from Slack to Mattermost. Hope to get the flatpak version soon!

yuya-oc:
Thanks @SemaiCZE for suggestion. Currently we strongly depends on [electron-builder](https://www.electron.build/configuration/linux) in order to generate linux packages. It seems that flatpak is not supported. So we need to add flatpak support from zero if it's needed. Would you add another issue for that?

https://github.com/electron-userland/electron-builder/issues/3212

Issue #852, opened 2018-08-24
App can't cope with "mandatory ASLR" in Windows 10 1803


licaon-kter:
- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

https://mattermost.atlassian.net/browse/MM-14174

**Environment**
- Mattermost Desktop App version: 4.1.2

**Steps to reproduce**
* Toggle Mandatory ASLR -> ON
![aslr](https://user-images.githubusercontent.com/2282799/44573482-6fa44a80-a78f-11e8-99da-c35e963560f4.png)
* Reboot
* Try to install or use the app

**Expected behavior**
App runs

**Observed behavior**
Crash of the app/installer: “Aplication Error 0xc0000005"
Eg. https://forum.mattermost.org/t/windows-10-client-not-starting-on-windows-1803-version/5028

**Possible fixes**
Toggle it to the default (for now?!) OFF

yuya-oc:
Reproduced also for me. And not reproduced when the app is not code-signed. So it seems that the problem occurs when Windows apps are code-signed on Linux or Mac.
https://github.com/electron-userland/electron-builder/issues/3188

@esethna @jasonblais Perhaps we might need to build Windows apps on Windows machine. Currently we are using Linux.

yuya-oc:
I'll take a look whether I can test with self-signed certificates.

jasonblais:
Thanks Yuya (and @licaon-kter!)

@cpanato Thoughts on this issue from dev ops perspective? Should we build the Windows apps on a Windows machine? Do we have one available if we wanted to?

yuya-oc:
Not reproduced the build which is built on CircleCI (Linux) with my self-signed certificates. Perhaps our jenkins' wine env might lead this issue because it might be related to wine version. https://github.com/electron-userland/electron-builder/issues/3188#issuecomment-412773642

I'll continue to research.

licaon-kter:
[...]

PR #877, opened 2018-10-23
Test code-signing with self-signed certificates

yuya-oc:
**Summary**
Test code-signing with self-signed certificates.

This PR is only for testing code-signing on jenkins. So this should NOT be merged.

**Issue link**
#852


yuya-oc:
v4.2.0 was directly code-signed by electron-builder instead of osslsigncode. However the problem was not resolved.

PR #900, opened 2018-11-29
MM-13145 Some improvments to navigation and messaging.

crspeller:
**Summary**
- Improved the wording of the bad certificate dialog to make it scarier and harder to just allow.
- Adding `will-navigate` and `new-window` handling to prevent the bug where the app would navigate to a random page.
- Added context isolation.

**Issue link**
https://mattermost.atlassian.net/browse/MM-13145

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -311,7 +311,7 @@ app.on('certificate-error', (event, webContents, url, error, certificate, callba
     }
     dialog.showMessageBox(mainWindow, {
       title: 'Certificate Error',
-      message: 'There is a configuration issue with the Mattermost server you are trying to connect to, or someone is trying to intercept your connection. You also may need to sign into the Wi-Fi you are connected to using your web browser.',
+      message: 'There is a configuration issue with this Mattermost server, or someone is trying to intercept your connection. You also may need to sign into the Wi-Fi you are connected to using your web browser.',
       type: 'error',
       buttons: [
         'More Details',
@@ -326,7 +326,7 @@ app.on('certificate-error', (event, webContents, url, error, certificate, callba
           detail,
           type: 'error',
           buttons: [
-            'Continue Insecurely',
+            'Trust Insecure Certificate',
             'Cancel Connection',
           ],
           cancelId: 1,
--- a/src/main.js
+++ b/src/main.js
@@ -6,6 +6,8 @@
 import os from 'os';
 import path from 'path';
 
+import {URL} from 'url';
+
 import electron from 'electron';
 const {
   app,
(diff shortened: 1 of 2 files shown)


crspeller:
@jasonblais Do you mind taking a look at the text changes I made to the certificate error dialog?

jasonblais:
Actually, @esethna would you be able to help out on the text reviews for desktop app?

jasonblais:
Ping @esethna

src/main/mainWindow.js
```diff
@@ -67,6 +67,7 @@ function createMainWindow(config, options) {
 
   mainWindow.webContents.on('will-attach-webview', (event, webPreferences) => {
     webPreferences.nodeIntegration = false;
+    webPreferences.contextIsolation = true;
```
yuya-oc: We can't use the context isolation.

src/main.js
```diff
@@ -650,3 +664,29 @@ app.on('ready', () => {
   // Open the DevTools.
   // mainWindow.openDevTools();
 });
+
+app.on('web-contents-created', (dc, contents) => {
+  contents.on('will-attach-webview', (event, webPreferences) => {
+    webPreferences.nodeIntegration = false;
+    webPreferences.contextIsolation = true;
```
yuya-oc: We can't use the context isolation.

src/main/mainWindow.js
```diff
@@ -67,6 +67,7 @@ function createMainWindow(config, options) {
 
   mainWindow.webContents.on('will-attach-webview', (event, webPreferences) => {
     webPreferences.nodeIntegration = false;
```
yuya-oc: Now it's configured in `app.on('web-contents-created')` event. So I think this part is no longer needed.

crspeller: Maybe not, but I want to keep it just to make sure.

src/main/mainWindow.js
yuya-oc: As far as I know, there is no easy workaround. webapp invokes notifications by itself though, to detect that event, I think there is no other way than `Notification()` constructor. The largest reason to extend `Notification` is that.
If it's missing, the app can't activate the main window when a notification is clicked (by design of electron), also can't switch tabs to the server which invoked the notification. If there is a way to hook notification.onclick, it would be a workaround.

On the other hand, by adding another websocket connection, the app can know that new messages have arrived. In that case, the app would be able to invoke notification by itself. However we would need to add a mechanism to suppress notifications in webapp.

crspeller: Alright. Going to create a ticket to investigate this possibility. For now I have removed it from this PR.

crspeller:
@yuya-oc This is ready for re-review.

src/main/mainWindow.js
crspeller: https://mattermost.atlassian.net/browse/MM-13609

wget:
I am cleaning temporary branches to avoid clashes on AppVeyor. Feel free to reopen pull requests as dedicated **local** branches.

ikkisoft:
Not sure if this is usable for your preload, but take a look at https://github.com/MarshallOfSound/electron-valence for the _contextIsolation_ problem

crspeller:
Thanks for the suggestion. Added to related ticket: https://mattermost.atlassian.net/browse/MM-13609

Issue #902, opened 2018-11-30
4.2 build error - webpack-cli\bin\config-yargs.js


6a7070:
I confirm (by marking "x" in the [ ] below: [x]):

- [X ] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X ] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X ] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
When compiling the MM desktop app from source, the build process errors out. The error appears related to config-yargs.js file for webpack-cli.

**Steps to reproduce**
- Operating System: Windows 10
- Mattermost Desktop App version: 4.2.0
- Mattermost Server version: N/A
- Clear steps to reproduce the issue:
```
cd C:\temp\desktop-4.2.0\src
yarn
cd ..
npm run build
<errors out - npm run build does not complete. error output below>
```

**Expected behavior**
npm run build will successfully build program.

**Observed behavior**
```
C:\temp\desktop-4.2.0>npm run build

> mattermost-desktop@4.2.0 build C:\temp\desktop-4.2.0
> npm-run-all build:*


> mattermost-desktop@4.2.0 build:main C:\temp\desktop-4.2.0
> webpack-cli --bail --config webpack.config.main.js

C:\temp\desktop-4.2.0\node_modules\webpack-cli\bin\config-yargs.js:89
                                describe: optionsSchema.definitions.output.properties.path.description,
                                                                           ^

TypeError: Cannot read property 'properties' of undefined
    at module.exports (C:\temp\desktop-4.2.0\node_modules\webpack-cli\bin\config-yargs.js:89:48)
    at C:\temp\desktop-4.2.0\node_modules\webpack-cli\bin\webpack.js:60:27
    at Object.<anonymous> (C:\temp\desktop-4.2.0\node_modules\webpack-cli\bin\webpack.js:515:3)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
[...]

PR #901, opened 2018-11-29
Bump dependencies

wget:
Before submitting, please confirm you've
 - [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
 - [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
 - [x] executed `npm run lint:js` for proper code formatting

Please provide the following information:

Bump deps in order to fix build issues on GNU/Linux distribution or on Windows with latest nodejs.

**Test Cases**
Executed them, but they are still broken.

**Additional Notes**

Issue #898, opened 2018-11-26
Upgrade dependencies for v4.3


yuya-oc:
#### Electron
- [x] Electron and Spectron (#892)
- [x] electron-builder (#906)

#### Build tools
- [x] Babel (#903)
- [x] Webpack (#903)

#### Test tools
- [x] ESLint (#906)
- [x] Other test tools (#906)

#### Others
- [x] Other devDependencies (#906)

#### React
- [x] React (#906)
- [x] React-Bootstrap (#906)

#### Others
- [x] Other dependencies (#906)

### build system
- [x] Remove yarn ([like discussed on the chat one month ago](https://community.mattermost.com/core/pl/ufs3971omjbg5n8sqgxjqzqqwo))

wget:
I added a link to the conversation we had about removing the deprecated yarn build tool. ^

yuya-oc:
All dependencies were upgraded.

PR #892, opened 2018-11-13
Upgrade to Electron v3

yuya-oc:
Before submitting, please confirm you've
 - [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
 - [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
 - [x] executed `npm run lint:js` for proper code formatting

Please provide the following information:

**Summary**
Upgrade Electron to v3.0.9.

Sadly I could not fix all of tests now. They passed on my PC, but it seems that the app can't start well in CircleCI (or docker) environment.

**Issue link**
N/A

**Test Cases**
A. General usecases should work.
B. Deep linking should work. (There is a breaking change in Electron v3)

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/1137#artifacts

PR #903, opened 2018-12-01
Upgrade babel, webpack and storybook stack

yuya-oc:

**Summary**
Upgrade babel, webpack and storybook stack.

This is a part of #898 in the release process. Those tools have major update, so there are some changes in config files. This PR eventually solves #901 and #902.

**Issue link**
#898, #901 , #902 

**Test Cases**
These cases are checked on CircleCI.

- The app can be built from scratch: `npm run package:(windows|mac|linux)`
- Storybook can be built: `npm run storybook`

**Additional Notes**
https://circleci.com/gh/yuya-oc/desktop/1172#artifacts


package.json
```diff
@@ -26,28 +26,31 @@
...
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-react": "^6.24.1",
+    "babel-loader": "^8.0.4",
     "chai": "^4.1.2",
     "cross-env": "^5.1.6",
-    "css-loader": "^0.28.11",
+    "css-loader": "^1.0.1",
     "devtron": "^1.4.0",
     "electron": "2.0.12",
```
wget: Please make sure to use `^` for electron, otherwise you are asking for a particular version of electron and I cannot build against a system one.

yuya-oc: Well, I think a particular version will be eventually used by yarn.lock or package-lock.json anyway. Is that okay? I'm not sure "I cannot build against a system one." though, would you describe in detail?

This PR focuses on build tools stack only, so I will update #892 or make another PR.

wget: Actually if you are asking to use a specific version of electron, this means on GNU/Linux distributions, I need to builtin that specific electron version to make the app work and avoid version mismatches.
While at the contrary, if you use ^2.0.12, the latest 2.0.14 from the 2.x branch will be pulled (at this time 2.0.14) and I will be able to use version 2.0.14 from my system and depend on it rather than build it in the final application distribution package :)

yuya-oc: I think that means that untested version of electron will be used in the final packages though, is that acceptable? In past versions of electron and electron-builder, I had met that kind of trouble (in recent versions, actually it's a rare case).

wget: It's acceptable because with the new build system , I plan to run CI builds every day. So these "potential untested" electron version will be something that will be taken care of.

And of the record, I don't think the `^` operator takes betas version into account, only full/real release, but this needs to be checked out. Any idea about this topic?

yuya-oc:
Rebased to solve conflicts.

package.json
wget: @yuya-oc Indeed. Then no more objection. We can merge! :D

PR #911, opened 2018-12-13
build a portable Windows target

warren-bank:

**Summary**

* adds the windows build target: "portable"
* adds the dependency: [electron-portable-paths](https://github.com/warren-bank/electron-portable-paths)
* in [`src/main.js`](https://github.com/warren-bank/electron-mattermost-desktop/blob/PR-01-win-portable/src/main.js#L73-L79):
  * uses the new dependency around the code that checks `argv['data-dir']`
  * the purpose of the imported [function](https://github.com/warren-bank/electron-portable-paths/blob/master/src/index.js) is to:
    * detect whether the electron app is a windows portable build
    * if so: call `app.setPath()` such that the paths to all special directories are relative to the `.exe`

**Issue link**

_N/A_

although:
* it should be noted that the resulting portable .exe is effected by [this issue](https://github.com/electron/electron/issues/11858#issuecomment-366050737)
  * erroneously creates the empty directory: `%APPDATA%\Mattermost\logs`

**Test Cases**

* [release w/ binary builds](https://github.com/warren-bank/electron-mattermost-desktop/releases/tag/v4.2.0.05-win-portable)

**Additional Notes**

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -22,6 +22,7 @@ const {
 import isDev from 'electron-is-dev';
 import installExtension, {REACT_DEVELOPER_TOOLS} from 'electron-devtools-installer';
 import {parse as parseArgv} from 'yargs';
+import {makePortable, setPortablePaths} from '@warren-bank/electron-portable-paths';
 
 import {protocols} from '../electron-builder.json';
 
@@ -69,7 +70,11 @@ let permissionManager = null;
 const argv = parseArgv(process.argv.slice(1));
 const hideOnStartup = shouldBeHiddenOnStartup(argv);
 
-if (argv['data-dir']) {
+if (argv.portable) {
+  makePortable(app);
+}
+
+if (!setPortablePaths(app, true, argv['data-dir']) && argv['data-dir']) {
   app.setPath('userData', path.resolve(argv['data-dir']));
 }
 
@@ -77,7 +82,7 @@ global.isDev = isDev && !argv.disableDevMode;
 
 let config = {};
 try {
-  const configFile = app.getPath('userData') + '/config.json';
+  const configFile = path.resolve(app.getPath('userData'), 'config.json');
   config = settings.readFileSync(configFile);
   if (config.version !== settings.version) {
     config = settings.upgrade(config);
@@ -89,7 +94,7 @@ try {
   if (!config.teams.length && config.defaultTeam) {
     config.teams.push(config.defaultTeam);
 
-    const configFile = app.getPath('userData') + '/config.json';
+    const configFile = path.resolve(app.getPath('userData'), 'config.json');
     settings.writeFileSync(configFile, config);
   }
 }
@@ -98,7 +103,7 @@ if (config.enableHardwareAcceleration === false) {
 }
 
 ipcMain.on('update-config', () => {
-  const configFile = app.getPath('userData') + '/config.json';
+  const configFile = path.resolve(app.getPath('userData'), 'config.json');
   config = settings.readFileSync(configFile);
   if (process.platform === 'win32' || process.platform === 'linux') {
     const appLauncher = new AutoLauncher();
@@ -420,11 +425,11 @@ app.on('ready', () => {
 
   if (!config.spellCheckerLocale) {
     config.spellCheckerLocale = SpellChecker.getSpellCheckerLocale(app.getLocale());
-    const configFile = app.getPath('userData') + '/config.json';
+    const configFile = path.resolve(app.getPath('userData'), 'config.json');
     settings.writeFileSync(configFile, config);
   }
 
-  const appStateJson = path.join(app.getPath('userData'), 'app-state.json');
+  const appStateJson = path.resolve(app.getPath('userData'), 'app-state.json');
   appState = new AppStateManager(appStateJson);
   if (wasUpdated(appState.lastAppVersion)) {
     clearAppCache();
@@ -457,7 +462,7 @@ app.on('ready', () => {
 
   mainWindow = createMainWindow(config, {
     hideOnStartup,
-    linuxAppIcon: path.join(assetsDir, 'appicon.png'),
+    linuxAppIcon: path.resolve(assetsDir, 'appicon.png'),
     deeplinkingUrl,
   });
 
@@ -664,7 +669,7 @@ app.on('ready', () => {
   });
   ipcMain.emit('update-dict');
 
-  const permissionFile = path.join(app.getPath('userData'), 'permission.json');
+  const permissionFile = path.resolve(app.getPath('userData'), 'permission.json');
   const trustedURLs = settings.mergeDefaultTeams(config.teams).map((team) => team.url);
   permissionManager = new PermissionManager(permissionFile, trustedURLs);
   session.defaultSession.setPermissionRequestHandler(permissionRequestHandler(mainWindow, permissionManager));
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -77,6 +77,7 @@
   "win": {
     "target": [
       "nsis",
+      "portable",
       "zip"
     ],
     "extraFiles": [
@@ -90,5 +91,8 @@
   },
   "nsis": {
     "artifactName": "${name}-setup-${version}-win.${ext}"
+  },
+  "portable": {
+    "artifactName": "${name}-portable-${version}-win.${ext}"
   }
 }
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -26,6 +26,11 @@
         "regenerator-runtime": "^0.12.0"
       }
     },
+    "@warren-bank/electron-portable-paths": {
+      "version": "3.0.8",
+      "resolved": "https://registry.npmjs.org/@warren-bank/electron-portable-paths/-/electron-portable-paths-3.0.8.tgz",
+      "integrity": "sha512-RavDQfvE15fv20BB4WELMTpaD5Sh6Nhi57QdjD/Gp3cTLIJENpyOfiWNyZ2NUxlzIv+LN+198rblwHl4082neA=="
+    },
     "ansi-regex": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
--- a/src/package.json
+++ b/src/package.json
@@ -9,6 +9,7 @@
   "homepage": "https://about.mattermost.com",
   "license": "Apache-2.0",
   "dependencies": {
+    "@warren-bank/electron-portable-paths": "^3.0.8",
     "auto-launch": "^5.0.5",
     "bootstrap": "^3.3.7",
     "electron-context-menu": "^0.10.1",


package.json
```diff
@@ -73,5 +73,8 @@
     "webpack-cli": "^3.1.2",
     "webpack-dev-server": "^3.1.10",
     "webpack-merge": "^4.1.4"
+  },
+  "dependencies": {
+    "@warren-bank/electron-win-portable-paths": "^0.1.0"
```
yuya-oc: dependencies should be added to `src/package.json` because we use two package.json structure. And `src/yarn.lock` also should be committed to restrict package version.

yuya-oc:
@jasonblais The added module, electron-win-portable-paths is licensed under GPL-v2. In the past discussion, I remember we can't accept GPL because the desktop app is licensed under Apache-2.0. Is that right?

@warren-bank Please proceed the CLA if you could agree. However the desktop would have a license conflict. If possible, would you like to consider changing electron-win-portable-paths' license to another one which is compatible with Apache-2.0 (for example, MIT, Apache-2.0, MPL-2.0, etc.)?

jasonblais:
Big thanks @warren-bank for the contribution!

@yuya-oc is correct - projects with an APGL or GPL license may not be included in Apache v2 products. [More information can be found here](https://www.apache.org/legal/resolved.html#category-x).

Would you be open to changing the license of `electron-win-portable-paths` to MIT or other Apache-v2 compatible license. If you have any questions or concerns, let us know.

warren-bank:
@yuya-oc 
* hi
* sorry about adding the dependency to the wrong `package.json`
  * I'll fix that immediately
* regarding the license of the dependency library
  * I'll change it whatever you prefer
  * that being said, I do prefer GPL licenses for copyleft
  * since [GPL-v3 is compatible with Apache-2.0](https://www.apache.org/licenses/GPL-compatibility.html).. would that be OK for desktop?
* regarding the [CLA agreement](https://www.mattermost.org/mattermost-contributor-agreement/)
  * I don't mean to be difficult
    * after-all, this PR was only intended as a friendly _hello_
      * ..so I could refer to it when I apply for [a job](https://jobs.lever.co/mattermost/d3373f0b-9fb6-4856-9443-696277177b23) with Mattermost
  * I wanted to give it a quick glance and then sign, but..
    * the phrasing of the legalese appears (to me) to grants permissions that are far too broadly scoped
      * "3. Grant of Patent License. With respect to any patents you own, or that you can license without payment to any third party, you hereby grant Mattermost a perpetual, irrevocable, non-exclusive, worldwide, no-charge, royalty-free license to: (i) make, have made, use, sell, offer to sell, ..."
    * I just can't agree to that
      * the way I understand it:
        * this is __not__ limited in scope to the code in the contribution
  * for the sake of contrast
    * here is the [Perl Foundation's CLA](https://www.perlfoundation.org/contributor-license-agreement.html)
    * I would gladly agree to _these_ terms without any concern
  * that being said, all of the logic is contained in the dependency
    * written as such for the sake of reusability
    * there is no code in this PR that isn't also contained in the dependency repo
      * (ie: under _Usage_ in the _README_)
    * you're welcome to use the dependency
      * (after I change the license to something more compatible)
      * with or without this PR..
      * even if I ultimately do not agree to the terms of the CLA

warren-bank:
MIT for the win.. not a problem

wget:
@warren-bank I'm reading your code and I'm a bit concerned wrt. localized environments.

https://github.com/warren-bank/electron-win-portable-paths/blob/master/src/index.js#L15
You are looping though all user folder assuming they are actually called like that. What happens if these folders are localized? What happens if the users decided to rename the folder? This seems like a really fragile solution.

I found a great nodejs lib allowing to get the user special folders in an OS agnostic way: https://www.npmjs.com/package/platform-folders
This libs is making use of the native OS API for each OS (C++/ObjectiveC).
Migrating to this lib would even allow to have the portable version working on macOS and Linux as well.

In the current situation, I discourage the Mattermost Desktop's project manager to merge this PR.

warren-bank:
@wget 

* that library does look great, but..
  * its _raison d'etre_ is to determine the real/native paths to "special folders"
* in contrast..
  * the purpose of this little helper function is to loop through all of the [keys](https://electronjs.org/docs/api/app#appgetpathname) associated with "special folders"
    * remap the value __from__ the real/native path __to__ a localized path that is relative to directory in which the executable is found
    * thereby making it more portable
* that being said, you could easily debate whether __all__ "special folders" _should_ be remapped in this way
  * or, whether some of them _should_ be left alone
  * maybe an additional optional parameter to the function?:
    * a blacklist.. an array of keys to leave alone

warren-bank:
* regarding the idea of making this kind of behavior an option for other platforms
  * I was considering the same
  * I would need to run a quick test to see if this works.. but I was thinking along the lines of:

```javascript
// add to dependency module exports
export function make_portable(app) {
  if (process.env.PORTABLE_EXECUTABLE_DIR) return false

  const rootPath = app.getAppPath()
  process.env.PORTABLE_EXECUTABLE_DIR = rootPath
  return true
}
```

```javascript
// in 'src/main.js':
if (argv['portable'])
  make_portable(app)

if (!setPortablePaths(app, argv['data-dir']) && argv['data-dir']) {
  app.setPath('userData', path.resolve(argv['data-dir']));
}
```

warren-bank:
quick update..
* I added the blacklist parameter in v0.2.0
* changed module format from ES6 to CommonJS
  * wrote a few quick tests to run in node.. and node hates ES6 modules
  * won't matter to you, since Webpack likes both
* changed module export from a single function to an Object
  * currently only 1 key: `setPortablePaths`
  * may add `makePortable` later..
    * still need to test how that works
  * I'll __need to update the PR__ to reflect this breaking change..

warren-bank:
interesting..

I tried the following:

```javascript
const make_portable = function(app) {
  if (process.env.PORTABLE_EXECUTABLE_DIR) return false

  const rootPath = path.dirname(
    path.resolve(
      app.getPath('exe')
    )
  )
  process.env.PORTABLE_EXECUTABLE_DIR = rootPath
//process.stdout.write('portable dir: ' + process.env.PORTABLE_EXECUTABLE_DIR + "\n")

  return true
}

if (argv['portable']) {
  make_portable(app);
}

if (!setPortablePaths(app, argv['data-dir']) && argv['data-dir']) {
  app.setPath('userData', path.resolve(argv['data-dir']));
}
```

* I built the windows targets: `npm run "package:windows"`
* unzipped the non-portable `zip` target
* ran with switch: `Mattermost.exe --portable`
* observed the following:
  * the "portable" data directory is created correctly: `./Mattermost`
  * the electron app crashes with the message:
    ```text
      Failed to read or upgrade config.json { Error: ENOENT: no such file or directory, open 'C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\Mattermost/config.json'
          at Object.fs.openSync (fs.js:577:3)
          at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:166:20)
          at Object.fs.readFileSync (fs.js:483:33)
          at Object.fs.readFileSync (ELECTRON_ASAR.js:563:29)
          at Object.readFileSync (C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:12280:73)
          at Module.<anonymous> (C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:224:70)
          at __webpack_require__ (C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:20:30)
          at C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:84:18
          at Object.<anonymous> (C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:87:10)
          at Object.<anonymous> (C:\workspace\electron-mattermost-desktop\release\unzip\mattermost-4.3.0-develop-win-x64\resources\app.asar\main_bundle.js:16671:3)
        errno: -4058,
        syscall: 'open',
        code: 'ENOENT',
        path:
         'C:\\workspace\\electron-mattermost-desktop\\release\\unzip\\mattermost-4.3.0-develop-win-x64\\Mattermost/config.json' }
    ```
* what's _interesting_ is that:
  * the incorrect `path.sep` is being used by electron
    * when it constructs its path to `config.json`
  * this doesn't occur in the `portable` target
    * though in both cases, `rootPath` is filtered through `path.resolve`

warren-bank:
ohhh..
* [found it](https://github.com/warren-bank/electron-mattermost-desktop/blob/b00313d087c923355a2dbb5c45a54825edcab6da/src/main.js#L81)
* not a bug in electron

warren-bank:
oh geesh..

* it must be because HEAD is unstable
  * the portable version crashes with the same error
* the version built from the last stable release (v4.2.0) runs without error

yuya-oc:
After running the portable app, I got following folder structure. There are user folders (`documents`, `desktop`, ...) and internal data files/folders (`config.json`, `Cache/`, ...) in the same level without distinction. It would seem that they are not organized for ordinary users. 

```
.
├── Mattermost/
│   ├── Cache/
│   ├── GPUCache/
│   ├── blob_storage/
│   ├── desktop/
│   ├── documents/
│   ├── downloads/
│   ├── logs/
│   ├── music/
│   ├── pictures/
│   ├── temp/
│   ├── videos/
│   ├── webrtc_event_logs/
│   ├── app-state.json
│   ├── bounds-info.json
│   └── config.json
└── mattermost-4.3.0-develop-portable.exe
```

So I think they are divided into more than one folder. Just for example, what about following structure?

```
.
├── Mattermost/
│   ├── data/
│   │   ├── Cache/
│   │   ├── GPUCache/
│   │   ├── blob_storage/
│   │   ├── temp/
│   │   ├── webrtc_event_logs/
│   │   ├── app-state.json
│   │   ├── bounds-info.json
│   │   └── config.json
│   ├── logs/
│   └── home/
│       ├── desktop/
│       ├── documents/
│       ├── downloads/
│       ├── music/
│       ├── pictures/
│       └── videos/
└── mattermost-4.3.0-develop-portable.exe
```

By the way, I did read the readme of electron-portable-paths though, I could not understand what will happen by "remapping" folders until I read its code. If it's well-documented, it would be very helpful in order to understand what the module will do.

src/main.js
```diff
@@ -69,15 +70,19 @@ let permissionManager = null;
 const argv = parseArgv(process.argv.slice(1));
 const hideOnStartup = shouldBeHiddenOnStartup(argv);
 
-if (argv['data-dir']) {
+if (argv['portable']) {
```
yuya-oc: ESLint error: ["portable"] is better written in dot notation (dot-notation)

`argv['portable']` -> `argv.portable`

src/main.js
```diff
@@ -69,15 +70,19 @@ let permissionManager = null;
...
+    makePortable(app);
```
yuya-oc: ESLint error: Expected indentation of 2 spaces but found 4 (indent)

warren-bank:
all good points.. thanks for the feedback; I'll make those changes by end of day.

warren-bank:
all suggested edits have been pushed to github.. please let me know what you think :)

yuya-oc:
I think now it's ready to test. Artifacts are available at https://circleci.com/gh/mattermost/desktop/1773#artifacts

On macOS, the application crashed.

```
$ ~/Downloads/Mattermost.app/Contents/macOS/Mattermost --portable
Failed to read or upgrade config.json { Error: ENOTDIR: not a directory, open '/Users/yuya-oc/Downloads/Mattermost.app/Contents/MacOS/Mattermost/data/config.json'
    at Object.fs.openSync (fs.js:577:3)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:166:20)
    at Object.fs.readFileSync (fs.js:483:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:563:29)
    at Object.readFileSync (/Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:12434:73)
    at Module.<anonymous> (/Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:210:70)
    at __webpack_require__ (/Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:20:30)
    at /Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:84:18
    at Object.<anonymous> (/Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:87:10)
    at Object.<anonymous> (/Users/yuya-oc/Downloads/Mattermost.app/Contents/Resources/app.asar/main_bundle.js:16825:3)
  errno: -20,
  syscall: 'open',
  code: 'ENOTDIR',
  path:
   '/Users/yuya-oc/Downloads/Mattermost.app/Contents/MacOS/Mattermost/data/config.json' }
```

On macOS, `.app` is a directory to bundle an app. And the actual executable file is located under `.app/Contents/macOS/`. So `app.getPath('exe')` would return it. This causes missing `Contents/MacOS/Mattermost/data/config.json`. And that will also break the code-signing verification because not-signed files will be located under `Mattermost.app`.

To provide same usability on all platforms, I think some tweaks would be needed for each platform, macOS and Linux. (Or, apply the portable feature only for Windows).

warren-bank:
I think this update should solve the issue with macOS bundles.. it adds 2 new things

1. special handler to detect a path inside a macOS bundle.. and modify the "portable" path accordingly
2. a file system check that the resulting "portable" path can be written to.. otherwise it is ignored and no "mapping" to it occurs
   * this check is platform agnostic

However, I don't have a mac on-hand to build or test this workaround. I'm really sorry to ask..

warren-bank:
I'll need to test the AppImage you built; if I remember right, it creates a sandbox that would have the same name as our "portable" directory..

warren-bank:
..pushed an update that _should_ work with AppImage
* my AppImage build target fails.. something about an HTTP2 connection timeout
* I'll need to get that working so I can test the effect of the update
  * maybe if I try building in Linux.. I've been cross-compiling from Windows

warren-bank:
current commit:
* works w/ AppImage (on Linux)
* _should_ work (but untested) w/ MacOS app bundle

yuya-oc:
Thanks @warren-bank and sorry for late response.

I tested Mac version of `--portable` flag. It looks good to me.

```shell
$ ./Mattermost.app/Contents/MacOS/Mattermost --portable
```

```
.
├── Mattermost/
│   ├── data/
│   ├── home/
│   └── logs/
└── Mattermost.app/
```

yuya-oc:
On Linux, I got following results.

**tar.gz**:
```
.
└── mattermost-desktop-4.3.0-develop-linux-x64/
    ├── Mattermost/
    └── mattermost-desktop*

(unrelated files are omitted)
```

**AppImage**:
```
.
├── Mattermost/
└── mattermost-desktop-4.3.0-develop-linux-x86_64.AppImage*
```

When using tar.gz version, the data directories are located under the extracted app directory. Sadly I can't say I'm familiar at Linux desktop apps though, is that suitable for Linux users?

yuya-oc:
Artifacts: https://circleci.com/gh/mattermost/desktop/1784#artifacts

warren-bank:
Hi @yuya-oc , happy holidays :)

Personally, this is the behavior that I would expect.. but I can't speak for anyone else.

My expectation is based on that it's similar to the conventions used by: PortableApps (windows), ThinApp portable sandbox (windows), AppImage special directories (linux, ex: `foo.AppImage.home/`,` foo.AppImage.config/`)...

What's nice is that it plays well with your existing command-line flag `--portable --data-dir "/path/to/different/portable/directory/for/Mattermost"` ..so whatever behavior is chosen for default can be modified by the end user

warren-bank:
It should be noted that when I tested `.deb`.. I ran into one issue.

If the package is installed by root: `sudo apt-get install ./mattermost.deb`.. then the portable directory can only be created when `mattermost-desktop --portable` is run by root, rather than a regular user. This is due to write permissions on the directory in which the `mattermost-desktop` executable is contained.

If the portable "Mattermost" directory is explicitly created at install and its permissions changed (777), then any user can write to it. But this extra step is not obvious.

If it doesn't exist and `mattermost-desktop --portable` is run as a regular user, then the `--portable` flag will effectively be ignored.

warren-bank:
Do you think it's fair to assume that when an installer is used (ex: windows setup, linux deb, etc).. portability is no-longer really a consideration?

The reasoning behind such an assumption is based on the fact that installers are inherently non-portable and modify the base OS.

Furthermore, the executable is ultimately installed to a path that the user cannot configure or change (without quite a bit of effort).
* windows setup:
  * `%LOCALAPPDATA%\Mattermost\mattermost.exe`
* linux deb:
  * `/opt/Mattermost/mattermost-desktop`

warren-bank:
[this](https://github.com/electron-userland/electron-builder/issues/2050#issuecomment-328321704) looks like a better solution..
* a post-install script for `.deb` that could:
  ```bash
    mkdir './Mattermost'
    chmod 777 './Mattermost'
   ```

warren-bank:
quick note:
* I tested adding bash scripts that run when the package manager installs/removes the .deb target
  * works great
  * would you like me to commit that feature?

yuya-oc:
Thanks for many feedback!

For the default directory for portable option, I agree to the design.

On the other hand, personally I think we don't have to consider about `--portable` if the package is installer target (windows setup, deb, etc.). They are installed to system via platform's mechanism, so it would be all right if config files are saved to user's directory. And the motivation to use `--portable` on their machine would be lesser than portable target.
Especially on deb, `/opt/` will be shared by all users on a machine. But `data/` contains session cookies, so it should not be recommended due to security reason.

By above reason, I feel post-install script would not be needed for us.

> I really wish electron-builder left some form of breadcrumb to enable the app to identify the build target at runtime.

100% agree.

warren-bank:
> I really wish electron-builder left some form of breadcrumb to enable the app to identify the build target at runtime.

doesn't hurt [to ask](https://github.com/electron-userland/electron-builder/issues/3579).. :)

wget:
> Do you think it's fair to assume that when an installer is used (ex: windows setup, linux deb, etc).. portability is no-longer really a consideration?
> 
> The reasoning behind such an assumption is based on the fact that installers are inherently non-portable and modify the base OS.

I agree with this point of view. Using an app installed via an installer could reasonably be considered as integrated to the system and therefore be not portable per definition.

**HOWEVER** .deb/packages are not considered as an installer per se. For Debian or Arch Linux, we might have a `mattermost-desktop-portable` version being installable via package manager but installed locally as a user. But I think it's the maintainer of such a package to figure this out, not our job.

> 
> Furthermore, the executable is ultimately installed to a path that the user cannot configure pre-installation, or change (without quite a bit of effort) post-installation.

This is not true any more. New msi installer will allow the end user to administratively install the application wherever he wants.
> 
> * windows setup:
>   
>   * `%LOCALAPPDATA%\mattermost\Mattermost.exe`

NSIS based installers won't be provided any more. The default install path will be in `Program Files`. But again this can be changed by the admin user.

> * linux deb:
>   
>   * `/opt/Mattermost/mattermost-desktop`

This is not exact science either. When we package the app for Linux we assume it will be installed in `/opt` as per the Linux File Hierarchy standard. However, the package maintainers are free to integrate the app further in the system. [This is what I do for Arch Linux](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mattermost-desktop&id=6612156ad04080865270e6d254b321f0e326a808#n64)

> 
> I really wish `electron-builder` left some form of breadcrumb to enable the app to identify the build target at runtime. Although, I'm honestly not sure what I would do with that information:
> 
> * ignore the `--portable` option for certain targets?
>   
>   * not a great solution either..

wget:
> quick note:
> 
> * I tested adding bash scripts that run when the package manager installs/removes the .deb target
>   
>   * works great
>   * would you like me to commit that feature?
>     
>     * [here](https://github.com/warren-bank/electron-starter-app/commit/11cbe43aea5e0c50be735f4342a5e0137c44391d) is a commit to a different repo that contains this feature (along with a few other unrelated tweaks).. should give you a good idea of how it wires together

I'm **completely** against this feature. Like explained above, this is not your job to trigger such a string on the end user's computer.

warren-bank:
> When the [new dependencies upgrade](https://github.com/mattermost/desktop/issues/898) has been merged, could we rebase &hellip; and retest everything carefully

sounds like a plan :+1: 

&hellip;please bump this thread again when there's a commit that's ready to use for a new base

yuya-oc:
@warren-bank Sorry for late response. Now #898 are solved, so it's ready to continue.

warren-bank:
note to self.. git cheatsheet:

```bash
git remote add mattermost 'https://github.com/mattermost/desktop.git'

git checkout master
git pull --ff-only mattermost master

git checkout 'PR-01-win-portable'
git branch 'PR-01-prebase-win-portable'

# save a reference to the non-squashed PR branch
git checkout 'PR-01-prebase-win-portable'
git push -u origin 'PR-01-prebase-win-portable'
git checkout 'PR-01-win-portable'

git reset --hard 5abe83b1e620ef69f0f2e2a1e923160ba6ccec89
git merge --no-commit --squash 'PR-01-prebase-win-portable'
git mergetool

# update "package-lock.json"
cd src
npm install
cd ..

# inspect updates before commit
git diff --no-ext-diff 5abe83b1e620ef69f0f2e2a1e923160ba6ccec89 >../diff.txt

git add --all .
git commit -m 'add "--portable" CLI option, add Windows "portable" target'

# update remote branch (changes pull request history)
git push -f origin 'PR-01-win-portable'
```

jasonblais:
@yuya-oc @wget Are there next steps for this PR?

hanzei:
@deanwhillier Could you help with next steps on this PR?

hanzei:
@deanwhillier Gentle reminder about the questions above

Willyfrog:
closing as codebase has changed way too much, will gladly help anyone who want to take on the effort of building such a target

PR #928, opened 2019-02-11
GPO: Configure server addresses and prevent adding new servers

JtheBAB:
This pull request add the GPO functionality to

- configure the server addresses
- control if the + for adding more server is available or not

Target for 4.3, if not we need to change slightly this policy before merge.


esethna:
@wget @yuya-oc any issues with merging this PR?

PR #945, opened 2019-03-08
MM 7970: Tap into system events and idle time to affect user online status based on user activity within the host OS

deanwhillier:

**Summary**
Implements  Electron system idle detection ([PowerMonitor](https://electronjs.org/docs/api/power-monitor#powermonitor)) to allow the Desktop app to maintain online status while the app is unfocused, but the user is still active on their computer.

Desktop app maintains an internal inactivity clock with a threshold of 5 minutes, during which updates are sent to the Webapp indicating the user is active. Once the clock exceeds the threshold, updates to the Webapp indicate the user is inactive. System events like screensaver, shutdown, lock, unlock etc. affect status sent to the Webapp.

This PR is part of several related PR's:
Desktop: This PR
Webapp: [PR 2466](https://github.com/mattermost/mattermost-webapp/pull/2466)
Redux: [PR 795](https://github.com/mattermost/mattermost-redux/pull/795)
Server: [PR 10469](https://github.com/mattermost/mattermost-server/pull/10469)

**Issue link**
[MM-7970](https://mattermost.atlassian.net/browse/MM-7970), supported by [MM-14582](https://mattermost.atlassian.net/browse/MM-14582)

**Test Cases**
1. Compile and run the Desktop app.
2. Run a local version of Webapp from PR linked above.
3. Add the local version of the Webapp to the Desktop app.
4. Leave the desktop app visible but in the background for ~10mins while 'using' your computer and observe the status indicator. (It should remain in the online state for the entire time)
5. Leave the desktop app visible but in the background for ~10mins while NOT using your computer and observe the status indicator. (It should go into away mode somewhere around the 6 minute mark).
6. Click on the desktop app to make it active again. Observe the status indicator, it should return to online status.

**Additional Notes**
This is the first phase of a potentially more involved implementation that will include the Webapp properly responding to inactive status caused by system events like shutdown, lock, unlock, screen saver etc. and respond more immediately to status updates.


src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+  // - updates internal idle time properties and conditionally triggers various tasks based on those values
+  updateIdleTime(idleTime) {
+    // store last and this idleTime for comparison
+    this.lastIdleTime = this.idleTime;
+    this.idleTime = idleTime;
+
+    // provide frequent activity updates
+    this.emit('update', {userIsActive: this.isActive, idleTime: this.idleTime});
+
+    if (this.isActive && this.idleTime > this.config.userActivityTimeout) {
```
lieut-data: I think these checks would be easier if we dispense with the `this.isActive` checks -- we already check the value in `updateUserActivityStatus` to avoid emitting spurious events anyway.

The last else condition, then, could be moved into `updateUserActivityStatus` instead, having it emit `sendStatusUpdate` /anyway/ even if the value hasn't changed. Or even `sendStatusUpdate` itself could debounce itself to avoid anyone calling it more often than needed?

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+      // update activity status to inactive after defined time
+      this.updateUserActivityStatus(false);
+    } else if (!this.isActive && !this.forceUserIsInactive && this.idleTime < this.config.userActivityTimeout && this.lastIdleTime > this.idleTime) {
```
lieut-data: It's a little unclear why we check `this.lastIdleTime > this.idleTime` -- isn't it enough that `idleTime < userActivityTimeout`?

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+    } else if (new Date().getTime() - this.lastStatusUpdate > this.config.statusUpdateThreshold) {
+      // provide periodic status updates
+      this.sendStatusUpdate();
+    }
+  }
+
+  // - updates user activity status if changed and triggers a status update
+  updateUserActivityStatus(userIsActive = false) {
+    if (userIsActive !== this.isActive) {
+      this.isActive = userIsActive === true;
```
lieut-data: if `userIsActive` is a boolean, there's no need to test `=== true` here. (If it's just a truthy value, this would faily anyway.)

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+      // update activity status to active if system idle time resets
+      this.updateUserActivityStatus(true);
+    } else if (new Date().getTime() - this.lastStatusUpdate > this.config.statusUpdateThreshold) {
+      // provide periodic status updates
+      this.sendStatusUpdate();
+    }
+  }
+
+  // - updates user activity status if changed and triggers a status update
+  updateUserActivityStatus(userIsActive = false) {
```
lieut-data: nit: parameter here could just be `isActive`, and it won't conflict with `this.isActive`.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+
+  // - sends an update with user activity status and current system idle time
+  sendStatusUpdate() {
+    this.lastStatusUpdate = new Date().getTime();
+    const status = {
+      userIsActive: this.isActive,
+      idleTime: this.idleTime,
+    };
+    if (this.callback) {
+      this.callback(status);
```
lieut-data: It doesn't seem like we actually use `callback` -- and this is an event emitter anyway. Should we just dispense with the `callback` and rely on the emitter semantics instead?

src/browser/webview/mattermost.js
```diff
@@ -194,6 +194,18 @@ function setSpellChecker() {
...
 ipcRenderer.on('set-spellchecker', setSpellChecker);
 
+ipcRenderer.on('user-is-active', () => {
+  if (window.desktopBridge && window.desktopBridge.updateUserActivityStatus) {
+    window.desktopBridge.updateUserActivityStatus(true);
+  }
+});
+
+ipcRenderer.on('user-is-inactive', () => {
+  if (window.desktopBridge && window.desktopBridge.updateUserActivityStatus) {
```
lieut-data: I know I'm biased here, but 🎉

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+import EventEmitter from 'events';
+
+import electron from 'electron';
+const {app} = electron;
+
+export default class UserActivityMonitor extends EventEmitter {
+  constructor(callback) {
+    super();
+
+    // store callback for later use
```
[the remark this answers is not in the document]
deanwhillier: Cleaned up all comments with ESDoc comments.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+    this.isActive = true;
+    this.forceUserIsInactive = false;
+    this.idleTime = 0;
+    this.lastIdleTime = 0;
+    this.lastStatusUpdate = 0;
+    this.systemIdleTimeIntervalID = -1;
+
+    // initialize config objec
+    this.config = {
+      regularUpdateFrequency: 1000, // every 1 second by default (in milliseconds)
```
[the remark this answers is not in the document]
deanwhillier: Good call on the naming! I actually had calculated out the values to begin with, but changed after a lint warning ... put back and am ignoring the warning.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+      statusUpdateThreshold: 60000, // every 60 seconds by default (in milliseconds)
+      userActivityTimeout: 300, // after 5 minutes by default (in seconds)
+    };
+
+    // create system event handlers
+    this.handleSuspend = () => {
```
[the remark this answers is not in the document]
deanwhillier: Got a weird error from Electron when I tried and used this approach to get around. Have now solved using binding instead.

src/main/UserActivityMonitor.js
deanwhillier: Good point, thx ... done.

src/main/UserActivityMonitor.js
deanwhillier: I was following a pattern already used in the desktop app, but you are correct, it's not really necessary. Removed.

src/browser/webview/mattermost.js
deanwhillier: Redid the bridge to mimick the existing functionality of Electron's ipcMain and ipcRenderer modules already used for this feature.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+    return this.forceUserIsInactive ? false : this.isActive;
+  }
+  get userIdleTime() {
+    return this.idleTime;
+  }
+
+  // class methods
+  // - begin monitoring system events and idle time at defined frequency
+  startMonitoring(config = {}) {
+    if (app.isReady()) {
```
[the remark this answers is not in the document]
deanwhillier: Done and done.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+        }
+      }, this.config.regularUpdateFrequency);
+    } else {
+      this.emit('error', new Error('UserActivityMonitor.startMonitoring can only be called after app is ready'));
+    }
+  }
+
+  // - stop monitoring system events and idle time
+  stopMonitoring() {
+    // stop monitoring system events
```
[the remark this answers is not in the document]
deanwhillier: As above, cleaned up with ESDoc comments

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+    electron.powerMonitor.off('unlock-screen', this.handleUnlockScreen);
+
+    // stop monitoring system idle time
+    clearInterval(this.systemIdleTimeIntervalID);
+  }
+
+  // - updates internal idle time properties and conditionally triggers various tasks based on those values
+  updateIdleTime(idleTime) {
+    // store last and this idleTime for comparison
+    this.lastIdleTime = this.idleTime;
```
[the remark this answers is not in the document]
deanwhillier: Not really needed for this initial version anymore, so removed.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+'use strict';
+
+import EventEmitter from 'events';
+
+import electron from 'electron';
+const {app} = electron;
+
+export default class UserActivityMonitor extends EventEmitter {
```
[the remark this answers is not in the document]
deanwhillier: Added ESDoc comments.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // - updates internal idle time properties and conditionally triggers various tasks based on those values
+  updateIdleTime(idleTime) {
+    // store last and this idleTime for comparison
+    this.lastIdleTime = this.idleTime;
+    this.idleTime = idleTime;
+
+    // provide frequent activity updates
+    this.emit('update', {userIsActive: this.isActive, idleTime: this.idleTime});
```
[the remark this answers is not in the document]
deanwhillier: Was left for possible flexibility, but its really not necessary so I removed it.

src/main/UserActivityMonitor.js
deanwhillier: Reorganized, thanks.

deanwhillier:
Thanks @lieut-data! The bridge has been rewritten to support 2 way communication and is patterned after the Electron ipcMain/ipcRenderer modules already in use on the desktop side of this feature.

I have not forgotten about tests – I just haven't had time to figure out how tests work in the desktop app yet. Am still going to try add some as suggested.

Apologies for the force push, I had some trouble with tests failing in areas I didn't touch.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,170 @@
...
+      idleTime: this.idleTime,
+    });
+  }
+
+  /**
+   * System event handlers
+   *
+   * @private
+   */
+  handleSuspend() {
```
hmhealey: It seems like there's only 2 different behaviours here. Would it make sense to just have two methods that are reused for different events? 0/5

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,170 @@
...
+   * Updates user activity status if changed and triggers a status update
+   *
+   * @param {boolean} isActive
+   * @private
+   */
+  updateUserActivityStatus(isActive = false) {
+    if (isActive !== this.isActive) {
+      this.isActive = isActive;
+      this.sendStatusUpdate();
+    } else if (new Date().getTime() - this.lastStatusUpdate > this.config.statusUpdateThresholdMs) {
```
hmhealey: `new Date().getTime()` could be shortened to `Date.now()`.

Also, is it possible that the "now" time here will be different than the "now" time in `sendStatusUpdate`? Perhaps we should get that time once and pass it as an argument to `sendStatusUpdate`. 0/5 on that though since I think it's a good practice, but I don't see any real issues that could result from it.

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,170 @@
...
+
+    this.handleSuspend = this.handleSuspend.bind(this);
+    this.handleResume = this.handleResume.bind(this);
+    this.handleShutdown = this.handleShutdown.bind(this);
+    this.handleLockScreen = this.handleLockScreen.bind(this);
+    this.handleUnlockScreen = this.handleUnlockScreen.bind(this);
+  }
+
+  get userIsActive() {
+    return this.forceUserIsInactive ? false : this.isActive;
```
hmhealey: 0/5 since it's totally nitpicking code styles, but I'd tend towards doing this without a ternary statement like
```suggestion
    return this.isActive && !this.forceUserIsInactive;
```

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,170 @@
...
+    this.lastStatusUpdate = 0;
+    this.systemIdleTimeIntervalID = -1;
+
+    this.config = {
+      regularUpdateFrequencyMs: 1 * 1000,
+      statusUpdateThresholdMs: 5 * 1000,
+      userActivityTimeoutSec: 5 * 60,
+    };
+
+    this.handleSuspend = this.handleSuspend.bind(this);
```
hmhealey: Do these need to be bound manually? I was under the assumption that we only did that in the apps because of how `React.Component` is written internally. Alternatively, we could define the methods as arrow functions since that makes manual binding unnecessary.

src/browser/js/DesktopBridge.js
```diff
@@ -0,0 +1,66 @@
...
+export default class DesktopBridge extends EventEmitter {
+  constructor() {
+    super();
+
+    this.ready = false;
+
+    this.desktop = null;
+    this.webapp = null;
+
+    window.desktopBridge = this;
```
hmhealey: Why use a global instance of the `DesktopBridge`? Couldn't this be something that's owned by the `MattermostView`?

src/browser/js/DesktopConnector.js
```diff
@@ -0,0 +1,32 @@
...
+ * Class to handle interfacing with the loaded webapp
+ */
+export default class DesktopConnector extends EventEmitter {
+  /**
+   * Sends data to the webapp for the provided event
+   * @param {string} event
+   * @param  {...any} data
+   */
+  send(event, ...data) {
+    if (window.desktopBridge && window.desktopBridge.ready) {
```
hmhealey: Similarly, instead of using the global instance of `DesktopBridge` here, couldn't the `DesktopBridge` just pass a reference to itself when creating the `DesktopConnector` like `this.registerDesktopConnector(new DesktopConnector(this));`? Then the `DesktopConnector` would be able to refer back to the bridge.

src/browser/js/DesktopBridge.js
deanwhillier: I believe we resolved this question when we talked last week?

src/main/UserActivityMonitor.js
deanwhillier: Without the binding, errors are thrown and fat arrow functions aren't supported by the current setup.

deanwhillier: Added inline comment to that effect

src/main/UserActivityMonitor.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    if (this.systemIdleTimeIntervalID >= 0) {
+      this.emit('error', new Error('User activity monitoring is already in progress'));
+      return;
+    }
+
+    this.config = Object.assign({}, this.config, config);
+
+    // NOTE: electron.powerMonitor cannot be referenced until the app is ready
```
hide212131: Hi. I am looking forward to releasing this feature.
As stated in the comments, isn't it necessary?
In my environment, I get an error without it.
```
// NOTE: electron.powerMonitor can not be referenced until the app is ready
electron.app.on ('ready', () => {
  electron.powerMonitor.on('suspend', this.handleSuspend);
  ...
})
```

deanwhillier:
Closing to submit new PR ... lots has happened since this original.

src/main/UserActivityMonitor.js
deanwhillier: Got the app running locally now in the background while I work the new PR for this and I'm staying active the whole time! 🙌

Yes, you are absolutely correct! 🙂 I put the comment in there to remind that the _startMonitoring_ method needs to be called once the app is ready because of the use of _electron.powerMonitor_ or you get the error you mentioned. In _main.js_, the _startMonitoring_ method is called after the app is ready. If the method happens to be called before the app is ready, there is a conditional at the top of the method to emit an error and return without starting the monitoring.

PR #946, opened 2019-03-08
MM-14446: consider subpath when evaluating if url is internal

lieut-data:
**Summary**
When clicking on an URL with `target=_blank`, the webview decides if it should launch an external browser or a new window within the Electron application. Update this logic to consider the application's configured subpath so as to treat links outside the subpath but on the same domain as external.

**Issue link**
https://mattermost.atlassian.net/browse/MM-14446

**Test Cases**
Added unit tests for helper function that tests if the target URL is internal. Tested with a subpath configured and without, verifying that links to different domains and different subpaths on the same domain are handled externally in the live build.


test/specs/utils/util_test.js
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2015-2016 Yuya Ochiai
```
crspeller: ?

lieut-data: I admit I don't know the policy here, having just copied the license header from another file. Should I omit @yuya-oc's copyright on this file?

crspeller: Well he didn't write it. No idea why it's on the others. I think legally it's fine because the CLA doesn't sign over copyright just an indefinite license.

lieut-data: Ah, now I know why:
```
/root/mattermost-desktop/test/specs/utils/util_test.js
  3:1  error  incorrect header  header/header
```

Fixing this by making the `header/header` rule apply to a hard-coded list of files, and going forward will only require the mattermost license on new files.

esethna:
Adding HH for a second review before merge (as Yuya is not very available this month)

esethna:
Given customer urgency we'll take this for a v4.2 dot release

hmhealey:
I just left this as ready to merge since I don't know the process for merging this for a 4.2 dot release

Issue #863, opened 2018-09-13
Mattermost opens both on fullscreen and on a smaller window when closing in fullscreen


carlosmcgregor:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

https://mattermost.atlassian.net/browse/MM-14176

**Summary**
When app is closed in fullscreen, opening it again leads to a fullscreen version _and_ a smaller window version. You cannot access the fullscreen version as it throws you again to the smaller window.

**Environment**
- Operating System: macOS High Sierra Version 10.13.6
- Mattermost Desktop App version: 4.1.2 (4.1.2.46)
- Mattermost Server version: 5.2.1

**Steps to reproduce**
1. Open Mattermost and make it fullscreen (green button on top left of window)
2. Press COMMAND + Q to quit Mattermost
3. Open Mattermost

**Expected behavior**

Mattermost opens either on fullscreen _or_ as a smaller window in a desktop.

**Observed behavior**

Mattermost opens both on fullscreen _and_ as a smaller window in a desktop, with the fullscreen being inaccessible.

<img width="365" alt="Desktop and Fullscreen Mattermost" src="https://user-images.githubusercontent.com/16442174/45503085-cfb95b80-b753-11e8-83c9-f246907b5e23.png">

**Possible fixes**

yuya-oc:
Sadly it does not reproduce for me (macOS 10.13.6, desktop app v4.1.2).

- Does it always happen by following the steps to reproduce?
- Would you try removing `~/Library/Application Support/Mattermost/bounds-info.json`?

carlosmcgregor:
* It always happens by following the steps to reproduce.
* I tried removing `~/Library/Application Support/Mattermost/bounds-info.json` and that forces Mattermost to open in windowed mode. However, the error persists when I attempt to recreate it. 

For reference, this is what I see in `bounds-info.json`:
```
{"x":0,"y":0,"width":1440,"height":900,"maximized":false,"fullscreen":true}
```

carlosmcgregor:
[...]

Issue #538, opened 2017-05-19
Don't restart the app in full screen mode


jasonblais:
I confirm (by marking "x" in the [ ] below): 

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md). 

---

Jira ticket: https://mattermost.atlassian.net/browse/MM-14097

**Summary**
Don't restart the app in full screen mode.

**Steps to reproduce**

Windows 10; Mattermost Desktop App v3.7; Mattermost Server v3.9; 

1) Open the app
2) Go to **View > Toggle Full Screen**
3) Quit the app, then restart it

**Expected behavior**
The app is not in full screen mode after restarting.

**Observed behavior**
The app is in full screen mode after restarting

amyblais:
Ticket for tracking: https://mattermost.atlassian.net/browse/MM-14097

PR #954, opened 2019-03-21
[MM-14176] Fix restart after quitting in full screen

deanwhillier:

**Summary**
Removes logic that saves fullscreen state when quitting, allowing the app to always open up in windowed mode.

**Issue link**
Github: #863
Jira: [MM-14176](https://mattermost.atlassian.net/browse/MM-14176)

**Additional Notes**
I was unable to replicate the exact issue described in the ticket in Mac OS Mojave, however, there was still some odd behaviour for me (App went fullscreen, but didn't get focus). This fix corrects that odd behaviour and should also correct the High Sierra issue.

### DIFF
--- a/src/main/mainWindow.js
+++ b/src/main/mainWindow.js
@@ -9,7 +9,6 @@ import {app, BrowserWindow} from 'electron';
 function saveWindowState(file, window) {
   const windowState = window.getBounds();
   windowState.maximized = window.isMaximized();
-  windowState.fullscreen = window.isFullScreen();
   try {
     fs.writeFileSync(file, JSON.stringify(windowState));
   } catch (e) {


hmhealey:
Is there any logic around using `windowState.fullscreen` when the app opens up that needs to be removed?

deanwhillier:
Not explicitly, but I’m guessing you’re thinking of someone that might already have fullscreen: true in their window state before upgrading? I’ll see about making sure it can never be set to true during startup.

deanwhillier:
Hmm, on closer inspection, the fix for #538 actually deals with #863. That fix already ensures that the app can't open in fullscreen, ever: https://github.com/mattermost/desktop/blob/deb86088742fbc6bc7d7c442bddcd245a9acd74e/src/main/mainWindow.js#L46

Would still be good to include this PR as it cleans up code that stored fullscreen in the saved window state, which is not necessary anymore.

deanwhillier:
Any ideas what's up with the failing checks below?

hmhealey:
Yeah, I think it makes sense to keep this to clean up that stored state

PR #959, opened 2019-04-10
[MM-14740] Consolidate configuration to support integration of MSI/GPO

deanwhillier:

**Summary**
This PR consolidates configuration handling (loading, combining default/build configs, updating, saving) in preparation for integrating Windows GPO data (function stub included). The new configuration object is instantiated in the main thread and any primary views loaded in the renderer thread. Each instance is setup with the ability to notify the other instance of changes in order to maintain synchronization between loaded config data in the main and renderer processes. This multi instance setup to needed to support existing tests and their access to config data.


deanwhillier:
Hmm, tests run fine locally ...

```

**Comment by enahum** [comment]
```
TypeError: Cannot read property 'should' of undefined
      at Context.should (test/specs/browser/settings_test.js:223:33)
      at processTicksAndRejections (internal/process/next_tick.js:81:5)
```

File: src/browser/components/SettingsPage.jsx [hunk 1]
```diff
@@ -65,99 +47,124 @@ export default class SettingsPage extends React.Component {
     this.handleShowUnreadBadge = this.handleShowUnreadBadge.bind(this);
     this.handleChangeUseSpellChecker = this.handleChangeUseSpellChecker.bind(this);
     this.handleChangeEnableHardwareAcceleration = this.handleChangeEnableHardwareAcceleration.bind(this);
+    this.saveSetting = this.saveSetting.bind(this);
+    this.updateSaveState = this.updateSaveState.bind(this);
```
enahum: Nothing to do now, but we should remove all this function bindings in favor of arrow functions to simplify the code

src/browser/components/SettingsPage.jsx
```diff
@@ -65,99 +47,124 @@ export default class SettingsPage extends React.Component {
...
+      Object.entries(savingState).forEach(([configType, currentState]) => {
+        if (currentState !== AutoSaveIndicator.SAVING_STATE_DONE) {
+          savingState[configType] = AutoSaveIndicator.SAVING_STATE_ERROR;
+          this.setState({savingState});
+        }
+      });
+    });
+
+    config.on('synchronize', () => {
+      ipcRenderer.send('reload-config');
```
enahum: Is this calling the event handler below? If that is the case why not just call `config.reload()` instead of another event

src/browser/index.jsx
```diff
@@ -15,22 +14,47 @@ import React from 'react';
...
+  deeplinkingUrl = remote.getCurrentWindow().deeplinkingUrl;
+}
+
+config.on('update', (configData) => {
+  teams.splice(0, teams.length, ...configData.teams);
+  requestingPermission.length = teams.length;
+});
+
+config.on('synchronize', () => {
+  ipcRenderer.send('reload-config');
```
enahum: Maybe I’m not following cause this is the second time, is this event being handled somewhere else and doing something else apart from config.reload()?

src/common/config/index.js
```diff
@@ -0,0 +1,357 @@
...
+    // remove unecessary data from default and build config
+    delete this.combinedData.defaultTeam;
+    delete this.combinedData.defaultTeams;
+
+    // IMPORTANT: properly combine teams from all sources
+    const combinedTeams = [];
+
+    // - start by adding default teams from buildConfig, if any
+    if (this.buildConfigData.defaultTeams && this.buildConfigData.defaultTeams.length) {
+      combinedTeams.push(...this.deepCopy(this.buildConfigData.defaultTeams));
```
enahum: I think you can safely remove deepCopy

src/browser/components/SettingsPage.jsx
deanwhillier: Converted in this component. Quick regex. :)

src/common/config/index.js
```diff
@@ -0,0 +1,358 @@
...
+
+  merge(base, target) {
+    return Object.assign({}, base, target);
+  }
+
+  copy(data) {
+    return Object.assign({}, data);
+  }
+
+  deepCopy(data) {
```
enahum: after removing the previous use of deepCopy is this still used anywhere?

src/browser/components/SettingsPage.jsx
deanwhillier: This is the part of the architecture that I'm not overly fond of as it is somewhat confusing. I have added inline comments to try to explain.

There are two processes running in the desktop app: the main process is run in Node.js and the renderer processes is run in Chromium (kinda like server<->client in web). The first `config.on('synchronize', () => {` function in both locations is used to tell the other process that an update occurred in this process and the second `[ipcMain|ipcRenderer].on('reload-config', () => {` function listens for changes to happen in the other process in order to reload config in this process so it can reload to keep in sync.

I initially only had the config loaded and managed in the main process with IPC functions available for the renderer process to use to retrieve and set config. Unfortunately that caused any test using config to fail as the simulated `config.json` files used in testing didn't seem to be available to the main process and so the renderer process could never get the config it was expecting. The quick solution was to load an instance of the same config class in both the main and renderer processes to be used directly and then add functionality to keep them in sync.

There is probably a better solution given enough time and if I can get debugging in tests to work, but I took this approach because we need it for MSI/GPO.

Does that make sense?

src/browser/index.jsx
deanwhillier: See above. :)

src/common/config/index.js
deanwhillier: Good catch, missed that on, thanks!

src/common/config/index.js
deanwhillier: Probably not, kept it for legacy purposes. Will remove.

src/browser/components/SettingsPage.jsx
enahum: yup it does. thanks for the explanation

deanwhillier:
Regarding tests, that failing test works fine locally and I'm not entirely sure why it wouldn't work here. I'm going to try something though ...

deanwhillier:
Okay, so test is fixed. The test was initially trying to 'click' on the `light` theme option, but that option is selected by default so I reversed the test to that it first 'clicks' on `dark` and then `light`. Test passes now. This didn't manifest before because the old way of handling the config wrote all config options every time, which meant that `light` was already set in the test config.json file.

wget:
Ok for me. Tested on Windows and Linux.
Windows (Appveyor) logs:
* https://ci.appveyor.com/project/mattermost/desktop/builds/23768406
* https://ci.appveyor.com/project/mattermost/desktop/builds/23768404
Linux logs (tests only):
* https://gist.github.com/wget/a863b81b2ec491d6b6c286daebbd5343

enahum:
Awesome!! Lets add the MSI stuff

PR #961, opened 2019-04-12
[MM-14740] Integrate GPO functionality

deanwhillier:

**Summary**
This PR integrates GPO functionality by retrieving supported values set in the Windows registry and appropriately merging them with the app config.

NOTE: This functionality needs to be tested in a Windows environment configured with the appropriate GPO capabilities.

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -24,6 +24,7 @@
     "semver": "^5.5.0",
     "simple-spellchecker": "^0.9.8",
     "underscore": "^1.9.1",
+    "winreg": "^1.2.4",
     "yargs": "^3.32.0"
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -80,6 +80,7 @@
         "src/common/config/buildConfig.js",
         "src/common/config/pastDefaultPreferences.js",
         "src/common/config/upgradePreferences.js",
+        "src/common/config/RegistryConfig.js",
         "src/common/osVersion.js",
         "src/common/config/defaultPreferences.js",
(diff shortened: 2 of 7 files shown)


src/package.json
```diff
@@ -24,6 +24,7 @@
     "semver": "^5.5.0",
     "simple-spellchecker": "^0.9.8",
     "underscore": "^1.9.1",
+    "winreg": "^1.2.4",
```
enahum: can we file a ticket to have all of this dependencies to be exact?

wget: @enahum Why do you need them to be exact? The API shouldn't change for 1.x.x, so `^` appears to be correct IMHO.
https://stackoverflow.com/a/22345808/3514658

wget: Or maybe you want to have `~` instead?

enahum: cause we use exact versions for dependencies everywhere and this one should not be the exception ;)

wget: @enahum Ok. Let's open a ticket for this asking to bumb dependencies as well (which will require a new testing session, so we are a bit short for the open release of today). Because right, now according to `npm` there are still some deps we use for which we haven't applied updates. In the meantime, I merge this PR :)
```

removed 1 package and audited 25559 packages in 55s
found 34 vulnerabilities (15 moderate, 19 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```

enahum: Yeap, as I said this has to be another ticket which I already created https://mattermost.atlassian.net/browse/MM-15208

PR #964, opened 2019-04-23
Update webpack-cli to fix error on npm run build

svelle:
Currently cloning and building the repo returns an error because of an outdated version of webpack-cli.
This PR updates the minimum required webpack version to 3.1.1 to enable building without errors again.


svelle:
Oh yes you're right. I feel kinda stupid now. Thanks for pointing this out 👍

PR #968, opened 2019-04-25
[MM-14740] Address potential race condition and cleanup main.js

deanwhillier:

**Summary**
This PR's primary purpose is to address a potential race condition between the Electron app's `ready` event that creates the app UI etc. and the loading of Windows registry config data needed by the app UI. This is handled using Promise.all – specifically, moving the app `ready` code into a separate function, moving the registryConfig `update` event code into the same function and waiting till both the app is ready and the registry config data has loaded before calling the new function to finish creating the app.

While making these changes, the PR also restructures the `main.js` file to clean it up and group related code together. The intent of this restructuring is to make it easier to see what `main.js` is doing without having to spend several hours combing through all the code to see what each piece does and when – specifically, making it easier to see what app feature events are being handled, what inter-process communication events are being handled, what renderer (mainWindow) events are being handled and more importantly, what order code is being executed in.

### DIFF
--- a/src/common/config/RegistryConfig.js
+++ b/src/common/config/RegistryConfig.js
@@ -9,6 +9,9 @@ import WindowsRegistry from 'winreg';
 const REGISTRY_HIVE_LIST = [WindowsRegistry.HKLM, WindowsRegistry.HKCU];
 const BASE_REGISTRY_KEY_PATH = '\\Software\\Policies\\Mattermost';
 
+/**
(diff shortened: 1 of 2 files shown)


deanwhillier:
This looks bigger than it actually is – most of the code is unchanged and has only been re-arranged to accommodate the Await/Async updates for loading registry config data and waiting for app ready event.

PR #971, opened 2019-05-13
[MM-11896] Do not use minimize shortcut on win32

steevsachs:

**Summary**
Adds condition to remove accelerator for Window -> Minimize when `process.platform === 'win32'`. 

By setting accelerator to `null` for other platform values, Minimize will use the default shortcut on those platforms.

**Issue link**
Fixes https://github.com/mattermost/mattermost-server/issues/9398

**Test Cases**
I was unable to come up with a good way to test this, and didn't see any comparable tests in the app.  I prefer to have a test here, and would appreciate any guidance to achieve that.

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -191,6 +191,9 @@ function createTemplate(mainWindow, config, isDev) {
     label: '&Window',
     submenu: [{
       role: 'minimize',
+
+      // empty string removes shortcut on Windows; null will default by OS
+      accelerator: process.platform === 'win32' ? '' : null,
     }, {
       role: 'close',
     }, separatorItem, ...teams.slice(0, 9).map((team, i) => {


wget:
@steevsachs Hi there. Why do we need to remove the shortcut accelerator of the Window menu?

steevsachs:
@wget Please see the discussion here https://github.com/mattermost/mattermost-server/issues/9398#issuecomment-491002725

The current behavior on Windows is that `ctrl+m` results in both minimizing the window _and_ sending the current message, which is a confusing UX. 

Resolving the issue with `ctrl+m` sending on Windows (because it produces an `Enter` keypress event) is problematic, so the proposed solution is to prevent the minimize instead.

yuya-oc:
Hi @steevsachs,

Thanks for your contributing! Just curious, what will happen when pressing Ctrl+M on Linux? In my guess, the problem will appear as well as Windows. So possibly we need to apply this patch also for Linux

(If it's difficult to test for you, feel free to let us know.)

steevsachs:
@yuya-oc `ctrl+m` on linux in the desktop client only minimizes, it does not send.

hanzei:
@wget Gentle reminder to review this

hanzei:
Thank you for your contribution @steevsachs :tada:

PR #990, opened 2019-06-13
[MM-15918] Pull reference counts from LHS only.

deanwhillier:

**Summary**
This PR updates the scraping mechanism that retrieves mentions from the DOM of the webapp for use in the mentions badge on the server tab (when visible). The mechanism now only pulls mentions from the LHS instead of anywhere in the DOM, which was previously causing the count in the tab to double whenever the suggestions drop-down was visible while the quick-switcher was open.

**Issue link**
[MM-15918](https://mattermost.atlassian.net/browse/MM-15918)

### DIFF
--- a/src/browser/webview/mattermost.js
+++ b/src/browser/webview/mattermost.js
@@ -90,7 +90,7 @@ function getUnreadCount() {
   }
 
   // mentionCount in sidebar
-  const elem = document.getElementsByClassName('badge');
+  const elem = document.querySelectorAll('#sidebar-left .badge');
   let mentionCount = 0;
   for (let i = 0; i < elem.length; i++) {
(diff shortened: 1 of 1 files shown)


src/browser/webview/mattermost.js
```diff
@@ -90,7 +90,7 @@ function getUnreadCount() {
   }
 
   // mentionCount in sidebar
-  const elem = document.getElementsByClassName('badge');
+  const elem = document.querySelectorAll('#sidebar-left .badge');
```
sudheerDev: Is this a temporary solution? If not i think we should comment in webapp that we use the ID in desktop for scraping so people don't accidentally change it and forget this.

deanwhillier: Good thought! Ideally it is just a temporary solution as we're planning on changing dom scraping to communicating with the webapp directly to get data as needed.

PR #993, opened 2019-06-20
[MM 7970] Maintain online status while the Desktop App is in the background ... and other things.

deanwhillier:

**Summary**
This PR and related Webapp & Server PR's support maintaining a user's online status while the Desktop app is in the background as well as automatically setting the user to away/online if the screensaver goes on/off, the user locks/unlocks their computer or shuts down their computer.

**PR's**
Desktop: [MM-7970](https://github.com/mattermost/desktop/pull/993) (this PR)
Webapp: [MM-7970](https://github.com/mattermost/mattermost-webapp/pull/2992)
Server: [MM-14582](https://github.com/mattermost/mattermost-server/pull/11312)

**Issue links**
Jira 1: [MM-7970](https://mattermost.atlassian.net/browse/MM-7970)
Jira 2: [MM-14582](https://mattermost.atlassian.net/browse/MM-14582)
Jira 3: [MM-16153](https://mattermost.atlassian.net/browse/MM-16153)

**Test Cases**
1. With the updated server and webapp running locally, build and run the updated Desktop App.
2. Create a server in the Desktop App pointing at the locally running instance of server/webapp
3. Put the Desktop App in the background and continue to use your computer without focusing on the Desktop App for more than the default 5min of inactivity the Webapp uses to determine if a user is away. The Desktop App should not change to away while actively using the computer.
4. Stop using your computer for more than the default 5min of inactivity the Webapp uses to determine if a user is away. The Desktop App should change to away after about 5mins while not using the computer. (The screensaver timer should be set to more than 5 minutes in order to observe this behaviour)
5. Log into another user on your local install from another machine/device to observe the status of the first user.
6. Manually engage the screensaver or lock the computer running the desktop app. After a short period of time (about a minute at most), the second machine/device should show the first user as away instead of online.
7. Cancel the screensaver or unlock the computer running the desktop app. The user's status icon in the Desktop App should be back to online and after a short period of time (about a minute at most), the second machine/device should show the first user as online again as well.

**Additional Notes**
Without updating user status functionality on the server, this setup basically just manages keeping a user's status `online` or bringing them back `online` if the server had determined they went `away` based on normal user (in)activity. The server still handles setting the user to `away`.

That said, locking the screen, engaging the screensaver or shutting down will force the user's status to change to `away` by setting the `manual` option on the server side to true. Unlocking the screen or cancelling the screensaver will also necessarily force the user's status to change to online by setting the 'manual' option on the server side to true. This means that if a user had manually set themselves to away and then either lock their screen or the screensaver comes on, they will be automatically set back to online when they unlock their screen or the screensaver ends, and they will need to manually set themselves to away again if desired. I'm not sure this can be helped without some additional work on the server-side ... open to suggestions.

### DIFF
--- a/src/browser/js/WebappConnector.js
+++ b/src/browser/js/WebappConnector.js
@@ -1,0 +1,18 @@
+// Copyright (c) 2015-2016 Yuya Ochiai
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import EventEmitter from 'events';
+
+/**
+* Provides the bridging infrastructure to connect the desktop and webapp together
+*/
+export default class WebappConnector extends EventEmitter {
+  constructor() {
+    super();
+
+    this.active = true;
+
+    window.webappConnector = this;
+  }
+}
--- a/src/browser/js/WebappConnector.js
+++ b/src/browser/js/WebappConnector.js
@@ -1,4 +1,3 @@
-// Copyright (c) 2015-2016 Yuya Ochiai
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -54,7 +54,6 @@
         "src/browser/updater.jsx",
         "src/browser/js/notification.js",
         "src/browser/js/badge.js",
-        "src/browser/js/WebappConnector.js",
         "src/browser/webview/mattermost.js",
         "src/browser/components/RemoveServerModal.jsx",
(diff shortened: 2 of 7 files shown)


src/browser/js/WebappConnector.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015-2016 Yuya Ochiai
```
lieut-data: Should we omit Yuya from new files?

deanwhillier: Sounds like it's not necessary for new files ... I'll update. Thx.

deanwhillier:
@lieut-data, I added some tests for UserActivityMonitor, let me know if that's what you had in mind.

PR #1000, opened 2019-07-08
Mm 16694 master validate urls

deanwhillier:

**Summary**
This PR validates deeplink and user-clicked urls before handling them in the Desktop app. This is intended to prevent the desktop app from running invalidly formatted URLs that could compromise the users computer.

**Issue link**
https://mattermost.atlassian.net/browse/MM-16694

**Test Cases**
See issue link for test cases.


deanwhillier:
Definitely interested in feedback to this approach. While I was not able to get the mentioned `whoami /all` output the ticket mentioned, I did get the new spawned window. This PR now prevents the spawned window for me.

src/main.js
```diff
@@ -756,10 +761,11 @@ function switchMenuIconImages(icons, isDarkMode) {
   }
 }
 
-function setDeeplinkingUrl(url) {
-  if (scheme) {
-    deeplinkingUrl = url.replace(new RegExp('^' + scheme), 'https');
+function getDeeplinkingUrl(url) {
+  if (!scheme || !Utils.isValidURL(url)) {
```
crspeller: Can we invert this condition? It's a bit unclear. Also it would be nice that `null` is the default.

deanwhillier:
This is also ready for review. While the exact exploit may not affect `master`, I replicated the same checks as the `release-4.2` branch. Tests still also need to be sorted out on this PR.

src/browser/components/MainPage.jsx
```diff
@@ -62,6 +58,27 @@ export default class MainPage extends React.Component {
...
+      const deeplinkURL = url.parse(deeplink);
+      let parsedDeeplink = null;
+      teams.forEach((team, index) => {
+        const teamURL = url.parse(team.url);
+        if (deeplinkURL.host === teamURL.host) {
+          parsedDeeplink = {
+            teamURL,
+            teamIndex: index,
+            originalURL: deeplinkURL,
+            url: `${teamURL.protocol}//${teamURL.host}${deeplinkURL.pathname}`,
```
hmhealey: Does "team" here actually refer to a server tab? I feel like this naming might date back to before a single user could be on multiple teams

src/main.js
```diff
@@ -378,6 +393,8 @@ function handleAppWebContentsCreated(dc, contents) {
 }
 
 function initializeAfterAppReady() {
+  app.setAppUserModelId('Mattermost.Desktop'); // Use explicit AppUserModelID
```
hmhealey: Just to double check, this ID is supposed to be changing from the previous one that referenced Squirrel, right?

src/browser/components/MainPage.jsx
deanwhillier: Yes, good point. There was some work done in another PR to change all `team` references to `server` where actually server, but a bunch of work had been done since making that PR almost impossible to merge. This still needs to be done, but I think it should be done as a whole instead piece by piece to avoid cross-over confusion.

src/main.js
deanwhillier: Yes, as part of investigating notifications for MSI, we decided to change to this. I removed all remaining reference to Squirrel as it wasn't all removed, so change this here as well.

deanwhillier: We discovered that this is apparently the convention to use for Windows for this property.

hanzei:
@deanwhillier Is this ready to merge?

deanwhillier:
Following a re-review it is, but the failing tests due to crashing Chrome needs to be sorted out. Tests work fine locally (using any of node 8, 10 or 12), but I don't know why they fail in CircleCI. @cpanato, @DSchalla, any chance you could please take a look? This seems to be happening for any PR's based off of master.

cpanato:
@deanwhillier @DSchalla why we need to look at it? we did not setup the pipeline and don't have a context in what is failing, at least me.

crspeller:
@deanwhillier Tests seem to be failing. Not sure what you want to do about that.

deanwhillier:
Added pipeline fixes from @Willyfrog, tests are now running again and passing!

norbertwenzel:
Thank you.

PR #1001, opened 2019-07-08
Mm 16694 4.2 validate urls

deanwhillier:

**Summary**
This PR validates deeplink and user-clicked urls before handling them in the Desktop app. This is intended to prevent the desktop app from running invalidly formatted URLs that could compromise the users computer.

**Issue link**
https://mattermost.atlassian.net/browse/MM-16694

**Test Cases**
See issue link for test cases.


deanwhillier:
Definitely interested in feedback to this approach. While I was not able to get the mentioned whoami /all output the ticket mentioned, I did get the new spawned window. This PR now prevents the spawned window for me.

Note: The package I used for URL validation in master won't work in 4.2 .. doesn't like the package's use or es6 spread syntax. I forked the package and changed to Object.assign.

hmhealey:
It's bizarre that Electron doesn't support ES6 without transpiling or anything since Node.js and browsers both support it...

deanwhillier:
Yup, fortunately it's just the 4.2 version, current master doesn't have the problem and is using the package directly.

hmhealey:
Oh okay, so the newer version of Electron supports it fine. That's good to know

deanwhillier:
Okay, I think this is ready to review now. I ran the reported exploit step-by-step and am unable to repro through deeplink exploit with the app closed or open or when clicking the exploit via a link directly in MM.

I still have to troubleshoot the existing tests, add new tests and the CI pipeline is still not working properly.

package.json
```diff
@@ -54,6 +54,7 @@
     "electron-builder-squirrel-windows": "~20.14.0",
     "electron-connect": "^0.6.3",
     "eslint": "^4.19.1",
+    "eslint-plugin-eslint-comments": "^3.1.2",
```
hmhealey: I assume this is copied over from master, but this is requiring a newer version than master if that matters

src/main/ParseArgs.js
```diff
@@ -0,0 +1,41 @@
...
+    if (deeplinkIndex !== -1) {
+      return args.slice(0, deeplinkIndex + 1);
+    }
+  }
+  return args;
+}
+
+function parseArgs(args) {
+  return yargs.
+    boolean('squirrel-uninstall').
```
hmhealey: I assume we also have to support these squirrel-specific args still?

package.json
deanwhillier: Good question, didn't go to Master for this. Just installed it on this branch because it was for some reason needing it, even though it seemingly hadn't previously? I will make sure master version matches! Good eye! 👁

src/main/ParseArgs.js
deanwhillier: Alas, yes. Unless we remove squirrel altogether, but then we'll have to solve for no notifications etc.

No longer in master, but we still have to solve for notifications etc. there. 😛

deanwhillier:
Okay, so the important part here is that the tests are now passing correctly again in CircleCI. Unfortunately Jenkins is till broken – if CircleCI can successfully build and test, there's no reason why Jenkins shouldn't be able to. I believe @lieut-data had indicated that it looked like Jenkins had an incorrect version of Node or something?

@DSchalla, @cpanato, any chance one of you could provide some assistance?

cpanato:
all fixed now

DSchalla:
@deanwhillier Problem was that the desktop pipeline currently is installing with npm install, but 4.2 still uses yarn. We temporarily changed the pipeline to use yarn again so it could succeed. Judging by Harrisons comments I don't know if you want to perform any changes, but otherwise this looks good. :tada:

deanwhillier:
That's great, thanks @cpanato and @DSchalla! No changes needed here from Harrisons comments.

PR #1008, opened 2019-08-02
remove jenkins

cpanato:
removing jenkins, circleci is doing the same for building the PRs

- after we merge this I will remove the jenkins job and set the check to use the one from circleci


cpanato:
this file is not used at all, the job was configured using other approach

PR #1029, opened 2019-08-30
[MM-18135] use no sandbox, separate linting, circle 2.1

Willyfrog:

**Summary**
- Separated linting and testing as a jobs.
- Fix build errors on CircleCI so people can rely on its results.
- Workflow prevents building if linting and testing jobs fail
- Minor optimizations.
- Migrated to circleci 2.1
- Upgraded electron to a newer minor version.

**Issue link**
[MM-18136](https://mattermost.atlassian.net/browse/MM-18136)
[MM-18135](https://mattermost.atlassian.net/browse/MM-18135)
[MM-17955](https://mattermost.atlassian.net/browse/MM-17955)
**Test Cases**

**Additional Notes**
- No MSI installer step yet
- we have to figure out if there is use for storybook, there seems to be several steps for that but I'm - not sure if its being used or stored.
- no s3 yet
- artifacts are generated on circleci (builds and storybook), but we are not doing anything with them


Willyfrog:
sometimes tests still fail, just re-run the build

.circleci/config.yml
```diff
@@ -57,5 +86,9 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - build
+      - lint
       - test
```
cpanato: can you make the `test` stage require `lint`?

.circleci/config.yml
```diff
@@ -57,5 +86,9 @@ workflows:
...
+      - build:
+          requires:
+            - lint
+            - test
```
cpanato: then you can remove the test here

.circleci/config.yml
Willyfrog: Since both are jobs that run under 5 min, this would reveal quicker any problem and would show both linting and testing. Having them sequentially would mean that we would go through first the linting problems and then the testing ones.

Is there a reason or improvement to run them sequentially?

lieut-data: I'm inclined to agree with @cpanato -- if the linting steps fail, there's probably no reason to think the testing step will be valid anyway: there's a decent chance it will fail for the same reasons.

test/modules/environment.js
```diff
@@ -49,10 +49,17 @@ module.exports = {
...
-    return new Application({
+    const options = {
       path: electronBinaryPath,
       args: [`${path.join(sourceRootDir, 'src')}`, `--data-dir=${userDataDir}`, '--disable-dev-mode'],
-    });
+
+      // enable this if chromedriver hangs to see logs
+      // chromeDriverLogPath: '../chromedriverlog.txt',
+    };
+    if (process.platform === 'darwin') {
```
lieut-data: Can you describe what this part of the change enables?

.circleci/config.yml
Willyfrog: In that case, should I run the linting and testing in the same job, there wouldn't be any gain in running them separately, would it?

test/modules/environment.js
Willyfrog: I added a comment :) thanks for pointing out it wasn't clear

.circleci/config.yml
Willyfrog: I put both steps as one, if we are not going to speed up showing as much errors, there is no point in penalising by having them in separate containers.

.circleci/config.yml
```diff
@@ -38,24 +75,11 @@ jobs:
...
-      - run: cp test-results.xml /tmp/test-results/
-      - store_test_results:
-          path: /tmp/test-results
-
 workflows:
   version: 2
   build_and_test:
     jobs:
-      - build
-      - test
```
cpanato: why you remove the test part?

Willyfrog: I merged testing and linting and changed the name to check (a bit more generic)

Issue #1036, opened 2019-09-18
Desktop Notification is not working


bitdessert:
I confirm (by marking "x" in the [ ] below: [x]):

- [ ] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [ ] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [ ] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

I minimize the application(mattermost desktop) and click desktop notification, but it will not run.

**Environment**
- Operating System: windows 10, windows 7, macOS(It runs but there is an error, eg. EnhancedNotification is not loaded.)
- Mattermost Desktop App version: 4.2.3 
- Mattermost Server version: 5.14.2 

Earlier versions worked as expected.(eg. mattermost desktop version 4.2.1)

**Steps to reproduce**
1. Someone minimizes the app.(on windows)
2. Send a message to someone from another pc.
3. Then, Someone was notified.
4. Someone click desktop notification
5. But, nothing happens.

**Expected behavior**
When someone clicks Desktop Notification, the app should launch.

**Observed behavior**

**Possible fixes**

amyblais:
@bitdessert We have this ticket open: https://mattermost.atlassian.net/browse/MM-18131. Let me know if this sounds like the same issue you're seeing.

bitdessert:
Yes.
So when can I install version 4.3?

Version 4.2.3 has other issues too.
eg. If terminate and launch the App(mattermost-desktop) in a Windows 10, the session may expire.

bitdessert:
Perhaps this problem is related to 'webPreferences.contextIsolation'.

deanwhillier:
This issue should be fixed by this PR: https://github.com/mattermost/desktop/pull/1040

amyblais:
Closing as this is fixed for the upcoming v4.3 release. v4.3 release is currently scheduled for October 16th,

PR #1040, opened 2019-09-20
[MM-18152] Desktop notifications

deanwhillier:

**Summary**

Adding `context-isolation` as part of a security improvement to the Desktop app broke Desktop Notifications as the Desktop App is no longer able to directly interact with / affect the loaded webapp.

This PR does a couple of things to re-enable full desktop notification support:

- replaces existing custom permissions handling with a simpler implementation that now works – the existing custom permissions handling broke in multiple places somewhere along the line and was silently denying all requests for things like notifications, camera/mic use etc. from the renderer process
- simpler custom permissions handling now supports only a subset of permission types and automatically grants permission when requested by the renderer process or any of the defined servers, all other requests are automatically denied
- Electron was updated to v5.x, including a Chrome update that now prevents HTML5 Notifications from being dispatched from insecure sites (http), handling notifications in the Desktop App renderer process bypasses this restriction re-enabling support for http servers
- Desktop App now registers itself with the webapp to indicate that it can handle notifications on behalf of the webapp
- Desktop App listens for notification dispatch requests from the webapp, dispatches the requested notifications, listens for the user to click on the notification and then updates the webapp so the webapp can properly navigate to the appropriate channel (prevents white flashing if the Desktop App handles the navigation)
- while dispatching notifications, the Desktop App once again handles bringing the app back into the foreground, whether minimized, closed or just in the background
- Desktop App should once again play a notification sound in Windows 8.1 and below

Implications:

- The v4.3 Desktop App now requires a version of the Server/Webapp that includes functionality to hand notifications over to the Desktop App in order to guarantee full desktop notification functionality. This functionality is planned for v5.16+ of the Server/Webapp and plans are to back-port to v5.15, v5.14 and to the current ESR.
- Desktop App versions v4.2.1 and earlier should maintain existing notification functionality
- Running v4.3 of the Desktop App on a non-supported Server/Webapp version will result in reduced notification functionality – clicking the notification won't bring the app into the foreground and sounds may not work on older versions of Windows

**Issue links**

Notification issues fixed as a result of this PR:

https://mattermost.atlassian.net/browse/MM-18152
https://mattermost.atlassian.net/browse/MM-18111
https://mattermost.atlassian.net/browse/MM-18131
https://github.com/mattermost/desktop/issues/1036

Other issues fixed as a result of this PR:

https://mattermost.atlassian.net/browse/MM-18637
https://mattermost.atlassian.net/browse/MM-18719

**Related Pull Requests**

- Has webapp changes: https://github.com/mattermost/mattermost-webapp/pull/3718

**Additional Notes**
Other minor bugs fixed as a result of this PR include:

- External links in posts are now validated as URI's instead of more specific URL's to allow custom protocol handling to work as intended.
- When quitting the app or shutting down the computer, the Desktop App was setting the user to manual away status, this has been corrected to maintain whatever status the user is in when quitting the app or shutting down.


src/browser/webview/mattermost.js
```diff
@@ -3,16 +3,13 @@
 // See LICENSE.txt for license information.
 'use strict';
 
-import {ipcRenderer, webFrame} from 'electron';
+/* eslint-disable no-magic-numbers */
```
Willyfrog: is this really needed? can we instead fix the error?

src/browser/webview/mattermost.js
```diff
@@ -3,16 +3,13 @@
...
 
-import EnhancedNotification from '../js/notification';
+import {ipcRenderer, webFrame, remote} from 'electron';
 
 const UNREAD_COUNT_INTERVAL = 1000;
-//eslint-disable-next-line no-magic-numbers
```
Willyfrog: I might accept this one better, in this particular case can be guessed quite quickly instead of defining multiple constants for hour, minute, etc.

src/main.js
```diff
@@ -571,10 +562,40 @@ function initializeAfterAppReady() {
...
+
+    // get the requesting webContents url
+    const requestingURL = webContents.getURL();
+
+    // is the target url trusted?
+    config.teams.forEach((team) => {
+      if (requestingURL.startsWith(team.url)) {
+        callback(true);
+      }
+    });
```
Willyfrog: what happens if it doesn't belong to any team?

src/browser/components/MattermostView.jsx
```diff
@@ -36,6 +38,7 @@ export default class MattermostView extends React.Component {
     };
 
     this.handleUnreadCountChange = this.handleUnreadCountChange.bind(this);
+    this.dispatchNotification = this.dispatchNotification.bind(this);
```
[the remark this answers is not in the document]
deanwhillier: Desktop App setup doesn't support arrow functions in that context yet. Fixable yes, but there is a lot that needs to be fixed ... we'll get to them in the coming releases. For now, still using binding.

src/browser/webview/mattermost.js
deanwhillier: There are a number of these all through the Destkop code that need to be looked at. I added this in the particular case to temporarily stop vscode from complaining while I was working on the notifications stuff. I'll make sure there is a ticket in the backlog to address along with a ticket to convert bind to arrow functions.

src/browser/webview/mattermost.js
deanwhillier: Ditto.

src/main.js
deanwhillier: hmmm, interesting, I had `callback(false)` in there at one point. Must have inadvertently removed while cleaning up. Good catch!

src/browser/components/MattermostView.jsx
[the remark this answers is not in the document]
deanwhillier: https://mattermost.atlassian.net/browse/MM-18754

src/browser/webview/mattermost.js
deanwhillier: https://mattermost.atlassian.net/browse/MM-18755

PR #1053, opened 2019-10-02
MM-19096 – simplify custom login process

deanwhillier:
**Summary**
The new functionality added to support 3rd party custom login mechanisms (SAML, OAUTH) following a security enhancement attempted to determine the 3rd party hostname and temporarily trust it for the duration of the login request. This temporary trust created some challenges during testing and it was subsequently decided that this aspect of the feature was too brittle given we ultimately can't control what an IDP does on their end to validate an authentication request, nor can we control the kinds of linkable resources an end user might be presented with during the login process.

This PR removes the temporary trust aspect of the feature and solely relies on appropriately flagging a custom login is in progress, during which any navigation that needs to occur until the user is returned to MM will be allowed to occur. Once the login process completes, the flag is reset and normal navigation handling resumes.

As a by-product of the change, this PR now allows links on IDP pages to function as intended, fixing the issue [reported here](https://mattermost.atlassian.net/browse/MM-18936).

**Issue links**
Primary: https://mattermost.atlassian.net/browse/MM-19096
Secondary: https://mattermost.atlassian.net/browse/MM-18936
Original feature: https://mattermost.atlassian.net/browse/MM-17255

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -761,10 +761,10 @@ function isTrustedURL(url) {
   const teamURLs = config.teams.reduce((urls, team) => {
     const parsedTeamURL = parseURL(team.url);
     if (parsedTeamURL) {
-      return urls.push(parsedTeamURL);
+      return urls.concat(parsedTeamURL);
     }
     return urls;
-  });
+  }, []);
   for (const teamURL of teamURLs) {
     if (parsedURL.origin === teamURL.origin) {
       return true;
--- a/src/main.js
+++ b/src/main.js
@@ -382,9 +382,8 @@ function handleAppWebContentsCreated(dc, contents) {
   contents.on('will-navigate', (event, url) => {
     const contentID = event.sender.id;
     const parsedURL = parseURL(url);
-    const urlIsTrusted = isTrustedURL(parsedURL);
 
-    if (urlIsTrusted) {
+    if (isTrustedURL(parsedURL)) {
       return;
     }
     if (customLogins[contentID].inProgress) {
@@ -402,8 +401,8 @@ function handleAppWebContentsCreated(dc, contents) {
   contents.on('did-start-navigation', (event, url) => {
     const contentID = event.sender.id;
     const parsedURL = parseURL(url);
-    const urlIsTrusted = isTrustedURL(parsedURL);
-    if (!urlIsTrusted) {
+
+    if (!isTrustedURL(parsedURL)) {
       return;
     }
 
--- a/src/main.js
+++ b/src/main.js
@@ -384,10 +384,14 @@ function handleAppWebContentsCreated(dc, contents) {
     const parsedURL = parseURL(url);
     const urlIsTrusted = isTrustedURL(parsedURL);
 
-    // don't prevent custom login attempts (oath, saml)
-    if (!urlIsTrusted && !customLogins[contentID].inProgress) {
-      event.preventDefault();
+    if (urlIsTrusted) {
+      return;
     }
+    if (customLogins[contentID].inProgress) {
+      return;
+    }
+
+    event.preventDefault();
   });
 
   // handle custom login requests (oath, saml):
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -763,40 +732,44 @@ function handleMainWindowWebContentsCrashed() {
...
-function isTrustedURL(url) {
+function parseURL(url) {
   if (!url) {
-    return false;
+    return null;
   }
-
-  let parsedUrl = url;
+  let parsedURL = url;
   if (typeof url === 'string') {
```
lieut-data: nit: if `url` isn't a string, we end up returning something that might not be an `URL`. Should we test for an `URL` and return that explicitly, but otherwise return `null`?

src/main.js
```diff
@@ -763,40 +732,44 @@ function handleMainWindowWebContentsCrashed() {
...
-
+function isTrustedURL(url) {
+  const parsedURL = parseURL(url);
+  if (!parsedURL) {
+    return false;
+  }
+  const trustedURLs = config.teams.map((team) => parseURL(team.url));
   for (const trustedURL of trustedURLs) {
-    if (parsedUrl.origin === trustedURL.origin) {
+    if (parsedURL.origin === trustedURL.origin) {
```
lieut-data: Might `trustedURL` be `null` here based on the `parseURL` semantics above?

src/main.js
```diff
@@ -383,60 +381,31 @@ function handleAppWebContentsCreated(dc, contents) {
...
+  //    - indicate custom login is in progress
+  // 2. are we finished with the custom login process?
+  //    - indicate custom login is NOT in progress
   contents.on('did-start-navigation', (event, url) => {
     const contentID = event.sender.id;
-    const parsedUrl = new URL(url);
-    const urlIsTrusted = isTrustedURL(parsedUrl);
-    const urlIsCustomLoginPath = isCustomLoginURL(parsedUrl);
+    const parsedURL = parseURL(url);
+    const urlIsTrusted = isTrustedURL(parsedURL);
```
lieut-data: nit: maybe early exit if `!urlIsTrusted`?

src/main.js
```diff
@@ -383,60 +381,31 @@ function handleAppWebContentsCreated(dc, contents) {
...
+    const urlIsCustomLoginPath = isCustomLoginURL(parsedURL);
     const previousPage = event.sender.history[event.sender.history.length - 1];
 
     if (urlIsTrusted && urlIsCustomLoginPath && !customLogins[contentID].inProgress && previousPage.endsWith('/login')) {
```
lieut-data: Will this `endsWith` check fail if there are query string parameters? Is it strictly necessary to check `previousPage` at all?

src/main.js
```diff
@@ -383,60 +381,31 @@ function handleAppWebContentsCreated(dc, contents) {
 
   contents.on('will-navigate', (event, url) => {
     const contentID = event.sender.id;
-    const parsedUrl = new URL(url);
-    const urlIsTrusted = isTrustedURL(parsedUrl);
-    const urlIsTrustedExternalLoginPath = parsedUrl.hostname === customLogins[contentID].externalHostname;
+    const parsedURL = parseURL(url);
+    const urlIsTrusted = isTrustedURL(parsedURL);
 
     // don't prevent custom login attempts (oath, saml)
```
lieut-data: nit: the double negatives here are hard to parse. Perhaps something like:
```js
    if (urlIsTrusted) {
        return;
    }

    // allow arbitrary navigation while in a trusted login flow
    if (customLogins[contentID].inProgress) {
       return;
    }

    // Otherwise, prevent the webview from navigating outside the trusted domain. This
    // prevents security issues such as...
    event.preventDefault();
```

src/main.js
deanwhillier: Good question. @crspeller, do you think it's necessary to ensure a SAML/OAUTH endpoint is accessed directly from the /login page in order to trigger the updated custom login process?

src/main.js
deanwhillier: I like this approach, just have to keep reminding myself to break my old habits.

src/main.js
deanwhillier: Reworking this a bit ...

deanwhillier:
Pushed some updates, left the logic to check for `/login` for now, pending @crspeller's thoughts.

src/main.js
deanwhillier: So the update to this one returns the url if it is an instance of URL and exits. If not, it tries to convert it to a URL and returns that if successful, otherwise returns null.

src/main.js
```diff
@@ -383,60 +381,38 @@ function handleAppWebContentsCreated(dc, contents) {
...
-        // redirect to starting url if set
-        if (customLogins[contentID].startingURL) {
-          event.sender.loadURL(customLogins[contentID].startingURL);
-          customLogins[contentID].startingURL = null;
-        }
-      }
+    const parsedURL = parseURL(url);
+    const urlIsTrusted = isTrustedURL(parsedURL);
+
+    if (urlIsTrusted) {
```
crspeller:
```suggestion
    if (isTrustedURL(parsedURL)) {
```

crspeller: This suggestion can be applied in a bunch of places here. I think it's leftovers from when this was more complicated.

src/main.js
crspeller: Probably not necessary. This can be removed. Worried about some kind of deep login link failing because of this.

deanwhillier: K, sounds good, thanks!

PR #1054, opened 2019-10-02
MM-19095 – Stop cursor bug fix from running in inactive tabs

deanwhillier:
**Summary**
A fix applied in [this PR](https://github.com/mattermost/desktop/pull/1025) to resolve an Electron cursor/focus bug has been messing up focus for the currently active tab when a server defined in an inactive tab is not able to load (offline). The problem is that the desktop app keeps trying to load the offline server and with each attempt, the cursor/focus fix is run again stealing focus from the active tab.

This PR adds a check to make sure a tab is active before running the fix.

**Issue link**
https://mattermost.atlassian.net/browse/MM-19095


lindalumitchell:
Nice hunch / fix! QA can test and verify after it's merged. Removing QA Review label.

PR #1056, opened 2019-10-05
[MM-10586] Desktop App Window/Tabs Update

devinbinnie:

**Summary**
New Server Tabs look and feel.

**Issue link**
https://mattermost.atlassian.net/browse/MM-10586


src/browser/components/MainPage.jsx
```diff
@@ -283,6 +294,19 @@ export default class MainPage extends React.Component {
...
+    // TODO Devin: Implement for Windows
+    const doubleClickAction = remote.systemPreferences.getUserDefault('AppleActionOnDoubleClick', 'string');
+    const win = remote.getCurrentWindow();
+    if (doubleClickAction === 'Minimize') {
+      win.minimize();
+    } else if (doubleClickAction === 'Maximize' && !win.isMaximized()) {
+      win.maximize();
+    } else if (doubleClickAction === 'Maximize' && win.isMaximized()) {
+      win.unmaximize();
+    }
```
Willyfrog: Maybe you want to have here some sort of strategy pattern, where you don't really care if it's win or mac, but define the behaviours on initialization and here you just apply whatever is configured.

src/browser/components/MainPage.jsx
```diff
@@ -319,24 +343,40 @@ export default class MainPage extends React.Component {
...
+          showAddServerButton={this.props.showAddServerButton}
+        />
       );
     }
 
+    let topBarClassName = 'topBar';
+    if (process.platform === 'darwin') {
+      topBarClassName += ' macOS';
+      if (this.state.isDarkMode) {
+        topBarClassName += ' darkMode';
```
Willyfrog: is this for debugging purposes?

src/browser/components/TabBar.jsx
```diff
@@ -71,13 +74,15 @@ export default class TabBar extends React.Component { // need "this"
           title='Add new server'
           draggable={false}
         >
-          <Glyphicon glyph='plus'/>
+          <div className='TabBar-tabSeperator'>
+            {'+'}
```
Willyfrog: are we no longer using fonts for icons?

devinbinnie: Not across the board, but the new design uses just a generic `+`

src/browser/components/MainPage.jsx
devinbinnie: Actually, I forgot. It should be a `state` variable, since it can change as the OS changes from light to dark mode.

andrewbrown00:
@devinbinnie do you know why the `Setup Cloud Test Server` label isn't available in under **Labels**?  Was trying to add the label so you'd know to spin up a test server whenever you need me to review.
image

devinbinnie:
> @devinbinnie do you know why the `Setup Cloud Test Server` label isn't available in under **Labels**? Was trying to add the label so you'd know to spin up a test server whenever you need me to review.
> image

We don't have an automated builder setup for desktop app currently, which I guess is a bit of a problem...

andrewbrown00:
@devinbinnie thanks for clarifying 😄 Who's the best person to bring this issue up with?

Willyfrog:
> @devinbinnie thanks for clarifying 😄 Who's the best person to bring this issue up with?

there is no `setup test server` but circle does build the app, although it doesn't sign it or package it. there are a couple of steps to get the build:

1. click on `show all checks`
1. on `build_and_test` click on `details`
1. click on `store_artifacts-1` which will open circle ci job
1. click on `artifacts` tab
1. select and download the app you will test

devinbinnie:
All features complete! Still needs a bit of work here and there, but is mostly ready!

.gitignore
```diff
@@ -14,3 +14,6 @@ test-results.xml
 test_config.json
 .idea
 testUserData
+
+src/browser/*.png
+src/browser/*.svg
```
Willyfrog: can you explain about this two lines? are images being generated within the src folder or something like that?

src/browser/webview/mattermost.js
```diff
@@ -47,6 +47,12 @@ window.addEventListener('load', () => {
   });
 });
 
+if (process.platform !== 'darwin') {
+  window.addEventListener('mousedown', () => {
+    ipcRenderer.sendToHost('closeMenu');
+  });
+}
```
Willyfrog: maybe I'm missing something, but are we sending a `closemenu` event every time the user clicks on linux and windows?
what's the reason for that behaviour?

.gitignore
devinbinnie: Yeah, it's using the webpack image loader, just like it generates the `*_.bundle.js` files. So I have to ignore them since they're generated.

src/browser/components/MainPage.jsx
devinbinnie: So there's two completely different systems at play here. The Mac one relies on this function and a `drag` region, whereas the Windows/Linux one is a custom toolbar npm package that handles all this. I'm not really sure how I'd want to 'inject' them separately, since the app doesn't really provides the correct points of insertion, nor do they require the same points.

I think if we were to change the application to be more dependency injection receptive, then it might make sense to modularize these.

src/browser/webview/mattermost.js
devinbinnie: It's so the `MattermostView` can send an event back to the index to close the top bar menu on Windows.

devinbinnie:
Hey @andrewbrown00,

I've fixed all the bugs listed above, except for the 3-dot menu since that's a new ask and the 2 listed below:
- Remove the lag/partially stuck tab when starting to drag a tab.
- Soften the transition when a tab goes back to its current spot if a user starts dragging but lets go before dropping into a new position, it has a brief flash.

Unfortunately there's nothing I can really do about those as they exist as part of the library.

devinbinnie:
@andrewbrown00 New menu has been added.
I've fixed the cursor issue, the font difference I can't really do anything about I don't think, I think it might just be the way it's being rendered. The other issues have been fixed, except the ones with the library, which I don't think are really worth doing at this point since it's more work than it's worth.

Let me know your thoughts :)

esethna:
Great work guys, some review feedback:

1. I discussed with @andrewbrown00, we should go with the standard Win 10 icons for the close expand and minimize buttons.

image

2. The ... icon feels heavily weighted compared to the "+", see spotify as an example of another ... icon style (0/5 which way we go but it feels off right now) @andrewbrown00 
image

3. The ... menu needs to be accessible. I'd propose ALT focuses the menu and then Enter opens the menu (same as Chrome and Slack).

image

4. @andrewbrown00 can we create tickets for the above known issues if we don't have them already?

Nice work, love this! @devinbinnie @andrewbrown00

esethna:
Adding @stevemudie and @lindy65 to help review this feature. @devinbinnie can you help make sure they have the latest builds for testing once items 1 & 3 are addressed from above. Reminder feature complete is monday for merging if we want this to ship with v4.4 in December.

lindy65:
Thanks @esethna @devinbinnie 

Tested on Win10 and, apart from Eric's issues mentioned in 1) and 3) above, found the following issues:

1. Upon opening the app in non-maximised view, the `minimise`, `maximise` and `close` icons overlap the last tab in view (fixed on maximising the app)

image

2. I am not seeing bold text on the selected tab as per final spec doc comment `Bold text only for selected tabs, unselected tabs use regular text.`  Text is slightly darker for me but not bold.

Spec doc screenshot:
image

What I see:
image

3. I am not seeing the `grabbing` cursor as per final spec doc comment `Cursor will change to the grabbing cursor while the user is dragging the tab`.  
   - On hovering over the tab I see a `hand pointer`
   - On mouse click on the tab I see a `pointer arrow`

no grab cursor

4. Not sure whether this is expected? I am seeing the cursor "float" as I move my mouse if I click on a tab and then move my mouse _below_ the tab bars.  The tab only stops "floating" once I click with my mouse on the tab bar again. (Testing as per final spec doc comment `The tab will stay fixed to the tab bar and only move horizontally left or right). I would expect the action to be "cancelled" if I move my mouse off the tab bar?

click and move mouse off the tab bar, tab floats

5. Tab colour does not change with theme - remains grey when I switch to Windows Dark theme

image

andrewbrown00:
@lindy65 thanks for the review!!!

**Feedback on the reviewed items**
1. That is not the expected behavior @devinbinnie what can we do here with the library? Can we overlay the app frame controls over the tabs with a visual cue the are sitting behind the controls? Or can we resize the tabs like Chrome?
 
2. We decided to go with "Regular" because **bold** was causing a small resize/flash in the tab when selected. **The doc and designs have been update to reflect the change**
3. We decided to always use the pointer cursor to maintain consistency across OS' with the different behaviors and cursors we we seeing with the control we have. **The doc and designs have been update to reflect the change**
4. That is not expected. The tab should continue to follow the cursor while the `mouse down` is still taking place.
5. The theme you set for the app should override the system theme mode.

andrewbrown00:
@esethna thanks for the review!!!

**Feedback on review items**
1. You might have an outdated build, the new WIN 10 app frame icons wen it yesterday
 > I discussed with @andrewbrown00, we should go with the standard Win 10 icons for the close expand and minimize buttons. These ones feel pixelated and aren't commonly seen
2. They do look heavier for you, my VM doesn't show them as heavy. I'll work with @devinbinnie  to resolve that.
3. Agreed
4. **Tickets created:**
   - https://mattermost.atlassian.net/browse/MM-20202
   - https://mattermost.atlassian.net/browse/MM-20203

stevemudie:
Observations on macOS Mojave 10.14.6
**Desktop Version 4.3.0-develop (4.3.0-develop.3916)**

Overall the ability to drag tabs around worked fine and was fluid and intuitive.

I did notice that you could down arrow and do what you’d like with the tab staying put where you left it even far to the right of other tabs. It snaps back to the mouse when you get to the top but won’t stay put unless you click it.

Dragging server tabs

Shrinking the window horizontally stacks the tabs but not in a clear way to allow you to click on them. At times they vanish outright and don’t respond as you’d expect. They do return without issue though when you drag the window wider.

Resizing horizontally

**Other things to note**:
- A server only shows an alert (!) icon on expired sessions - as expected - but not on servers yet to be signed into. Is that expected as well?
 - I was unable to copy text from the 'About' modal window even though by way of any means.
 - The current tab only looks darker in text as Lindy noted instead of bold.
 - I didn’t have any issues with cursor changes it remained standard at all times.

lindy65:
Thanks @andrewbrown00,

> 2. We decided to go with "Regular" because **bold** was causing a small resize/flash in the tab when selected. **The doc and designs have been update to reflect the change**

Ah, ok - thanks - will this be updated in the final spec doc?

> 3. We decided to always use the pointer cursor to maintain consistency across OS' with the different behaviors and cursors we we seeing with the control we have. **The doc and designs have been update to reflect the change**

Thanks :)

> 4. That is not expected. The tab should continue to follow the cursor while the `mouse down` is still taking place.

@devinbinnie - are you going to address this issue on this current PR or would you like me to create a separate ticket?

> 5. The theme you set for the app should override the system theme mode.

Apologies, my wording wasn't clear...  The "Windows Dark" theme I spoke about above _was_ set in the app (one of the options in Account Settings > Display > Theme) but the tabs and tab bar did not adopt the dark colour of the theme...

devinbinnie:
From @lindy65:

> are you going to address this issue on this current PR or would you like me to create a separate ticket?

I'm looking into it but it might be another issue that will require me to replace the Drag and Drop library.

> Apologies, my wording wasn't clear... The "Windows Dark" theme I spoke about above was set in the app (one of the options in Account Settings > Display > Theme) but the tabs and tab bar did not adopt the dark colour of the theme...

That is expected, if you want to set dark mode on the app:
- On Mac: Set your computer to Dark Mode in System Preferences
- On others: Go to 3-dot menu > View > Toggle Dark Mode

From @andrewbrown00 

> That is not the expected behavior @devinbinnie what can we do here with the library? Can we overlay the app frame controls over the tabs with a visual cue the are sitting behind the controls? Or can we resize the tabs like Chrome?

Resizing won't work since the tabs are only as wide as the text is already. I'm not sure what you mean about a visual cue, since they'll still want a way to access those tabs. I'll need some design input here as to what to with tab overflow.

>They do look heavier for you, my VM doesn't show them as heavy. I'll work with @devinbinnie to resolve that.

Let me know what you want to do here.

From @stevemudie 

> A server only shows an alert (!) icon on expired sessions - as expected - but not on servers yet to be signed into. Is that expected as well?

Yes, that is expected.

> I was unable to copy text from the 'About' modal window even though by way of any means.

If you mean the About Mattermost dialog from the Mac topbar, then yes there's no way to copy that text. That's a separate issue from this PR. If you mean from within Mattermost, I didn't have any issues doing so, can you send a video if so?

> The current tab only looks darker in text as Lindy noted instead of bold.

Correct, that is expected.

> I didn’t have any issues with cursor changes it remained standard at all times.

That is also expected.

stevemudie:
@devinbinnie Thank you for your feedback.

As for the copy/paste issue. It's not from within Mattermost other than being part of the menu bar at the top. In all previous versions I've been able to copy text without issue as I can with all other apps to my knowledge.

Two GIFs below. One of the current app and one showing the develop build here.

I use a utility called _PopClip_ to copy as you'll see but standard keyboard copy/paste commands have the same effect on either/both apps.

**Current app**
Version 4 3 1 (4 3 1 3679)





**Develop version**
Version 4 3 0-develop (4 3 0-develop 3916)

Please let me know if you need me to write it up as a bug.

devinbinnie:
@stevemudie Yeah go ahead and make a bug, that same issue is happening on the current master.

esethna:
@devinbinnie @andrewbrown00 I think it's worth investigating other drag/drop libraries (react dnd) to try and solve a lot of these issues at once. I think DB has a ticket for this already?

andrewbrown00:
> @devinbinnie @andrewbrown00 I think it's worth investigating other drag/drop libraries (react dnd) to try and solve a lot of these issues at once. I think DB has a ticket for this already?

@devinbinnie let's chat about this tomorrow and get an idea on the mana and if the library will resolve the issues. We can also talk through how we'll handle the tab overflow.

devinbinnie:
@deanwhillier @esethna @andrewbrown00 @lindy65 @stevemudie 
This is now ready for review. All issues above have been addressed.

lindy65:
Hi @devinbinnie @andrewbrown00 @esethna 

I've retested on Win10 on the latest build as per my previous feedback as well as Eric's issues 1) and 3) above and:

From Eric above:
> 1. I discussed with @andrewbrown00, we should go with the standard Win 10 icons for the close expand and minimize buttons.

This is fixed

From Eric above:
> The ... menu needs to be accessible. I'd propose ALT focuses the menu and then Enter opens the menu

This is also fixed

From my feedback above:
> 4. Not sure whether this is expected? I am seeing the cursor "float" as I move my mouse if I click on a tab and then move my mouse below the tab bars. The tab only stops "floating" once I click with my mouse on the tab bar again. (Testing as per final spec doc comment `The tab will stay fixed to the tab bar and only move horizontally left or right). I would expect the action to be "cancelled" if I move my mouse off the tab bar?

I've filed https://mattermost.atlassian.net/browse/MM-20652 as a separate ticket to address this issue.

I will leave the QA review label on this PR until @stevemudie has done his review and he can then remove it if no issues are found.

deanwhillier:
Hey @devinbinnie, looks like there may be an issue with Catalina and dark mode. I have dark mode enabled, but the top bar of the app is showing as light for me right now: 
image

deanwhillier:
On Windows, mousing over the close window 'X' button on the top right should change the icon to white from black.

Current desktop app: 
image

Current File Explorer app: 
image

deanwhillier:
Is the dark highlight under the Windows menu and Mac window controls intentional or related to dark mode glitches?

Windows: 
image

Mac: 
image

devinbinnie:
> Is the dark highlight under the Windows menu and Mac window controls intentional or related to dark mode glitches?
> 
> Windows:
> image
> 
> Mac:
> image

Under Windows it's intentional, since we need a way to show that the 3-dot menu is focused. On Mac it's a glitch, but it should be fixable.

devinbinnie:
@deanwhillier Fixed the above issues other than Catalina dark mode, I have a ticket to investigate that: https://mattermost.atlassian.net/browse/MM-20757

src/browser/components/MainPage.jsx
```diff
@@ -405,29 +533,102 @@ export default class MainPage extends React.Component {
...
+            className='three-dot-menu'
+            onClick={this.openMenu}
+            tabIndex={0}
+            ref={this.threeDotMenu}
+          >
+            <DotsVerticalIcon/>
+          </button>
+          {tabsRow}
+          <span className='overlay-gradient'/>
+          <span className='title-bar-btns'>
```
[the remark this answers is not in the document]
devinbinnie: I can do that for the above elements, but the 3-dot menu element needs to stay to ensure the flexbox renders properly.

src/browser/index.html
```diff
@@ -4,6 +4,10 @@
 <head>
   <meta charset="UTF-8">
   <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.min.css">
+  <link 
+    href="https://fonts.googleapis.com/css?family=Open+Sans:400,600&display=swap" 
+    rel="stylesheet"
+  >
```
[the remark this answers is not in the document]
devinbinnie: Probably bad things. I will fix this.

src/common/config/RegistryConfig.js
```diff
@@ -72,6 +72,7 @@ export default class RegistryConfig extends EventEmitter {
         teams.push({
           name: team.name,
           url: team.value,
+          order: team.order,
```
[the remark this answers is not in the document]
devinbinnie: How are these fields mapped to the GPO fields? According to the definitions we have `"Value name" => name` and `"value" => url`

devinbinnie:
@deanwhillier Addressed all your feedback except the GPO bit, which I left a question for.
@esethna I've addressed #2, #1 is going to take some deeper investigation, #3 will need @andrewbrown00 for.

devinbinnie:
@esethna All feedback addressed, waiting on final approval.

andrewbrown00:
@esethna answers to your previous comment:

1. @devinbinnie will handle this
2. New icons have been provided and are in the latest build
image
3. Separate ticket created https://mattermost.atlassian.net/browse/MM-21329
4. Resolved by `#2`

esethna:
@devinbinnie is this ready for final review?

devinbinnie:
> @devinbinnie is this ready for final review?

Yep!

esethna:
We can merge this

Willyfrog:
re-running the workflow as it doesn't make any sense to fail there

devinbinnie:
@Willyfrog could you do a quick re-review before we merge to make sure everything is up to standards?

src/browser/components/MattermostView.jsx
```diff
@@ -193,6 +193,12 @@ export default class MattermostView extends React.Component {
       case 'onNotificationClick':
         self.props.onNotificationClick();
         break;
+      case 'mouse-move':
+        this.handleMouseMove(event.args[0]);
+        break;
+      case 'mouse-up':
+        this.handleMouseUp();
+        break;
```
Willyfrog: why do we need to handle movement and release of button?

src/browser/components/TabBar.jsx
```diff
@@ -2,12 +2,17 @@
...
+import {Nav, NavItem} from 'react-bootstrap';
+import {Container, Draggable} from 'react-smooth-dnd';
+import PlusIcon from 'mdi-react/PlusIcon';
 
 export default class TabBar extends React.Component { // need "this"
   render() {
-    const tabs = this.props.teams.map((team, index) => {
+    const orderedTabs = this.props.teams.concat().sort((a, b) => a.order - b.order);
+    const tabs = orderedTabs.map((team) => {
+      const index = this.props.teams.indexOf(team);
```
Willyfrog: you can get the index out of the map function as a parameter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/browser/components/TabBar.jsx
```diff
@@ -29,37 +34,53 @@ export default class TabBar extends React.Component { // need "this"
...
+        badgeDiv = (
+          <div className='TabBar-dot'/>
+        );
       }
-      const id = 'teamTabItem' + index;
 
-      // draggable=false is a workaround for https://github.com/mattermost/desktop/issues/667
-      // It would obstruct https://github.com/mattermost/desktop/issues/478
-      return (
+      const id = 'teamTabItem' + index;
```
Willyfrog:
```suggestion
      const id = `teamTabItem${index}`;
```
this is optional, just for consistency

src/main.js
```diff
@@ -461,6 +465,19 @@ function handleAppWebContentsCreated(dc, contents) {
...
+      mainWindow.webContents.send('focus-three-dot-menu');
+      return;
+    }
+
+    // Hack to detect keyPress so that alt+<key> combinations don't default back to the 3-dot menu
+    if (input.key === 'Alt' && input.type === 'keyDown') {
+      altLastPressed = true;
+    } else {
+      altLastPressed = false;
+    }
```
Willyfrog: why is ALT a shorcut? wouldn't that interfere not only with alt+up/down but with anything regarding the OS from the user if he has any shortcut using that? whouldn't it make more sense to use a separate shortcut for that that doesn't use a modifier key as the only one?
non-comprehensive list of things that might break:
- alt+f4 on windows (as an example of OS related)
- alt+right click for mark as unread (as an example of MM related)

src/main/Validator.js
```diff
@@ -61,6 +61,29 @@ const configDataSchemaV1 = Joi.object({
   spellCheckerLocale: Joi.string().regex(/^[a-z]{2}-[A-Z]{2}$/).default('en-US'),
 });
 
+const configDataSchemaV2 = Joi.object({
+  version: Joi.number().min(2).default(1),
```
Willyfrog:
```suggestion
  version: Joi.number().min(2).default(2),
```
also, I'm not really sure if min should be `1`, as I don't know if the validator is running before or after the upgrade

src/main/Validator.js
```diff
@@ -61,6 +61,29 @@ const configDataSchemaV1 = Joi.object({
...
+  notifications: Joi.object({
+    flashWindow: Joi.any().valid(0, 2).default(0),
+    bounceIcon: Joi.boolean().default(false),
+    bounceIconType: Joi.any().allow('').valid('informational', 'critical').default('informational'),
+  }),
+  showUnreadBadge: Joi.boolean().default(true),
+  useSpellChecker: Joi.boolean().default(true),
+  enableHardwareAcceleration: Joi.boolean().default(true),
+  autostart: Joi.boolean().default(true),
+  spellCheckerLocale: Joi.string().regex(/^[a-z]{2}-[A-Z]{2}$/).default('en-US'),
```
Willyfrog: i wouldn't change it in this PR as it has grown big enough, but shouldn't we consider using system's locale as default? we should create a ticket for that

package.json
```diff
@@ -64,11 +65,14 @@
...
     "eslint-plugin-react": "^7.11.1",
     "file-loader": "^2.0.0",
+    "image-webpack-loader": "5.0.0",
+    "mdi-react": "^6.2.0",
     "mocha": "^5.2.0",
     "mocha-circleci-reporter": "0.0.3",
     "npm-run-all": "^4.1.5",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
+    "react-smooth-dnd": "github:devinbinnie/react-smooth-dnd#a72095d3fa1fd6ffc3a6b797174d019a07d7b011",
```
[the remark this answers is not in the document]
devinbinnie: Yes, though i'm not sure how to go about adding a fork to the MM repo.

scripts/fix_catalina_dark_mode_debug.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
devinbinnie: It's only used for debug mode, I think the name's okay.

src/browser/components/MattermostView.jsx
devinbinnie: This is a fix for the tab dragging and dropping, since if you leave the bounds of the DnD area before dropping, it doesn't drop the tab. This does it manually/

src/main.js
devinbinnie: It's a manual override to do something the OS already does, which is open a window context menu  when you press ALT. It was part of @esethna's suggestions for accessibility.

I agree it would be better to use a different shortcut, but that's why we're checking for keyup and keydown to make sure it's not being used as a combo.

src/main/Validator.js
devinbinnie: min should be 2 for the v2 config, it should fail on that and go through the upgrade process.

src/main/Validator.js
devinbinnie: Probably a good idea!

package.json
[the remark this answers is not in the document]
hmhealey: You'll have to transfer ownership of it to someone who has admin access to the GitHub organization. I'll message you more details.

PR #1061, opened 2019-10-08
[MM-19266] User activity monitor updates

deanwhillier:
**Summary**
In this PR:
- re-work of the mechanism for determining user activity status and updating the server via the webapp
- removing system events (login/out screensaveer on/of etc.) to be re-considered for a future release – prompted by stability issues that aren't readily resolvable given the current implementation of status management in MM

**Issue link**
https://mattermost.atlassian.net/browse/MM-19266
https://mattermost.atlassian.net/browse/MM-19302

### DIFF
--- a/src/main/UserActivityMonitor.js
+++ b/src/main/UserActivityMonitor.js
@@ -38,9 +38,9 @@ export default class UserActivityMonitor extends EventEmitter {
    * Begin monitoring system events and idle time at defined frequency
    *
    * @param {Object} config - overide internal configuration defaults
-   * @param {number} config.updateFrequencyMs
-   * @param {number} config.inactiveThresholdSec
-   * @param {number} config.statusUpdateThresholdMs
+   * @param {number} config.updateFrequencyMs - internal update clock frequency for monitoring idleTime
+   * @param {number} config.inactiveThresholdSec - number in seconds the idleTime needs to reach to internally be considered inactive
+   * @param {number} config.statusUpdateThresholdMs - minimum amount of time before sending a new status update
    * @emits {error} emitted when method is called before the app is ready
    * @emits {error} emitted when this method has previously been called but not subsequently stopped
    */
--- a/src/main/UserActivityMonitor.js
+++ b/src/main/UserActivityMonitor.js
@@ -15,25 +15,19 @@ export default class UserActivityMonitor extends EventEmitter {
     super();
 
     this.isActive = true;
-    this.forceInactive = false;
     this.idleTime = 0;
-    this.lastStatusUpdate = 0;
+    this.lastSetActive = null;
     this.systemIdleTimeIntervalID = -1;
 
     this.config = {
-      internalUpdateFrequencyMs: 1 * 1000, // eslint-disable-line no-magic-numbers
+      updateFrequencyMs: 1 * 1000, // eslint-disable-line no-magic-numbers
+      inactiveThresholdSec: 60, // eslint-disable-line no-magic-numbers
       statusUpdateThresholdMs: 60 * 1000, // eslint-disable-line no-magic-numbers
-      activityTimeoutSec: 5 * 60, // eslint-disable-line no-magic-numbers
     };
-
-    // NOTE: binding needed to prevent error; fat arrow class methods don't work in current setup
-    // Error: "Error: async hook stack has become corrupted (actual: #, expected: #)"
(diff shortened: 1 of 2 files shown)


src/main/UserActivityMonitor.js
```diff
@@ -77,51 +65,52 @@ export default class UserActivityMonitor extends EventEmitter {
...
-      this.isActive = isActive;
-      this.sendStatusUpdate(now, isSystemEvent);
-    } else if (now - this.lastStatusUpdate > this.config.statusUpdateThresholdMs) {
-      this.sendStatusUpdate(now, isSystemEvent);
+
+    if (isActive && (this.lastSetActive == null || now - this.lastSetActive >= this.config.statusUpdateThresholdMs)) {
+      this.sendStatusUpdate(false);
+      this.lastSetActive = now;
+    } else if (!isActive) {
+      this.lastSetActive = null;
```
Willyfrog: shouldn't it keep at whatever value was before? otherwise it'll set as inactive next time.

src/main/UserActivityMonitor.js
```diff
@@ -44,9 +38,9 @@ export default class UserActivityMonitor extends EventEmitter {
    * Begin monitoring system events and idle time at defined frequency
    *
    * @param {Object} config - overide internal configuration defaults
-   * @param {nunber} config.internalUpdateFrequencyMs
-   * @param {nunber} config.statusUpdateThresholdMs
-   * @param {nunber} config.activityTimeoutSec
+   * @param {number} config.updateFrequencyMs
+   * @param {number} config.inactiveThresholdSec
+   * @param {number} config.statusUpdateThresholdMs
```
Willyfrog: missing description

src/main/UserActivityMonitor.js
deanwhillier: If you mean `this.lastSetActive = null` then no, that's being done on purpose to allow for an immediate update when the desktop app detects the user going from inactive to active. Basically allows the desktop app to set you back to online immediately when you start interacting with your computer after being away.

src/main/UserActivityMonitor.js
```diff
@@ -77,51 +65,52 @@ export default class UserActivityMonitor extends EventEmitter {
...
    * @param {boolean} isActive
    * @param {boolean} isSystemEvent – indicates whether the update was triggered by a system event (log in/out, screesaver on/off etc)
    * @private
    */
-  updateUserActivityStatus(isActive = false, isSystemEvent = false) {
+  setActivityState(isActive = false, isSystemEvent = false) {
+    this.isActive = isActive;
+
+    if (isSystemEvent) {
+      this.sendStatusUpdate(true);
```
lieut-data: Do system events still occur? If so, should we be returning early here, or do we still want to fall through and set `lastSetActive` and potentially invoke `sendStatusUpdate` a second time?

src/main/UserActivityMonitor.js
```diff
@@ -15,25 +15,19 @@ export default class UserActivityMonitor extends EventEmitter {
...
     this.isActive = true;
-    this.forceInactive = false;
     this.idleTime = 0;
-    this.lastStatusUpdate = 0;
+    this.lastSetActive = null;
     this.systemIdleTimeIntervalID = -1;
 
     this.config = {
-      internalUpdateFrequencyMs: 1 * 1000, // eslint-disable-line no-magic-numbers
+      updateFrequencyMs: 1 * 1000, // eslint-disable-line no-magic-numbers
```
lieut-data: Nit: it’s annoying having mixed units. Worth converging on milliseconds?

src/main/UserActivityMonitor.js
deanwhillier: Gah, I'm seeing double ... thanks for catching that! So, for v4.3, system events have been removed, but I left the infrastructure in for later. I could remove it all the way back to just before transitioning to the webapp. The webapp is expecting a boolean for `manual`, which is set based on `isSystemEvent` from this file ... I could hardcode to `false` for now. Thoughts?

src/main/UserActivityMonitor.js
deanwhillier: Ya, I agree. I guess since I'm already overriding eslint's `no-magic-numbers`, another one for converting to seconds won't hurt.

deanwhillier:
Thanks @lieut-data and thanks for your help in working this out with me!

PR #1079, opened 2019-10-17
Update CHANGELOG.md to clarify known issues

esethna:


esethna:
@amyblais 0/5 if we move the known issues section to the top of the changelog or if you have another method to communicate an upcoming dot release

amyblais:
@esethna I think we should first decide the fixes for v4.3.1 and then format the upcoming v4.3.1 fixes at the top similar to how we do in the server changelog: https://github.com/mattermost/docs/pull/3083/files.

esethna:
@amyblais , makes sense, updated the format. I think all listed should go into the dot release.

esethna:
@deanwhillier please help merge when you can

PR #1093, opened 2019-10-25
Fixed email links being rejected by will-navigate

devinbinnie:

**Summary**
The changes we made to enforce the use of trusted URLs has shut off the `mailto` protocol. This PR adds an exception for it.

**Issue link**
https://mattermost.atlassian.net/browse/MM-19596

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -389,6 +389,9 @@ function handleAppWebContentsCreated(dc, contents) {
     if (isTrustedURL(parsedURL) || isTrustedPopupWindow(event.sender)) {
       return;
     }
+    if (parsedURL.protocol === 'mailto:') {
+      return;
+    }
     if (customLogins[contentID].inProgress) {
       return;
     }


devinbinnie:
Agreed @Willyfrog, we need to think about how to deal with other protocols so that most links work correctly.

deanwhillier:
@Willyfrog, I believe properly answering your question will need to include the security team as the `will-navigate` listener was originally implemented by them as a security enhancement.

How did you link the `ftp` URI above? When I paste it into a new message either directly, or using a markdown link it works properly for me. It looks like links in posts end up with `target="blank"`, which is actually specifically handled by the Desktop app before reaching the `will-navigate` listener. If the clicked URL is not within the server domain, the desktop app is supposed to attempt to open the link externally (browser, ftp client etc.). Additionally, that same process includes a protocol check that asks for user permission if the protocol for the link has not been previously trusted (when not using `http(s)`). So in my case, the first time I clicked on the `ftp` link, the app asked me if I wanted to open it in an external application and after I clicked the 'yes and save' option, the link was opened in my ftp client and I wasn't asked again.

The problem this PR addresses is that `mailto:` links don't seem to get `target="_blank"` by the Webapp and so the existing 'new window' handler for links in the Desktop app isn't getting triggered and is thereby causing the `will-navigate` listener to trigger, which stops it dead in its tracks.

That said, Devin and I were thinking that maybe some common protocol's like `mailto:` and `ftp:` should be automatically trusted, whether handled by the 'new window' handler or the `will-navigate` listener. IMO, anything other than common protocol's should probably still be approved by the user before being used for the first time, but @crspeller and/or @DSchalla would be best to answer that question.

So, in my mind the questions are:
* Should the webapp make sure that any links that aren't local to the current hostname and/or using the `http(s)` protocols, should force the link to open in a new window to be resolved outside the webapp?
* If not entirely the webapp's responsibility, should we enhance the Desktop app to catch any triggered link, evaluate it and determine an appropriate course of action regardless of whether the link is handled by the 'new window' or the `will-navigate` listeners, only resorting to blocking the link as a last resort?
* Both? (probably)
* Should common protocols be automatically trusted by the Desktop app? (mailto, ftp, others?)

Willyfrog:
So the main reason for the ticket and my question is exactly that, to start the conversation rolling and see if it makes sense to do anything of the above

PR #1105, opened 2019-11-13
[MM-19963] set working directory on start

Willyfrog:
**Summary**
Ensure working directory is set to where the executable is to prevent problems

**Issue link**
[MM-19963](https://mattermost.atlassian.net/browse/MM-19963)
**Test Cases**
run from the command line from a directory different from where it is installed
use procmon or some other process monitoring tool that can provide information of the working directory and check that is properly set.

Screenshot 2019-11-13 at 12 41 16

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -89,11 +89,14 @@ const customLogins = {};
  */
 async function initialize() {
   process.on('uncaughtException', criticalErrorHandler.processUncaughtExceptionHandler.bind(criticalErrorHandler));
+  global.willAppQuit = false;
 
   // prevent using a different working directory, which happens on windows running after installation.
-  process.chdir(path.dirname(process.execPath));
-
-  global.willAppQuit = false;
+  const expectedPath = path.dirname(process.execPath);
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -89,9 +89,15 @@ const customLogins = {};
  */
 async function initialize() {
   process.on('uncaughtException', criticalErrorHandler.processUncaughtExceptionHandler.bind(criticalErrorHandler));
-
   global.willAppQuit = false;
 
+  // prevent using a different working directory, which happens on windows running after installation.
```
jupenur: The `chdir` call should probably come after `initializeArgs`. Command-line `--data-dir` resolves relative paths against the cwd on line 149, and there the cwd should still come from the shell.

Maybe move this into `initializeBeforeAppReady` instead? That should still be early enough to address the original issue.

jgilliam17:
Reassigning to @stevemudie to verify on Windows machine.
Thanks Steve!

Willyfrog:
done cherrypicking on release-4.3.0

PR #1106, opened 2019-11-18
[MM 19649] Fix blank window when auto-starting the app on login

deanwhillier:
**Summary**
Prior to the upgrade to Electron v5 (from v2), the app would immediately minimize the app window once created. It would then wait for the window's 'ready-to-show' event fired and call the `.show()` method to show the UI in the window while the window remained minimized. Following the upgrade of Electron, calling the `.show()` method un-minimizes the Window, but the UI doesn't get shown resulting in a blank window (I suspect this has something to do with not being told the Window is visible and/or not getting a resize event to re-render the UI). Although the below ticket only refers to Windows, the reality is that it can be repro'd on other OS's by using the supported `--hidden` flag when starting the app from command line.

This PR sets the initial `show` option when creating the window based on whether the auto-launch setting is configured or not, i.e. the app UI is not hidden by default when the app is started by auto-launch since the entire window is not visible. This avoids the blank screen when the app is auto launched (or started with the `--hidden` flag as the UI is never hidden to begin with.

The PR also tweaks the code that restores the window when a user clicks on the tray and/or running app icon.

Updated non-macOS app behaviour – when the app is auto-launched:

1. if the app is configured to show a tray icon, upon login only the tray icon is shown. The user can click on the tray icon to reveal the window. Note: in Windows, the tray icon will always be shown, i.e. there is no setting to turn the tray icon off in Windows
2. if the app is configured to NOT show a tray icon, upon login the app window is immediately minimized into the running app icon. The user can click on the running app icon to reveal the window.

Updated macOS app behaviour – when the app is auto-launched, it will always minimize into the app icon whether there is a tray icon present or not.

Note: in all cases, if the app was maximized when it was last quit, it will be automatically be maximized again when the user clicks the tray icon or app icon after auto-start.

**Issue link**
https://mattermost.atlassian.net/browse/MM-19649

### DIFF
--- a/src/main/Validator.js
+++ b/src/main/Validator.js
@@ -8,6 +8,8 @@ const defaultOptions = {
   stripUnknown: true,
 };
 
+const defaultMinWindowXPos = -100;
+const defaultMinWindowYPos = -100;
 const defaultWindowWidth = 1000;
 const defaultWindowHeight = 700;
 const minWindowWidth = 400;
@@ -23,8 +25,8 @@ const argsSchema = Joi.object({
 });
 
 const boundsInfoSchema = Joi.object({
-  x: Joi.number().integer().min(0),
-  y: Joi.number().integer().min(0),
+  x: Joi.number().integer().min(defaultMinWindowXPos).default(0),
+  y: Joi.number().integer().min(defaultMinWindowYPos).default(0),
   width: Joi.number().integer().min(minWindowWidth).required().default(defaultWindowWidth),
   height: Joi.number().integer().min(minWindowHeight).required().default(defaultWindowHeight),
   maximized: Joi.boolean().default(false),
--- a/src/main/mainWindow.js
+++ b/src/main/mainWindow.js
@@ -68,6 +68,10 @@ function createMainWindow(config, options) {
 
   mainWindow.once('ready-to-show', () => {
     mainWindow.webContents.setZoomLevel(0);
+
+    // handle the initial state of the app on launch
+    // 1. When not configured to hide on startup, immediately show contents and optionally maximize as needed
+    // 2. When configured to hide on startup and there is no tray icon, show contents and minimize to the app icon
     if (!hideOnStartup) {
       mainWindow.show();
       if (windowIsMaximized) {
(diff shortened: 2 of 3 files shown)


src/main/Validator.js
```diff
@@ -23,8 +25,8 @@ const argsSchema = Joi.object({
 });
 
 const boundsInfoSchema = Joi.object({
-  x: Joi.number().integer().min(0),
-  y: Joi.number().integer().min(0),
+  x: Joi.number().integer().min(defaultMinWindowXPos).default(0),
```
deanwhillier: This was changed to prevent the window size from resetting when maximized in Windows as the reported x/y coordinate when maximized in Windows is -8/-8.

src/main/Validator.js
```diff
@@ -23,8 +25,8 @@ const argsSchema = Joi.object({
...
+  y: Joi.number().integer().min(defaultMinWindowYPos).default(0),
```
deanwhillier: This was changed to prevent the window size from resetting when maximized in Windows as the reported x/y coordinate when maximized in Windows is -8/-8.

src/main/mainWindow.js
```diff
@@ -63,34 +66,49 @@ function createMainWindow(config, options) {
...
-  if (options.hideOnStartup) {
-    if (windowOptions.maximized) {
-      mainWindow.maximize();
-    }
-
-    // on MacOS, the window is already hidden until 'ready-to-show'
-    if (process.platform !== 'darwin') {
+  // handle hiding the app when launched by auto-start
+  if (hideOnStartup) {
+    if (trayIconShown && process.platform !== 'darwin') {
```
deanwhillier: No longer ignoring hideOnStartup when in macOS because this is also true when `--hidden` flag is passed via command line so we also need to minimize for macOS.

deanwhillier:
Added @stevemudie instead of @jgilliam17 for testing on Windows. @jgilliam17, feel free to add yourself as well if you would like to test on Mac. Although the Jira ticket only references Windows, the issue is actually cross-platform.

deanwhillier:
@amyblais, this is the PR for the other Desktop app dot release ticket.

PR #1124, opened 2019-12-10
[MM-14135] allow negative window app positions

Willyfrog:
**Summary**
Allow for negative window position, so the app can be set to a monitor to the left or on top of the primary display

**Issue link**
[MM-14135](https://mattermost.atlassian.net/browse/MM-14135)

**Test Cases**

*** Case 1 ***
On a multimonitor setup, set as main display the rightest monitor (or the bottom one if they are stacked)
open mattermost
move the window to any monitor not on the main display
quit the app
reopen

expected: window position is as we left it. Notice that if we leave the window partially outside of the screen it might reposition on reopen.

*** Case 2 ***
Same as on Case 1, but disconnect the extra display before launching
expected: window goes to main display.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -997,23 +997,22 @@ function clearAppCache() {
   }
 }
 
-function getValidWindowPosition(state, screen) {
+function isWithinDisplay(state, display) {
+  // given a display, check if window is within it
+  return (state.x > display.maxX || state.y > display.maxY || state.x < display.minX || state.y < display.minY);
+}
+
+function getValidWindowPosition(state) {
   // Check if the previous position is out of the viewable area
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -1024,15 +1023,20 @@ function getValidWindowPosition(state, screen) {
 }
 
 function resizeScreen(screen, browserWindow) {
+  console.log('resize screen called!');
```
devinbinnie: Do we need to keep all of these `console.log` statements?

Willyfrog: definitely not, not sure why the linter didn't complain :(

PR #1125, opened 2019-12-10
[MM-20350] Block Basic auth for non trusted URLs

devinbinnie:

**Summary**
This PR fixes an issue where Basic Authentication challenges can be fired from the Mattermost app from arbitrary domains. This fix locks down the domains to only trusted URLS (ie. the URLs used by the Mattermost servers you have configured).

**Issue link**
https://mattermost.atlassian.net/browse/MM-20350

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -352,6 +352,9 @@ function handleAppGPUProcessCrashed(event, killed) {
 
 function handleAppLogin(event, webContents, request, authInfo, callback) {
   event.preventDefault();
+  if (!isTrustedURL(request.url)) {
+    return;
+  }
   loginCallbackMap.set(JSON.stringify(request), callback);
   mainWindow.webContents.send('login-request', request, authInfo);
 }


src/main.js
```diff
@@ -352,6 +352,9 @@ function handleAppGPUProcessCrashed(event, killed) {
 
 function handleAppLogin(event, webContents, request, authInfo, callback) {
   event.preventDefault();
+  if (!isTrustedURL(request.url)) {
+    return;
+  }
```
deanwhillier: One quick question, should the conditional be above `event.preventDefault();`? That's supposed to prevent the login request from it's default of cancelling authentication: https://electronjs.org/docs/api/app#event-login

devinbinnie: I don't think it matters, it just needs to happen before the event is sent back to the renderer. Unless there's something in Electron that stays open as a result of escaping that method this way.

deanwhillier: Ya, that's the part I'm not sure about.

Willyfrog: but if it is a trusted URL (like when we ever implement white list of urls, or if the server is behind a proxy), would this prevent auth as it is before checking for it?

devinbinnie: No, that's what `preventDefault()` is for, because the default action of Electron is to cancel authentication (see Dean's link above)

PR #1126, opened 2019-12-11
[MM-19205] Allow the use of v2 configs by v1 config desktop apps

devinbinnie:
**Summary**
This PR changes the config to allow newer versions of the config to attempt to use the version of the config currently supported by the version of the app being used (ie. if you have a v2 config, the app will try to treat it as a v1 config)

**Issue link**
https://mattermost.atlassian.net/browse/MM-19205

### DIFF
--- a/src/common/config/index.js
+++ b/src/common/config/index.js
@@ -209,11 +209,9 @@ export default class Config extends EventEmitter {
       configData = this.readFileSync(this.configFilePath);
 
       // validate based on config file version
-      switch (configData.version) {
-      case 1:
+      if (configData.version > 0) {
         configData = Validator.validateV1ConfigData(configData);
-        break;
-      default:
+      } else {
         configData = Validator.validateV0ConfigData(configData);
       }
       if (!configData) {


devinbinnie:
This will work for the move from v2 -> v1 since the schemas are so similar.
If any keys are REMOVED, we will need an explicit upgrade path.
I've chosen not to support v0 as the settings are vastly different, and it only supported one server.

jgilliam17:
@devinbinnie I tested this on the test app using the steps listed in the linked jira issue and it is not working when downgrading during step 7.  - I am logged out from all the servers when going back to 4.2.3

devinbinnie:
> @devinbinnie I tested this on the test app using the steps listed in the linked jira issue and it is not working when downgrading during step 7. - I am logged out from all the servers when going back to 4.2.3

@jgilliam17 Sorry the ticket is unclear, we are unable to fix the logout issue between 4.2 and 4.3 due to the Electron upgrade. It's in the comments.

As long as your configuration (ie. the servers themselves) are preserved between 4.2, 4.3 and master (which will be 4.4), then this should be considered fixed. Let me know if you have more questions.

PR #1130, opened 2019-12-12
[MM-21075] Prevent known teams to open in a new app window

Willyfrog:
**Summary**
When a link to a team is posted and the user has that server on the list of tabs on the desktop app, it should only open in the browser, it shouldn't open a new desktop window with the server

**Issue link**

[MM-21075](https://mattermost.atlassian.net/browse/MM-21075)

**Test Cases**

Repro steps:
  1) Add mysql.test.mattermost.com as a server on the desktop app as well as another server
  2) On the other server, post a link to https://mysql.test.mattermost.com/
  3) Click on that link

Observed: A new desktop app window opens to mysql.test and a browser tab for mysql.test also opens

Expected: Nothing should open another desktop app window. Current expected is that it opens in browser tab.

**Additional Notes**
It is not expected to change tabs within the app

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -432,6 +432,10 @@ function handleAppWebContentsCreated(dc, contents) {
       log.info(`Untrusted popup window blocked: ${url}`);
       return;
     }
+    if (isTrustedTeam(url)) {
+      log.info('this is a known team, preventing to open a new window');
+      return;
+    }
     if (popupWindow && popupWindow.getURL() === url) {
       log.info(`Popup window already open at provided url: ${url}`);
       return;
@@ -852,7 +856,7 @@ function parseURL(url) {
   }
 }
 
-function isTrustedURL(url) {
+function isTrustedTeam(url) {
   const parsedURL = parseURL(url);
   if (!parsedURL) {
     return false;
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -432,6 +432,10 @@ function handleAppWebContentsCreated(dc, contents) {
       log.info(`Untrusted popup window blocked: ${url}`);
       return;
     }
+    if (isKnownTeam(url)) {
+      log.info('this is a known team, preventing to open a new window');
+      return;
```
devinbinnie: Just curious, does this stop the link from being opened altogether?

Willyfrog: no, it still opens in the browser

src/main.js
```diff
@@ -432,6 +432,10 @@ function handleAppWebContentsCreated(dc, contents) {
       log.info(`Untrusted popup window blocked: ${url}`);
       return;
     }
+    if (isKnownTeam(url)) {
```
hmhealey: This logic doesn't make sense to me. The if-block right before this is checks `!isTrustedURL(url)`, but since `isTrustedURL` just calls `isKnownTeam`, this code block seems to work like
```javascript
if (!isKnownTeam(url)) {
    return;
}
if (isKnownTeam(url)) {
    return;
}
```
When is it supposed to continue past both of these if statements, and does it actually work in that case?

Willyfrog: yeah, good catch, I totally misread a part

PR #1137, opened 2019-12-17
[MM-20363] prevent using api to navigate within the app's window

Willyfrog:
**Summary**
Image proxy api call was being used to navigate the app to an untrusted url, this PR prevents using any api endpoint to navigate the window

**Issue link**
[MM-20363](https://mattermost.atlassian.net/browse/MM-20363)

**Test Cases**

1. Connect the desktop app to a server that has Image Proxying disabled.
2. Post the following message to a chat: `[let's hack](/api/v4/image?url=https://example.com/)`
3. Click on the link.


src/main.js
```diff
@@ -430,10 +430,12 @@ function handleAppWebContentsCreated(dc, contents) {
     event.preventDefault();
     if (!isTrustedURL(url)) {
       log.info(`Untrusted popup window blocked: ${url}`);
+      shell.openExternal(url);
```
jupenur: Can we be sure that `url` has a trusted scheme here? I.e `http`, `https`, `mattermost`, or something from `allowedProtocols.json`.

Willyfrog: You mean for opening in the user browser? should we be restrictive about that?

src/main.js
```diff
@@ -430,10 +430,12 @@ function handleAppWebContentsCreated(dc, contents) {
...
       return;
     }
     if (isTeamUrl(url) === true) {
       log.info(`${url} is a known team, preventing to open a new window`);
+      shell.openExternal(url);
```
saturninoabril: Tested and looks good, only that I'm seeing belong in the log which is not meaningful
```bash
[17:47:10.227] [info] http://localhost:8065/api/v4/image?url=https://example.com/ is a known team, preventing to open a new window
```

src/main.js
jupenur: Yes. The problem is `shell.openExternal` doesn't _just_ open the default web browser, it's a potential execution sink. See e.g. https://github.com/doyensec/electronegativity/wiki/OPEN_EXTERNAL_JS_CHECK

src/main.js
Willyfrog: I'll probably have to update the log message, as it doesn't apply in this case, originally was trying to prevent only teams from displaying

src/main.js
Willyfrog: i'm removing this line, as apart from what you mention it also means that links open twice in the browser

src/main.js
Willyfrog: added a separate case for api, so it is properly handled

Willyfrog:
I've been looking at this the whole day, so I'm holding it for now.
regarding the double window, the `new-window` event is being called twice from a single click, I've yet to find the source of the second event. This is only happening on community-* can't repro on my local server.

> Opening /api/ URLs in the browser is allowed?

Not sure where should that happen if it's not on the browser, I don't think we should allow them on the app for sure. Any thoughts on this one?

> Opening "known team" URLs is not allowed (at all?)?

currently not, on a future patch from devin it will trigger adding a server or opening the server's tab if it's already configured.

> Things like /admin_console/ open in the main window by default, but can also be opened in a popup?

That was the behaviour not long ago and we decided to remove that. 

> [url with .. in it](/admin_console/../api/v4/image?url=http://example.com)

this is quite interesting, but I think we should create a separate bug ticket for this

jupenur:
> This is only happening on community-* can't repro on my local server.

Fwiw I was seeing this locally in an up-to-date master branch of the server. Haven't tried community.

Willyfrog:
ok, so I think I'm contemplating all the cases, but please point it out if I missed any that you might think of.
what's not being addresed here:
- image from proxy taking over screen
- urls with `..` in it

what should work:
- plugins open in a new window. Although I'm a bit against this, it was the previous behaviour so I keep it.
- urls not from the server should open in browser
- api redirections  (via image proxy) should open in browser
- navigation as usual
- authentication via external parties. (this might not work on subpath servers, I think we have to take care of that but haven't yet tested)
- download files. (again it might not work from subpath)

regarding subpath issues, my plan is to open a couple of tickets and investigate them as soon as we have time for that. I haven't touched that part, but I think it's not taking subpath into account, but didn't want this PR to grow bigger.

src/browser/components/MattermostView.jsx
```diff
@@ -132,13 +132,16 @@ export default class MattermostView extends React.Component {
...
           window.open(e.url, remote.app.getName(), 'nodeIntegration=no, contextIsolation=yes, show=yes');
+        } else {
+          e.preventDefault();
+          shell.openExternal(e.url);
         }
       } else {
-        // if the link is external, use default browser.
-        shell.openExternal(e.url);
+        // if the link is external, use default os' application.
+        ipcRenderer.send('confirm-protocol', destURL.protocol, e.url);
```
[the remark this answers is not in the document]
Willyfrog: we might want to add http/s as default safe, would that make sense? that way everything would go through the same mechanisms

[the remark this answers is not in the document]
jupenur: I'd be fine with that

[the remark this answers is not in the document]
Willyfrog: updated to handle http/s by default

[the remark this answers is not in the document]
jupenur: Could we simplify this code just a bit further by removing [this check](https://github.com/mattermost/desktop/blob/dabc2f571d7d1ab69446c3615853411721b077bd/src/browser/components/MattermostView.jsx#L110-L113) and switching [the comparison in `Utils.isInternalURL`](https://github.com/mattermost/desktop/blob/dabc2f571d7d1ab69446c3615853411721b077bd/src/utils/util.js#L40) to use origins instead of just hosts? And I guess `` `${scheme}:` `` would have to be added to the protocol whitelist, too. What do you think?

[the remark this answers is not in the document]
Willyfrog: ticket created for the URL vs url: https://mattermost.atlassian.net/browse/MM-21835

PR #1151, opened 2020-01-15
[MM-21341] Changed keyboard shortcut logic to pull focused web contents if an explicitly focused tab is not found

devinbinnie:

**Summary**
When using the JIRA plugin connect functionality, the desktop app will open a new window for sign-in. The copy/paste shortcuts weren't working due to the logic grabbing the Mattermost tab that opened the window rather than the window itself.

This PR changes the logic to ensure the Mattermost tab is focused, otherwise it should grab whatever web contents are focused and send the command there.

**Issue link**
https://mattermost.atlassian.net/browse/MM-21341

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -77,10 +77,6 @@ export default class MainPage extends React.Component {
 
   getTabWebContents(index = this.state.key || 0, teams = this.props.teams) {
     const allWebContents = remote.webContents.getAllWebContents();
-    const openDevTools = allWebContents.find((webContents) => webContents.getURL().includes('chrome-devtools') && webContents.isFocused());
-    if (openDevTools) {
-      return openDevTools;
-    }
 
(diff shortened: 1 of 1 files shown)


devinbinnie:
> looking good, although on [here](https://github.com/mattermost/desktop/pull/1137/) I think it would prevent the jira plugin from opening within the app (and sends it to the browser), which would also prevent this problem.

Fair point actually. Either way, my new logic is cleaner, so this is still an improvement.

amyblais:
@devinbinnie This wasn't cherry-picked so that is why the Jira ticket was re-opened. We should probably push this to v4.5.

devinbinnie:
@amyblais I can cherry-pick it now unless you feel it's too late.

amyblais:
@devinbinnie It's not a regression in v4.4 so I think it's too late.

devinbinnie:
> @devinbinnie It's not a regression in v4.4 so I think it's too late.

Okay let's push it.

PR #1165, opened 2020-01-22
[MM-19919] Allow links to other servers to go directly to that server in the app

devinbinnie:

**Summary**
When clicking a link from Server A that links to Server B, even if you have Server B configured, it would open the link in your browser. This PR allows those links to be opened directly in the tab for that server.

**Issue link**
https://mattermost.atlassian.net/browse/MM-19919

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -15,6 +15,8 @@ import DotsVerticalIcon from 'mdi-react/DotsVerticalIcon';
 
 import {ipcRenderer, remote} from 'electron';
 
+import Utils from '../../utils/util';
+
 import restoreButton from '../../assets/titlebar/chrome-restore.svg';
 import maximizeButton from '../../assets/titlebar/chrome-maximize.svg';
 import minimizeButton from '../../assets/titlebar/chrome-minimize.svg';
@@ -370,15 +372,7 @@ export default class MainPage extends React.Component {
   };
 
   handleInterTeamLink = (linkUrl) => {
-    let selectedTeam;
-    this.props.teams.forEach((team, index) => {
-      if (linkUrl.href.startsWith(team.url)) {
-        selectedTeam = {
-          team,
-          index,
-        };
-      }
-    });
+    const selectedTeam = Utils.getServer(linkUrl, this.props.teams);
     if (!selectedTeam) {
       return;
     }
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -369,6 +369,23 @@ export default class MainPage extends React.Component {
     this.handleSelect(key);
   };
 
+  handleInterTeamLink = (linkUrl) => {
+    let selectedTeam;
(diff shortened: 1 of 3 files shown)


src/browser/components/MainPage.jsx
```diff
@@ -369,6 +369,23 @@ export default class MainPage extends React.Component {
     this.handleSelect(key);
   };
 
+  handleInterTeamLink = (linkUrl) => {
+    let selectedTeam;
+    this.props.teams.forEach((team, index) => {
```
Willyfrog: maybe we should modify https://github.com/mattermost/desktop/blob/master/src/utils/util.js#L85 to return the index. seems an operation that can be done multiple times throughout the code

src/browser/components/MattermostView.jsx
```diff
@@ -126,8 +126,14 @@ export default class MattermostView extends React.Component {
           shell.openExternal(e.url);
         }
       } else {
-        // if the link is external, use default os' application.
-        ipcRenderer.send('confirm-protocol', destURL.protocol, e.url);
+        const parsedURL = Utils.parseURL(e.url);
```
Willyfrog: not a comment to change this, but we should unify url and URL uses into a single lib (there is a ticket suggesting that)

src/browser/components/MainPage.jsx
devinbinnie: Yeah I think that's worth doing, I'll make the change.

src/browser/components/MainPage.jsx
```diff
@@ -369,6 +371,15 @@ export default class MainPage extends React.Component {
     this.handleSelect(key);
   };
 
+  handleInterTeamLink = (linkUrl) => {
+    const selectedTeam = Utils.getServer(linkUrl, this.props.teams);
+    if (!selectedTeam) {
+      return;
+    }
+    this.refs[`mattermostView${selectedTeam.index}`].handleDeepLink(linkUrl.href);
```
jwilander: Any chance the key could be undefined in `this.refs`?

src/browser/components/MattermostView.jsx
```diff
@@ -126,8 +126,14 @@ export default class MattermostView extends React.Component {
...
+        const serverURL = Utils.getServer(parsedURL, this.props.teams);
+        if (serverURL !== null && Utils.isTeamUrl(serverURL.url, parsedURL)) {
```
jwilander: Should `serverURL !== null` be `serverURL != null` so that it handles undefined as well?

src/browser/components/MainPage.jsx
devinbinnie: Shouldn't be, we do that check above to make sure that we find a team first. Then if `selectedTeam` is null that means we couldn't find one, so that's our null check.

src/browser/components/MattermostView.jsx
devinbinnie: `getServer()` explicitly returns null, so I think we're okay here.

PR #1169, opened 2020-01-27
[MM-20832] prevent dialog stacking, move to promises

Willyfrog:

**Summary**
whenever there is no connection stablished, a new connection retry is launched every few seconds, if that fails a new dialog was being presented to the user, which might end up in a stack of errors. this PR only allows one dialog per instance and error-type.

Also, on electron6 the api for the dialog moved to promises, so I updated the code for that.

**Issue link**
[MM-20832](https://mattermost.atlassian.net/browse/MM-20832)
 
**Test Cases**
change the date on your machine to a year in the future, connect to community and the dialog should show up. waiting for several minutes shouldn't create new dialogs.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -337,12 +337,17 @@ function handleSelectedCertificate(event, server, cert) {
 }
 
 function handleAppCertificateError(event, webContents, url, error, certificate, callback) {
-  if (certificateStore.isTrusted(url, certificate)) {
+  const parsedURL = new URL(url);
+  if (!parsedURL) {
+    return;
+  }
+  const origin = parsedURL.origin;
+  if (certificateStore.isTrusted(origin, certificate)) {
     event.preventDefault();
     callback(true);
   } else {
     // update the callback
-    const errorID = `${url}:${error}`;
+    const errorID = `${origin}:${error}`;
 
     // if we are already showing that error, don't add more dialogs
     if (certificateErrorCallbacks.has(errorID)) {
@@ -350,8 +355,8 @@ function handleAppCertificateError(event, webContents, url, error, certificate, 
       certificateErrorCallbacks.set(errorID, callback);
       return;
     }
-    const extraDetail = certificateStore.isExisting(url) ? 'Certificate is different from previous one.\n\n' : '';
-    const detail = `${extraDetail}URL: ${url}\nError: ${error}`;
+    const extraDetail = certificateStore.isExisting(origin) ? 'Certificate is different from previous one.\n\n' : '';
+    const detail = `${extraDetail}origin: ${origin}\nError: ${error}`;
 
     certificateErrorCallbacks.set(errorID, callback);
     dialog.showMessageBox(mainWindow, {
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -344,37 +345,56 @@ function handleAppCertificateError(event, webContents, url, error, certificate,
...
+        ({response}) => {
+          if (response === 0) {
+            dialog.showMessageBox(mainWindow, {
+              title: 'Certificate Not Trusted',
+              message: `Certificate from "${certificate.issuerName}" is not trusted.`,
+              detail,
+              type: 'error',
+              buttons: ['Trust Insecure Certificate', 'Cancel Connection'],
+              cancelId: 1,
+            }).then(({response: responseTwo}) => {
```
lieut-data: 0/5, but would it make sense to chain the promises together vs. nesting?

src/main.js
```diff
@@ -344,37 +345,56 @@ function handleAppCertificateError(event, webContents, url, error, certificate,
...
+              if (responseTwo === 0) {
+                certificateStore.add(url, certificate);
+                certificateStore.save();
+                certificateErrors.get(errorID)(true);
```
lieut-data: Without context, it's not immediately obvious why a "certificate error" is callable. Is there a naming opportunity here?

src/main.js
Willyfrog: changed to certificateErrorCallbacks, originally only stored if the error was being handled, but I later changed my mind into using it to store the callback

src/main.js
Willyfrog: Definitely, changed to certificateErrorCallbacks

src/main.js
```diff
@@ -345,56 +345,54 @@ function handleAppCertificateError(event, webContents, url, error, certificate,
...
+            title: 'Certificate Not Trusted',
+            message: `Certificate from "${certificate.issuerName}" is not trusted.`,
+            detail,
+            type: 'error',
+            buttons: ['Trust Insecure Certificate', 'Cancel Connection'],
+            cancelId: 1,
+          });
+        }
+        return {response};
       }).then(
```
lieut-data: nit: might want to outdent this line, to line up with line 366.

jgilliam17:
@Willyfrog I tested using the steps from the jira issue and it works fine. Upon return to the app and reconnecting the internet red websocket banner is self dismissed.

While testing using `test cases` listed in the description here in the PR 
`change the date on your machine to a year in the future, connect to community and the dialog should show up. waiting for several minutes shouldn't create new dialogs.`
There is an immediate "Your session has expired"
I can log back in using email & password, app displays "Please check connection, Mattermost unreachable" 
Not able to log in using OneLogin, "Cannot connect to Mattermost" asking user to check connection, URL and option to open in browser.
Is this expected?

Willyfrog:
I think it is, by moving forward into the future, any certs will probably be expired and most connections will be rejected by default as not secure. Have you tried using email+password?

jgilliam17:
> I think it is, by moving forward into the future, any certs will probably be expired and most connections will be rejected by default as not secure. Have you tried using email+password?

Yes, I tried email+password and I can log in afterwards with app's banner displaying "Please check connection, Mattermost unreachable"
No popup dialogs though, which I guess is expected with the banner already present.

Willyfrog:
Mannually cherry-picked

PR #1164, opened 2020-01-22
[MM-21974] improve CSS for client certificate screens

Willyfrog:
**Summary**
Current CSS for selecting the certificate and showing the info is not looking good. this attempts to fix it.

**Issue link**

[MM-21974](https://mattermost.atlassian.net/browse/MM-21974)

PR #1170, opened 2020-01-27
[MM-22006] fix name mapping

Willyfrog:

**Summary**
there are two aliases for the fields, we'll try to get the one that is filled.

**Issue link**
[MM-22006](https://mattermost.atlassian.net/browse/MM-22006)


deanwhillier:
@Willyfrog, I think the mapping problem is that lines 64 & 68 need to be swapped. I noticed and corrected that while doing some tweaks here: https://github.com/mattermost/desktop/pull/1164/commits/ff9d234e765532ca3de0547312ee3324aa6e656c#

That said, I think the changes you made here are still valid.

PR #1142, opened 2019-12-31
[MM-21469] refactor: e2e test improvements

hunterlester:
Before submitting, please confirm you've
 - [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
 - [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
 - [x] executed `npm run lint:js` for proper code formatting

Please provide the following information:

**Summary**
Improvements for e2e testing stability

**Issue link**
Fixes https://github.com/mattermost/desktop/issues/1144
https://mattermost.atlassian.net/browse/MM-21469
https://mattermost.atlassian.net/browse/MM-20822

**Test Cases**
- Security suite enabled as well as other misc skipped tests
- Enabled previously skipped test in Close Button suite
71280565-6f2eac00-2310-11ea-9185-9034835af7f4

**Additional Notes**
  - Settings page rendered and controlled via MainPage
  - MainPage driven by state instead of props
  - Single config source for renderers
  - Enables Spectron security suite and other misc previously skipped tests
  - SettingsPage's backToIndex refactored to update state instead of load URL
  - Adds new suite for e2e Menu tests
  - Adds missing prop from MM-10586
  - Enables keyboard shortcuts in settings modal
  - Settings page changes addresses https://mattermost.atlassian.net/browse/MM-20822

Originally intended to start working with TestCafe but realized that it's missing important capabilities which are present in Spectron. Then I thought to have both TestCafe and Spectron tests to complement each other but this is not practical or very nice for the next developer figuring out where and which framework to use to write their next test.

My new goal is to improve the desktop applications in order to stabilize e2e testing with Spectron as much as possible.
I'll continue to enable previously skipped tests and add more tests to provide a strong foundation for preventing regressions as we make big changes like migrating from `webview` to `BrowserView`, for which I'm devising a plan.

PR #1171, opened 2020-01-29
[MM-22073] Added new titlebar to Settings page

devinbinnie:

**Summary**
When I added the new titlebar, I neglected to add it for the Settings page. It is now added.

**Issue link**
https://mattermost.atlassian.net/browse/MM-22073

**Test Cases**
- On Windows, when you open the settings page, you should see a titlebar.
- Each of the titlebar buttons on Windows in the settings page should function the same as they do on the main view.
- On Mac, you should see the titlebar with the normal titlebar buttons.

**Additional Notes**
This wasn't tested on Mac, though I believe things should be working.

### DIFF
--- a/src/browser/components/SettingsPage.jsx
+++ b/src/browser/components/SettingsPage.jsx
@@ -6,15 +6,23 @@
 /* eslint-disable react/no-set-state */
 
 import React from 'react';
+import PropTypes from 'prop-types';
 import {Button, Checkbox, Col, FormGroup, Grid, HelpBlock, Navbar, Radio, Row} from 'react-bootstrap';
 
 import {ipcRenderer, remote} from 'electron';
 import {debounce} from 'underscore';
+import DotsVerticalIcon from 'mdi-react/DotsVerticalIcon';
 
 import Config from '../../common/config';
 
+import restoreButton from '../../assets/titlebar/chrome-restore.svg';
+import maximizeButton from '../../assets/titlebar/chrome-maximize.svg';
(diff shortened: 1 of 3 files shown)


hunterlester:
Looking good on macOS Mojave.

I also kindly advise that we may be encouraging redundancy and complexity by continuing to render settings as a separately loaded page. If we move the `SettingsPage` component into `MainPage`, then event listeners and window behavior will not need to be replicated.
Much simpler to manager in one place when we need to make updates in the future.

See this [PR](https://github.com/mattermost/desktop/pull/1142) as an example of moving settings into MainPage.
In this case, MM-22073 could be resolved by either CSS positioning, height, etc, instead of replicating rendered state and behavior.

devinbinnie:
> Looking good on macOS Mojave.
> 
> I also kindly advise that we may be encouraging redundancy and complexity by continuing to render settings as a separately loaded page. If we move the `SettingsPage` component into `MainPage`, then event listeners and window behavior will not need to be replicated.
> Much simpler to manager in one place when we need to make updates in the future.
> 
> See this [PR](https://github.com/mattermost/desktop/pull/1142) as an example of moving settings into MainPage.
> In this case, MM-22073 could be resolved by either CSS positioning, height, etc, instead of replicating rendered state and behavior.

Definitely a good idea, in an ideal world I'd like to actually pull out the TopBar into its own component as well so there's even less code duplication.

For now I just wanted to make sure the functionality intact, but we'll be discussing this for sure.

jgilliam17:
Thanks in advance @ogi-m for testing on Windows.

PR #1176, opened 2020-01-30
Make plugins oauth and oauth2 urls to be considered outside links

larkox:
**Summary**
Make plugins oauth and oauth2 urls to be considered outside links. This is to permit plugins to authenticate on a browser, and avoid [Google restrictions on log-in](https://security.googleblog.com/2019/04/better-protection-against-man-in-middle.html).
**Issue link**
https://mattermost.atlassian.net/browse/MM-22113

### DIFF
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -84,9 +84,9 @@ function isPluginUrl(serverUrl, inputURL) {
 
 function isPluginOAuth(inputURL) {
   const parsedURL = parseURL(inputURL);
-  const pluginOauthRegexes = /^\/plugins\/[\w.-]+\/oauth[2]?\//i;
+  const regex = /^\/plugins\/[\w.-]+\/oauth[2]?\//i;
 
-  return pluginOauthRegexes.test(parsedURL.pathname)
+  return regex.test(parsedURL.pathname)
 }
 
 function getServer(inputURL, teams) {
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -85,8 +85,8 @@ function isPluginUrl(serverUrl, inputURL) {
 function isPluginOAuth(inputURL) {
   const parsedURL = parseURL(inputURL);
   const pluginOauthRegexes = [
-    /^\/plugins\/[A-Za-z0-9\.]+\/oauth\//i,
-    /^\/plugins\/[A-Za-z0-9\.]+\/oauth2\//i,
+    /^\/plugins\/[A-Za-z0-9.]+\/oauth\//i,
+    /^\/plugins\/[A-Za-z0-9.]+\/oauth2\//i,
   ];
 
   for (const regexPath of pluginOauthRegexes) {
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -84,18 +84,9 @@ function isPluginUrl(serverUrl, inputURL) {
 
 function isPluginOAuth(inputURL) {
   const parsedURL = parseURL(inputURL);
-  const pluginOauthRegexes = [
-    /^\/plugins\/[A-Za-z0-9.]+\/oauth\//i,
-    /^\/plugins\/[A-Za-z0-9.]+\/oauth2\//i,
-  ];
+  const pluginOauthRegexes = /^\/plugins\/[\w.-]+\/oauth[2]?\//i;
 
-  for (const regexPath of pluginOauthRegexes) {
-    if (parsedURL.pathname.match(regexPath)) {
-      return true;
-    }
-  }
-
-  return false;
+  return pluginOauthRegexes.test(parsedURL.pathname)
 }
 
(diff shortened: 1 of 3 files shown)


hunterlester:
Nice! Reading up on plugins docs to manually test with Zoom plugin as you suggested.
Do you have a way of testing without paid Zoom account?
Meanwhile I'm also thinking about best way to minimally test, especially to protect the behavior with e2e testing.
Let me know if you have any thoughts.

src/utils/util.js
```diff
@@ -82,6 +82,22 @@ function isPluginUrl(serverUrl, inputURL) {
   return server.origin === parsedURL.origin && parsedURL.pathname.toLowerCase().startsWith(`${server.subpath}plugins/`);
 }
 
+function isPluginOAuth(inputURL) {
```
abdusabri: Hey @larkox, thanks for being so active in the community - and a release MVP as well! 🙂

I've a few recommendations on the implementation of this function (`isPluginOAuth`):

- Having `[A-Za-z]` and the `i` flag is redundant, so we could remove the flag in this case (in the suggested implementation below, I've used `\w`, which is equivalent to `[A-Za-z0-9_]`)
- Since we are only interested in knowing whether the pattern matches the `RegExp` or not (we are not using the matches from the `match` function), it is better to use `RegExp.test()`, which is faster than the `match` function. Using `test` gives the bonus of returning its result directly from `isPluginOAuth` instead of adding conditions and returning `true` or `false` explicitly
- We could use an optional `2` at the end of the `RegExp`, like `oauth[2]?` and avoid the need for the array and the for loop
- I think hyphens and underscores are also possible in the url before `/oauth/`, right?
- I don't know if the scope of this function should handle edge cases for `.` in the url, like 2 consecutive ones `..`, or a url like `/plugins/./oauth/`

I suggest an implementation like the following (not handling edge cases for `.`)

```javascript
function isPluginOAuth(inputURL) {
  const parsedURL = parseURL(inputURL);
  return /^\/plugins\/[\w-.]+\/oauth[2]?\//.test(parsedURL.pathname);
}
```
Hope this helps 😊. BYW, I'm not familiar with MM plugins yet, just the `RegEx` stuff got me a bit curious 🙃

larkox:
Thanks for the comments, but finally this was not the real solution. I will upload a new PR with the actual solution.

larkox:
I reopened this PR due to the conversation [here](https://github.com/mattermost/desktop/pull/1178).

crspeller:
Not much to say on the code here. But I would like @DHaussermann to test this out with the JIRA plugin at least to see how it works.

DHaussermann:
@larkox I'm not seeing the authentication for Zoom and other plugins work consistently. I've tested this on 2 windows machines and in some cases, it opens a browser windows that shows `Not authorized`. I can see this for Zoom as well as GitHub and GitLab when I tested. It does not reproduce on 1 on my machines.
ZoomLink
GitHubLink



Jira is working but, Jira is opening a 2nd window for the desktop app and not a browser window. 
Jira


0/5 - I think in cases where it is working there are pre-existing browser session that are authenticated or there are credentials saved in the browser (or possibly the desktop app its self).


Can you see if you can repo this? Maybe you have some thoughts on a way to isolate the issue?

Steps (windows):
- Uninstall Mattermost desktop
- Clear all application data for Mattermost desktop
- Ensure no relevant credentials are stored in your browser and close all browser sessions
- Launch .exe from the CircleCi zip package for this PR
- Add a server name and address where Zoom is installed and working
- Login as a user not mapped to any Zoom account
- Click the connection link
Expected: Zoom login page
Observed: `Not authorized`

larkox:
I just realised how naive this solution was.
You get the `Not Authorized` message because you are not logged in to Mattermost on the browser. You are logged in on the App. It was working on my system because I was also logged in on the browser.
Therefore, unless we can somehow send the session to the browser, it would not be possible to use this solution.

Right now I am out of ideas. Do you know any other possible solution for this?

hunterlester:
@larkox It's been a busy week so I haven't been able to make time, but I'm still wanting to test with BrowserView and see if the restriction is solely for webview.
I don't have a paid Zoom account, so I may try Github plugin.
Just need a minimum setup for testing.

DHaussermann:
@hunterlester if it helps - I was able to to create a developer account on Zoom and create an oAuth2 application that way. The only issue is that my dev application only works with the one set of Google credentials that I registered with so, it's not shareable as no other users can be authorized.

crspeller:
@larkox @deanwhillier Looks like from the discussion here this is not going to work. Should we close this while we figure out an alternate solution? Not sure if there is already discussion ongoing about that.

larkox:
I only kept this open to have a hub for discussing other solutions. Nevertheless I have no problem closing it.

PR #1178, opened 2020-01-31
Add user-agent to bypass Google security concerns

larkox:
**Summary**
Add user-agent to bypass Google security concerns when login in with oAuth. Passing this user agent removes the message shown on this issue: https://mattermost.atlassian.net/browse/MM-22113 More info on Google message: https://security.googleblog.com/2019/04/better-protection-against-man-in-middle.html

**Issue link**
https://mattermost.atlassian.net/browse/MM-22113

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -500,7 +500,7 @@ function handleAppWebContentsCreated(dc, contents) {
           popupWindow = null;
         });
       }
-      popupWindow.loadURL(url, { userAgent: 'Chrome' });
+      popupWindow.loadURL(url, {userAgent: 'Chrome'});
     }
   });
 
--- a/src/main.js
+++ b/src/main.js
@@ -500,7 +500,7 @@ function handleAppWebContentsCreated(dc, contents) {
           popupWindow = null;
         });
       }
-      popupWindow.loadURL(url);
+      popupWindow.loadURL(url, { userAgent: 'Chrome' });
     }
   });
 


lieut-data:
I'm not sure there's a whole lot of value for me to add here -- adding @jupenur for another security-related perspective. My only thought might be if we should specify a "more complete" user agent string that correctly reflects the underlying application.

crspeller:
@larkox Reading the link you posted doesn't sound like oauth though electron is supported by Google anymore. Changing the user agent is just tricking whatever code is checking. No guarantee they will not check other things in the future to detect electron. 

Adding @deanwhillier for his Desktop perspective.

larkox:
Thanks for the comments @crspeller . I did another [PR](https://github.com/mattermost/desktop/pull/1176) where I just opened a new browser, instead of keep inside the app, but the user experience from my point of view was really bad. This solution (or anything in the same line) has a way better user experience, but I understand it is a not very reliable.

jupenur:
I'd be inclined to agree with @crspeller on this: Changing the UA string seems like the wrong approach and not future-proof. Especially considering [Blink is planning on deprecating UA strings entirely](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ) in favor of UA-CH. Google is likely to adopt UA-CH for OAuth as soon as it ships.

levb:
Closing as unacceptable, considering https://github.com/mattermost/desktop/pull/1176.

PR #1186, opened 2020-02-07
[MM-19636] make file names consistent

Willyfrog:
**Summary**
Some filenames for windows added a `v` in front of the version name, this is removed now
removed fixed versions from xml file
get release version from package.json instead of tag

**Issue link**
[MM-19636](https://mattermost.atlassian.net/browse/MM-19636)

**Additional Notes**
for -develop versions, it'll add the number of commits to the build version.


Willyfrog:
manually cherry-picked

Issue #1189, opened 2020-02-11
Duplicate UI Server Tab issue on 4.4.0-rc1 when using GPO


nyclabels:
I confirm (by marking "x" in the [ ] below: [x]):

- [X] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [X] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [X] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
When a Mattermost Team is defined by Group Policy Object, multiple duplicate tabs are created across the top (one per every minimize/open of the client).

**Environment**
- Operating System: Windows 10 Pro x64 1903
- Mattermost Desktop App version: 4.4.0-rc1
- Mattermost Server version: 5.16.3

**Steps to reproduce**
Install client.
Configure GPO to define a mattermost server.
Log in.
Minimize/Open Mattermost Window

**Expected behavior**
There is a single tab on the top for the GPO-assigned server, like with Client 4.3.2.

**Observed behavior**
Tabs will continue to duplicate every time the window is minimized and opened.

PR #1191, opened 2020-02-12
[MM-22494] Fixed misuse of config.set

devinbinnie:

**Summary**
On apps using GPO configurations, when launching the app for the first time, or when adding a new team using the + button, GPO teams would sometimes be duplicated. This was caused by running `config.set('teams', <some array>)` with the GPO teams in the array, thus setting them as local teams as well.

This PR ensures that all uses of `config.set('teams', <some array>)` do not include the GPO teams.

**Issue link**
https://mattermost.atlassian.net/browse/MM-22494
https://github.com/mattermost/desktop/issues/1189

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -768,13 +768,13 @@ export default class MainPage extends React.Component {
           });
         }}
         onSave={(newTeam) => {
-          this.props.teams.push(newTeam);
-          this.setState({
-            showNewTeamModal: false,
-            key: this.props.teams.length - 1,
+          this.props.localTeams.push(newTeam);
+          this.props.onTeamConfigChange(this.props.localTeams, () => {
+            self.setState({
+              showNewTeamModal: false,
+              key: this.props.teams.length - 1,
+            });
           });
-          this.render();
-          this.props.onTeamConfigChange(this.props.teams);
         }}
       />
     );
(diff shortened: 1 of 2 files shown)


devinbinnie:
> I've just tested on win and it seems I can't reorder the tabs, is this intended or has to do with GPO?

Yes it's intended AND has to do with GPO. We turned off dragging for users using GPO tabs because we don't have a reliable way of writing the order back to the registry, or storing it in local config.

This is something I'm hoping to address at a later date.

devinbinnie:
> what i did:
> 
> * configured GPO
> * using the artifacts from [here](https://app.circleci.com/jobs/github/mattermost/desktop/5504/artifacts)
> * started the app
> 
> Observation:
> 
> * As expected just one tab is displayed
> 
> Next step:
> 
> * closed the app and started again
> 
> Observation:
> 
> * As expected just one tab is displayed
> 
> Next step:
> 
> * added manually a second server
> 
> Observation:
> 
> * As expected both tab are displayed, also after closing and starting the app.
> 
> Just one note:
> It is still possible to drag the tab to another position but it jumps back to the original position as soon as you release the mouse. If possible i would freeze the tab completely otherwise the users could be confused.

Thanks @JtheBAB! Unfortunately there's no easy way to disable the initial drag without unwrapping the component from the dragging library, which will cause a lot of CSS headaches.

JtheBAB:
Ok. Then i guess it is needed to add it to known issue section of the release notes.

lindy65:
Thanks for testing @JtheBAB :)

I've removed `QA-review` label and added `QA-review-done`

@amyblais - please see `JtheBAB's` comment above about adding the issue to Known Issues for release. Thanks!

PR #1193, opened 2020-02-14
[MM-22585] Ctrl/Cmd+<num> now switches to tabs based on order

devinbinnie:

**Summary**
When switching the order of the tabs, and using `Ctrl+i` or `Cmd+i`, where `i` is the 1-based order of the tab to switch to, the app was opening the wrong tab if the order of the tabs did not match the storage order.

This PR now makes sure that the hotkey switches to the correct tab.

**Issue link**
https://mattermost.atlassian.net/browse/MM-22585

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -153,7 +153,8 @@ export default class MainPage extends React.Component {
 
     // can't switch tabs sequentially for some reason...
     ipcRenderer.on('switch-tab', (event, key) => {
-      this.handleSelect(key);
+      const nextIndex = this.props.teams.findIndex((team) => team.order === key);
+      this.handleSelect(nextIndex);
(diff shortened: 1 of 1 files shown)


src/browser/components/MainPage.jsx
```diff
@@ -153,7 +153,8 @@ export default class MainPage extends React.Component {
 
     // can't switch tabs sequentially for some reason...
     ipcRenderer.on('switch-tab', (event, key) => {
-      this.handleSelect(key);
+      const nextIndex = this.props.teams.findIndex((team) => team.order === key);
```
Willyfrog: what would happen if I hit cmd+3 but there are only 2 teams?

devinbinnie: Just fails silently

PR #1213, opened 2020-02-26
[MM-22363] - Add desktop notification on download complete

nevyangelova:
**Summary**
Added a desktop notification when a file is being downloaded. I had some issues installing a Windows VM on my current computer so I couldn't test on windows. Any feedback is welcomed. Also I tried to add a custom cursor for better UX but I didn't find anything in the docs about custom styling. Advice is welcomed too :D 

**Issue link**
https://mattermost.atlassian.net/browse/MM-22363


nevyangelova:
@esethna how do you have two server tabs open? I can't seem to log into two different teams but on the same server in two different tabs, because it would load one of the teams at all. When I log into the same team on two tabs, I only get one download notification. When I click, it open the location of the file in the finder.  From what I found online about the Electron notifications there aren't customisable styles. I tried to change the cursor or highlight the text but its not in the API.
Maybe the other devs @Willyfrog @deanwhillier can advise? 

Screen Shot 2020-02-27 at 14 49 36


Screen Shot 2020-02-27 at 14 18 52

Willyfrog:
from the [docs](https://www.electronjs.org/docs/api/shell#shellshowiteminfolderfullpath) it says it will highlight it if possible, so I guess it does happen on some platforms but not in all or that there is something else preventing the highlight

Willyfrog:
@esethna might be a [bug with explorer](https://www.winhelponline.com/blog/file-explorer-not-highlight-selected-files-windows-10/)?

nevyangelova:
@esethna apologies I hadn't built the app, but after building and in the artifacts from the CI we get the correct icon. I can't test yet on Windows unfortunately because I can't install the Windows VM on my current laptop. After I get the new Macbook I can install it and investigate what I can do with the highlight. @Willyfrog is currently testing the behaviour on Windows to see if he would get the correct number of notifications.

Willyfrog:
Want to do some more testing, but as far as I got: on the exe it is a notification per server (got 3) and it doesn't highlight the file.

esethna:
I got the same behaviour as Guillermo ^

nevyangelova:
Waiting for my new laptop so I can set up the Windows environment and fix this :)

nevyangelova:
@esethna @devinbinnie @Willyfrog I tried to research why we are not selecting the file in the file explorer on windows and there isn't much information online apart from 'it should work out of the box with ```showItemInFolder``` , there is one known issue https://github.com/electron/electron/issues/11617 and it seems that when the path includes slashes and possibly spaces it won't work. I tried using ```normalize()``` and apart from that have no other ideas and I am not sure this is possible. Potentially it can work for some files depending on the path. Maybe someone can advise?

src/browser/components/MainPage.jsx
```diff
@@ -598,6 +602,39 @@ export default class MainPage extends React.Component {
...
+  showDownloadCompleteNotification = async (event, item) => {
+    const title = 'Download Complete';
+
+    const notification = await this.dispatchNotification(title, item.title);
+    notification.onclick = () => {
+      shell.showItemInFolder(item.defaultPath.normalize());
+    };
+  }
+
+  dispatchNotification = async (title, body, channel, teamId, silent) => {
```
Willyfrog: why the teamId? if this is only for download I'd rename to `dispatchDownloadNotification` or something that can be distinguished from the other one.
Should we have all notification code together? I'm 0/5 on this if we don't have a more general case.

src/browser/components/MainPage.jsx
```diff
@@ -598,6 +602,39 @@ export default class MainPage extends React.Component {
...
+    }
+    const notification = new Notification(title, {
+      body,
+      tag: body,
+      icon: appIconURL,
+      requireInteraction: false,
+      silent,
+    });
+    notification.onclick = () => {
+      this.webviewRef.current.send('notification-clicked', {channel, teamId});
```
Willyfrog: why send channel and teamId if it's undefined? shouldn't we use a different event? are we notifying the webapp for some reason?

src/main.js
```diff
@@ -739,6 +739,15 @@ function initializeAfterAppReady() {
       defaultPath: os.homedir() + '/Downloads/' + filename,
       filters,
     });
+
+    item.on('done', (doneEvent, state) => {
+      if (state === 'completed') {
+        mainWindow.webContents.send('download-complete', {
+          title: filename,
+          defaultPath: os.homedir() + '/Downloads/' + filename,
+        });
```
Willyfrog: should we have any notification if state was other than completed?

src/main.js
```diff
@@ -739,6 +739,15 @@ function initializeAfterAppReady() {
       defaultPath: os.homedir() + '/Downloads/' + filename,
       filters,
     });
+
+    item.on('done', (doneEvent, state) => {
+      if (state === 'completed') {
+        mainWindow.webContents.send('download-complete', {
```
Willyfrog: if you create a custom channel, you can add all the info that requires the display format (see jira ticket for that)

you probably need to get that info from the webcontents object in the `will-download` event, that way you can figure out which server initiated the download process.

src/browser/components/MainPage.jsx
devinbinnie: Agreed, this is repeated code from `MattermostView.jsx`, so if we could find a way to generalize sending notifications between the normal case and the few corner cases I would be in favour of that.

0/5 on if it happens for this PR for not, but if not we should create a ticket for it.

src/browser/components/MainPage.jsx
devinbinnie: I think this is just leftover code from the other notification block.

nevyangelova: yes, I was thinking of extracting it to a util function, is there a preferred place to do that?

src/main.js
nevyangelova: Probably a good idea to have error handling but not sure how to test it.

src/browser/components/MainPage.jsx
devinbinnie: You can add it to `src/util/util.js`

src/utils/util.js
```diff
@@ -139,14 +140,43 @@ function equalUrlsIgnoringSubpath(url1, url2) {
   return url1.origin.toLowerCase() === url2.origin.toLowerCase();
 }
 
+const dispatchNotification = async (title, body, channel, teamId, silent) => {
```
Willyfrog: instead of having the channel and team id, why not pass a callback for the `onclick` function? that way you don't have to ignore the parameters and it's a bit more malleable for future usage.

nevyangelova: I checked and it seems that we do need to pass the channel and teamId, because of 
```
ipcRenderer.on('notification-clicked', (event, {channel, teamId}) => {
  window.postMessage(
    {
      type: 'notification-clicked',
      message: {
        channel,
        teamId,
      },
    },
    window.location.origin
  );
});
```
we will need to go through the whole webapp and will need to refactor all notifications to rely on the custom onClick, instead of the teamId and channel. Maybe in a separate ticket?

nevyangelova:
@ogi-m I updated the PR to transform the URL to server name. Also I can't reproduce, for me the link opens correctly.

ogi-m:
@nevyangelova I see the server name shown now, but clicking on the notification still doesn't do anything for me on the 2 different Window 10 PCs that I've tried.

src/browser/components/MainPage.jsx
```diff
@@ -602,13 +602,20 @@ export default class MainPage extends React.Component {
   };
 
   showDownloadCompleteNotification = async (event, item) => {
-    const serverName = process.platform === 'win32' ? item.serverInfo.origin : 'Download Complete';
+    const regexResult = (/https?:\/\/(.+).mattermost.com\/?$/).exec(item.serverInfo.origin);
```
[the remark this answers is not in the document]
nevyangelova: There wasn't a servername parameter exposed, all I got was the server url.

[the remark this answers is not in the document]
Willyfrog: but 3 lines below (608) we are accessing `serverName`:
```js
       serverName = item.serverInfo.serverName; 
```
if it isn't there, that line would fail, right?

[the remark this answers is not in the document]
nevyangelova: Sorry, maybe this is not the best name for the variable, but this is actually the server URL and if the regex doesn't match the URL (ie, its not a mattermost url) the serverName would actually be the url.

PR #1214, opened 2020-02-27
[MM-22720] Assume maximize setting when AppleActionOnDoubleClick isn't explicitly set

devinbinnie:

**Summary**
In some cases (like mine), the `AppleActionOnDoubleClick` setting isn't explicitly set even though System Preferences says so. This PR assumes that if it is not set to default to the 'zoom' or 'maximize' setting.

**Issue link**
https://mattermost.atlassian.net/browse/MM-22720


Willyfrog:
is this also expected to work that way on windows/linux? [getUserDefaults](https://www.electronjs.org/docs/api/system-preferences#systempreferencesgetuserdefaultkey-type-macos) is an osx only feature.

devinbinnie:
> is this also expected to work that way on windows/linux? [getUserDefaults](https://www.electronjs.org/docs/api/system-preferences#systempreferencesgetuserdefaultkey-type-macos) is an osx only feature.

Windows handles this differently, but I'm not sure how linux does.

PR #1240, opened 2020-03-18
[MM-22434] update joi, convert to json before validating

Willyfrog:

**Summary**

When loading the certificate.json file, it was failing validation, this was due to several parts, one it was missing the protocol on it to validate as uri. also it wasn't loaded as an object, so i couldn't load.

Improved code and upgraded joi to next version.

**Issue link**

[MM-22434](https://mattermost.atlassian.net/browse/MM-22434)

**Test Cases**

with a server with an untrusted certificate, go throught the dialog and save the connection as valid. upon restart, no dialog should appear.

### DIFF
--- a/src/main/certificateStore.js
+++ b/src/main/certificateStore.js
@@ -32,18 +32,16 @@ function getHost(targetURL) {
 function CertificateStore(storeFile) {
   this.storeFile = storeFile;
   let storeStr;
-  let jsonStore;
   try {
     storeStr = fs.readFileSync(storeFile, 'utf-8');
-    jsonStore = JSON.parse(storeStr);
-    const result = Validator.validateCertificateStore(jsonStore);
+    const result = Validator.validateCertificateStore(storeStr);
     if (!result) {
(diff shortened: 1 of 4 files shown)


src/main/certificateStore.js
```diff
@@ -26,27 +25,25 @@ function areEqual(certificate0, certificate1) {
...
 
 function CertificateStore(storeFile) {
   this.storeFile = storeFile;
   let storeStr;
+  let jsonStore;
   try {
     storeStr = fs.readFileSync(storeFile, 'utf-8');
-    storeStr = Validator.validateCertificateStore(storeStr);
-    if (!storeStr) {
+    jsonStore = JSON.parse(storeStr);
```
devinbinnie: Do we need to do this twice? Seems like overkill since `validateCertificateStore` is already making sure we have an object.

Willyfrog: fixed, it is also cleaner, thanks!

PR #1243, opened 2020-03-24
[MM-20742] Null checking on URL pathname when deep linking

devinbinnie:

**Summary**
When deep linking to the root of a server and the URL doesn't end with a `/`, the URL `pathname` would be `null` and would be converted to a string as such. This PR does a null check returning a blank string if `pathname` is `null`.

**Issue link**
https://mattermost.atlassian.net/browse/MM-20742

**Test Cases**
1. Open Mattermost Desktop and add the Community server
2. Go to your browser and enter `mattermost://community.mattermost.com`
3. Press Enter and allow to open in Mattermost
4. The deep link should pop the community server up, rather than generate a Channel Not Found error.

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -70,8 +70,8 @@ export default class MainPage extends React.Component {
             teamURL,
             teamIndex: index,
             originalURL: deeplinkURL,
-            url: `${teamURL.protocol}//${teamURL.host}${deeplinkURL.pathname || ''}`,
-            path: deeplinkURL.pathname || '',
+            url: `${teamURL.protocol}//${teamURL.host}${deeplinkURL.pathname || '/'}`,
+            path: deeplinkURL.pathname || '/',
           };
         }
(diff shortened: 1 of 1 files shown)


src/browser/components/MainPage.jsx
```diff
@@ -70,8 +70,8 @@ export default class MainPage extends React.Component {
             teamURL,
             teamIndex: index,
             originalURL: deeplinkURL,
-            url: `${teamURL.protocol}//${teamURL.host}${deeplinkURL.pathname}`,
-            path: deeplinkURL.pathname,
+            url: `${teamURL.protocol}//${teamURL.host}${deeplinkURL.pathname || ''}`,
+            path: deeplinkURL.pathname || '',
```
Willyfrog: if you use `URL` instead, pathname will always have the value of '/' as a minimum. and I know i've relied on that on the past. Until we unify use of url and URL, I'd advise to use '/' in case it is undefined.

Issue #1224, opened 2020-03-03
Desktop client opens to blank white Window when using GPO-set teams 


nyclabels:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Mattermost opens to white window when only a GPO-defined team is configured.

**Environment**
- Operating System: Windows 10 Pro x64 1903 
- Mattermost Desktop App version: 4.4.0 
- Mattermost Server version: 5.16.0

**Steps to reproduce**
Install Mattermost 4.4
Configure GPO to define a default team
Open Mattermost from being fully closed

**Expected behavior**
Mattermost loads the default team where you last left off (worked as intended in 4.3.2)

**Observed behavior**
Mattermost opens to a white window with the option to select the relevant team.

Window that opens (Team name redacted): 
![MATTERMOST](https://user-images.githubusercontent.com/17073218/75825346-1e591c00-5d73-11ea-9109-be79718352e4.png)

Once you click on the correct team name, it seems to load up just fine and run like "normal" (ie, just like 4.3.2 did).

**Possible fixes**
I have no idea, I just report things.

amyblais:
@nyclabels Do you see any console errors https://docs.mattermost.com/install/desktop.html#reporting-issues?

nyclabels:
Yes, actually I do!

index_bundle.js:20096 Uncaught TypeError: Cannot read property 'focusOnWebView' of undefined
    at focusListener (index_bundle.js:20096)
    at CallbacksRegistry.apply (C:\Program Files\Mattermost\Desktop\resources\electron.asar\renderer\callbacks-registry.js:41)
    at C:\Program Files\Mattermost\Desktop\resources\electron.asar\renderer\api\remote.js:281
    at EventEmitter.<anonymous> (C:\Program Files\Mattermost\Desktop\resources\electron.asar\renderer\api\remote.js:271)
    at EventEmitter.emit (events.js:200)
[...]

PR #1246, opened 2020-03-25
[MM-23082] Fixes whitescreen issue on app upgrade / GPO / whitelabeled app deployments

justledbetter:
If sort order was not present in buildConfig.js (or previous configuration), we will assume reasonable defaults.

**Summary**
This fixes an issue where enterprises deploying with buildConfig.js or GPO settings were getting a white screen on app startup.  It might have also affected users that upgraded their application from previous releases due to a configuration file format change.

**Issue link**
https://github.com/mattermost/desktop/issues/1224

**Test Cases**
* Define default servers in src/common/config/buildConfig.js, but leave out the order:0 parameter
* Launch application
* Confirm that application loads and displays tab, as expected.

**Additional Notes**

### DIFF
--- a/src/common/config/buildConfig.js
+++ b/src/common/config/buildConfig.js
@@ -25,7 +25,7 @@ const buildConfig = {
     }
   */],
   helpLink: 'https://about.mattermost.com/default-desktop-app-documentation/',
-  enableServerManagement: false,
+  enableServerManagement: true,
   enableAutoUpdater: true,
 };
 
--- a/src/common/config/buildConfig.js
+++ b/src/common/config/buildConfig.js
@@ -20,8 +20,7 @@ const buildConfig = {
   defaultTeams: [/*
     {
       name: 'example',
-      url: 'https://example.com',
-      order: 0
+      url: 'https://example.com'
     }
   */],
   helpLink: 'https://about.mattermost.com/default-desktop-app-documentation/',
--- a/src/common/config/buildConfig.js
+++ b/src/common/config/buildConfig.js
@@ -9,6 +9,7 @@
  * @prop {Object[]} defaultTeams
  * @prop {string} defaultTeams[].name - The tab name for default team.
  * @prop {string} defaultTeams[].url - The URL for default team.
+ * @prop {string} defaultTeams[].order - Sort order for team tabs (0, 1, 2)
  * @prop {string} helpLink - The URL for "Help->Learn More..." menu item.
  *                           If null is specified, the menu disappears.
  * @prop {boolean} enableServerManagement - Whether users can edit servers configuration.
@@ -19,11 +20,12 @@ const buildConfig = {
   defaultTeams: [/*
     {
       name: 'example',
-      url: 'https://example.com'
-    }*/
-  ],
+      url: 'https://example.com',
+      order: 0
+    }
+  */],
   helpLink: 'https://about.mattermost.com/default-desktop-app-documentation/',
-  enableServerManagement: true,
+  enableServerManagement: false,
   enableAutoUpdater: true,
 };
 
--- a/src/common/config/index.js
+++ b/src/common/config/index.js
@@ -341,6 +341,7 @@ export default class Config extends EventEmitter {
       if (y.order == null) {
         y.order = 0;
       }
+
       // once we ensured `order` exists, we can sort numerically
       return x.order - y.order;
     });
--- a/src/common/config/index.js
+++ b/src/common/config/index.js
@@ -289,8 +289,8 @@ export default class Config extends EventEmitter {
       combinedTeams.push(...this.localConfigData.teams);
     }
 
-    combinedTeams = this.filterOutDuplicateTeams(combinedTeams)
-    combinedTeams = this.sortUnorderedTeams(combinedTeams)
+    combinedTeams = this.filterOutDuplicateTeams(combinedTeams);
+    combinedTeams = this.sortUnorderedTeams(combinedTeams);
 
     this.combinedData.teams = combinedTeams;
     this.combinedData.localTeams = this.localConfigData.teams;
@@ -336,7 +336,11 @@ export default class Config extends EventEmitter {
     let i = 0;
 
     newTeams = newTeams.filter((newTeam) => {
-      if (!newTeam.order || newTeam.order < 0) { newTeam.order = i++; } return newTeam; // eslint-disable-line max-nested-callbacks
+      if (!newTeam.order || newTeam.order < 0) {
+        newTeam.order = i++;
+      }
+
+      return newTeam; // eslint-disable-line max-nested-callbacks
     });
     return newTeams;
   }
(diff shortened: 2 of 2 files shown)


justledbetter:
This pull request has not caused any new tests to fail - Confirmed that existing codebase v4.4.0 is throwing the same errors as what is reported above.

src/common/config/index.js
```diff
@@ -324,6 +327,24 @@ export default class Config extends EventEmitter {
...
+    newTeams = newTeams.filter((newTeam) => {
+      if (!newTeam.order || newTeam.order < 0) {
+        newTeam.order = i++;
+      }
+
+      return newTeam; // eslint-disable-line max-nested-callbacks
+    });
+    return newTeams;
+  }
+
```
Willyfrog: I think we can refine this algorithm, people can add random order (not negative but I can setup 281 as order on a two server setup), so anything depending on user input needs to be resilient.  With that in mind I have a few questions:

what would happen if I had 5 servers and the last one has the order value of 3?

I'm not convinced about using filter here, the only way to filter is if newteam is falsy (null, undefined in this case) and that shouldn't happen, but in case it does it would fail on line 339. How about transition it to either map or reduce?

do we mind having gaps in the order (e.g.: user added an order value of 5 in a 3 server setup)? My guess is that it is going to be fine, but we would have to double check any code depending on order value is expecting gaps. Another option would be to remove the gaps if they exist.

src/common/config/buildConfig.js
```diff
@@ -19,11 +20,12 @@ const buildConfig = {
   defaultTeams: [/*
     {
       name: 'example',
-      url: 'https://example.com'
-    }*/
-  ],
+      url: 'https://example.com',
+      order: 0
```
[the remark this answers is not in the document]
justledbetter: I guess this was more of a documentation change for me than anything.  I'm fine with leaving the new setting out of this file (but leaving it in the comment) if that is more readable on your end.

src/common/config/buildConfig.js
```diff
@@ -19,11 +20,12 @@ const buildConfig = {
...
+    }
+  */],
   helpLink: 'https://about.mattermost.com/default-desktop-app-documentation/',
-  enableServerManagement: true,
+  enableServerManagement: false,
```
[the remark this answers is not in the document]
justledbetter: This one is completely my fault - enableServerManagement should be false.  I've corrected this and pushed the change up.

src/common/config/index.js
justledbetter: I based this code off of the surrounding functions, but am open to other ideas.  You are correct that users that specify nonsequential order flags may clash with this; If the config file specifies order:1 on one entry and leaves two others blank, you would end up with [0, 1, 1].

If order is merely consumed as a sort indicator, this might be "close enough" to what users expect to see; If it needs to be unique (I need to dig through the code and see), this could lead to a crash.

src/common/config/buildConfig.js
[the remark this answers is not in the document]
justledbetter: Fixed in code.

src/common/config/buildConfig.js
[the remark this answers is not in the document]
justledbetter: Fixed in code.

src/common/config/index.js
justledbetter: I've pushed a change to the code that I think addresses this.  It makes two passes on the user's defaultTeams object, first to sort the list numerically by [order], and then again to renumber the entire list from 0..max.  This ensures the user's specified sort order is preserved, while also making sure that we have an item with order:0, which seems to be required for the first tab to be auto-raised on startup.

(There may be another way to fix this, but I'm sticking with the only section of code I've figured out so far.  :) )

src/common/config/index.js
```diff
@@ -324,6 +327,34 @@ export default class Config extends EventEmitter {
...
+    // Make a best pass at interpreting sort order. If an order is not specified, assume it is 0.
+    //
+    const newTeams = teams.sort((x, y) => {
+      if (x.order == null) {
+        x.order = 0;
+      }
+      if (y.order == null) {
+        y.order = 0;
+      }
+      return (x.order === y.order) ? 0 : ((x.order > y.order) ? 1 : -1); // eslint-disable-line no-nested-ternary
```
Willyfrog: Sorting numbers can be efficiently done by simply substracting.
```suggestion
      // once we ensured `order` exists, we can sort numerically
      return x.order - y.order;
```

src/common/config/index.js
```diff
@@ -324,6 +327,34 @@ export default class Config extends EventEmitter {
...
+    });
+
+    // Now re-number all items from 0 to (max), ensuring user's sort order is preserved. The
+    // new tabbed interface requires an item with order:0 in order to raise the first tab.
+    //
+    let i = 0;
+    newTeams.forEach((team) => {
+      team.order = i++;
+    });
```
Willyfrog: We can remove handling the `i` by using its actual index
```suggestion
    newTeams.forEach((team, i) => {
      team.order = i;
    });
```

justledbetter:
Built and confirmed that everything works as expected after @Willyfrog's changes.

Added one commit to re-remove the optional 'order' field from the sample config file (and also to respawn the build_and_test job).  Everything should come up clean now!

src/common/config/index.js
```diff
@@ -324,6 +327,35 @@ export default class Config extends EventEmitter {
...
+
+      // once we ensured `order` exists, we can sort numerically
+      return x.order - y.order;
+    });
+
+    // Now re-number all items from 0 to (max), ensuring user's sort order is preserved. The
+    // new tabbed interface requires an item with order:0 in order to raise the first tab.
+    //
+    newTeams.forEach((team, i) => {
+      team.order = i;
```
devinbinnie: Is the order here being saved back to `config.json`? Or does that not happen until the order is changed by dragging tabs?

justledbetter: I'm not sure (I've only understood the code enough to fix the bug.)  How can I test that?

devinbinnie: Open up `config.json` in a text editor, then open Mattermost and see if the order changes on load. I assume it would only happen if there were new GPO teams it had to account for.

justledbetter: Done - Just confirmed that the config file is not updated until the user changes it (add/remove server, move tabs around, etc).

devinbinnie: Okay cool, can you also confirm that when moving a tab and then restarting the app the tabs appear in the same order as they were before restarting?

justledbetter: Yep, I tested that when I ran through the previous test. It remembered the user's desired tab sort order.

devinbinnie: Great, thanks :)

Willyfrog:
great! for GPO you can use one of https://app.circleci.com/pipelines/github/mattermost/desktop/1101/workflows/ae2431dc-c2b1-479b-a9da-35590a8d31ce/jobs/5910/artifacts and GPO docs. @ogi-m let me know if you need any help setting it up

Willyfrog:
@justledbetter I'll do the cherry picking, don't worry
thanks for your contribution! :D

PR #1248, opened 2020-03-26
[MM-23153] Fixed the issue where back/forward navigation not in the main window causes the app to crash

devinbinnie:

**Summary**
This PR stops the crash that happens when you try to navigate back/forward with an OAuth window open, now the back/forward buttons will correctly work with the focused window.

**Issue link**
https://mattermost.atlassian.net/browse/MM-23153

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -214,7 +214,7 @@ function createTemplate(mainWindow, config, isDev) {
         if (focusedWindow === mainWindow) {
           mainWindow.webContents.send('go-back');
         } else if (focusedWindow.webContents.canGoBack()) {
-          focusedWindow.goBack();
+          focusedWindow.webContents.goBack();
         }
       },
     }, {
@@ -224,7 +224,7 @@ function createTemplate(mainWindow, config, isDev) {
         if (focusedWindow === mainWindow) {
           mainWindow.webContents.send('go-forward');
         } else if (focusedWindow.webContents.canGoForward()) {
-          focusedWindow.goForward();
+          focusedWindow.webContents.goForward();
         }
       },
     }],
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -210,22 +210,14 @@ function createTemplate(mainWindow, config, isDev) {
     submenu: [{
       label: 'Back',
       accelerator: process.platform === 'darwin' ? 'Cmd+[' : 'Alt+Left',
-      click: (item, focusedWindow) => {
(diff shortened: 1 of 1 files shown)


src/main/menus/app.js
```diff
@@ -210,22 +210,14 @@ function createTemplate(mainWindow, config, isDev) {
...
       label: 'Back',
       accelerator: process.platform === 'darwin' ? 'Cmd+[' : 'Alt+Left',
-      click: (item, focusedWindow) => {
-        if (focusedWindow === mainWindow) {
-          mainWindow.webContents.send('go-back');
-        } else if (focusedWindow.webContents.canGoBack()) {
-          focusedWindow.goBack();
-        }
+      click: () => {
+        mainWindow.webContents.send('go-back');
```
larkox: If I hit back on the popup, wouldn't this make the mainWindow to go back? I feel that is not what the user experience should feel like.
Wouldn't it be better to leave this as `if (focusedWindow === mainwindow) do nothing`? (Or even better if the button is not there, but I understand that is not possible / way more complicated)

devinbinnie: It's a menu shortcut, so we can't hide the button. We could do nothing on `focusedWindow !== mainWindow` I suppose, but I wanted the users to still be able to go back/forward with the main window. I'm 0/5 on this.

devinbinnie:
> taking a look at the problem (i was testing oauth for a plugin so it was easy) I see the error is:
> `TypeError: focusedWindow.goBack is not a function`
> 
> doesn't this browser window have a webcontent to use the goback?

Yeah, but it sounds like we actually just don't want to allow the OAuth window to go back/forward in th first place.

Willyfrog:
> Yeah, but it sounds like we actually just don't want to allow the OAuth window to go back/forward in th first place.

@larkox is that so? or only because is crashing?

larkox:
My first (and obvious) complain was the crashing. Regarding whether going back is useful or desirable on an OAuth flow, I have not the expertise to answer, but it sounds that should not be there.

devinbinnie:
Direction from @esethna is to allow it to work on any focused window.

src/main/menus/app.js
larkox: Yes, I meant `focusedWindow !== mainWindow`. Nevertheless if this is the desired behaviour, I have no more complains.

amyblais:
@devinbinnie Can you help cherry-pick this to v4.4 as well?

PR #1249, opened 2020-03-30
[MM-23195] fix logic to detect if windows is outside of display boundaries

Willyfrog:

**Summary**
logic for detecting if the window was within a display was backwards. It only happened on restore, not on first launch.

**Issue link**
[MM-23195](https://mattermost.atlassian.net/browse/MM-23195)

**Test Cases**

**Additional Notes**

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -155,7 +155,7 @@ function initializeConfig() {
 function initializeAppEventListeners() {
   app.on('second-instance', handleAppSecondInstance);
   app.on('window-all-closed', handleAppWindowAllClosed);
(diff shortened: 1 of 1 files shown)


Willyfrog:
while I created [a ticket](https://mattermost.atlassian.net/browse/MM-23801) for completely removing all this code, in the meantime we need to fix this.

Willyfrog:
@jgilliam17 it shouldn't jump around. fixed

if position is hiddes the window for more than 50%, it sends to the center on the same display.

PR #1269, opened 2020-04-16
[MM-23215] prevent mouse event to go through and click on window

Willyfrog:

**Summary**
when clicking on the close button with the new tabbar, on windows (not sure why only this one) it goes through and prevents the app from losing focus, which the webapp interprets as having the window visible

**Issue link**
[MM-23215](https://mattermost.atlassian.net/browse/MM-23215)

**Test Cases**
have two users on the same channel, one on the app and another somewhere else
close the app 
mention the user on the app from the other one
expected: notification appears
**Additional Notes**

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -520,17 +520,20 @@ export default class MainPage extends React.Component {
     }
   }
 
-  handleClose = () => {
+  handleClose = (e) => {
+    e.stopPropagation(); // since it is our button, the event goes into MainPage's onclick event, getting focus back.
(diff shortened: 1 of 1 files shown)


Willyfrog:
> It's strange that closing it doesn't cause it blur regardless of the user being able to click on the window when closing it.

it does cause blur, but right after that, it gets focus again because the mouse click closed and clicked on the window. preventing to go through the button into the window stops the focusing.

Willyfrog:
switching @jgilliam17 to @stevemudie as this has to be tested mainly on windows

PR #1292, opened 2020-05-12
[MM-23228] warn when setting tray icon on linux

Willyfrog:

**Summary**
On linux, there is a known problem which crashes the desktop. Since it is an upstream problem in the library and a fix is already provided on some modern distros, we are warning that this might be the case and providing a link to useful information for users to try and overcome the limitation.

**Issue link**

[MM-23228](https://mattermost.atlassian.net/browse/MM-23228)

**Test Cases**

Open settings on linux: when trying to turn the `minimize to tray option` on, a warning should appear with a link to a documentation page.

This should not happen if platform is other than linux or while turning it off.

**Additional Notes**
I created a component to override security on UI-based links to go to external sites.

It currently is missing the actual desktop channel post, as I want to have a first review on functionality, as the post can be edited later on. It is linking to the jira ticket for now.

**Screenshots**
Screenshot 2020-05-12 at 14 48 18

### DIFF
--- a/src/browser/components/DestructiveConfirmModal.jsx
+++ b/src/browser/components/DestructiveConfirmModal.jsx
@@ -19,7 +19,9 @@ export default function DestructiveConfirmationModal(props) {
       <Modal.Header closeButton={true}>
         <Modal.Title>{title}</Modal.Title>
       </Modal.Header>
-      {body}
+      <Modal.Body>
+        {body}
+      </Modal.Body>
       <Modal.Footer>
         <Button
           bsStyle='link'
--- a/src/browser/components/MattermostView.jsx
+++ b/src/browser/components/MattermostView.jsx
@@ -79,7 +79,6 @@ export default class MattermostView extends React.Component {
     // Open link in browserWindow. for example, attached files.
     webview.addEventListener('new-window', (e) => {
       if (!Utils.isValidURI(e.url)) {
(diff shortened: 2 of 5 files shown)


devinbinnie:
This feels like something we shouldn't be drawing a lot of attention to. It sounds like this crash is pretty rare and won't affect many users, plus I don't think we officially support Linux Mint.

I'm all for trying to help out our users, but I feel Eric might be right and we'll be showing a prompt that will likely scare more users than help them out.

I feel like we could address this as a Known Issue in our release notes somewhere, and provide the same links there that users could go check out, rather than code this message into the app, which should be permanently fixed at some point in the future.

deanwhillier:
@esethna, @devinbinnie, this ticket was groomed and the general approach was approved. That said, I think the exact approach could use a followup conversation and some UX thoughts. Putting a WIP label on for now and putting the ticket back into grooming for some refinement.

Willyfrog:
Just to address some comments,

I currently can't think of a way to differentiate. I'm only aware of it being fixed in latest Ubuntu.

The report affects many Linux distros, not only mint although it was originally detected there.

Other than that, yeah, let's groom it again 👍

Willyfrog:
if we work on this again, we'll better start off from scratch as the code base has changed a lot. Closing

PR #1295, opened 2020-05-18
[MM-22648] basic auth external sites

Willyfrog:

**Summary**
currently basic web authentication was only possible agains server uris, with this we enable to request permission to use it on external sites.

Additionally fixes two bugs found during development:
- stop using an object `stringified` as a key, order of an object is not guaranteed and as such the key will be different.
- when a user failed to login using the basic auth modal, the modal will reappear with the data as he left, but internal state would be cleared, if the user didn't change one of the fields (the username generally) it would be sent blank, failing to authenticate. component is now using state to show that both fields have been cleared.

**Issue link**

[MM-22648](https://mattermost.atlassian.net/browse/MM-22648)
fixes https://github.com/mattermost/mattermost-server/issues/14474

**Test Cases**

see jira ticket for steps

**Additional notes**
1. Added babel to main compilation so it can use a more modern JS

1. I've added the `ExternalLink` component that I created in https://github.com/mattermost/desktop/pull/1292 since that one hasn't been merged yet but the same functionality is needed here. Expect a conflict if merging the other PR, although is the same code so it should be a simple one.

**Screenshots**
Screenshot 2020-05-19 at 00 05 01
Screenshot 2020-05-19 at 00 05 10


Willyfrog:
failing test are unrelated to changes, sadly since moving towards electron7 it is getting more flakier

Willyfrog:
@esethna I added a video showing how it works into [its ticket](https://mattermost.atlassian.net/browse/MM-22648?focusedCommentId=88205) (sorry it was too big for GH) hope this helps, as setting up the testing environment is a bit cumbersome

nevyangelova:
Do we give any user feedback? What happens when they enter a wrong password, or the URL can't be parsed, or any steps fail along the way? Might be nice to add some feedback :P apart from that LGTM!

Willyfrog:
> Do we give any user feedback? What happens when they enter a wrong password, or the URL can't be parsed, or any steps fail along the way? Might be nice to add some feedback :P apart from that LGTM!

definitely, I'll create a ticket for that, as I only fixed the weird situation when it made you think you wrote the wrong username/pass but it was actually the code sending it wrong.

jupenur:
Unsure if related to this change or not, but when I connect to a new server that's behind Basic authentication, hit `Cancel` on the authentication modal, then hit `Ctrl+R` to reload, I get `ERR_INVALID_URL`. Instead I would expect the modal to reapper.

Willyfrog:
> Unsure if related to this change or not, but when I connect to a new server that's behind Basic authentication, hit `Cancel` on the authentication modal, then hit `Ctrl+R` to reload, I get `ERR_INVALID_URL`. Instead I would expect the modal to reapper.

I'd say it is a separate bug, we are basically ignoring the cancel while we should be calling the callback without arguments (as [stated on the documentation](https://www.electronjs.org/docs/api/app#event-login))
I'll create a new bug: https://mattermost.atlassian.net/browse/MM-25323

src/browser/components/MainPage.jsx
```diff
@@ -836,6 +839,9 @@ export default class MainPage extends React.Component {
           onLogin={this.handleLogin}
           onCancel={this.handleLoginCancel}
         />
+        <PermissionModal
+          webContentsId={remote.getCurrentWindow().webContents.id}
```
devinbinnie, line 843: Can the `PermissionModal` not get this itself? Doesn't feel right passing props into a component not used by the parent.

Willyfrog: done, I also realised that we are currently downloading a shim for object.fromEntries already, so i removed the custom code.

jupenur:
Overlapping permission requests cause the app to crash (?) turning the entire window blank. To repro, post the following where `example.com` is a server running Basic auth:

```
![one](http://example.com/one)
![two](http://example.com/two)
```

Hit `Deny` on the permission request to observe the crash.

jupenur:
It also seems the permission request doesn't get triggered for IP addresses. I tried ``, ``, and `` -- all simply failed to load.

Willyfrog:
regarding posts using images behind basic auth, there is something being cached, as I see a request being rejected with 401 on the proxy once we cancel the first, so there is no login process there. I'm seeing a similar behaviour on the browser (on firefox it is never answering back)
I've fixed https://mattermost.atlassian.net/browse/MM-25323 in the process to see if that one was partially responsible, but seems to keep working the same way.

I've alos addressed @esethna comments

jupenur:
The crash still reproduces. Logs show this:

image

Offending line is at https://github.com/mattermost/desktop/pull/1295/files#diff-5f8b4435e102e79f5613703f241edf7aR108

Willyfrog:
thanks @jupenur that was superhelpfull! already patched, let me know if you find any other problem.

Willyfrog:
Screenshot 2020-05-22 at 11 46 10
Screenshot of the changes requested by @esethna so @andrewbrown00 can check

Willyfrog:
Addressed all but the last point, as this would require work on all modals. Filling a [ticket for it](https://mattermost.atlassian.net/browse/MM-25491) (feel free to add details to it if you want something else as it is a bit generic)
Screenshot 2020-05-25 at 12 10 07

** Additional Notes **
I've added the `ExternalLink` component that I created in https://github.com/mattermost/desktop/pull/1292 since that one hasn't been merged yet but the same functionality is needed here. Expect a conflict if merging the other PR, although is the same code so it should be a simple one.

PR #1301, opened 2020-05-27
[MM-25429]/[MM-25433] fix finder focus, align buttons, keep state consistent

Willyfrog:

**Summary**

Focus was being lost, due to click went through and was also captured by the tabbar, which then autofocused itself. Since the searchbar was probably misplaced, I moved it to where it seemed to look right. Also fixed misalignment of buttons mentioned on the ticket.

**Issue link**

[MM-25429](https://mattermost.atlassian.net/browse/MM-25429)
[MM-25433](https://mattermost.atlassian.net/browse/MM-25433)

**Test Cases**

Steps:

1. open searchbar (cmd+f)
1. click outside
1. click on searchbar

Result:

Focus should remain on searchbar

**Additional Notes**

@andrewbrown00 should we create a separate story for styling the search bar in dark mode?

**Screenshots**
mac
Screenshot 2020-05-27 at 11 21 44
win
Screenshot 2020-05-27 at 11 35 34


Willyfrog:
don't do dev review yet, as I found partially the problem but there seems to be some similar effects (although not always)

Willyfrog:
now it's good to review

andrewbrown00:
> @andrewbrown00 should we create a separate story for styling the search bar in dark mode?

@Willyfrog yes, let's create separate tickets to handle the styling for dark modes.

Also, will you post a link for me to download and test the desktop app with the changes.

Willyfrog:
> Did it you move it in this PR?

Yes, that's the reason why I asked your review, it felt out of place and it was partially causing a problem (shouldn't happen now)

Other than that I didn't do anything else to the element, but I'll address your feedback.

Willyfrog:
> 
>     * [x]  Move the control back in to the header so it doesn't cover the channel header control. Did it you move it in this PR?

As mentioned above, I did move, having the square border at the top made me think it had moved from below.

>     * [x]   Invert the arrow control states, they are backwards. They default state should not have a fill and the gray should be the hover state.

done. Please, notice that I added a small margin on the left to separate from search box and added round borders. search box already had a border and it felt weird. let me know if you want it back.

>     * [ ]  Add the gray hover state to the `X` close control. Make sure the corner radius matches what's on the arrows and is the same for all corners.

added. it is separated from the arrow controls, so it has round border on all corners. if you want the three of them to form a single unit, let me know.

>     * [ ]  Something funky is going on with the count. It show one count at first, then when I go to the next the count changes. Sometimes it would change to a random count that didn't match the results. I wasn't able to consistency repo it.

Can't do anything about it, as far as I know. When you move with the arrows the channel might scroll up/down revealing new content, so the search count changes as there is new text.

>     * [ ]  There's an extra click needed when I get to the top or bottom of the list ( i.e. 1/8 or 8/8) when clicking the arrows. Clicking the arrows should always move to the next item without and extra click.

fixed, it was restarting the search.

>     * [ ]  Can we easily add the `⌘ G` / `Ctrl G` keyboard shortcut to go through results like browsers do?

i can add that easily as a hack, but to proper implement that (like having a menu item in the top bar that greyes out if finder is not accesisble) I might need a few architectural changes. I'll queue an item for the hangouts and add a ticket for it.

andrewbrown00:
@Willyfrog I'd like to circle back on this to look at the overall experience to:
* Make sure the controls all work as expected, i.e. right now I can't paste in to the input
* Should we extend the feature to support `⌘ F` and `⌘ Shift F`? One could open the browser search control as we do today and the other could open the global search and add the `in: channelName` modifier based on the channel the user is in
image
* ...

PR #1335, opened 2020-07-02
[MM-26665] fix security issues from dependabot

Willyfrog:
**Summary**
upgrade minimist and kind-of deps for security issues

**Issue link**

[MM-26665](https://mattermost.atlassian.net/browse/MM-26665)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,832 +11,920 @@
       "dev": true
     },
     "@babel/code-frame": {
-      "version": "7.0.0",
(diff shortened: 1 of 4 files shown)


amyblais:
@Willyfrog Is this for a dot release? The Jira ticket refers to v4.5 milestone.

Willyfrog:
since it was security, I was originally thinking about doing a dot release, but as @jupenur mentioned it might not be needed. so we can add this to 4.6 instead.
I'll change the ticket

PR #1344, opened 2020-07-09
Replace npm install with npm ci on CircleCI

jupenur:

**Summary**
Builds on CircleCI currently use `npm install` for dependencies, when they should be using `npm ci`. That way `package-lock.json` does not get modified during build and we can be sure of the versions we ship. This PR fixes that.

**Issue link**

**Test Cases**

**Additional Notes**
Not sure if this actually works or if there's some reason we actually have to use `npm install`, but I guess we'll see when the checks run. Also not sure how to test the PowerShell change without cutting a release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ commands:
         default: ""
     steps:
       - run: apt-get update && apt-get -y install << parameters.apt_opts >>
-      - run: npm install
+      - run: npm ci
 
   win_make:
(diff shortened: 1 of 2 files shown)


jupenur:
This messes with the caching we currently have, but I don't know if there are any good alternatives. There doesn't seem to be any flag to stop `npm ci` from deleting `node_modules/` nor to make `npm i` respect the lockfile. Suggestions welcome.

PR #1345, opened 2020-07-13
[MM-26832] don't show the window until it is ready

Willyfrog:

**Additional Notes**
I've yet to test on windows and ubuntu 20.20 so I'm setting the wip flag until then. but feel free to review in the meantime, it won't take long 😄 

**Summary**

Fixes a race condition when using the `--hidden` flag in the cli (or from using autolaunch) that was causing it will try to show and hide the window.

**Issue link**

[MM-26832](https://mattermost.atlassian.net/browse/MM-26832)

**Test Cases**
Two possibilities for testing the same

1. Autostart:
  - set autostart on login from the preferences
  - restart OS

2. Terminal
- launch the app from the command line with the `--hidden` flag

Expected:
- app should show minimized

### DIFF
--- a/src/main/mainWindow.js
+++ b/src/main/mainWindow.js
@@ -44,7 +44,7 @@ function createMainWindow(config, options) {
     windowOptions = {width: defaultWindowWidth, height: defaultWindowHeight};
   }
 
-  const {hideOnStartup, trayIconShown} = options;
+  const {hideOnStartup} = options;
   const {maximized: windowIsMaximized} = windowOptions;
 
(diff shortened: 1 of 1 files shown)


Willyfrog:
this prevents an initial load of the webapp code, and the user is not aware of having unread mentions (the small red dot on the mm tray icon)

Willyfrog:
suppossedly the flag for painting is turned on by default, but I noticed that on ubuntu 19.04 having it explicitly set to true makes the app render 🤷‍♂️

Issue #1328, opened 2020-06-19
Unable to debug "ERR_BAD_SSL_CLIENT_AUTH_CERT"


bbodenmiller:
Trying to sign in to Mattermost Desktop using Smart Badge with certificate. I'm getting the following error message:

![image](https://user-images.githubusercontent.com/1192780/85181668-5c535080-b23b-11ea-8eaf-f495cbe7a544.png)

I've tried debugging via Developer tools for application wrapper and current server but neither return any useful details. Security sections are all showing items as valid.

Sign in works fine from Chrome. How can this be debugged and fixed?

bbodenmiller:
@Willyfrog @lieut-data @devinbinnie @jgilliam17 this seems to have started around https://github.com/mattermost/desktop/pull/1169. Ideas?

Willyfrog:
is this with the latest version?
that code is for handling certificate problems, do you see any error being raised or in the console logs?

bbodenmiller:
@Willyfrog yes this is with 4.5.0. I do not see any further errors in Developer tools for application wrapper and current server. Should I be looking for logs somewhere else?

Willyfrog:
if you look where the config is saved (depends on OS and install method), there might be some extra logs.

Do you get a dialog like the one shown here: https://github.com/mattermost/desktop/pull/1148#issuecomment-575390778

in the network tab, are there differences between the one on the desktop (current server) and the one on chrome?

bbodenmiller:
@Willyfrog Yes...

![image](https://user-images.githubusercontent.com/1192780/85438197-9c1e7e80-b540-11ea-81f5-796a41e89432.png)

then get prompted for PIN

then get

![image](https://user-images.githubusercontent.com/1192780/85181668-5c535080-b23b-11ea-8eaf-f495cbe7a544.png)

App Wrapper dev tools show

![image](https://user-images.githubusercontent.com/1192780/85440503-fb7d8e00-b542-11ea-8e20-e2d5f5fafcf6.png)

I found a file at %localappdata%\programs\mattermost-desktop with `[0619/183653.213:ERROR:registration_protocol_win.cc(131)] TransactNamedPipe: The pipe has been ended. (0x6D)`. Not finding any other logs with details in meaningful details in `C:\Users\xxxx\AppData\Roaming\Mattermost`

Willyfrog:
A couple of things to try:
[...]

PR #1347, opened 2020-07-15
change certificate store key to webcontents id

Willyfrog:
**Note**: This is a test build for investigating some issues with certificates. but if changing the key actually fixes the problem it might be an actual PR, so keeping this around for context and ease of access.

**Summary**
Adds logs to the select client certificate process
change key from using url to webcontents id, so having multiple server tabs requesting the same url for authentication doesn't cause any problems.

related issues:  #1328

### DIFF
--- a/src/browser/components/MainPage.jsx
+++ b/src/browser/components/MainPage.jsx
@@ -142,9 +142,10 @@ export default class MainPage extends React.Component {
       this.loginRequest(event, request, authInfo);
     });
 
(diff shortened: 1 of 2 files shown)


Willyfrog:
this is waiting on a customer to see if it improves having multiple tabs authenticate against the same server

emilyacook:
Hey @Willyfrog should I close this or set it to frozen?

Willyfrog:
I'll just close it. I don't think we are going to implement this for the moment

PR #1351, opened 2020-07-19
[MM-467] Notification sounds

rvillablanca:
**Summary**

This PR is the continuation of another PR in the webapp to support custom notification sounds and adds the necessary functionality to play correctly the sounds.

**Ticket**
[MM-467](https://mattermost.atlassian.net/browse/MM-467])

Related webapp PR: https://github.com/mattermost/mattermost-webapp/pull/5721


rvillablanca:
@esethna I need a definition here before implementing this, so let me explain the current behaviour

* Windows
    * lower or equal to 8.1 plays a custom sound
    * greater than 8.1 plays the native notification sound
* Linux is working the same as webapp in browser
* Mac: I have not done tests because I don't have one but my suspect is there is no custom sound for mac, meaning that it is always playing a native sound

Now we are adding 6custom sounds for the webapp, that means that for desktop app in Windows and Mac It would have 7 sound (6 custom + native). So the idea is the webapp can offer the selection the custom sounds only when the desktop app version is greater than x.y.z (up to you) as an older version of the app won't have the custom sounds bundled in the desktop app so it should not be able to play them.

If the desktop app is newer then the webapp will show the option to select custom sounds so until here there is no problem. My concern is as we are giving the option to choose custom sounds, when an user selects one, then the native sounds (for windows > 8.1 and mac) will be lost because it was overwritten by the custom selection. So I would like to know if this is ok or we should do another thing to handle this. Keep in mind the selection is only for the 6 custom sounds and the native one is not present in the list. 

Maybe we should add the native option when the desktop app is detected, of course it complicates a bit more the things but I'm here to solve this :wink: 

@Willyfrog I you see I said something wrong please let me know

Willyfrog:
@rvillablanca i haven't tested it, but there is the `silent` option in the options object passed to the constructor, so if the notification is going to be custom, we might want to use that to prevent having both sounds. See [specs](https://developer.mozilla.org/es/docs/Web/API/Notification/Notification).

rvillablanca:
Hi @Willyfrog this is ready to be reviewed, I left some comments [here](https://github.com/mattermost/mattermost-webapp/pull/5721#issuecomment-662256646) that are related with this PR :wink:

src/browser/js/notification.js
```diff
@@ -28,17 +41,21 @@ export default class EnhancedNotification extends OriginalNotification {
       Reflect.deleteProperty(options, 'icon');
     }
 
+    const customSound = !options.silent;
```
Willyfrog, line 44:
```suggestion
    const isWin7 = (process.platform === 'win32' && osVersion.isLowerThanOrEqualWindows8_1() && DEFAULT_WIN7);
    const customSound = !options.silent && ((options.data && options.data.soundName) || isWin7)
```

src/browser/js/notification.js
```diff
@@ -28,17 +41,21 @@ export default class EnhancedNotification extends OriginalNotification {
...
       title,
       options,
     });
 
-    if (process.platform === 'win32' && osVersion.isLowerThanOrEqualWindows8_1()) {
-      if (!options.silent) {
-        playDing();
-      }
+    if (customSound) {
+      playSound(options.data.soundName);
```
Willyfrog, line 58:
```suggestion
      playSound(customSound);
```

src/browser/js/notification.js
```diff
@@ -8,14 +8,27 @@ import {throttle} from 'underscore';
...
+import ripple from '../../assets/sounds/ripple.mp3';
+import upstairs from '../../assets/sounds/upstairs.mp3';
+
+export const notificationSounds = new Map([
+  ['Bing', bing],
+  ['Crackle', crackle],
+  ['Down', down],
+  ['Hello', hello],
+  ['Ripple', ripple],
+  ['Upstairs', upstairs],
```
Willyfrog, line 24: should we also have the 'Ding' sound that is used in win7?

src/browser/js/notification.js
```diff
@@ -8,14 +8,27 @@ import {throttle} from 'underscore';
...
+]);
```
Willyfrog, line 25:
```suggestion
]);
const DEFAULT_WIN7 = 'Ding';
```

src/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "mattermost-desktop",
   "productName": "Mattermost",
   "desktopName": "Mattermost.desktop",
-  "version": "4.6.0-develop",
+  "version": "4.9.0-develop",
```
Willyfrog, line 5: I think you upgraded the version by mistake (most likely your own manual testing :) )

it is also the reason why the test are failing to build

src/utils/util.js
```diff
@@ -204,7 +204,8 @@ const dispatchNotification = async (title, body, silent, handleClick) => {
     tag: body,
     icon: appIconURL,
     requireInteraction: false,
-    silent
+    silent,
+    data: {soundName}
```
Willyfrog, line 208: why not send data as the original argument in the message passing from the webapp? that way, if we change what's inside data because we want to add some more metadata/options we won't have to change all the signatures as you had to do.

src/browser/js/notification.js
rvillablanca: If we are goint to support the old win7 sound (Ding) I think we should

src/utils/util.js
rvillablanca: Ok I got it :)

rvillablanca: Ok I got it

src/browser/js/notification.js
```diff
@@ -28,17 +45,23 @@ export default class EnhancedNotification extends OriginalNotification {
       Reflect.deleteProperty(options, 'icon');
     }
 
+    const isWin7 = (process.platform === 'win32' && osVersion.isLowerThanOrEqualWindows8_1() && DEFAULT_WIN7);
+    const customSound = !options.silent && ((options.data.soundName !== 'None' && options.data.soundName) || isWin7);
+    console.log("Play sound: " + customSound);
```
Willyfrog, line 50:
```suggestion
```

src/package.json
Willyfrog: this has to go back to its previous value

hanzei:
@esethna @jgilliam17 Gentle reminder to review this PR when you have time :)

esethna:
Thanks @rvillablanca, nice work!

jgilliam17:
Verified on Windows 7 - working as expected

hanzei:
Blocking until https://github.com/mattermost/mattermost-webapp/pull/5721 is merged.

esethna:
This can be merged now @hanzei

PR #1374, opened 2020-09-14
[MM-28620] allow navigating links to admin_console

Willyfrog:
**Summary**

Navigate to admin pages when the link clicked tries to open them in a new window.

**Issue link**
[MM-28620](https://mattermost.atlassian.net/browse/MM-28620)
**Test Cases**

Go to admin console, scroll down to `Custom Terms of Service (beta)` and click on `go to Site Configuration > Customization.`

Expected: app loads the page.

**Additional Notes**

since we are stopping window creation and loading a new page, it is expected to have a loading flash.

### DIFF
--- a/src/browser/components/MattermostView.jsx
+++ b/src/browser/components/MattermostView.jsx
@@ -94,7 +94,7 @@ export default class MattermostView extends React.Component {
         } else if (destURL.path.match(/^\/help\//)) {
           // continue to open special case internal urls in default browser
           shell.openExternal(e.url);
-        } else if (Utils.isTeamUrl(this.props.src, e.url, true)) {
+        } else if (Utils.isTeamUrl(this.props.src, e.url, true) || Utils.isAdminUrl(this.props.src, e.url)) {
           e.preventDefault();
           this.webviewRef.current.loadURL(e.url);
         } else if (Utils.isPluginUrl(this.props.src, e.url)) {
--- a/src/main.js
+++ b/src/main.js
@@ -487,7 +487,7 @@ function handleAppWebContentsCreated(dc, contents) {
     const parsedURL = Utils.parseURL(url);
     const server = Utils.getServer(parsedURL, config.teams);
 
-    if ((server !== null && Utils.isTeamUrl(server.url, parsedURL)) || isTrustedPopupWindow(event.sender)) {
+    if ((server !== null && Utils.isTeamUrl(server.url, parsedURL)) || isTrustedPopupWindow(event.sender) || Utils.isAdminUrl(server.url, parsedURL)) {
       return;
     }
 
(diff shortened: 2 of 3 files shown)


nevyangelova (approved):

hmhealey (approved):

hmhealey:
Looks good, but we're probably going to need to extend this at some point. I think plugins can host some content as part of the webapp, so we're probably going to be asked about that at some point.

Willyfrog:
> Looks good, but we're probably going to need to extend this at some point. I think plugins can host some content as part of the webapp, so we're probably going to be asked about that at some point.

they might need it, but right now plugins are the only urls capable of opening in a new window so I'd consider to be good for now. But it might need extending to other pages as well.

jgilliam17:
@Willyfrog 
There is a link in system console that is crashing the desktop app.

Quick solution would be to just `unlink` it and show as plain text, since it is only given as an example here.
We would need to fix the way app handles opening links like this from System Console.
The same link, posted in any channel, opens in the default browser on click.

System Console>Authentication>OAuth 2.0>GitLab
Screen Shot 2020-09-28 at 3 32 26 PM

Screen Shot 2020-09-28 at 3 44 16 PM

jgilliam17:
There is one place where the link (links are identical) behaves differently, top one opens, bottom crashes the app.
I tested on rc.test
`https://rc.test.mattermost.com/_redirect/integrations/bots`
Screen Shot 2020-09-28 at 4 11 58 PM

Willyfrog:
great catch!

In fact that link is missing the target=_blank part, but the app shouldn't crash for that.
Added a check to prevent crashing

hmhealey:
Not directly related to this PR, but I'm not even sure if that OAuth link in the System Console should be a clickable link. I think that URL is an API endpoint for the OAuth login process, not something that a user would hit.

jgilliam17:
@Willyfrog Can you please check why some links are not opening 
like this one:
System Console>Plugins>Zoom>Documentation (on the very bottom or same screen - link to More info under Enable OAuth)

Willyfrog:
> @Willyfrog Can you please check why some links are not opening
> like this one:
> System Console>Plugins>Zoom>Documentation (on the very bottom or same screen - link to More info under Enable OAuth)

that's because it is a link to outside of the server without target=_blank, so it tries to navigate within the app's window which we don't allow. To fix it, the plugin should add `rel="noopener noreferrer" target="_blank"` to its links

jgilliam17 (approved):
Thank you @Willyfrog 
Tested, looks good to merge.
- Verified external links and the ones pointing to the system console or any of the already configured servers on the desktop app work as expected.
- Issue found with some outside links failing to open, filed separately for follow up.
https://mattermost.atlassian.net/browse/MM-29239

PR #1376, opened 2020-09-17
[MM-25789] - Update default settings for new installations

nevyangelova:

**Summary**
Adds the showing of tray icon and minimise icon to tray default options for Linux and Mac OS as well. For the variable `minimizeTray` I only saw that it's used in Linux context that's why I updated the default options directly. But if I missed something let me know.

**Issue link**
https://mattermost.atlassian.net/browse/MM-25789

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -699,7 +699,7 @@ function initializeAfterAppReady() {
 
   mainWindow = createMainWindow(config.data, {
     hideOnStartup,
-    trayIconShown: process.platform === 'win32' || config.showTrayIcon,
+    trayIconShown: config.showTrayIcon,
     linuxAppIcon: path.join(assetsDir, 'appicon.png'),
     deeplinkingUrl,
   });
@@ -1145,10 +1145,7 @@ function getDeeplinkingURL(args) {
 }
 
 function shouldShowTrayIcon() {
-  if (process.platform === 'win32') {
-    return true;
-  }
-  if (['darwin', 'linux'].includes(process.platform) && config.showTrayIcon === true) {
+  if (config.showTrayIcon === true) {
     return true;
   }
   return false;
--- a/src/main.js
+++ b/src/main.js
@@ -699,7 +699,7 @@ function initializeAfterAppReady() {
 
   mainWindow = createMainWindow(config.data, {
     hideOnStartup,
-     trayIconShown: process.platform === 'win32' || config.showTrayIcon, 
+    trayIconShown: process.platform === 'win32' || config.showTrayIcon,
     linuxAppIcon: path.join(assetsDir, 'appicon.png'),
     deeplinkingUrl,
   });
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -699,7 +699,7 @@ function initializeAfterAppReady() {
 
   mainWindow = createMainWindow(config.data, {
     hideOnStartup,
-    trayIconShown: process.platform === 'win32' || config.showTrayIcon,
+    trayIconShown: config.showTrayIcon,
```
Willyfrog, line 702: changing this line, allows win32 users to not show the icon. Was that intended?

src/main.js
```diff
@@ -1145,10 +1145,7 @@ function getDeeplinkingURL(args) {
 }
 
 function shouldShowTrayIcon() {
-  if (process.platform === 'win32') {
-    return true;
-  }
-  if (['darwin', 'linux'].includes(process.platform) && config.showTrayIcon === true) {
+  if (config.showTrayIcon === true) {
```
Willyfrog, line 1148: same here

Willyfrog (changes requested):
I'm not sure you wanted to allow windows users to disable their tray icon.

src/main.js
nevyangelova: oh I understood that since showTrayIcon will be initially true this logic will affect the windows users as well

src/main.js
nevyangelova: Doesn't this now include all operating systems? Since the only needed condition is that `showTrayIcon` is true?

Willyfrog: but having the default to true would be different than forcing win32 to be true, right? so it should be something like:
```suggestion
  if (config.showTrayIcon === true || process.platform === 'win32') {
```

src/main.js
Willyfrog, line 702:
```suggestion
     trayIconShown: process.platform === 'win32' || config.showTrayIcon, 
```

Willyfrog (changes requested):

nevyangelova:
Thanks @Willyfrog I think I understand the issue you mentioned now. Sorry for the confusion :(

Willyfrog (approved):
LGTM! thanks :D

Willyfrog:
seems like my editing caused a missmatch of number of spaces, which prevents linting from running properly, sorry about that

jgilliam17 (approved):
Thank you @nevyangelova 
Tested, looks good to merge.
On macOS - Verified 'Show Mattermost icon in the menu bar' is enabled by defaults for new installs 
On Ubuntu - Verified ‘Show icon in the notification area’ and ‘Leave app running in the notification area when application window is closed’ are enabled by default for new installs

devinbinnie (approved):

PR #1379, opened 2020-10-01
[MM-27332] show window at autolaunch

Willyfrog:

**Summary**

on autostart don't minimize the window

**Issue link**

[MM-27332](https://mattermost.atlassian.net/browse/MM-27332)

**Test Cases**

Case 1
on windows/linux
- open preferences and tick the autostart on login
Expected:
upon restart, the app open with visible window

Case 2
on windows/linux
- open prefereces and untick same option
Expected
upon restart, the app doesn't start automatically

### DIFF
--- a/src/browser/components/SettingsPage.jsx
+++ b/src/browser/components/SettingsPage.jsx
@@ -699,8 +699,6 @@ export default class SettingsPage extends React.Component {
           {'Start app on login'}
           <HelpBlock>
             {'If enabled, the app starts automatically when you log in to your machine.'}
-            {' '}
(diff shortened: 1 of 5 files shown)


nevyangelova (approved):

devinbinnie (approved):

jgilliam17:
Thank you @Willyfrog 
- Works as expected on Windows for both use cases listed.
- On Ubuntu (VM latest version) the app does not start on login when autostart is selected.

Willyfrog:
merging since I tested that scenario and worked for me, we can chat later to see if there was some other problem involved or do a second RC without this PR in case we missed something

PR #1385, opened 2020-10-05
Update NOTICE.txt

amyblais:


NOTICE.txt
```diff
@@ -21,7 +342,1033 @@ Launch node applications or executables at login (Mac, Windows, and Linux)
...
+
+This product contains 'font-awesome' by Fort Awesome.
+
+The iconic SVG, font, and CSS framework
+
+* HOMEPAGE:
+  * https://github.com/FortAwesome/Font-Awesome
+
+* LICENSE: Font Awesome Free License
+-------------------------
```
amyblais, line 1300: If this is the correct one, then we need to contact the repo owners because the license is not one of the open source ones https://opensource.org/licenses/alphabetical.

NOTICE.txt
```diff
@@ -21,7 +342,1033 @@ Launch node applications or executables at login (Mac, Windows, and Linux)
...
+
+This product contains 'hapi/joi' by Sideway Inc.
+
+The most powerful schema description language and data validator for JavaScript.
+
+* HOMEPAGE:
+  * https://github.com/sideway/joi
+
+* LICENSE: Copyright (c) 2012-2020, Sideway Inc, and project contributors.
+Copyright (c) 2012-2014, Walmart.
```
amyblais, line 1346: If this is the correct one, then we need to contact the repo owners because the license is not one of the open source licenses https://opensource.org/licenses/alphabetical.

NOTICE.txt
```diff
@@ -43,21 +1390,33 @@ SOFTWARE.
...
 
-The most popular front-end framework for developing responsive, mobile first projects on the web.
+jq is a lightweight and flexible command-line JSON processor.
 
 * HOMEPAGE:
-  * https://getbootstrap.com/
+  * https://github.com/stedolan/jq
+
+* LICENSE:
+
```
amyblais, line 1403: If this is the correct one, then we need to contact the repo owners because there is no license listed on their repo.

NOTICE.txt
Willyfrog: which one isn't open source? there is a different license depeding on the asset (src, icons or fonts) but I believe all of them are open.

NOTICE.txt
Willyfrog: I believe it is MIT licensed.
its license is not in the usual place but in [copying](https://github.com/stedolan/jq/blob/master/COPYING)

Willyfrog:
nice work, amy! while reviewing the file I thought it might be easier to read in github if we use markdown, so I added the changes to make it happen, let me know what you think, if you prefer the old txt, we can revert the changes.

NOTICE.txt
amyblais: I see now that there are different licenses listed.

amyblais:
Thanks @Willyfrog, the NOTICE files in other repos (e.g. https://github.com/mattermost/mattermost-mobile/blob/master/NOTICE.txt and https://github.com/mattermost/mattermost-webapp/blob/master/NOTICE.txt) are also txt files instead of markdown and I have no idea if it's by design or not. @jasonblais do you know if it's ok to change this into a markdown file?

jasonblais:
@Willyfrog I see your perspective and agree Markdown renders more nicely in the GitHub UI. However, for the NOTICE file, .txt is preferred. If a legal team reviews it, they may download the raw .txt file and be confused by the Markdown formatting (like ##, **, etc). That is also why LICENSE file is in .txt instead of .md.

cc @amyblais

Willyfrog:
> @Willyfrog I see your perspective and agree Markdown renders more nicely in the GitHub UI. However, for the NOTICE file, .txt is preferred. If a legal team reviews it, they may download the raw .txt file and be confused by the Markdown formatting (like ##, **, etc). That is also why LICENSE file is in .txt instead of .md.
> 
> cc @amyblais

Sure, I'll revert the changes.

Willyfrog:
reverted and added licenses for joi and jq, for Font Awesome I didn't do anything as I'm not sure how to display that one.

Willyfrog (approved):

NOTICE.txt
```diff
@@ -8,6 +8,327 @@ This document includes a list of open source components used in Mattermost Deskt
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+---
+
+## @babel/core
```
hmhealey, line 46: Are we sure we actually need to include all of these things that are used to build the web app but aren't actually included in it? Things like babel are only used to compile code for the app and ESLint is only used to check the styling. I don't think we've included those in the NOTICE.txt elsewhere since they aren't part of the finished product.

amyblais: Makes sense, I'm seeing now that some of those are dev dependencies which don't need to be included. I'll work on removing them. https://github.com/mattermost/desktop/blob/master/package.json#L41

hmhealey (approved):
Thanks. That makes it a lot smaller too, so that's nice

amyblais:
@Willyfrog I don't think we need to cut RC-2 just for this since it doesn't impact the product, but let me know if I'm wrong.

Willyfrog:
I agree taht there is no need for an extra RC

PR #1400, opened 2020-10-28
[MM-31221][BrowserView] first modal: adding a server while in a server view

Willyfrog:

**Summary**
first working modal with browserviews, architecture might need to change, but right now it works, prevents clicking through and stays on top if you change tab. Also prevents adding additional modals if the key matches.

it also adds the server ;)

requires the urlview one to be merged before this one


src/main/modalView.js
```diff
@@ -0,0 +1,96 @@
...
+    this.status = SHOWING;
+    if (this.view.webContents.isLoading()) {
+      this.view.webContents.once('did-finish-load', () => {
+        this.view.webContents.focus();
+      });
+    } else {
+      this.view.webContents.focus();
+    }
+
+    // uncomment if something goes wrong with modals
```
nevyangelova, line 59: can we have an environment variable for this?

src/renderer/css/components/AddServerModal.css
```diff
@@ -0,0 +1,181 @@
...
+  line-height: 1.4;
+  width: 1000px;
+  position: absolute;
+  top: 25%;
+  left: 50%;
+  transform: translate(-50%, -25%);
+  font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
+}
+
+form {
```
nevyangelova, line 31: can we avoid non-class styling? This gives us a lot of trouble especially in webapp when adding new features and the styles overtake, since they are element specific?

src/renderer/css/components/AddServerModal.css
```diff
@@ -0,0 +1,181 @@
...
+}
+
+.help-block {
+  display: block;
+  margin-top: 5px;
+  margin-bottom: 10px;
+  color: #737373;
+}
+
+button {
```
nevyangelova, line 82: same here

nevyangelova (approved):
Thanks! Added a few non-blocking comments :)

devinbinnie (approved):

Willyfrog:
added a couple of env vars to enable modal and settings dev tools. since this should only be used by devs, I don't think we need anything more specific, but let me know if you feel otherwise.

src/main/modalManager.js
```diff
@@ -0,0 +1,91 @@
...
+    requestModal.reject(data);
+    filterActive();
+  }
+}
+
+function filterActive() {
+  modalQueue = modalQueue.filter((modal) => modal.isActive());
+}
+
+export default {
```
devinbinnie, line 88: Do you need this if you're already exporting the functions above?

src/main/modalManager.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  });
+}
+
+function handleModalResult(event, data) {
+  const requestModal = findModalByCaller(event);
+  if (requestModal) {
+    requestModal.resolve(data);
+    filterActive();
+  }
```
devinbinnie, line 73: If there are multiple modals in the queue, would we not want to show the next one in the queue after the last one was dismissed?

nevyangelova (approved):
LGTM :)

devinbinnie (approved):

Willyfrog:
@devinbinnie added the fix for production, can you re-review?

devinbinnie (approved):

PR #1387, opened 2020-10-14
Bv menus

Willyfrog:
**Summary**

Most menus working, there are a few that still need some extra work

PR #1403, opened 2020-10-30
[MM-22691][Browserview] fix tray icon

Willyfrog:
**Summary**
fix path for images so tray can load, also fix that on osx theme change it changes as well.

**Issue link**
[MM-22691](https://mattermost.atlassian.net/browse/MM-22691)

**Test Cases**
change settings to display tray menu and restart the app
open osx preferences and switch from dark to light and back

Expected: icon changes to be displayed correctly

**Additional Notes**
requires https://github.com/mattermost/desktop/pull/1387


src/main/main.js
```diff
@@ -163,13 +162,19 @@ function initializeArgs() {
...
-  WindowManager.setConfig(config.data, config.showTrayIcon, deeplinkingUrl);
+async function initializeConfig() {
+  const loadConfig = new Promise((resolve, reject) => {
+    config = new Config(app.getPath('userData') + '/config.json');
+    config.once('update', (configData) => {
+      config.on('update', handleConfigUpdate);
+      config.on('synchronize', handleConfigSynchronize);
+      config.on('darkModeChange', handleDarkModeChange);
+      handleConfigUpdate(configData);
+      resolve();
```
Willyfrog, line 173: in windows, there is a bigger delay until the registry is read, and this causes the config to not be loaded in time, making other parts fail

devinbinnie: Do we need the second `await` in this function? Couldn't we just return the promise and the `await` in `initialize()` will take care of it?

webpack.config.renderer.js
```diff
@@ -42,6 +43,19 @@ module.exports = merge(base, {
       ignoreOrder: true,
       chunkFilename: '[id].[contenthash].css',
     }),
+    new CopyPlugin({
+      patterns: [{
+        from: 'assets/windows/*.ico',
+        context: 'src',
+      }, {
+        from: 'assets/linux/*/*.png',
```
devinbinnie, line 51: I think you can reduce this by doing:
```suggestion
        from: 'assets/**/*.png',
```

devinbinnie (approved):

nevyangelova (approved):

src/main/main.js
Willyfrog: config is being used by `initializeBeforeAppReady` so we have to wait for it to finish. In windows it needs to go to the registry and get the info from there and it ends up being a race condition, so it feel safer this way.

PR #1409, opened 2020-11-05
[MM-22810] Update loading screen with new design & animation

deanwhillier:
**Summary**
This PR introduces a redesigned loading screen with background visuals and animated loading icon. The loading screen supports dark/light mode to automatically match the host OS. Animation uses a combination of SVG for visuals and CSS animations for ... animations 🤷‍♂️. New components are function components instead of class components as the direction we plan on taking moving forward.

Two custom hooks have been added for working with CSS `transitionend` and `animationend` events.

The npm `classnames` package has been added to make working with classes in jsx a bit easier.

There are also a few tweaks based on eslint errors in some of the existing files I worked in.

**Issue link**
[MM-22810](https://mattermost.atlassian.net/browse/MM-22810)


deanwhillier:
Dark Mode:
ezgif-2-d4126d0877a0

deanwhillier:
Multi-tab layering is messed at the moment, don't review just yet. I'll ping when I have it resolved.

deanwhillier:
@Willyfrog, @devinbinnie, @matthewbirtch, should be good to go now.

matthewbirtch (approved):
This is awesome, @deanwhillier. Thanks for tackling this!

src/browser/hooks/useAnimationEnd.js
```diff
@@ -0,0 +1,38 @@
...
+
+function useAnimationEnd(
+  ref,
+  callback,
+  animationName,
+  allowEventBubbling = true,
+) {
+  React.useEffect(() => {
+    if (!ref.current) {
+      return;
```
Willyfrog, line 14: if ref changed and current is no longer available, should we remove the listener? or can it be reasigned?

src/package.json
```diff
@@ -12,6 +12,7 @@
     "@hapi/joi": "^16.1.8",
     "auto-launch": "^5.0.5",
     "bootstrap": "^3.3.7",
+    "classnames": "^2.2.6",
```
Willyfrog, line 15: we should have added this long ago! 👍

Willyfrog (changes requested):
I really like the effect, thanks to both Matt and you for this :D

I added a few notes and changes, but overall looking good

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
```diff
@@ -0,0 +1,85 @@
...
+  LOADED: 'loaded', // animation graphics fade out
+  COMPLETE: 'complete', // animation graphics are removed from the DOM
+};
+
+const ANIMATION_COMPLETION_DELAY = 500;
+
+function LoadingAnimation({
+  loading = false,
+  darkMode = false,
+  onLoadAnimationComplete = () => {}} // eslint-disable-line no-empty-function
```
devinbinnie, line 24: Can we just make this optional or `null` or `undefined` and just do a null check instead of executing an empty function?

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
```diff
@@ -0,0 +1,85 @@
...
+) {
+  const loadingIconContainerRef = React.useRef(null);
```
devinbinnie, line 26: I think this should be:
```suggestion
  const loadingIconContainerRef = React.createRef();
```

src/browser/components/LoadingScreen/LoadingScreen.jsx
```diff
@@ -0,0 +1,71 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import useTransitionEnd from '../../hooks/useTransitionEnd.js';
+
+import LoadingAnimation from '../LoadingAnimation';
+
+function LoadingScreen({loading = false, darkMode = false}) {
+  const loadingScreenRef = React.useRef(null);
```
devinbinnie, line 13:
```suggestion
  const loadingScreenRef = React.createRef();
```

devinbinnie (changes requested):

src/browser/components/LoadingAnimation/LoadingIcon.jsx
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+function LoadingAnimation() {
+  return (
```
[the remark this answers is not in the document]
deanwhillier: I'd actually rather have the SVG be in a `.svg` file and inlined during build but didn't have time to implement that given the infrastructural changes needed. Instead I opted for a separate private component because the SVG is large and cumbersome and for me makes this component more difficult to understand when combined. Since the control of the SVG is handled almost exclusively by CSS with some basic JS state changes (via class names), I opted for keeping the SVG entirely separate. The index.js file exposes this file as the public representation of the component allowing the SVG to remain private.

src/browser/components/LoadingAnimation/index.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export {default} from './LoadingAnimation.jsx';
```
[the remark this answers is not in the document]
deanwhillier: TL'DR, I used this ticket to initially test newer conventions including:
* using function component approach with hooks vs. class components
* components contained in like-name folders (incl. supporting assets) ... I did initially have the CSS in these folders as well, but the current infrastructure is not entirely conducive to this approach
* using an `index.js` file to define the public API, though admittedly that is more for react module style of code organization where there are several public exports

I know this is not the convention in the Desktop App at this point and we certainly aren't using a module based approach to organizing code right now, so happy to revert to the current approach without an index.js file to define a public API for a component/module.

Thoughts?

src/browser/components/LoadingScreen/LoadingScreen.jsx
```diff
@@ -0,0 +1,71 @@
...
+    >
+      <LoadingAnimation
+        loading={loading}
+        darkMode={darkMode}
+        onLoadAnimationComplete={handleLoadAnimationComplete}
+      />
+    </div>
+  ) : null;
+
+  return loadingScreen;
```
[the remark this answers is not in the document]
deanwhillier: It certainly would, thanks!

src/browser/components/LoadingScreen/LoadingScreen.jsx
```diff
@@ -0,0 +1,71 @@
...
+  const [fadeOutIsComplete, setFadeOutIsComplete] = React.useState(true);
+
+  React.useEffect(() => {
+    // reset internal state if loading restarts
+    if (loading) {
+      resetState();
+    }
+    if (!loading) {
+      setLoadingIsComplete(true);
+    }
```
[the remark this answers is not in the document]
deanwhillier: This one I'm never sure about. I've seen so many instances of people decrying the use of `else` and related thoughts that I'm never sure which way to go anymore 📦 

I personally prefer using the else is this case, so will switch back. Thanks!

src/browser/components/LoadingScreen/index.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export {default} from './LoadingScreen.jsx';
```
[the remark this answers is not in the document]
deanwhillier: See reply above.

src/browser/hooks/useAnimationEnd.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    function handleAnimationend(event) {
+      if (!allowEventBubbling && event.target !== ref.current) {
```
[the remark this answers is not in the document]
deanwhillier: TL;DR, I think the name would probably be better as `listenToEventBubbling` as it's not about allowing/preventing, but listening to events for all descendent elements vs just the targeted element and ignoring events from descendent elements. I will update the name accordingly. Thx.

More specifically:

This prop is not about how bubbling works or about preventing it but rather listening to it or not. I added this option thinking about broader use cases than this PR as hooks are a functional mechanism to be able to reuse code in multiple components. Scenarios are as follows:

If there is a CSS animation reused on multiple elements that are part of a shared ancestry (i.e. one element is the descendent of another) ...
   1. In one instance, I only want to know if the animation has completed on the targeted (ancestor) element and not a descendent element. In that case I can set the hook prop to not listen to event bubbling, thereby ignoring events from descendent elements that might complete first and only responding to events directly from the targeted element.
   2. In another instance, it might be easier to listen for the completion of an animation event on a descendent element from the targeted (ancestor) element because of multiple levels of components or because descendent elements are added/removed continually so registering listeners becomes a challenge to manage or there are numerous elements that all use the same animation where registering a single listener on the ancestor element is more efficient.

Does that make sense? Regardless, I'll change the name of the prop.

src/browser/hooks/useAnimationEnd.js
deanwhillier: The final returned arrow function in the `useEffect` closure function is used to unregister the event listener appropriately. Hooks are very interesting in how they work! Because of the array of props provided to the `useEffect` call, if the ref changes (or any of the other provided props), the returned arrow function will automatically be run (thereby unregistering the event handler) before the `useEffect` handler function is rerun to  again register the event listener on the changed ref. If `ref.current` no longer exists, nothing will be registered because of the initial conditionals in the `useEffect` closure function. If none of the provide props change, the `useEffect` closure function does not get re-run.

src/browser/components/LoadingAnimation/index.js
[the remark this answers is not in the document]
deanwhillier: See response to same comment from Guillermo.

src/browser/components/LoadingScreen/index.js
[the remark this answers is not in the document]
deanwhillier: See response to same comment from Guillermo.

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
deanwhillier: Yup. I've seen this done all kinds of ways, but not set on one in particular. :)

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
deanwhillier: `React.createRef()` is primarily for class-based components where you store the reference on the class instance. `React.useRef(...)` is the new hook used with function components. The former recreates the ref each time it is called, the latter returns the previously created ref and so is best for function components.

At least that is my understanding and why I used here as I purposely created these as function components to test the use of hooks. Chatted with Guillermo about doing so before I submitted the PR.

src/browser/components/LoadingScreen/LoadingScreen.jsx
deanwhillier: See response above.

src/browser/components/LoadingAnimation/index.js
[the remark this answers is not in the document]
Willyfrog: >     * using function component approach with hooks vs. class components

totally onboard

>     * components contained in like-name folders (incl. supporting assets) ... I did initially have the CSS in these folders as well, but the current infrastructure is not entirely conducive to this approach

for supporting this, we would have to change more things in webpack, not sure how it'll end :/

>     * using an `index.js` file to define the public API, though admittedly that is more for react module style of code organization where there are several public exports

yeah, I totally see it if there are multiple options to import (see my Notifications PR which has several modules and some functions tying all in the index which is the one handling the exports). I like the idea of having folders to group things together (and yes that's probably a react convention). My concern was having a file with 3 lines of code of which 2 are the header. 

But I guess in the end it doesn't matter as webpack will handle that and won't even affect the resulting build

> I know this is not the convention in the Desktop App at this point and we certainly aren't using a module based approach to organizing code right now, so happy to revert to the current approach without an index.js file to define a public API for a component/module.

I'm fine with changing conventions if we go for a better one. In fact right now there is no convention and coders have to figure out where the file might be or rely on some external search, so having one is for the best.

src/browser/components/LoadingScreen/LoadingScreen.jsx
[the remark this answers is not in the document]
Willyfrog: if there is a return in the if branch, there is no need for the else, but in this case is making the person think less.
Anyways, I never liked removing the else branch, I just do it as it pleases the linter rules (which are a group decision)

src/browser/hooks/useAnimationEnd.js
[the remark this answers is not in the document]
Willyfrog: ah! then it makes sense, but I think we should add it to the effect's documentation as I'm not sure people will get it right away. 

This makes me think it might make sense to have a common effects library like the common components? If so, this might be far more useful in the webapp

src/browser/hooks/useAnimationEnd.js
Willyfrog: Yes, your PR made me refresh what I read about hooks, and you are correct that it will run the functions on `ref` changing, but the function returned first checks if it's null and then returns (I should probably have added the comment to that second check instead of this one 😓 see line 30, sorry for the confusion)

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
Willyfrog: I'm 0/5 on this (either way is fine to me), but if we decide it is ok to have an empty-function we should remove the linting rule.

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
Willyfrog: yes, I think `createRef` was not supposed to work with hooks, I think I read that somewhere, but I can't seem to find it.

src/browser/hooks/useAnimationEnd.js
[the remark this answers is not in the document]
deanwhillier: True and agreed. I usually like to add JSDoc's but completely forgot for this PR 😳

RE: the library, yes we absolutely could consider that.

src/browser/hooks/useAnimationEnd.js
deanwhillier: Right. Honestly I thought the same at first and didn't have the conditional, but ran into an error somewhere along the line where ref.current was null at that point so there was nothing to remove. That said, it must have been something else, because an arrow function will never be returned if ref.current is null to begin with so you are of course correct – good catch! :)

src/browser/components/LoadingAnimation/index.js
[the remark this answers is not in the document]
deanwhillier: I moved the `LoadingScreen.jsx` file out of the folder and deleted the folder & index file but left the other one given the added private icon.

src/browser/hooks/useAnimationEnd.js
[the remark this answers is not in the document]
deanwhillier: I added some ESDocs

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
devinbinnie: Strange that they would duplicate that functionality.

src/browser/components/LoadingAnimation/LoadingAnimation.jsx
devinbinnie: Still looking to have this changed. 3/5, since we're violating the linting rule, and I think a null check is a more common practice.

devinbinnie (changes requested):

Willyfrog (approved):
Other than deciding on the empty function, it looks good to me

devinbinnie (approved):

deanwhillier:
@esethna, @jgilliam17, I can provide a build of this PR to test if you need. Let me know!

esethna:
Duuuude, this is awesome! Thanks @matthewbirtch @deanwhillier!

esethna (approved):

esethna:
@deanwhillier does this replace our current MM spinner on the desktop app?

deanwhillier:
@esethna, just the desktop loading spinner for this round. Webapp is a different story. Some of the assets can be reused though.

jgilliam17:
Thank you @deanwhillier 
This looks great! Both light and dark loading screens look good.

The loading screen supports dark/light mode to automatically match the host OS - I am finding this is only true on macOS.
On Windows and Ubuntu the user set mode `Toggle Dark Mode` overrides the host OS setting.
See attached image for the systemwide setting being dark, while app setting is light - the app respects the desktop app setting, not OS.
If this is the intended implementation, we may need to document and set correct user expectations per host OS.

Screenshot 2020-11-16 171943

Willyfrog:
in 4.6 it is intended that way, once we get browserview, the app will also try to match windows settings

jgilliam17:
Thanks for confirming @Willyfrog 
Looks good to merge.

jgilliam17 (approved):

PR #1411, opened 2020-11-06
[browserview] Electron notifications

Willyfrog:
**Summary**

working electron's notification. This does not contain the native modules for windows.

**Additional Notes**
There is a known bug: if a mention is received in a background tab that was never shown and that tab is in the channel where the mention happens, the notification is not sent. [MM-30339](https://mattermost.atlassian.net/browse/MM-30339)


devinbinnie (approved):

nevyangelova (approved):
Thanks, LGTM

PR #1417, opened 2020-11-18
[BrowserView] Native modules & registry access

Willyfrog:
**Summary**

initial work to get native modules working.
this also creates environment variables for installing and building the desktop app, this will need to be updated whenever we upgrade the desktop version

Also fixes reading foreign characters from the registry (like spanish `ñ` or german `ß`)

**Additional Notes**
requires https://github.com/mattermost/desktop/pull/1403


nevyangelova (approved):
LGTM! :)

src/common/config/RegistryConfig.js
```diff
@@ -3,11 +3,20 @@
...
 import {EventEmitter} from 'events';
+import log from 'electron-log';
 
-import WindowsRegistry from 'winreg';
+let registry;
+try {
+  // eslint-disable-next-line global-require
+  registry = require('registry-js');
+} catch (e) {
+  if (process.platform === 'win32') {
```
devinbinnie, line 13: How come the check for `win32` is only performed here?

devinbinnie (approved):

src/common/config/RegistryConfig.js
Willyfrog: there is no registry for win32, so only if we are running win32 will show the error. I don't expect it to work on mac or linux.

PR #1428, opened 2020-12-14
Desktop Theme Integration with Webapp

devinbinnie:
**Summary**
This PR adds integration of webapp themes into the desktop. It will replace the existing light/dark theme with whatever theme the webapp is using.

Based on the existing BrowserView code since we're moving to that.
Done for the 2020 MM Hackathon.

What's done:
- Desktop app changes theme when the webapp theme changes via Account Settings or between teams
- Desktop app changes theme when different servers are selected
- Theme is pulled down directly from the server on initial load

What's not done:
- Websocket stuff to update the theme on prefs change (I really wanted to do this :()
- A bug that pops up parsing the JSON sometimes
- Exact choice of colour mapping from webapp theme > desktop theme (just took a general stab at it)
- Various other edge cases


esethna:
This is cool! @devinbinnie which theme color variable from webapp is used for the server tab bar?

devinbinnie:
> This is cool! @devinbinnie which theme color variable from webapp is used for the server tab bar?

It's using stuff from the sidebar, so I think that one is the sidebar header background.

src/main/MattermostView.js
```diff
@@ -54,6 +54,73 @@ export class MattermostView extends EventEmitter {
...
+      log.error(error);
+      WindowManager.sendToRenderer('use-default-theme');
+    });
+    req.end();
+  };
+
+  handleUserTheme = (resp) => {
+    resp.on('data', (data) => {
+      const prefsRaw = `${data}`;
+      const prefs = JSON.parse(prefsRaw);
```
Willyfrog, line 97:
```suggestion
let prefs;
try {
      prefs = JSON.parse(prefsRaw);
} catch e {
      log.error(`Error loading user theme for ${this.server.name}: ${e}`);
      WindowManager.sendToRenderer('use-default-theme');
      return;
}
```

src/main/MattermostView.js
```diff
@@ -54,6 +54,73 @@ export class MattermostView extends EventEmitter {
...
+      if (!Array.isArray(prefs)) {
+        log.error(prefs);
+        WindowManager.sendToRenderer('use-default-theme');
+        return;
+      }
+
+      const themePrefs = prefs.filter((pref) => pref.category === 'theme');
+      if (themePrefs && themePrefs.length) {
+        // use custom theme
+        const themePref = themePrefs.find((pref) => pref.name === '');
```
Willyfrog, line 107:
```suggestion
      const themePref = themePrefs.find((pref) => pref.category === 'theme' && pref.name === '');
      if (themePref) {
```

src/main/MattermostView.js
```diff
@@ -54,6 +54,73 @@ export class MattermostView extends EventEmitter {
...
+        const themeData = JSON.parse(themePref.value);
+        WindowManager.sendToRenderer('use-custom-theme', themeData);
```
Willyfrog, line 109:
```suggestion
        try {
            const themeData = JSON.parse(themePref.value);
            WindowManager.sendToRenderer('use-custom-theme', themeData);
        } catch e {
            console.log(`error parsing theme data for ${this.server.name}: ${e}`);
            WindowManager.sendToRenderer('use-default-theme');
        }
```

Willyfrog (changes requested):
love this change, the first one to talk with the server 🙌 

added a few comments for handling errors, messages and code structure, but looking mostly good.

One question: if a user is connected to multiple tabs, what would be the theme applied? does it change when the user switches tabs?

esethna:
@matthewbirtch thoughts on this from a UX perspective?

matthewbirtch:
@devinbinnie how can I test this on my local? I'm having a hard time seeing the sidebar theme colors working for the server tab bar, but need to see it for myself.

Also, the server tab bar is more of an OS window bar, so shouldn't it match the OS dark/light settings?

devinbinnie:
Closing this until we decide whether we want to do it or not.

PR #1431, opened 2020-12-17
[MM-31266] fix access url when it's not a mm server

Willyfrog:
**Summary**

fix checking server's url param when it is not a team url during attempt to navigate in the main window.

**Issue link**
[MM-31266](https://mattermost.atlassian.net/browse/MM-31266)

**Test Cases**

login via OneLogin or gitlab shouldn't crash the app


devinbinnie (approved):

nevyangelova (approved):
LGTM!

PR #1433, opened 2020-12-22
[MM-31225][MM-31217][MM-31219][Browserview] fix linux compilation + other fixes

Willyfrog:
**Summary**
- removed native module as it was not compiling for linux 32bit, this means that we are back to using winreg
- change how webpack copies files and resources. Same for electron-build this results in a build without `/dist` within the `app.asar` file. This fixes production builds not finding asset files like the tray icon or the preload script.
- moved dev tools into detached windows so it doesn't get hidden by browser-view
- added an error screen if no server is configured. this shouldn't be final, but as a quick workaround so it doesn't look like a crash
- fixed a crash when there were no servers configured
**Issue link**

[MM-31225](https://mattermost.atlassian.net/browse/MM-31225)
[MM-31217](https://mattermost.atlassian.net/browse/MM-31217)
[MM-31219](https://mattermost.atlassian.net/browse/MM-31219)

**Additional Notes**
current QA build is from this branch, without it compiling in linux/win might fail


src/common/config/RegistryConfig.js
```diff
@@ -131,22 +122,24 @@ export default class RegistryConfig extends EventEmitter {
...
+          if (name) { // looking for a single entry value
+            const registryItem = results.find((item) => item.name === name);
+            resolve(registryItem && registryItem.value ? registryItem.value : null);
+          } else { // looking for an entry list
+            resolve(results);
+          }
+        });
       } catch (e) {
         log.error(`There was an error accesing the registry for ${name}: ${e}`);
-        reject(e);
```
devinbinnie, line 149: How come we aren't using the usual failure mode of `Promise` here anymore?

src/renderer/components/MainPage.jsx
```diff
@@ -114,16 +115,19 @@ export default class MainPage extends React.Component {
 
   getTabStatus() {
     // TODO: should try to make this a bit safer in case we get into a weird situation
```
devinbinnie, line 117: Still need this TODO?

devinbinnie (approved):

nevyangelova (approved):

src/common/config/RegistryConfig.js
Willyfrog: originally it wasn't using the resolve/reject, only resolve. Since I was going back, i also went back to that. I think I prefer to use the resolve/reject mechanism, but since it was kind of manual revert I went back to the original.

if you prefer the resolve/reject I'll add it back

src/renderer/components/MainPage.jsx
Willyfrog: it could always be safer.... but no, no need for it anymore :) thanks for catching that

devinbinnie (approved):

nevyangelova (approved):

PR #1434, opened 2020-12-23
[MM-31224] fix reloading servers and other tab issues

Willyfrog:
**Summary**

Change server load, so it reuses browserviews, preventing white pages flashes 
Fixed some errors from the renderer that were turning the tabs white

what's not fixed:
- remove `saving...` label, when changing any settings the label remains. Created ticket [MM-31554](https://mattermost.atlassian.net/browse/MM-31554)
- Sometimes it doesn't update which tab is in focus.

**Issue link**

[MM-31224](https://mattermost.atlassian.net/browse/MM-31224)

### DIFF
--- a/src/main/viewManager.js
+++ b/src/main/viewManager.js
@@ -23,41 +23,45 @@ export class ViewManager {
     this.urlView = null;
   }
 
+  loadServer = (server, mainWindow) => {
+    const srv = new MattermostServer(server.name, server.url);
+    const view = new MattermostView(srv, mainWindow, this.viewOptions);
+    this.views.set(server.name, view);
+    view.setReadyCallback(this.activateView);
+    view.load();
+    view.on(UPDATE_TARGET_URL, this.showURLView);
+  }
+
   // TODO: we shouldn't pass the main window, but get it from windowmanager
   // TODO: we'll need an event in case the main window changes so this updates accordingly
   load = (mainWindow) => {
-    this.configServers.forEach((server) => {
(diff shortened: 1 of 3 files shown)


src/main/viewManager.js
```diff
@@ -23,41 +23,45 @@ export class ViewManager {
...
-    for (const view of oldviews.values()) {
-      // try to restore view to the same tab if possible, but if not use initial one
-      if (view.isVisible) {
-        log.info(`will try to restore ${view.name}`);
-        newView = this.views.get(view.name);
+    const sorted = this.configServers.sort((a, b) => a.order - b.order);
+    let setFocus;
+    sorted.forEach((server) => {
+      const recycle = oldviews.get(server.name);
+      if (recycle) {
```
devinbinnie, line 49: Would it be worth checking parameters here like the URL? I'm not sure if the tab would load a new URL here if it was changed in the settings. Unless we're manually kicking it on the save.

devinbinnie (approved):

nevyangelova (approved):
LGTM but I couldn't figure out the error in CI

src/main/viewManager.js
Willyfrog: fixed

Willyfrog:
> LGTM but I couldn't figure out the error in CI

seems like the deps are failing, i'll have to take a look into that, not sure if it is temporary or permanent

nevyangelova (approved):

devinbinnie (approved):

PR #1453, opened 2021-01-25
[MM-31467] Move protocol handling over from original MattermostView into web contents handler

devinbinnie:
**Summary**
Protocol handling wasn't migrated over from the original `MattermostView`, so I've done that and made it work with the `webContents` object rather than `webview`.

### DIFF
--- a/src/main/allowProtocolDialog.js
+++ b/src/main/allowProtocolDialog.js
@@ -7,7 +7,7 @@ import fs from 'fs';
 
 import path from 'path';
 
-import {app, dialog, ipcMain, shell} from 'electron';
+import {app, dialog, shell} from 'electron';
 
 import {protocols} from '../../electron-builder.json';
 
@@ -24,7 +24,7 @@ function addScheme(scheme) {
   }
 }
 
-function init(mainWindow) {
+function init() {
   fs.readFile(allowedProtocolFile, 'utf-8', (err, data) => {
     if (!err) {
       allowedProtocols = JSON.parse(data);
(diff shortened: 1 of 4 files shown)


src/main/main.js
```diff
@@ -593,17 +592,52 @@ function handleAppWebContentsCreated(dc, contents) {
...
+
     const server = urlUtils.getServer(parsedURL, config.teams);
 
     if (!server) {
       shell.openExternal(url);
       return;
     }
+
+    // Public download links case
+    // TODO: This doesn't work correctly right now, needs to be refactored
```
devinbinnie, line 620: @Willyfrog Realized this wasn't migrated over either, it's meant to handle public links to files hosted on your MM server. But it doesn't seem to be working correctly, complaining about the URL type we're using. Something we should verify how we want to handle and hopefully fix the issue.

Funny enough it doesn't really work in the current desktop app because we don't use the same path that matched the original regex anymore, so we currently just show the image in the same `webview` and you have to press Back to get out of it.

src/main/allowProtocolDialog.js
```diff
@@ -37,51 +37,49 @@ function init(mainWindow) {
...
-    }).then(({response}) => {
-      switch (response) {
-      case 1: {
-        allowedProtocols.push(protocol);
-        function handleError(err) {
-          if (err) {
-            console.error(err);
-          }
+function handleDialogEvent(protocol, URL) {
+  if (allowedProtocols.indexOf(protocol) !== -1) {
```
nevyangelova, line 44: is this a save enough check? What if another protocol is removed or this ends up being at this position?

devinbinnie: `indexOf` only returns `-1` if the requested item isn't in the array, so it will only pass this check if the protocol is in `allowedProtocols`.

`allowedProtocols` is a singleton in this context, meaning that there will only ever be one instance of it in the app, and it's only modified by this function or upon loading the initial list of allowed protocols.

nevyangelova (approved):

src/main/main.js
Willyfrog: is this related to https://mattermost.atlassian.net/browse/MM-32333 ?

Willyfrog (approved):
it might need https://github.com/mattermost/desktop/pull/1457 as I run into a couple of situations where it crashed, but not your fault, and there is already a PR for it.

PR #1455, opened 2021-01-26
[MM-31650] Restore focus to active server on modal and settings window closure + other fixes

devinbinnie:
**Summary**
When closing the settings window or a modal, we weren't giving focus back to the active `BrowserView`. This PR adds that.

Other fixes:
- When a modal was open, you can still interact with the app in the background. This PR disabled the tab bar so that focus is always with the modal until it is closed.
- Restored the original `NewTeamModal` component so that clicking the `+` button shows a modal that looks like it used to.

**Ticket Link**
https://mattermost.atlassian.net/browse/MM-31650


Willyfrog (approved):
Looking good,

regarding newServer.jsx, I'm not really sure how to handle that, if we should send the modal request back to the main window (and maybe close the settings page?) or use the same browserview modals in the settings page. Not that it needs to change right now, anyway

nevyangelova (approved):
LGTM

PR #1456, opened 2021-01-27
[browser view] MM-32277 bump version, exe, cache errrors

Willyfrog:

**Summary**
- update version to 5.0.0-development
- remove cache from src/package-lock.json, also other references to that file
- enable copying exe files to artifacts

**Issue link**

[MM-32277](https://mattermost.atlassian.net/browse/MM-32277)

### DIFF
--- a/scripts/cleanCache.command
+++ b/scripts/cleanCache.command
@@ -1,0 +1,3 @@
+#!/bin/bash
+BASE_DIR=~/Library/ApplicationSupport/Mattermost
+realpath $BASE_DIR/* | grep -v .json | xargs -I{} rm -rf "{}"
--- a/scripts/cleanCache.command
+++ b/scripts/cleanCache.command
@@ -1,3 +1,0 @@
-#!/bin/bash
-BASE_DIR=~/Library/ApplicationSupport/Mattermost
-realpath $BASE_DIR/* | grep -v .json | xargs -I{} rm -rf "{}"
--- a/scripts/cp_artifacts.sh
+++ b/scripts/cp_artifacts.sh
@@ -21,12 +21,9 @@ if [[ -f "${SRC}/mattermost-desktop-${VERSION}-win-x64.zip" ]]; then
     SOMETHING_COPIED=$((SOMETHING_COPIED + 2))
 fi
 
-# make sure the var exists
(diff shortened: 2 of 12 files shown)


deanwhillier (approved):

scripts/cleanCache.command
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+BASE_DIR=~/Library/ApplicationSupport/Mattermost
```
devinbinnie, line 2: Should this be a variable to support Windows/Linux? Or do we not need this script to run automatically

devinbinnie (approved):

scripts/cleanCache.command
Willyfrog: I added this one by mistake, I want to create this as a tool for people in case they run into problems (mostly devs) while doing upgrades/downgrades. 
and yes, it would need to be changed depending on OS and available tools.

Also, I want it to be kind of 'click and forget' and that's basically the reason why I didn't add it before. I'll remove it

scripts/cp_artifacts.sh
```diff
@@ -20,12 +20,13 @@ if [[ -f "${SRC}/mattermost-desktop-${VERSION}-win-x64.zip" ]]; then
...
-# if [[ -f "${SRC}/mattermost-desktop-setup-${VERSION}-win.exe" ]]; then
-#     echo -e "Copying win-no-arch\n"
-#     cp "${SRC}/mattermost-desktop-setup-${VERSION}-win.exe" "${DEST}/"
-#     SOMETHING_COPIED=$((SOMETHING_COPIED + 4))
-# fi
+
+if [[ ${MM_WIN_INSTALLERS-0} -eq 1 && -f "${SRC}/mattermost-desktop-setup-${VERSION}-win.exe" ]]; then
+    echo -e "Copying win-no-arch\n"
+    cp "${SRC}/mattermost-desktop-setup-${VERSION}-win.exe" "${DEST}/"
+    cp "${SRC}/mattermost-desktop-setup-${VERSION}-win.exe" "${DEST}/"
```
deanwhillier, line 27: Should this be here twice?

Willyfrog: definitely no, good catch!

devinbinnie (approved):

PR #1458, opened 2021-01-29
[MM-32424] fix server devtools

Willyfrog:
**Summary**

toggle devtools tries to restore to the last known position of the devtools, so it might end below the browserview, checking and using open with detach mode prevents this from happening. Also removes the option to reattaching the window.

**Issue link**

[MM-32424](https://mattermost.atlassian.net/browse/MM-32424)


Willyfrog:
wrong base branch, left a ton of conflicts for no reason, I'll reopen

PR #1459, opened 2021-01-29
[MM-32424] fix server devtools being hidden by browserview

Willyfrog:
**Summary**

toggle devtools tries to restore to the last known position of the devtools, so it might end below the browserview, checking and using open with detach mode prevents this from happening. Also removes the option to reattaching the window.

**Issue link**

[MM-32424](https://mattermost.atlassian.net/browse/MM-32424)

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -195,9 +195,9 @@ function createTemplate(config) {
       if (focusedWindow) {
         // toggledevtools opens it in the last known position, so sometimes it goes below the browserview
         if (focusedWindow.isDevToolsOpened()) {
-          focusedWindow.openDevTools({mode: 'detach'});
-        } else {
           focusedWindow.closeDevTools();
+        } else {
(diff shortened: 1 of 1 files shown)


src/main/menus/app.js
```diff
@@ -193,7 +193,12 @@ function createTemplate(config) {
     })(),
     click(item, focusedWindow) {
       if (focusedWindow) {
-        focusedWindow.toggleDevTools({mode: 'detach'});
+        // toggledevtools opens it in the last known position, so sometimes it goes below the browserview
+        if (focusedWindow.isDevToolsOpened()) {
```
devinbinnie, line 197: Might be misunderstanding, but are we opening the dev tools only if they're already opened?

Willyfrog: no, you are right, tested it, but linter was complaining about negative conditions and I failed to copy properly 🤦

devinbinnie (approved):

hmhealey (approved):

PR #1460, opened 2021-01-29
[MM-31570] update mentions/unreads/session on jewel, tray and dock

Willyfrog:

**Summary**

update's app with number of mentions and if there is any unread or expired session.

**Issue link**
[MM-31570](https://mattermost.atlassian.net/browse/MM-31570)
**Test Cases**
using different servers (versions below v5.28 and above) and new sidebar configurations
- receive and read mentions
- mark channel as unread (either by messaging or using mark as unread)
- disconnect other user

badge, tab and dock should update accordingly

**Additional Notes**
while the initial change to support figure out how to get the mentions was easy, moving this information around wasn't so. Detecting if the server has unread channels varies depending on the server version and sidebar. For version up to v5.28, it will be noted by using an asterisk in the titlebar, after that we'll wait for favicon to change to notice if it was set as unread. Since name is obscured by webpack, we have to figure out what does it mean to have one favicon or the other and memoize it. So the first time a new favicon appears, we'll have to do a quick scan of the page and see if one of the teams is unread and if there is only one team search for an unread channel. 

Added a new appState to handle information and distribute, I was planning to do something similar once we started listening to the WS but seemed the right solution for this case. Any other state we need to have can be tied to this as well in the future

Also moved all the tray related info to a single file, removing it from being in multiple places. this process already started with browserview but now is more consistent. We should do the same with the badge, but there is still a few things in other files, but was totally out of this already-too-big PR.

One thing is left to improve, currently detecting if session is disconnected we are using an interval to figure it out. having some event we could listen would be a much better experience

Also removed the unread count, it was used only in tooltips and to get that number acurately we would have to go back to using a lot of resources to keep it synched. If it is really important we can look into figuring some alternative to it.


Willyfrog:
I still have to test on windows/linux but shouldn't change too significantly

src/main/preload/mattermost.js
```diff
@@ -116,12 +119,17 @@ window.addEventListener('message', ({origin, data: {type, message = {}} = {}} =
     // it will be captured by itself too
     break;
   case 'dispatch-notification': {
-    const {title, body, channel, teamId, silent, data} = message;
-    ipcRenderer.send(NOTIFY_MENTION, title, body, channel, teamId, silent, data);
+    const {title, body, channel, teamId, silent, data: messageData} = message;
+    ipcRenderer.send(NOTIFY_MENTION, title, body, channel, teamId, silent, messageData);
```
devinbinnie, line 123: Shouldn't this be `data`?

devinbinnie (approved):

src/main/MattermostView.js
```diff
@@ -167,4 +187,56 @@ export class MattermostView extends EventEmitter {
...
+    if (this.server && serverName === this.server.name) {
+      if (favicon) {
+        this.faviconMemoize.set(favicon, result);
+      }
+      if (favicon === null || favicon === this.currentFavicon) {
+        appState.updateUnreads(serverName, result);
+      }
+    }
+  }
+}
```
deanwhillier, line 242: Add newline
```suggestion
}

```

src/common/communication.js
```diff
@@ -46,3 +46,11 @@ export const UPDATE_TARGET_URL = 'update_target_url';
...
 
 export const GET_DOWNLOAD_LOCATION = 'get_download_location';
+
+export const UPDATE_MENTIONS = 'update_mentions';
+export const UPDATE_UNREADS = 'update_unreads';
+export const IS_UNREAD = 'is_unread';
+export const UNREAD_RESULT = 'unread_result';
+export const SESSION_EXPIRED = 'session_expired';
+export const UPDATE_TRAY = 'update_tray';
+export const UPDATE_BADGE = 'update_badge';
```
deanwhillier, line 56: Add newline
```suggestion
export const UPDATE_BADGE = 'update_badge';

```

deanwhillier (changes requested):
A few minor things for consideration.

src/main/main.js
```diff
@@ -194,7 +192,7 @@ function initializeBeforeAppReady() {
     app.disableHardwareAcceleration();
   }
 
-  trayImages = getTrayImages(config.trayIconTheme);
+  getTrayImages(config.trayIconTheme);
```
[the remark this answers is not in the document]
Willyfrog: I'll rename to refresh, but the function still returns the tray images in case they are needed

Willyfrog (commented):
addressed comments
I'll be able to push once I get my admin rights back (i did a rebase with `browser-view` branch and now I can't force-push)

deanwhillier (approved):

PR #1461, opened 2021-01-29
[MM-31215][MM-31387] Fixes for bad tab navigation and dragging

devinbinnie:
**Summary**
I've reworked how we notify the front-end wrapper that the tabs have changed externally, and made a few fixes around making sure the right tabs are picked at any given time.

What should be working now:
- Tab navigation via keyboard shortcuts
- Clicking on a server in Settings and loading it correctly
- Seamless tab dragging without reloading any servers

**Issue link**
https://mattermost.atlassian.net/browse/MM-31215
https://mattermost.atlassian.net/browse/MM-31387

### DIFF
--- a/src/main/viewManager.js
+++ b/src/main/viewManager.js
@@ -51,7 +51,7 @@ export class ViewManager {
       if (recycle && recycle.isVisible) {
         setFocus = recycle.name;
       }
-      if (recycle && urlUtils.parseURL(recycle.server.url).toString() === urlUtils.parseURL(server.url).toString()) {
+      if (recycle && recycle.server.url === urlUtils.parseURL(server.url)) {
         oldviews.delete(recycle.name);
         this.views.set(recycle.name, recycle);
       } else {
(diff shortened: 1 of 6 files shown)


src/main/viewManager.js
```diff
@@ -50,7 +51,7 @@ export class ViewManager {
       if (recycle && recycle.isVisible) {
         setFocus = recycle.name;
       }
-      if (recycle && recycle.server.url === server.url) {
+      if (recycle && urlUtils.parseURL(recycle.server.url).toString() === urlUtils.parseURL(server.url).toString()) {
```
devinbinnie, line 54: This was needed because one of these URLs would have a `/` embedded at the end, so in order to avoid semantics I just reparsed them. Definitely open to a better solution here if needed.

Willyfrog: maybe we should try and normalize in its original input (when added to the config and when read from the config) so we don't have to care about that in any other place.

Willyfrog (approved):

PR #1463, opened 2021-02-01
[MM-20227][MM-31388] move to roles and fix focus

Willyfrog:
**Summary**

Copying anything from the about window was crashing and the only way to make it work is to use the role instead of a custom function. I tried to use commands while focusing on the renderer but seems I can't copy anything from the tabs or zoom in there (which is the expected and desirable) so I moved some commands to use roles, and removed any click function that was already with a role property (it gets ignored)

additionally it fixes focus issues when starting or alt-tabbing

**Issue link**

[MM-20227](https://mattermost.atlassian.net/browse/MM-20227)
[MM-31388](https://mattermost.atlassian.net/browse/MM-31388)

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -7,8 +7,6 @@ import {app, Menu, session, shell, webContents} from 'electron';
 
 import * as WindowManager from '../windows/windowManager';
 
-const ZOOM_DIFFERENTIAL = 0.5;
-
 function createTemplate(config) {
   const separatorItem = {
     type: 'separator',
@@ -63,10 +61,6 @@ function createTemplate(config) {
     platformAppMenu = platformAppMenu.concat([
       separatorItem, {
         role: 'quit',
-        accelerator: 'CmdOrCtrl+Q',
-        click() {
-          app.quit();
-        },
       }]);
   }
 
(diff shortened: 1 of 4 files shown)


src/main/menus/app.js
```diff
@@ -158,31 +121,11 @@ function createTemplate(config) {
...
-    label: 'Zoom In',
-    accelerator: 'CmdOrCtrl+SHIFT+=',
-    click() {
-      const focused = webContents.getFocusedWebContents();
-      const level = focused.getZoomLevel();
-      if (level <= 3) {
-        focused.setZoomLevel(level + ZOOM_DIFFERENTIAL);
-      }
-    },
+    role: 'zoomIn',
```
devinbinnie, line 126: Zooming seems to be broken when the focus isn't on the `BrowserView`. I've fixed a few of those cases in my latest PR, but I'm not sure what's going to happen here if you were to click on a tab and then click Zoom, since the in-focus window is the tab bar, not the MattemostView.

nevyangelova (commented):
Does role mean events in this case or? I was confused because I thought roles will be referred to as user roles :D

src/main/menus/app.js
Willyfrog: yeah, that was mi experience the first time I did that, but while trying several things this was working properly. I could switch tabs and zoom, and it was the content that was zoomed not the tab. Did you manage to break it? if so, do you have repro steps? I tried several ways and it was always working for me

Willyfrog:
> Does role mean events in this case or? I was confused because I thought roles will be referred to as user roles :D

role means predefined action/label, so 'copy' will behave as the OS dictates.

src/main/menus/app.js
devinbinnie: on macOS catalina, if you go to another window and then come back and then use the zoom in keyboard shortcut, it will zoom the tabs in rather than the mattermost view. I think it could be fixed by ensuring focus goes to the mattermost view when refocusing the window rather than the main window web contents.

Willyfrog:
@devinbinnie added a fix to the focus issue, should be working now for you

nevyangelova (approved):

devinbinnie (approved):

PR #1468, opened 2021-02-05
[MM-32333] Open public links in the user's default browser

devinbinnie:
**Summary**
We were currently handling public links by treating the `/files` route as another team, which was opening images in the BrowserView. This PR switches over to opening public links in the user's default browser so that all files can be handled correctly and the app will not get into a weird state.

**Issue link**
https://mattermost.atlassian.net/browse/MM-32333

### DIFF
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -550,19 +550,21 @@ function handleAppWebContentsCreated(dc, contents) {
     }
 
     // Public download links case
-    // TODO: This doesn't work correctly right now, needs to be refactored
-    // if (parsedURL.pathname.match(/^(\/api\/v[3-4]\/public)*\/files\//)) {
-    //   downloadURL(url, (err) => {
-    //     if (err) {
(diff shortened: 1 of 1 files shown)


deanwhillier (approved):

src/main/main.js
```diff
@@ -550,19 +550,21 @@ function handleAppWebContentsCreated(dc, contents) {
...
-    //       });
-    //       log.error(err);
-    //     }
-    //   });
-    //   return;
-    // }
+    // TODO: We might be handling different types differently in the future, for now
+    // we are going to mimic the browser and just pop a new browser window for public links
+    if (parsedURL.pathname.match(/^(\/api\/v[3-4]\/public)*\/files\//)) {
+      // downloadURL(url, (err) => {
```
Willyfrog, line 556: We should remove this if we are not going to use it.

Also, if I recall correctly, there is an import downloadURL at the beginning of the file which we should also remove

Willyfrog (approved):
Minor cosmetic nitpick, but good to go

PR #1469, opened 2021-02-08
[MM-27545][browserview] Fix spelling with electron

Willyfrog:
**Summary**
seems electron context menu supports a bit better browserviews, updating it allows us to benefit from electron's spelling system

**Test Cases**
misspell word
right click on it
there should be an option to learn and an option to correct

Screen Shot 2021-02-08 at 12 51 14

**Ticket**
[MM-27545](https://mattermost.atlassian.net/browse/MM-27545)


devinbinnie (approved):
8136 lines deleted in `package-lock` for just a package upgrade? :O

Willyfrog:
> 8136 lines deleted in `package-lock` for just a package upgrade? :O

yes, not sure how that happended but if I run `npm run clean` and then `npm i` and I got the same `package-lock.json`

nevyangelova:
> > 8136 lines deleted in `package-lock` for just a package upgrade? :O
> 
> yes, not sure how that happended but if I run `npm run clean` and then `npm i` and I got the same `package-lock.json`

That doesn't seem right and maybe its related to the failing test, maybe try removing the node_modules and package-lock.json manually as the clean command might not be sufficient 🤷

nevyangelova (approved):

PR #1472, opened 2021-02-10
[MM-31340] Resize browser view and show back button when on non-team URL

devinbinnie:
**Summary**
When we are on a non-team URL (eg. OAuth page, OneLogin), we need to show a back button so that the user can get back into Mattermost if need be. So we needed to set the size of the browser view to accommodate for a bar with a Back button and show the Back button when relevant.

**Issue link**
https://mattermost.atlassian.net/browse/MM-31340


src/main/utils.js
```diff
@@ -22,17 +23,17 @@ export function shouldBeHiddenOnStartup(parsedArgv) {
...
 
-export function getAdjustedWindowBoundaries(width, height) {
+export function getAdjustedWindowBoundaries(width, height, hasBackBar = false) {
   return {
     x: 0,
-    y: TAB_BAR_HEIGHT,
+    y: TAB_BAR_HEIGHT + (hasBackBar ? BACK_BAR_HEIGHT : 0),
     width,
-    height: height - TAB_BAR_HEIGHT,
+    height: height - TAB_BAR_HEIGHT - (hasBackBar ? BACK_BAR_HEIGHT : 0),
```
nevyangelova, line 36: I know this was the case before, but why are we removing the bars height?

nevyangelova (commented):
LGTM 💯

src/main/utils.js
devinbinnie: We want the browser view to be the same size as the window, but without the height of the topbar (or in this case, the back bar as well) included.

eg. Window is 500px tall, tab bar is 40px and the backbar is 30px. So we'd want the browser to be rendered at the bottom of the backbar, at a height of 430px.

nevyangelova (approved):

src/main/MattermostView.js
```diff
@@ -187,6 +188,19 @@ export class MattermostView extends EventEmitter {
     return WindowManager.getMainWindow.webContents;
   }
 
+  handleDidNavigate = (event, url) => {
+    const isUrlTeamUrl = urlUtils.isTeamUrl(this.server.url, url);
+    if (isUrlTeamUrl) {
+      this.setBounds(getWindowBoundaries(this.window));
```
Willyfrog, line 194: seems like we are always calling setbounds with getWindowBoundaries as a parameter, maybe we should have that call into setbounds and pass only the needed parameters?

I'm 0/5 on this one, just noticed that they are always together

src/main/windows/windowManager.js
```diff
@@ -360,3 +362,12 @@ export function restore() {
   const focused = BrowserWindow.getFocusedWindow();
   focused.restore();
 }
+
+export function handleHistory(event, offset) {
+  if (status.viewManager) {
+    const activeView = status.viewManager.getCurrentView();
+    if (activeView && activeView.view.webContents.canGoToOffset(offset)) {
+      activeView.view.webContents.goToOffset(offset);
```
Willyfrog, line 370: I'd add a try/catch around this line, on some nested redirections it might cause trouble.

In case of failure we should log the error, try to go to the server's url and turn off the bar.

Willyfrog (changes requested):
just one small change request and a suggestion. other than that looks good 👍

Willyfrog (approved):
pre-aproving so you can merge once the change is made

src/main/MattermostView.js
devinbinnie: I'm fine with keeping them separate and having `setBounds` be more generic. I don't think there's any case currently where we wouldn't want to size the view to the size of the window, but there's no harm in having that option.

PR #1475, opened 2021-02-15
[MM-28541] restore deeplinking

Willyfrog:
**Summary**

Allow app to handle deeplinking
fixed a couple of errors due to different casing for the word URL
changed how MattermostView notifies it has successfully loaded the requested page

**Issue link**

[MM-28541](https://mattermost.atlassian.net/browse/MM-28541)

**Additional Notes**

### DIFF
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -14,7 +14,7 @@ import 'airbnb-js-shims/target/es2015';
 import Utils from 'common/utils/util';
 import urlUtils from 'common/utils/url';
 
-import {DEVELOPMENT, PRODUCTION, SECOND} from 'common/utils/constants';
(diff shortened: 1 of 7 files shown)


nevyangelova (approved):

src/main/main.js
```diff
@@ -412,24 +402,26 @@ function handleAppGPUProcessCrashed(event, killed) {
   console.log(`The GPU process has crashed (killed = ${killed})`);
 }
 
+function openDeepLink(deeplinkingUrl) {
```
devinbinnie, line 405: Should we not use this above in the `handleAppSecondInstance` handler?

devinbinnie (approved):

src/main/main.js
Willyfrog: definitely, good catch!

PR #1477, opened 2021-02-15
Fix draw badge

Willyfrog:
**Summary**
on windows, the badge code is no longer in the renderer process and it doesn't have direct access to the document to draw in a canvas.  To fix that we reach out to the webcontents of the main window and use it to get a canvas to draw.

@jupenur since we inject code into the webcontents, I wonder if that might be exploitable.

### DIFF
--- a/src/main/windows/windowManager.js
+++ b/src/main/windows/windowManager.js
@@ -209,18 +209,25 @@ function createDataURL(text, small) {
   }
 
   // since we don't have a document/canvas object in the main process, we use the webcontents from the window to draw.
-  return win.webContents.executeJavascript(`
-  window.drawBadge = fucntion ${drawBadge};
-  window.drawBadge(${text}, ${small});
-  `);
+  const code = `
+    window.drawBadge = ${drawBadge};
+    window.drawBadge(${text}, ${small});
+  `;
+  return win.webContents.executeJavaScript(code);
 }
 
-export function setOverlayIcon(badgeText, description) {
+export async function setOverlayIcon(badgeText, description) {
   if (process.platform === 'win32') {
-    const overlay = badgeText ? nativeImage.createFromDataURL(createDataURL(badgeText)) : null;
-    if (status.mainWindow) {
-      status.mainWindow.setOverlayIcon(overlay, description);
+    let overlay = null;
+    if (status.mainWindow && badgeText) {
+      try {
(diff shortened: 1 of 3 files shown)


jupenur (approved):
> since we inject code into the webcontents, I wonder if that might be exploitable

The code doesn't expose any new APIs to the webContents, so there isn't much to exploit. An attacker would be able to replace the badge with an arbitrary image, but that's not much of a security issue really.

src/main/windows/windowManager.js
```diff
@@ -178,12 +178,56 @@ export function flashFrame(flash) {
...
+function createDataURL(text, small) {
+  const win = status.mainWindow;
+  if (!win) {
+    return null;
+  }
+
+  // since we don't have a document/canvas object in the main process, we use the webcontents from the window to draw.
+  const code = `
+    window.drawBadge = ${drawBadge};
+    window.drawBadge(${text}, ${small});
```
jupenur, line 214: How do the variables get expanded here? Shouldn't they be wrapped in quotes or `JSON.stringify`?

Willyfrog: the `small` one wouldn't be a problem as long as it is a boolean or undefined.
the text one should be surrounded, yes, but since my tests were involving numbers it didn't matter as it was used as an integer. but I'll fix those to prevent any errors

src/main/windows/windowManager.js
```diff
@@ -178,12 +178,57 @@ export function flashFrame(flash) {
   }
 }
 
-export function setOverlayIcon(overlayDataURL, description) {
+function drawBadge(text, small) {
+  const scale = 2; // should rely display dpi
```
devinbinnie, line 182: 0/5 on if we do this, but I wonder if it's safe and secure to instead have a template for this HTML and CSS so that it's not all in JavaScript, and then we render that template when we need to.

devinbinnie (approved):

nevyangelova (approved):

src/main/windows/windowManager.js
Willyfrog: should we add that as a HW ticket? or do you prefer to fix it in this PR?

devinbinnie: It's not required right now, would be 100% on board with making a HW ticket for that. I'll do that now.

PR #1482, opened 2021-02-17
[MM-31399] Use webapp ESLint config in desktop app and resolve inconsistencies

devinbinnie:
**Summary**
The desktop app was using its own ESLint configuration based on an older version of the webapp's ESLint config. This has been updated and some changes have been made:
- Indents are now 4 spaces wide instead of 2 spaces wide (this has been a long time coming)
- `setState` is now allowed throughout all React components
- Console statements are not allowed in any main process file, they will only be allowed in preload scripts, external scripts and renderer process files

Some outstanding issues:
- We still have a lot of TODO statements from the first pass of conversion to `BrowserView`. These will stay and be removed later on once we're sure all functionality is migrated.
- Some files are too long, and we will need to shorten them in time.

**Issue link**
https://mattermost.atlassian.net/browse/MM-31399


src/main/viewManager.js
```diff
@@ -19,258 +19,258 @@ const FINDER_WIDTH = 310;
...
+        } else {
+            this.showInitial();
+        }
     }
 
-    // TODO: send event to highlight selected tab
-  }
+    showInitial = () => {
+        if (this.configServers.length) {
+        // TODO: handle deeplink url
```
Willyfrog, line 75: i believe this comment is not needed anymore

src/main/viewManager.js
```diff
@@ -19,258 +19,258 @@ const FINDER_WIDTH = 310;
...
+            if (element) {
+                this.showByName(element.name);
+            }
+        }
 
-  showByName = (name) => {
-    const newView = this.views.get(name);
-    if (newView.isVisible) {
-      return;
+        // TODO: send event to highlight selected tab
```
Willyfrog, line 82: this seems unnecessary, showbyname already handles that

src/renderer/components/TeamList.jsx
```diff
@@ -10,185 +10,183 @@ import TeamListItem from './TeamListItem.jsx';
...
 
 TeamList.propTypes = {
-  onTeamsChange: PropTypes.func,
-  showAddTeamForm: PropTypes.bool,
-  teams: PropTypes.array,
-  addServer: PropTypes.func,
-  updateTeam: PropTypes.func,
-  toggleAddTeamForm: PropTypes.func,
-  setAddTeamFormVisibility: PropTypes.func,
-  onTeamClick: PropTypes.func,
```
Willyfrog, line 193: this one was removed but it is still being called in the code (see line 101)

src/renderer/js/notification.js
```diff
@@ -15,57 +15,57 @@ import dingDataURL from '../../assets/ding.mp3'; // https://github.com/mattermos
...
+    //     handler();
+    //   };
+    // }
 
-  // get onclick() {
-  //   return super.onclick;
-  // }
+    // get onclick() {
+    //   return super.onclick;
+    // }
```
Willyfrog, line 70: this can be removed, maybe eve the whole file?

Willyfrog (approved):
LGTM, added a few non-blocking comments (well one is, but should be easy enough :) )

great work, the codebase will now be much easier to maintain 🙌

src/main/menus/app.js
```diff
@@ -8,242 +8,242 @@ import {app, Menu, session, shell, webContents} from 'electron';
...
+                    focused.goBack();
+                }
+            },
+        }, {
+            label: 'Forward',
+            accelerator: process.platform === 'darwin' ? 'Cmd+]' : 'Alt+Right',
+            click: () => {
+                const focused = webContents.getFocusedWebContents();
+                if (focused.canGoForward()) {
+                    focused.goForward();
```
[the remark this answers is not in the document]
devinbinnie: I don't think there's really a difference between what we do here and what happens in that logic. I think we just use `goToOffset` in the `windowManager` because we want to be able to handle back and forward with one event.

0/5 of whether we change this or not.

.eslintrc-webapp.json
```diff
@@ -66,24 +59,142 @@
...
     {
-      "files": ["tests/**"],
+      "files": ["**/*.tsx", "**/*.ts"],
+      "extends": [
+        "plugin:@typescript-eslint/recommended"
+      ],
+      "rules": {
+        "camelcase": 0,
+        "no-shadow": 0,
+        "import/no-unresolved": 0, // ts handles this better
```
hmhealey, line 78: Weird that ESLint is fine with these comments since JSON doesn't technically allow them. I guess it doesn't actually read JSON since I don't think the other repos have a file extension on these files

package.json
```diff
@@ -87,6 +91,7 @@
     "npm-run-all": "^4.1.5",
     "spectron": "^12.0.0",
     "style-loader": "^0.23.1",
+    "typescript": "4.1.3",
```
hmhealey, line 94: This is adding TypeScript support or is that coming more later?

hmhealey (approved):

PR #1484, opened 2021-02-20
[MM-21716] fix appicon path resolution

Willyfrog:
**Summary**

fix path used to setup win/linux icon for notifications

**Test Cases**
mention the user

PR #1486, opened 2021-02-22
[MM-25355] Throttle notifications for Windows by channel id

devinbinnie:
**Summary**
On Windows, when receiving multiple notifications sequentially, the application would show them in order but there would be a delay between each notification and if there was a large number of notifications, the application would spam the user.

This PR stops the spam of notifications by only storing the last notification from each channel id in the Notifications Center and auto-dismissing all previous notifications.

**Issue link**
https://mattermost.atlassian.net/browse/MM-25355

### DIFF
--- a/src/main/notifications/index.js
+++ b/src/main/notifications/index.js
@@ -27,13 +27,13 @@ export function displayMention(title, body, channel, teamId, silent, webcontents
         data,
     };
 
-    const mention = new Mention(options, channel);
+    const mention = new Mention(options, channel, teamId);
 
     mention.on('show', () => {
         // On Windows, manually dismiss notifications from the same channel and only show the latest one
         if (process.platform === 'win32') {
             for (const notification of currentNotifications.values()) {
-                if (notification.channel.id === mention.channel.id) {
+                if (notification.channel.id === mention.channel.id && notification.teamId === mention.teamId) {
                     notification.close();
                     currentNotifications.delete(notification);
                 }
--- a/src/main/notifications/index.js
+++ b/src/main/notifications/index.js
@@ -11,6 +11,8 @@ import * as windowManager from '../windows/windowManager';
 import {Mention} from './Mention';
 import {DownloadNotification} from './Download';
 
+const currentNotifications = new Set();
+
 export function displayMention(title, body, channel, teamId, silent, webcontents, data) {
     if (!Notification.isSupported()) {
         log.error('notification not supported');
@@ -24,9 +26,20 @@ export function displayMention(title, body, channel, teamId, silent, webcontents
         silent,
         data,
     };
-    const mention = new Mention(options);
+
+    const mention = new Mention(options, channel);
 
     mention.on('show', () => {
+        // On Windows, manually dismiss notifications from the same channel and only show the latest one
+        if (process.platform === 'win32') {
+            for (const notification of currentNotifications.values()) {
+                if (notification.channel.id === mention.channel.id) {
+                    notification.close();
+                    currentNotifications.delete(notification);
+                }
+            }
+            currentNotifications.add(mention);
+        }
         const notificationSound = mention.getNotificationSound();
         if (notificationSound) {
             windowManager.sendToRenderer(PLAY_SOUND, notificationSound);
--- a/src/main/notifications/index.js
+++ b/src/main/notifications/index.js
@@ -11,7 +11,7 @@ import * as windowManager from '../windows/windowManager';
 import {Mention} from './Mention';
 import {DownloadNotification} from './Download';
 
-const currentNotifications = new Set();
+const currentNotifications = new Map();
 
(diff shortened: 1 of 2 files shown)


devinbinnie:
Notes:
- @esethna Added you as reviewer to confirm that this change fixes the issue in your eyes.
- Includes a similar change to https://github.com/mattermost/desktop/pull/1484, which I'll merge into this branch once that PR is merged.

esethna:
@devinbinnie looks like build failed

devinbinnie:
> @devinbinnie looks like build failed

Build always fails right now, I think there's an issue with the CI. Just wanted your take on the approach as described above.
There will be a build you can test out if this gets merged.

src/main/notifications/index.js
```diff
@@ -24,9 +26,20 @@ export function displayMention(title, body, channel, teamId, silent, webcontents
...
         data,
     };
-    const mention = new Mention(options);
+
+    const mention = new Mention(options, channel);
 
     mention.on('show', () => {
+        // On Windows, manually dismiss notifications from the same channel and only show the latest one
+        if (process.platform === 'win32') {
+            for (const notification of currentNotifications.values()) {
```
hmhealey, line 35: Instead of using a `Set` and iterating over it, you could use a `Map` mapping channel IDs to pending notifications instead. 0/5 on that though

src/main/notifications/index.js
```diff
@@ -24,9 +26,20 @@ export function displayMention(title, body, channel, teamId, silent, webcontents
...
+                if (notification.channel.id === mention.channel.id) {
+                    notification.close();
+                    currentNotifications.delete(notification);
```
hmhealey, line 38: Do we need to do anything to remove notifications from `currentNotifications` when they're dismissed?

Willyfrog (changes requested):
some thoughts:
- we could listen to the 'close' event (which is not guaranteed but it happens most of the time) to remove the channel from the queue to allow new notifications to arrive even if the user is not doing anything.
- since that event is not guaranteed, I'd also add a timer to remove it after a while, otherwise we might get some reports about missing notifications.
- also, I would recomment usign the tabName+team+channel as the key to prevent having two channels with the same name on different servers/teams obscuring each other.

src/main/notifications/index.js
devinbinnie: Isn't that what `currentNotifications.delete` is doing? Should be removing it from the `Set`. Do you mean destroying the object?

src/main/notifications/index.js
Willyfrog: I agree with @hmhealey using a composite key in a map makes more sense, as we want to have a unique element for the key, but the value will be different.

src/main/notifications/index.js
```diff
@@ -24,9 +26,20 @@ export function displayMention(title, body, channel, teamId, silent, webcontents
...
+
+    const mention = new Mention(options, channel, teamId);
 
     mention.on('show', () => {
+        // On Windows, manually dismiss notifications from the same channel and only show the latest one
+        if (process.platform === 'win32') {
+            for (const notification of currentNotifications.values()) {
+                if (notification.channel.id === mention.channel.id && notification.teamId === mention.teamId) {
+                    notification.close();
+                    currentNotifications.delete(notification);
```
Willyfrog, line 38: if the notification was closed by user action, will a second close on it fail somehow?

Willyfrog (approved):
as it is, should work ok, added a couple of comments for consideration :)

esethna:
> This PR stops the spam of notifications by only storing the last notification from each channel id in the Notifications Center and auto-dismissing all previous notifications.

@devinbinnie This only applies to Windows correct? I think this approach is fine but will need thorough testing cc// @jgilliam17

src/main/notifications/index.js
esethna: @jgilliam17 on this test case ^

devinbinnie:
> > This PR stops the spam of notifications by only storing the last notification from each channel id in the Notifications Center and auto-dismissing all previous notifications.
> 
> @devinbinnie This only applies to Windows correct? I think this approach is fine but will need thorough testing cc// @jgilliam17

Correct, and agreed on the testing.

jgilliam17:
@devinbinnie Please give me a ping here or in a DM to let me know which desktop daily build this is going to be on. Thanks.

src/main/notifications/index.js
hmhealey: I mean, is there anything to do if the notification is dismissed by the user before this triggers? I think that's the same thing Guillermo is asking about as well.

devinbinnie: Nope, if it's dismissed then `close()` just doesn't do anything. In terms of the `Set` we always just get rid of the last notification regardless.

hmhealey (approved):

Willyfrog:
@jgilliam17 this should be in the daily build already

PR #1487, opened 2021-02-22
[MM-33141] Fixed use of bad context in TeamList

devinbinnie:
**Summary**
When I migrated over the ESLint configuration from the webapp, I had changed the use of `self` here to `this`. I thought I had tested it, but turns out that the context here wasn't the class, so I've reverted to `self`.

Funny enough, the ESLint warning seems to be gone anyhow...

**Issue link**
https://mattermost.atlassian.net/browse/MM-33141

### DIFF
--- a/src/renderer/components/TeamList.jsx
+++ b/src/renderer/components/TeamList.jsx
@@ -64,18 +64,6 @@ export default class TeamList extends React.PureComponent {
         this.props.onTeamsChange(teams);
     }
 
-    handleTeamEditing = (teamName, teamUrl, teamIndex, teamOrder) => {
-        this.setState({
-            showEditTeamForm: true,
-            team: {
-                url: teamUrl,
-                name: teamName,
-                index: teamIndex,
-                order: teamOrder,
-            },
-        });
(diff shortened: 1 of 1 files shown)


hmhealey (approved):

src/renderer/components/TeamList.jsx
```diff
@@ -84,32 +72,39 @@ export default class TeamList extends React.PureComponent {
...
+            });
+        };
+    }
 
+    render() {
+        const teamNodes = this.props.teams.map((team, i) => {
             return (
                 <TeamListItem
                     index={i}
                     key={`teamListItem_${team.name}`}
```
Willyfrog, line 102: seems these two are not used by the component, we can remove them

Willyfrog (approved):
minor garbage collection, but looking good :)

src/renderer/components/TeamList.jsx
devinbinnie: Have to keep `key` to satisfy linter but we can remove `index`.

PR #1488, opened 2021-02-22
[MM-25122] Use modded version of winreg that supports UTF-8

devinbinnie:
**Summary**
I found a modded version of `node-winreg` that actually supports UTF-8 correctly by changing the code page of the console when the call to `reg.exe` is executed. This allows special characters to work correctly with GPO servers.

**Issue link**
https://mattermost.atlassian.net/browse/MM-25122

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19960,6 +19960,11 @@
       "resolved": "https://registry.npmjs.org/winreg/-/winreg-1.2.4.tgz",
       "integrity": "sha1-ugZWKbepJRMOFXeRCM9UCZDpjRs="
     },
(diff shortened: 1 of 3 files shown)


hmhealey (approved):
Seems fine to me, although it's kind of annoying that we have to switch to a much less used fork of that library. I'm probably just worrying about that more since I'm in the middle of dependency updates right now

Willyfrog (approved):
oh! that's awesome! I didn't see that mentioned anywhere, so that's great with minimal changes.

@hmhealey you are right, but I guess this works for now, we should probably look into something that looks a bit more maintained

PR #1490, opened 2021-02-25
[MM-31342] fix "save image as" context menu crash

Willyfrog:
**Summary**

upgrade electron-context-menu so it upgrades electron-dl which was causing the crash.

**Issue link**
[MM-31342](https://mattermost.atlassian.net/browse/MM-31342)
**Test Cases**

click on a profile picture and use `save image` or `save image` as context menu options

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7970,13 +7970,13 @@
       }
     },
     "electron-context-menu": {
-      "version": "2.4.0",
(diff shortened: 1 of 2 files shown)


devinbinnie (approved):

nevyangelova (approved):
Best kind of fixes :D

PR #1494, opened 2021-02-25
[MM-32809] Remove Toggle Dark Mode menu item for Windows, enable toggling on Linux

devinbinnie:
**Summary**
@Willyfrog has already done the work to use the native dark mode option provided by Electron, so I'm just cleaning up by removing the menu item for Windows.

EDIT: The toggle wasn't working for non-Windows/macOS machines so I've enabled that as well.

**Issue link**
https://mattermost.atlassian.net/browse/MM-32809

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -3,7 +3,7 @@
 // See LICENSE.txt for license information.
 'use strict';
 
-import {app, Menu, session, shell, webContents} from 'electron';
+import {app, ipcMain, Menu, session, shell, webContents} from 'electron';
 
 import * as WindowManager from '../windows/windowManager';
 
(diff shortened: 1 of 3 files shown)


src/main/menus/app.js
```diff
@@ -146,17 +146,6 @@ function createTemplate(config) {
         },
     }];
 
-    if (process.platform !== 'darwin') {
```
Willyfrog, line 149: shouldn't we just check for linux?

Willyfrog (changes requested):

src/main/menus/app.js
devinbinnie: Right lol

devinbinnie: I think we actually want to check for !windows and !macOS, in case someone decides to boot the app up on BSD or sunOS :P

Willyfrog: yep, I agree. Even if we don't oficially support them, we shouldn't make it harder to run on some OSes :)

Willyfrog (approved):
LGTM

Willyfrog (approved):

hmhealey (approved):

PR #1495, opened 2021-02-25
[MM-33238] Check for admin URL when toggling back bar

devinbinnie:
**Summary**
When switching between tabs, we check if the back bar should be displayed. Unfortunately, we didn't account for the `admin_console` route and you would see the back displayed when switching back to a tab on the `admin_console`.

Now we check for the `admin_console` route specifically.
NOTE: Are there any other routes we should be checking for that don't need the back button?

**Issue link**
https://mattermost.atlassian.net/browse/MM-33238

### DIFF
--- a/src/main/MattermostView.js
+++ b/src/main/MattermostView.js
@@ -132,7 +132,7 @@ export class MattermostView extends EventEmitter {
         const request = typeof requestedVisibility === 'undefined' ? true : requestedVisibility;
         if (request && !this.isVisible) {
             this.window.addBrowserView(this.view);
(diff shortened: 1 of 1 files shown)


Willyfrog:
>  NOTE: Are there any other routes we should be checking for that don't need the back button?

maybe the plugin marketplace?
not sure if plugins might also have problems with that (like incident response)

devinbinnie:
> > NOTE: Are there any other routes we should be checking for that don't need the back button?
> 
> maybe the plugin marketplace?
> not sure if plugins might also have problems with that (like incident response)

Plugin Marketplace is a full-screen modal so it doesn't have its own route.

Willyfrog (approved):

hmhealey (approved):

Willyfrog (approved):
LGTM

PR #1496, opened 2021-02-26
[MM-33032] Use `hidden` titleBarStyle value to fix macOS Catalina click issue

devinbinnie:
**Summary**
If you have a tab near the center of the tab bar on macOS, and try to click on the top-most portion of it, the tab would not register a click event and thus not respond.

This is due to the way that Electron was insetting the traffic light buttons, causing some kind of overlap over the tabs. This PR changes the `titleBarStyle` to `hidden` and manually sets the inset, which seems to solve the issue.

**Issue link**
https://mattermost.atlassian.net/browse/MM-33032


Willyfrog (approved):
LGTM!

nevyangelova (approved):

PR #1499, opened 2021-03-01
[MM-33334] Restore keyboard shortcuts for menu items moved to roles

devinbinnie:
**Summary**
When we moved a bunch of the menu shortcuts over to roles, some of the keyboard shortcuts disappeared. This PR restores them.

**Issue link**
https://mattermost.atlassian.net/browse/MM-33334

### DIFF
--- a/src/main/menus/app.js
+++ b/src/main/menus/app.js
@@ -61,6 +61,7 @@ function createTemplate(config) {
         platformAppMenu = platformAppMenu.concat([
             separatorItem, {
                 role: 'quit',
+                accelerator: 'CmdOrCtrl+Q',
             }]);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/menus/app.js
```diff
@@ -117,10 +124,13 @@ function createTemplate(config) {
     }, separatorItem, {
         label: 'Actual Size',
         role: 'resetZoom',
+        accelerator: 'CmdOrCtrl+0',
     }, {
         role: 'zoomIn',
+        accelerator: 'CmdOrCtrl+SHIFT+=',
```
Willyfrog, line 130: I'm not entirely convinced about this one, this is assuming all layouts are US or similar?
```suggestion
        accelerator: 'CmdOrCtrl+SHIFT++',
```
not sure if the above works or needs escaping somehow

Willyfrog (approved):
looking good except for a minor concern on one

src/main/menus/app.js
devinbinnie: It's the same shortcut from the current desktop app, it reflects the fact that in order to get the `+` symbol, you would need to press `SHIFT+=`, unless you're using the numpad (in my case I actually don't have a numpad)

nevyangelova (approved):

src/main/menus/app.js
Willyfrog: I know, that happens in US, UK and French layout (among others), but German and Spanish layouts (and there might be others as well) don't work that way and might even be confusing (I can't press shift on a __shifted__ key)
image
image

devinbinnie: Fair point, I don't know if we can detect keyboard layout but we could do an override if so.
Otherwise we might need some UX input on this.

devinbinnie: @Willyfrog Can I merge as-is for now to fix the base issue? I think we have an outstanding ticket to investigate the keyboard shortcut.

PR #1501, opened 2021-03-04
[MM-33434] Upgrade to Electron v11, some other dependency upgrades

devinbinnie:
**Summary**
Upgraded to Electron v11

Notable changes:
- Issue with `electron-context-menu` as described below. Had to bind a couple functions from the `webContents` that were missing.
- Removal of `destroy()` from `BrowserView`, it is now self memory managed
- `contextIsolation` default was deprecated, so I've enforced it everywhere it makes sense.

**Issue link**
https://mattermost.atlassian.net/browse/MM-33434


Willyfrog (changes requested):
spectron should be upgraded with each major version of electron, in this case: ^13.0.0 See [Version Mapping](https://github.com/electron-userland/spectron#version-map)

also, not to fix it here, but they added [resized and moved events to browserview](https://github.com/electron/electron/pull/26454) which might reduce some code/glitches when resizing that we had in the past.

Looking good and looking forward to testing a native build on M1 🎉

Willyfrog (approved):

wget:
@Willyfrog And if you need M1 devices to test, these are as expensive as 2.4 EUR / day at Scaleway :)

Screenshot_20210305_123307

Willyfrog:
> @Willyfrog And if you need M1 devices to test, these are as expensive as 2.4 EUR / day at Scaleway :)
> 
> Screenshot_20210305_123307

Thanks for the recommendation, I already have one for testing. But that seems like a really nice alternative and I'll keep it in mind

hmhealey (approved):

PR #1504, opened 2021-03-10
[MM-32805] Merge master, migrate LoadingScreen to BrowserView

devinbinnie:
**Summary**
Merging the rest of `master` back into `browser-view` to keep things synced.
Also have added @deanwhillier's loading screen back in using Browser.

NOTE: Might want to review with 'ignore whitespace' on to avoid the linting changes.

**Issue link**
https://mattermost.atlassian.net/browse/MM-32805


src/main/main.js
```diff
@@ -523,6 +524,16 @@ function initializeAfterAppReady() {
         });
     });
 
+    // Code for preventing user-agent checking. See: https://mattermost.atlassian.net/browse/MM-31626
```
Willyfrog, line 527: I'm not sure this is needed, I think it wasn't happening on browser-view

in that regard, we are currently changing the UA for BrowserViews as it wasn't being added by electron, but we might want to check that's still the case. if so, then the patch for gitlab might still be needed

Willyfrog (approved):
did you run into any problems with the loading screen? if so, what was the problem?

src/main/main.js
devinbinnie: Yeah, looks like it works just fine so I'll remove this.

src/renderer/hooks/useAnimationEnd.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+function useAnimationEnd(
+    ref,
+    callback,
+    animationName,
+    listenForEventBubbling = true,
+) {
+    React.useEffect(() => {
+        if (!ref.current) {
+            return undefined;
```
Willyfrog, line 23: I think `undefined` can't be reasigned anymore, but just in case
```suggestion
            return;
```

devinbinnie: That breaks the linter, since we're returning either `void` or a function. Returning `undefined` is a different type and denotes that the function always returns something (in this case, nothing)

deanwhillier (commented):
A couple comments, but otherwise LGTM.

If the loading screen is needed, it needs to remain visible until it returns indicating that the loading animation is complete. I can see that is still how it works, however, I'm noticing that after first loading the app and then switching to another server tab, there is a flash of the loading screen on first load that should either not be there at all, or if it is needed becuase that tab is actually still loading, it needs to stay visible until the primary animation is complete. I think in the webview version, tab switches didn't show the loading screen because they all loaded when the app started.

I was working on a mechanism to get the loading screen to show when refreshing the desktop app but just realized I didn't get that in. I'm guessing what I was trying doesn't apply anymore given all the changes for browserview so not sure how easy that would be to add now.

Speaking of reloading a tab, the app is crashing for me every time with this error when I reload:

```
Application: Mattermost 5.0.0-develop [commit: b1b9602
]
Platform: Darwin 20.3.0 x64
TypeError: Cannot read property 'setInitialized' of undefined
    at ViewManager.setServerInitialized (/Users/deanwhillier/Projects/desktop/release/mac/Mattermost.app/Contents/Resources/app.asar/index.js:39264:12)
    at IpcMainImpl.handleReactAppInitialized (/Users/deanwhillier/Projects/desktop/release/mac/Mattermost.app/Contents/Resources/app.asar/index.js:38749:24)
    at IpcMainImpl.emit (events.js:315:20)
    at Object.<anonymous> (electron/js2c/browser_init.js:157:9051)
    at Object.emit (events.js:315:20)
```

deanwhillier:
FWIW, this is where I had gotten for reloading (yes, it's fairly hacky):

```javascript
    // if a navigation event starts and ends with the same URL, assume a reload occurred
    let prevURL;
    webview.addEventListener('will-navigate', (e) => {
      prevURL = e.url;
    });
    webview.addEventListener('did-navigate', (e) => {
      console.log(self.props.src, e.url);
      if (prevURL && prevURL === e.url) {
        prevURL = null;
        self.setState({
          errorInfo: null,
          reloadTimeoutID: null,
          isWebviewLoaded: false,
        });
      }
    });
```

devinbinnie:
> A couple comments, but otherwise LGTM.
> 
> If the loading screen is needed, it needs to remain visible until it returns indicating that the loading animation is complete. I can see that is still how it works, however, I'm noticing that after first loading the app and then switching to another server tab, there is a flash of the loading screen on first load that should either not be there at all, or if it is needed becuase that tab is actually still loading, it needs to stay visible until the primary animation is complete. I think in the webview version, tab switches didn't show the loading screen because they all loaded when the app started.
> 
> I was working on a mechanism to get the loading screen to show when refreshing the desktop app but just realized I didn't get that in. I'm guessing what I was trying doesn't apply anymore given all the changes for browserview so not sure how easy that would be to add now.
> 
> Speaking of reloading a tab, the app is crashing for me every time with this error when I reload:
> 
> ```
> Application: Mattermost 5.0.0-develop [commit: b1b9602
> ]
> Platform: Darwin 20.3.0 x64
> TypeError: Cannot read property 'setInitialized' of undefined
>     at ViewManager.setServerInitialized (/Users/deanwhillier/Projects/desktop/release/mac/Mattermost.app/Contents/Resources/app.asar/index.js:39264:12)
>     at IpcMainImpl.handleReactAppInitialized (/Users/deanwhillier/Projects/desktop/release/mac/Mattermost.app/Contents/Resources/app.asar/index.js:38749:24)
>     at IpcMainImpl.emit (events.js:315:20)
>     at Object.<anonymous> (electron/js2c/browser_init.js:157:9051)
>     at Object.emit (events.js:315:20)
> ```

Looks like I was missing some reload logic, just needed to make sure I followed the same logic to load the `webContents` that we use on initial load and make sure that I show the loading screen explicitly.

@deanwhillier Have another look :)

deanwhillier (approved):
LGTM, thanks @devinbinnie.

Note: I'm still seeing the flashing loading screen upon first switching to a different server tab. Non blocking atm though.

Reload is now working and now crashing!

devinbinnie:
> LGTM, thanks @devinbinnie.
> 
> Note: I'm still seeing the flashing loading screen upon first switching to a different server tab. Non blocking atm though.
> 
> Reload is now working and now crashing!

Is it just a single flash but the loading screen still shows? I think that has to do with us calling `showByName` more than once or something along those lines. Will have to look into it.

deanwhillier:
Nah, it's a brief flash of the loading screen showing for a split second but then disappearing. Only on switching to a new tab for the first time after the app launches.

PR #1506, opened 2021-03-11
[MM-33668] Restore tests to browser-view branch

Willyfrog:
**Summary**
separated test in three categories: lint, unit and e2e
lint uses original source code
unit makes a test compilation of required code
e2e makes a test compilation (which enables remote and disables contextisolation)

**Issue link**

[MM-33668](https://mattermost.atlassian.net/browse/MM-33668)

**Additional Notes**
many tests have been commented to speed up having some testing enabled. we'll need some effort to re-code some of the remaining tests


Willyfrog:
adding @jupenur as I had to modify the dependency-check security step

src/main/windows/mainWindow.js
```diff
@@ -72,6 +72,7 @@ function createMainWindow(config, options) {
             contextIsolation: false,
             disableBlinkFeatures: 'Auxclick',
             spellcheck,
+            enableRemoteModule: process.env.NODE_ENV === 'test',
```
devinbinnie, line 75: `remote` is being deprecated in Electron 12, and removed in Electron 14. They're replacing it, so will this work with the new module?

devinbinnie (approved):

src/main/windows/mainWindow.js
Willyfrog: it is deprecated as part of `electron`, and moved towards `@electron/remote` so I hope we don't have to implement any big changes.

anyways there are already a number of complains that it requires remote, so I hope Spectron devs address that soon and most likely it will mean we'll need to expose some of the electron APIs in the preload

jupenur (approved):

hmhealey (approved):

Issue #1399, opened 2020-10-26
Changing language forces menu focus and makes it hard to write multi-language messages


steve-todorov:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

If you use multiple languages you would normally swap between them via `ALT + SHIFT`.

However, because of an Electron bug (https://github.com/electron/electron/issues/22213), this focuses the menu and thus you need to either hit Escape like 15 times or manually click with the mouse in the message input. 

As you can imagine, this becomes an **EXTREME PAIN** when your workflow requires language switch. 
I literally need to slow down my entire flow to avoid and go around this bug.

Mattermost as well as https://github.com/RocketChat/Rocket.Chat.Electron/issues/1355, Skype, Slack and other Electron based applications also have this issue and there are a dozen of opened tickets about this in the Electron github issue tracker.

**Environment**
- Operating System: Ubuntu 20.04 LTS
- Mattermost Desktop App version:  Mattermost Version: 4.6.1 
- Mattermost Server version: Mattermost Version: 5.28.1 

**Steps to reproduce**

Setup an additional input language and configure it to switch via  `ALT + SHIFT`.
Start typing and switch fast which is common practice since sometimes you write `нещо на Български език :D` where part of the message is in Bulgarian and part of it an emoji which is in English.

**Expected behavior**

Be able to write a message like a normal person without having to use the mouse to 

**Observed behavior**

`ALT + SHIFT` causes menu focus and severely complicates writing and usage.

![image](https://user-images.githubusercontent.com/2303663/97231716-95509280-17e4-11eb-85e1-b8e774202f9e.png)

**Possible fixes**

Usually this issue goes away if you can set the application to work in full screen.
This hides the toolbar menu therefore it never gets focused.
However with the latest desktop version `5.28.1` of Mattermost there's a top toolbar which is always shown even in full screen and the workaround no longer work.

[...]

PR #1509, opened 2021-03-15
Cleanup of BrowserView migration, some bug fixes

devinbinnie:
**Summary**
Cleaned up some orphaned files, removed and fixed some TODOs, fixed a couple bugs, added some missing constants.

Still some TODOs outstanding, @Willyfrog perhaps we can go through those and either make tickets, make the changes, or remove them.


src/main/views/MattermostView.js
```diff
@@ -202,6 +207,22 @@ export class MattermostView extends EventEmitter {
         return WindowManager.getMainWindow.webContents;
     }
 
+    handleInputEvents = (_, input) => {
+        // Handler for pressing the Alt key to focus the 3-dot menu
+        if (input.key === 'Alt' && input.type === 'keyUp' && this.altLastPressed) {
+            this.altLastPressed = false;
+            WindowManager.focusThreeDotMenu();
+            return;
+        }
```
Willyfrog, line 216: we might want to review this edge-case: https://github.com/mattermost/desktop/issues/1399

src/renderer/components/ErrorView.jsx
```diff
@@ -8,28 +8,12 @@ import React from 'react';
...
-    //   event.preventDefault();
-    //   console.log('TODO: send to main');
-    //   ipcRenderer.send(OPEN_EXTERNAL, props.url);
-    // }
-
-    let retry = null;
-    if (props.retry) {
-        const seconds = (Date.now() - props.retry) / SECOND;
-        retry = <div className='retry-info'><p>{`Trying to load again in ${seconds}`}</p></div>;
-    }
```
Willyfrog, line 32: I was hoping to fix the timer as visually was wrong, but functionality was working ok. Any reason to remove it? it would prevent constant reloads of the server if it can't connect.

src/renderer/index.jsx
```diff
@@ -13,19 +13,11 @@ if (process.env.NODE_ENV === 'production') {
     module.hot.accept();
 }
 
-// TODO: enable again, but for the moment seems to conflict with electron-webpack
-// window.eval = global.eval = () => { // eslint-disable-line no-multi-assign, no-eval
-//   import 'bootstrap/dist/css/bootstrap.min.css';
-//     throw new Error(`Sorry, ${remote.app.name} does not support window.eval() for security reasons.`);
-//   };
```
Willyfrog, line 20: we should look into this, well maybe do a full security evaluation that we are not opening any new possibilities for attacks

Willyfrog (commented):
looking good, and thanks for tackling this :)

re: TODOs totally agree, we should go through them and create tickets as needed. some might just not apply anymore.

added a few questions, but other than that looks good

src/main/views/MattermostView.js
devinbinnie: I'm actually surprised this still happens, since my code should only work on keypress for `Alt` effectively. But i'll take a look on Ubuntu.

src/renderer/components/ErrorView.jsx
devinbinnie: Currently the new loading screen is showing instead of that "Cannot connect to Mattermost" screen. That screen now only shows if we have no connection. UX seems better, though we do provide the user with less information.

I would envision that we include that timer on the new loading screen if at all. @deanwhillier thoughts?

src/renderer/index.jsx
devinbinnie: Created a ticket to deal with this properly: https://mattermost.atlassian.net/browse/MM-33828
Some other changes are involved, I might work on this this week.

src/renderer/components/ErrorView.jsx
Willyfrog: we might not have to include the timer, but keep the retries at least? so it doesn't end in some sort of "eternal loading" screen

devinbinnie: Oh the retries are still happening behind the scenes, there's just no indication to the user.

Willyfrog (approved):

src/main/views/MattermostView.js
devinbinnie: Can't see to reproduce that issue on this branch, so I'm assuming it's been fixed.

hmhealey (approved):

PR #1516, opened 2021-03-22
[MM-34088] remove listeners before destroy

Willyfrog:

**Summary**
there seems to be a race condition when at some point the server is not there when trying to access it. this should probably fix it. but without repro steps is hard to tell

**Issue link**
[MM-34088](https://mattermost.atlassian.net/browse/MM-34088)

### DIFF
--- a/src/main/views/MattermostView.js
+++ b/src/main/views/MattermostView.js
@@ -14,6 +14,8 @@ import {getWindowBoundaries, getLocalPreload} from '../utils';
 import * as WindowManager from '../windows/windowManager';
 import * as appState from '../appState';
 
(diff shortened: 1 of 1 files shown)


devinbinnie (approved):
Would it make sense to null check `server` as well? Not sure if it's necessary or not.

nevyangelova (approved):

Willyfrog:
> Would it make sense to null check `server` as well? Not sure if it's necessary or not.

That would definitely prevent it. But we would have to ensure it doesn't reappear anywhere else. The problem I see is that it is a race condition and it is complicated to repro. So far this is my best guess as where it might happen.

PR #1517, opened 2021-03-22
[MM-33828] Define CSP for all BrowserWindows, enable contextIsolation for main/settings window, other fixes

devinbinnie:
**Summary**
Some security fixes:
- Defined a Content Security Policy for all BrowserWindows created (using index.html, which they all should be using). Currently enables `unsafe-inline` for styles, which was required for either `react-smooth-dnd` or `react-bootstrap`. Follow-up ticket: https://mattermost.atlassian.net/browse/MM-34152
- Enabled `contextIsolation` and disabled `nodeIntegration` for the main and settings windows, so that neither of them can access Node modules. Done for security. Added a preload script that uses `contextBridge` to expose ONLY the required functionality for the renderer process to work (ie. ipcRenderer, some platform information). We could probably go a step further and lock this down to only our functions so that there's even less ways arbitrary calls could run.
- Fixed an issue with the error log when the webContents are killed.

**Issue link**
https://mattermost.atlassian.net/browse/MM-33828

### DIFF
--- a/src/main/windows/settingsWindow.js
+++ b/src/main/windows/settingsWindow.js
@@ -4,17 +4,19 @@
 import {BrowserWindow} from 'electron';
 import log from 'electron-log';
 
-import {getLocalURLString} from '../utils';
+import {getLocalPreload, getLocalURLString} from '../utils';
 
 export function createSettingsWindow(mainWindow, config, withDevTools) {
+    const preload = getLocalPreload('mainWindow.js');
     const spellcheck = (typeof config.useSpellChecker === 'undefined' ? true : config.useSpellChecker);
     const settingsWindow = new BrowserWindow({
         ...config.data,
         parent: mainWindow,
         title: 'Desktop App Settings',
         webPreferences: {
-            nodeIntegration: true,
-            contextIsolation: false,
+            nodeIntegration: false,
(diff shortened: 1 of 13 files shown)


deanwhillier (approved):
LGTM, but with one minor thing.

src/main/windows/settingsWindow.js
```diff
@@ -4,17 +4,19 @@
...
     const spellcheck = (typeof config.useSpellChecker === 'undefined' ? true : config.useSpellChecker);
     const settingsWindow = new BrowserWindow({
         ...config.data,
         parent: mainWindow,
         title: 'Desktop App Settings',
         webPreferences: {
-            nodeIntegration: true,
-            contextIsolation: false,
+            nodeIntegration: false,
+            contextIsolation: true,
```
Willyfrog, line 18: we have removed the tests for the settings, but they should be back at some point.
```suggestion
            nodeIntegration: process.env.NODE_ENV === 'test',
            contextIsolation: process.env.NODE_ENV !== 'test',
```

Willyfrog (approved):
minor comment, but looking good and more secure :)

jupenur (approved):

PR #1519, opened 2021-03-25
[MM-34067] Explicitly destroy web contents in non-permanent BrowserViews

devinbinnie:
**Summary**
When we upgraded to Electron v11, `BrowserView.destroy()` was removed, with Electron electing to have Node manage the objects once removed. However, as indicated by this post: https://github.com/electron/electron/pull/23578#issuecomment-742150830, if you don't explicitly destroy the `webContents` object inside, a zombie will persist.

This was causing issues when hovering links or opening modals, as we would create the BrowserView and then simply set its reference to null. Doing this multiple times would result in a lot of zombies.

This PR explicitly destroys the `webContents` object on each `BrowserView` destroy to avoid this issue.

**Issue link**
https://mattermost.atlassian.net/browse/MM-34067

### DIFF
--- a/src/main/views/MattermostView.js
+++ b/src/main/views/MattermostView.js
@@ -171,6 +171,7 @@ export class MattermostView extends EventEmitter {
         if (this.window) {
             this.window.removeBrowserView(this.view);
         }
+        this.view.webContents.destroy();
         this.window = null;
         this.server = null;
         this.isVisible = false;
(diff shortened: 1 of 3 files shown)


lieut-data (commented):
🎉

Willyfrog (approved):
nice cacth!

since this is a workaround until https://github.com/electron/electron/issues/26929 gets fixed, can we add a comment reminding us to remove it once it is fixed upstream?

hmhealey (approved):
There's something really satisfying about saying we're going to be just destroying something

PR #1529, opened 2021-04-07
[MM-33902] connection issues, add server through command line, non-mattermost servers

Willyfrog:
**Summary**

Make animation finish when it is an error, or on anything other that is not a mattermostt server if it is configured as such
Added two new flags to add a server from the comman line. one adds a regular mattermost server, the other adds a non-mattermost server.

current `--help` looks like:

```shell
$ mattermost --help
Options:
  --version, -v          Prints the application version.               [boolean]
  --dataDir, -d          Set the path to where user data is stored.     [string]
  --disableDevMode, -p   Disable development mode. Allows for testing as if it
                         was Production.                               [boolean]
  --serverName, -s       Add url a new mattermost server to the configuration
                                                                        [string]
  --otherServerName, -S  Add a server that might not be a Mattermost one, this
                         feature is unsupported and you might experience
                         unexpected behaviour.                          [string]
  --help                 Show help                                     [boolean]
```

**Issue link**
[MM-33902](https://mattermost.atlassian.net/browse/MM-33902)

**Test Cases**

***Fail to connect***

1. add a server that doesn't respond
2. wait a few seconds (15s) until the 3 retries fail
expected an error page should show

***Add a mattermost server***

1. on the command line, launch the app with the following flags `-s <name of the server> <url of the server>`
given that name is unique and server url is unique, it will be added to the list of tabs and it will persist after quiting the app

Expected: server tab appears with the provided name, restarting the app persists.

***Add a non-mattermost server***

1. on the command line, launch the app with the following flags `-S <name of the server> <url of the server>` (notice this is a capital S)

given that name is unique and server url is unique, it will be added to the list of tabs and it will persist after quiting the app

Expected: server tab appears with the provided name, restarting the app persists. if it is not a mattermost server it will show once it's loaded.

Another way to test this is to add the key `isNonMattermost` to `true` to any server configured in `config.json`

**Additional Notes**

@esethna @lieut-data one of the problems with the new loading screen was that it waits until the webapp says it's ready. this helps hide the webapp loading (the user would be experiencing two load screens which is not nice), so I took the opportunitty to add the flag we mentioned offline to detect wether the server is a supported mattermost server or any other page.

We also need to do a PR to docs/release notes to try and clarify this feature as not officially supported, as it might have unexpected consecuences.

### DIFF
--- a/src/common/communication.js
+++ b/src/common/communication.js
@@ -38,6 +38,7 @@ export const MODAL_INFO = 'modal-info';
 export const MODAL_CANCEL = 'modal-cancel';
 export const MODAL_RESULT = 'modal-result';
 export const MODAL_SEND_IPC_MESSAGE = 'modal-send-ipc-message';
+export const MODAL_INVOKE_IPC_MESSAGE = 'modal-invoke-ipc-message';
 export const MODAL_OPEN = 'modal-open';
 export const MODAL_CLOSE = 'modal-close';
 export const NOTIFY_MENTION = 'notify_mention';
@@ -75,3 +76,5 @@ export const SELECT_NEXT_TAB = 'select-next-tab';
 export const SELECT_PREVIOUS_TAB = 'select-previous-tab';
 export const ADD_SERVER = 'add-server';
 export const FOCUS_THREE_DOT_MENU = 'focus-three-dot-menu';
+
+export const TEST_SERVER = 'test-server';
--- a/src/common/config/buildConfig.js
+++ b/src/common/config/buildConfig.js
@@ -21,7 +21,8 @@ const buildConfig = {
     defaultTeams: [/*
     {
       name: 'example',
-      url: 'https://example.com'
+      url: 'https://example.com',
+      isNonMattermost: false
     }
   */],
     helpLink: 'https://about.mattermost.com/default-desktop-app-documentation/',
--- a/src/common/config/index.js
+++ b/src/common/config/index.js
@@ -9,6 +9,8 @@ import {EventEmitter} from 'events';
 import {ipcMain, nativeTheme, app} from 'electron';
 import log from 'electron-log';
 
+import urlUtils from 'common/utils/url';
+
 import * as Validator from '../../main/Validator';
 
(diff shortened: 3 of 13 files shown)


esethna (changes requested):
@Willyfrog sorry if this was already discussed, but if we are able to add a server that is non-mattermost via command line, why not allow those types of "servers" to load if they are already configured? (Rather than breaking peoples experience after upgrading)

That said, it makes sense that moving forward (ie when new servers added after upgrading) we would not want to allow adding non-mattermost servers though.

I'd also propose that if someone tries to add a non-mattermost server we should actually error them out on the add server screen rather than them waiting 15seconds for the connection error
image

Message:
> Please add the URL to a Mattermost workspace

Willyfrog:
> @Willyfrog sorry if this was already discussed, but if we are able to add a server that is non-mattermost via command line, why not allow those types of "servers" to load if they are already configured? (Rather than breaking peoples experience after upgrading)
> 

it is a choice we have to make, either we annoy people with non-mattermost servers already configured or people with a properly configured server get a double loading screen.

After submiting I still kept thinking about the issue, we can probably improve a little bit the non-mm server experience by providing a timeout to the loading screen. It'll take longer to appear to load, but at least it will still happen.

> That said, it makes sense that moving forward (ie when new servers added after upgrading) we would not want to allow adding non-mattermost servers though.
> 
> I'd also propose that if someone tries to add a non-mattermost server we should actually error them out on the add server screen rather than them waiting 15seconds for the connection error
> image
> 

I was planning to add that in a future PR, but I think you are right that it should be addressed now to give feedback to people.

> Message:
> 
> > Please add the URL to a Mattermost workspace

lieut-data:
> one of the problems with the new loading screen was that it waits until the webapp says it's ready. this helps hide the webapp loading (the user would be experiencing two load screens which is not nice), so I took the opportunitty to add the flag we mentioned offline to detect wether the server is a supported mattermost server or any other page.

Is there an opportunity to simplify things and simply have a timeout after which the loading screen unconditionally closes? Perhaps this would allow us to defer the "non-Mattermost server" support and complications that arise.

Willyfrog:
as discussed offline, I'm closing this one. separating mattermost servers from other will have to wait for v5.0

a simpler solution for this is in https://github.com/mattermost/desktop/pull/1540

PR #1532, opened 2021-04-08
prevent eager filtering

Willyfrog:
**Summary**

grep -v seems to be using a regex to filter, this should prevent treating `.` as a wildcard


metanerd (approved):
LGTM!

hmhealey (approved):

PR #1544, opened 2021-04-15
[MM-34494] Encapsulate context menu in a class object and add it to each view as necessary

devinbinnie:
**Summary**
Instead of using one context menu object and re-assigning it based on active view, we now create a new `electronContextMenu` object for each view so that the views are able to manage their own menus and each one will have one.

**Issue link**
https://mattermost.atlassian.net/browse/MM-34494


Willyfrog (approved):
looks good, but seeing this makes me think again if we should create our own context-menu library 🤔

hmhealey (approved):

PR #1546, opened 2021-04-16
[MM-34890][MM-34897] Fix some fullscreen issues

devinbinnie:
**Summary**
Fix 2 full-screen issues:
- Fullscreen status wouldn't restore when the app was restarted
- When opening the settings window in fullscreen mode, the window would get stuck in fullscreen and there was no way of getting out of it, so I've prevented the window from full-screening at all on opening.

**Issue link**
https://mattermost.atlassian.net/browse/MM-34890
https://mattermost.atlassian.net/browse/MM-34897

### DIFF
--- a/src/main/windows/mainWindow.js
+++ b/src/main/windows/mainWindow.js
@@ -6,10 +6,10 @@ import fs from 'fs';
 import path from 'path';
 import os from 'os';
 
-import {app, BrowserWindow} from 'electron';
+import {app, BrowserWindow, ipcMain} from 'electron';
 import log from 'electron-log';
 
-import {SELECT_NEXT_TAB, SELECT_PREVIOUS_TAB} from 'common/communication';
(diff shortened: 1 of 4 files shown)


src/main/windows/mainWindow.js
```diff
@@ -52,6 +52,7 @@ function createMainWindow(config, options) {
     }
 
     const {maximized: windowIsMaximized} = windowOptions;
+    const fullscreen = process.platform === 'darwin' && windowIsMaximized && windowOptions.x === 0 && windowOptions.y === 0;
```
Willyfrog, line 55: since we save the info in the file, why not save the fullscreen state so we don't have to guess if it is fullscreen or maximized?

Willyfrog (changes requested):
mostly looks good, I think we can improve the bounds-info.json content so we don't have to guess.

lieut-data (approved):

Willyfrog (approved):

Willyfrog:
@devinbinnie can you cherry-pick this into 4.7?

PR #1560, opened 2021-04-21
[MM-34836] Cleanse URLs of extra slashes on parsing, and add the image proxy case

devinbinnie:
#### Summary
When working with URLs that have multiple slashes in the pathname, if we tried to match them using one of our `is___URL` functions they wouldn't match due to the extra slashes.

This PR adds some code to cleanse the extra slashes out of URLs (since apparently the URL library doesn't do that on parsing), and explicitly defines the image proxy case such that those links will open in a browser.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34836

```release-note
NONE
```

### DIFF
--- a/src/common/utils/url.js
+++ b/src/common/utils/url.js
@@ -40,7 +40,7 @@ function parseURL(inputURL) {
         return inputURL;
     }
     try {
-        return new URL(inputURL);
+        return new URL(inputURL.replace(/([^:]\/)\/+/g, '$1'));
     } catch (e) {
         return null;
     }
--- a/src/main/views/webContentEvents.js
+++ b/src/main/views/webContentEvents.js
@@ -136,6 +136,12 @@ const generateNewWindowListener = (getServersFunction, spellcheck) => {
             return;
         }
 
+        // Image proxy case
+        if (parsedURL.pathname.match(/^\/api\/v[3-4]\/image/)) {
+            shell.openExternal(url);
+            return;
+        }
+
         if (parsedURL.pathname.match(/^\/help\//)) {
             // Help links case
             // continue to open special case internal urls in default browser
--- a/test/unit/index.js
+++ b/test/unit/index.js
@@ -2,3 +2,4 @@
 // See LICENSE.txt for license information.
 
 import './trusted_origins_test';
+import './url_test';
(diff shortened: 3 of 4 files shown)


src/common/utils/url.js
```diff
@@ -40,7 +40,7 @@ function parseURL(inputURL) {
         return inputURL;
     }
     try {
-        return new URL(inputURL);
```
Willyfrog, line 43: can we add a unit test for the regex? so we can see what's a valid one and what are we filtering? that way if we find some other weird urls that cause us trouble we can make sure we don't reopen old problems

src/main/views/webContentEvents.js
```diff
@@ -136,6 +136,12 @@ const generateNewWindowListener = (getServersFunction, spellcheck) => {
             return;
         }
 
+        // Image proxy case
+        if (parsedURL.pathname.match(/^\/api\/v[3-4]\/image/)) {
+            shell.openExternal(url);
+            return;
+        }
+
```
Willyfrog, line 144: maybe we should send to `openExternal` anything that comes from the API? is there anything in API that should be displayed on the desktop?

Willyfrog (approved):
looking good, i left a couple of suggestions

src/main/views/webContentEvents.js
devinbinnie: I think we're blocking anything else that comes from the API, or it opens in a popup window. Not sure what the popup case is. I'd prefer to just handle individual cases as they come up and block other stuff unless it makes sense to allow it through. Just to reduce other potential security problems.

Willyfrog (approved):

src/main/views/webContentEvents.js
```diff
@@ -136,6 +136,12 @@ const generateNewWindowListener = (getServersFunction, spellcheck) => {
             return;
         }
 
+        // Image proxy case
+        if (parsedURL.pathname.match(/^\/api\/v[3-4]\/image/)) {
```
hmhealey, line 140: Strictly speaking, this API never existed in API v3, but we could always leave it for consistency

hmhealey (approved):

PR #1568, opened 2021-04-26
Make daily links in s3 look always the same to help with rainforest

Willyfrog:
#### Summary

In order to help rainforest testing, we need to provide the same link for every daily build, so we rename the files removing its version and adding daily so they can be easily recognized

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,9 +307,19 @@ jobs:
       - attach_workspace:
           at: ./dist
       - run:
+          name: "install renaming utility"
+          command: |
+            apt-get install rename
+      - run:
           name: "Normalize folder names"
           command: |
             mv ./dist/macos-release ./dist/macos
+      - run:
+          name: "Rename to daily for consistency"
+          command: |
+            rename 's/\d+\.\d+\.\d+/daily/' macos/*
+            rename 's/\d+\.\d+\.\d+/daily/' linux/*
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -306,10 +306,20 @@ jobs:
...
+          command: |
+            apt-get install rename
       - run:
           name: "Normalize folder names"
           command: |
             mv ./dist/macos-release ./dist/macos
+      - run:
+          name: "Rename to daily for consistency"
+          command: |
+            rename 's/\d+\.\d+\.\d+/daily/' macos/*
```
metanerd, line 320: I cannot find the exact file name we are trying to replace here (also not in the bucket AFAIS). Maybe it makes sense to write a comment what gets replaced with what. 
Is this replacing e.g. `4.7.0` to `daily`? 
Also maybe it is possible to already create it with this name? If not, then this approach is valid.

metanerd: Ah, I got it now, nevermind.

metanerd (approved):

.circleci/config.yml
Willyfrog: I thought about it, but we only need this change to help with rainforest tests, I prefer to have the version on it as it helps clarify what are you downloading from the channel.

devinbinnie (approved):

PR #1583, opened 2021-05-04
[MM-35454] Fix default downloads path for Windows and Linux, only set for the app if the path is not blank

devinbinnie:
#### Summary
When I had forced the entire app to use the downloads path we use in our config, it revealed that we were defaulting to a path used only by Mac. So now we support Windows and Linux (and don't set a default for other OSes) and will only set the path if it exists at all.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35454

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/common/config/defaultPreferences.js
+++ b/src/common/config/defaultPreferences.js
@@ -13,11 +13,9 @@ const getDefaultDownloadLocation = () => {
         return `/Users/${process.env.USER || process.env.USERNAME}/Downloads`;
     case 'win32':
         return `C:\\Users\\${process.env.USER || process.env.USERNAME}\\Downloads`;
-    case 'linux':
-        return `/home/${process.env.USER || process.env.USERNAME}/Downloads`;
     default:
-        return '';
-    };
+        return `/home/${process.env.USER || process.env.USERNAME}/Downloads`;
+    }
 };
 
 const defaultPreferences = {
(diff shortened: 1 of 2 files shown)


src/common/config/defaultPreferences.js
```diff
@@ -6,6 +6,20 @@
...
+const getDefaultDownloadLocation = () => {
+    switch (process.platform) {
+    case 'darwin':
+        return `/Users/${process.env.USER || process.env.USERNAME}/Downloads`;
+    case 'win32':
+        return `C:\\Users\\${process.env.USER || process.env.USERNAME}\\Downloads`;
+    case 'linux':
+        return `/home/${process.env.USER || process.env.USERNAME}/Downloads`;
+    default:
+        return '';
```
Willyfrog, line 19: most other OSs use a similar structure as linux. defaulting to `''` might not be the best option.

Willyfrog (commented):
minor comment on defaulting for other OSes

jwilander (approved):
LGTM other than @Willyfrog's comment

src/common/config/defaultPreferences.js
devinbinnie: I would prefer the absolute default to be '' because then there's no value and therefore the app default won't be set. I think what I could do is add some of the other platform types as a case here, like BSD and those ones.

devinbinnie: Looks like all other platforms are Unix-based, so the `/home/user/Downloads` path makes sense as the default.

Willyfrog (approved):
LGTM!

PR #1598, opened 2021-05-11
[MM-35588] Ensure that the window exists before calling minimize/maximize/restore

devinbinnie:
#### Summary
We had a crash reported when minimize was called. It's not reproducible easily, but to make sure it doesn't happen again, I'm making sure we check if the window exists before calling maximize/minimize/restore.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35588

#### Release Note
```release-note
NONE
```


jwilander (approved):
LGTM

PR #1611, opened 2021-05-25
add rpm build option to electron builder

svelle:
adds an option to also build rpm packages for deployment in RPM based distros like Fedora and RHEL

we should probably mark this as unoffical or something like that for now.

But tested and working on Fedora 34 with latest master.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -47,7 +47,8 @@
     "target": [
       "deb",
       "tar.gz",
-      "appimage"
+      "appimage",
+      "rpm"
     ],
     "extraFiles": [
       {


Willyfrog:
Looks good, we might want to modify a few things to support this on nightly and releases as well, but the first step is to build it.

We might want to add a few [options specific for RPM packages](https://www.electron.build/configuration/linux#linuxtargetspecificoptions-apk-freebsd-pacman-p5p-and-rpm-options) like category, maintainer and icon.

svelle:
> Looks good, we might want to modify a few things to support this on nightly and releases as well, but the first step is to build it.
> 
> We might want to add a few [options specific for RPM packages](https://www.electron.build/configuration/linux#linuxtargetspecificoptions-apk-freebsd-pacman-p5p-and-rpm-options) like category, maintainer and icon.

Agree on all of that. I pretty much just added the flag and checked if it built, installed and ran :smile:

devinbinnie (approved):

Willyfrog (approved):

PR #1619, opened 2021-06-07
[MM-19601] Update to v12.0.10, fix alt issue in linux

### LINKED ISSUE #1080: Shift+Alt moves focus to top menu instead of changing keyboard layout
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).

---

**Summary**
When doing Alt+Shift, top menu opens instead of changing keyboad language (the same issue Skype had few weeks ago). 

**Environment**
- Operating System: Ubuntu 18.04
- Mattermost Desktop App version: 4.3.0 

**Steps to reproduce**
Open Desktop app, type Alt+Shift

**Expected behavior**
Keyboard layout changes.

**Observed behavior**
Focus is now on top menu of Mattermost window.

### LINKED ISSUE #1399: Changing language forces menu focus and makes it hard to write multi-language messages

Willyfrog:
#### Summary
Update to electron v12.0.10

fixes #1080 and fixes  #1399 as well as some other issues (see [release notes of electron v12.0.10 and v12.0.9](https://www.electronjs.org/releases/stable?version=12))

tested `alt+shift` doesn't trigger menu, `alt` by itself does trigger it.
OS ubuntu 21.04

#### Ticket Link

[MM-19601](https://mattermost.atlassian.net/browse/MM-19601)

#### Release Note
```release-note
update to v12.0.10, fixed alt+shift problem on Linux.
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,24301 @@
 {
   "name": "mattermost-desktop",
   "version": "4.8.0-develop",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "7zip-bin": {
+  "packages": {
+    "": {
+      "name": "mattermost-desktop",
+      "version": "4.8.0-develop",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@hapi/joi": "^16.1.8",
+        "auto-launch": "^5.0.5",
+        "bootstrap": "^3.3.7",
+        "brace-expansion": "^2.0.0",
+        "classnames": "^2.2.6",
+        "electron-context-menu": "^2.5.0",
+        "electron-devtools-installer": "^3.1.1",
+        "electron-is-dev": "^2.0.0",
+        "electron-log": "^4.3.2",
+        "electron-updater": "4.3.8",
+        "font-awesome": "^4.7.0",
+        "prop-types": "^15.6.2",
+        "react": "^16.6.3",
+        "react-bootstrap": "~0.32.4",
+        "react-dom": "^16.6.3",
+        "react-smooth-dnd": "github:mattermost/react-smooth-dnd#af6b471295007274560a375799622c1cd52d678a",
+        "react-transition-group": "^2.5.0",
+        "semver": "^5.5.0",
+        "underscore": "^1.9.1",
+        "valid-url": "^1.0.9",
+        "winreg-utf8": "^0.1.1",
+        "yargs": "^15.3.1"
+      },
+      "devDependencies": {
+        "@babel/core": "^7.2.0",
+        "@babel/plugin-proposal-class-properties": "^7.8.3",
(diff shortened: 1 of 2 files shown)


Willyfrog:
I'm slightly concerned about changing the package-lock to this new version since we didn't have any discussions regarding node v16 and npm v7 is it ok or should we keep the prev. version?

Willyfrog:
tested alt+shift doesn't trigger menu, alt by itself does trigger it.
OS ubuntu 21.04

nevyangelova (approved):

devinbinnie (approved):

devinbinnie:
> I'm slightly concerned about changing the package-lock to this new version since we didn't have any discussions regarding node v16 and npm v7 is it ok or should we keep the prev. version?

I think it makes sense to use Node v16 and npm v7 for development on the desktop app, since we're using it for the webapp. So I think the changes are fine.

PR #1625, opened 2021-06-09
[MM-36308] New way to get signature

Willyfrog:
#### Summary

Seems opensuse.org changed the signature two days ago and it wasn't working for us. changing how we acquire that signature seems to work.

This is already working on 4.7 as I needed to cut the RC

#### Ticket Link

[MM-36308](https://mattermost.atlassian.net/browse/MM-36308)

```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ commands:
         type: string
         default: ""
     steps:
-      - run: wget -qO - https://download.opensuse.org/repositories/Emulators:/Wine:/Debian/xUbuntu_18.04/Release.key | apt-key add -
+      - run: curl -fsSL https://download.opensuse.org/repositories/Emulators:/Wine:/Debian/xUbuntu_18.04/Release.key | gpg --dearmor | tee /etc/apt/trusted.gpg.d/wine.gpg > /dev/null
       - run: apt-get update && apt-get -y install << parameters.apt_opts >>
       - run: npm ci
 


metanerd (approved):
LGTM

amyblais:
@jupenur Kind reminder for review on this.

.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ commands:
         type: string
         default: ""
     steps:
-      - run: wget -qO - https://download.opensuse.org/repositories/Emulators:/Wine:/Debian/xUbuntu_18.04/Release.key | apt-key add -
+      - run: curl -fsSL https://download.opensuse.org/repositories/Emulators:/Wine:/Debian/xUbuntu_18.04/Release.key | gpg --dearmor | tee /etc/apt/trusted.gpg.d/wine.gpg > /dev/null
```
jupenur, line 67: Unrelated to this change, but how do we actually know this is the right key?

Willyfrog: Do you mean that we should check that the key is from opensuse? 


if there is a way to check that, we can add that to the code

if that can't be done, the only other way to do that would be to add the key as context in circleci and use it or (probably better option) create our own dockerfile with the key already present and use that as a base.

jupenur: > Do you mean that we should check that the key is from opensuse?

Yeah we'd need to compare the fingerprint since download.opensuse.org redirects to third-party mirrors that could be hosted by anyone.

>  the only other way to do that would be to add the key as context in circleci

This sounds like a good idea to me.

jwilander:
@Willyfrog we came across this ticket in web platform planning. Is this something still needed?

cc @devinbinnie

Willyfrog:
> @Willyfrog we came across this ticket in web platform planning. Is this something still needed?
> 
> cc @devinbinnie

I believe this PR is not needed anymore, but we should implement the changes suggested in the conversation with @jupenur

PR #1628, opened 2021-06-10
[MM-36030] Allow for duplicate servers with different names to behave normally

devinbinnie:
#### Summary
When adding a server with a different name but the same URL, the app would crash due to differences between what the config was reporting as the server list, and what views actually exists. Now the config will filter on both the server name and URL, effectively allowing users to access the same server with multiple different names,

NOTE: You can still add the same server (ie. same name and URL) twice, it just won't cause a crash. We should be adding some validation on server creation to remedy this in a future ticket.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36030

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/common/config/index.js
+++ b/src/common/config/index.js
@@ -336,7 +336,7 @@ export default class Config extends EventEmitter {
         let newTeams = teams;
         const uniqueURLs = new Set();
         newTeams = newTeams.filter((team) => {
-            return uniqueURLs.has(team.url) ? false : uniqueURLs.add(team.url);
+            return uniqueURLs.has(`${team.name}:${team.url}`) ? false : uniqueURLs.add(`${team.name}:${team.url}`);
         });
         return newTeams;
     }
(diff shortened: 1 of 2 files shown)


Willyfrog (approved):

Willyfrog:
> NOTE: You can still add the same server (ie. same name and URL) twice, it just won't cause a crash. We should be adding some validation on server creation to remedy this in a future ticket.

Once we move the servers from the settings to its own dropdown we can probably add this, as there will be only one place to add/modify servers

jwilander (approved):
LGTM

amyblais:
@devinbinnie @Willyfrog Can this be cherry-picked, or is this fix risky?

devinbinnie:
@amyblais I'd like to see some more testing since it does change the way we handle duplicates, but it fixes a crash so I think it might be worth cherry-picking anyways.

amyblais:
I think it would be good to cherry-pick this, but I know we're getting closer to releasing 4.7. @jgilliam17 What are your thoughts?

amyblais:
@devinbinnie Please help cherry-pick ^

Willyfrog:
cherry-picked directly

PR #1629, opened 2021-06-14
[MM-36326] Added DND and quiet hours packages to check if OS is in DND mode

devinbinnie:
#### Summary
Our application never took into account the DND settings for the OS when decided whether or not to send notifications. Most OSes would block the notification, but some would still allow the sound to play.

This PR checks to see if we are in DND mode for Mac/Windows and blocks sending the notification to begin with.

NOTE: This uses native node modules!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36326

#### Release Note
```release-note
Fixed an issue where notifications or sound notifications would still play when the OS was set to Do Not Disturb or Focus Assist.
```


devinbinnie:
@amyblais Did we want to cherry-pick this?

amyblais:
> @amyblais Did we want to cherry-pick this?

I'd say not needed at this point, it doesn't look like a regression or super urgent.

webpack.config.base.js
```diff
@@ -37,6 +40,27 @@ module.exports = {
             static: path.resolve(__dirname, './src/assets'),
         },
     },
+    module: {
+        rules: [
+            {
+                test: /\.js$/,
```
Willyfrog, line 46: do we need to do this on the renderer as well? if not, we probably want to have this only in the config for the main process.

package.json
```diff
@@ -64,7 +66,6 @@
     "@storybook/react": "^4.0.11",
     "@typescript-eslint/eslint-plugin": "4.15.0",
     "@typescript-eslint/parser": "4.15.0",
-    "7zip-bin": "^4.1.0",
     "awesome-node-loader": "^1.1.1",
```
Willyfrog, line 69: we might want to remove this dep if we don't use it in webpack

src/main/notifications/index.js
```diff
@@ -72,3 +79,15 @@ export function displayDownloadCompleted(fileName, path, serverInfo) {
...
+function getDoNotDisturb() {
+    if (process.platform === 'win32') {
+        return getFocusAssist().value;
+    }
+
+    if (process.platform === 'darwin') {
+        return getDarwinDoNotDisturb();
+    }
+
+    return false;
```
Willyfrog, line 92: we might be able to take a look at gsettings for getting the state in linux (see https://github.com/vector-im/element-web/issues/7758#issuecomment-671433520, it mentions dconf but I think gnome3 migrated to gsettings)

It wont be perfect but we can probably check every now and then for the settings and update accordingly so most linux users won't experience this problem. it might be out of scope of this PR (which is already a great improvement), if so, we can fill a second ticket for it.

Willyfrog (commented):
this is great! we now can use native modules!

I added a couple of comments regarding linux and where to position things.

Willyfrog (changes requested):
also, you might want to take a look at the tests.

src/main/notifications/index.js
devinbinnie: I think we should make it a separate ticket, purely because I don't think this is standard and it might be a little finicky to test and get right. I can make a ticket for it.

Willyfrog (approved):
LGTM!

devinbinnie:
There's a problem with the result of `electron-builder`, looks like it's building the native modules for Linux only. Need to investigate a bit more.

hmhealey (approved):

devinbinnie:
Unfortunately there is still an inherent problem with using native node modules, and that is that they must be built on the OS that they are for. We aren't currently set up to do that and there are multiple CI issues resulting from this change.

Since this change isn't a huge priority and the amount of work to setup the CI to build these properly is a considerable effort, I think it's best to close this PR and backlog this ticket until such time as we can make time to support this. I still have a vested interest in doing this though so I'll make sure to keep it bookmarked for a future date.

PR #1650, opened 2021-07-06
Bootstrap v4 dark mode for modals and settings window

devinbinnie:
#### Summary
There's no provided dark mode for Bootstrap, so we used this package: https://github.com/heminei/bootstrap-dark

The modals and settings window are using default Bootstrap colours for the most part, so we're adding a simple dark theme so that users of the app in dark mode can keep everything easy on the eyes.

#### Release Note
```release-note
Added dark mode for settings and modals.
```


Willyfrog (approved):

src/renderer/modals/certificate/certificate.tsx
```diff
@@ -5,14 +5,27 @@ import React from 'react';
...
 import 'renderer/css/components/CertificateModal.css';
 
+import darkStyles from 'renderer/css/lazy/modals-dark.lazy.css';
+
+import SelectCertificateModal from './certificateModal';
+
+window.addEventListener('message', async (event) => {
+    if (event.data.type === DARK_MODE_CHANGE) {
+        if (event.data.data) {
+            darkStyles.use();
```
hmhealey, line 21: I didn't know about this functionality from `style-loader`. That's really cool actually.

Would it perhaps be possible to not repeat this event handler in multiple files? Perhaps there could be a helper in common for importing the correct CSS files for theming as a one-liner instead of repeating this each time? 0/5 though.

hmhealey (approved):

src/renderer/modals/certificate/certificate.tsx
devinbinnie: I'm actually not sure if it works that way or not, but maybe the import tree would pick it up if it was in a helper. I'll give it a shot.

hmhealey (approved):

cwarnermm:
@devinbinnie - Is there any outstanding documentation needed for this PR from your perspective?

PR #1651, opened 2021-07-07
[MM-36459] Drag and drop for dropdown

devinbinnie:
#### Summary
Ability to drag and drop the items in the server dropdown, using the handle on the left side of the button.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36459

```release-note
NONE
```


src/main/preload/dropdown.js
```diff
@@ -34,12 +35,15 @@ window.addEventListener('message', async (event) => {
...
+    case UPDATE_TEAMS:
+        ipcRenderer.invoke(UPDATE_TEAMS, event.data.data);
+        break;
     default:
         console.log(`got a message: ${event}`);
         console.log(event);
     }
 });
 
-ipcRenderer.on(UPDATE_TEAMS_DROPDOWN, (event, teams, activeTeam, darkMode, expired, mentions, unreads) => {
```
Willyfrog, line 43: 075 but would it make sense to add a GPO field to the server info? that way you could potentially reorder non-gpo servers while disallowing drag and drop on the others.

This might be making everything more complex so it could be something to do in a future PR

Willyfrog (approved):
LGTM

hmhealey (approved):

src/main/preload/dropdown.js
devinbinnie: Yeah there's a bit more complexity to that, I think we need to evaluate how we deal with GPO teams in the first place. I'd like to actually maybe add them to the user stored config (or at least their ordering) so that they can be re-ordered. But that's another problem to solve another time.

matthewbirtch:
Hey @devinbinnie a few things I noticed:
1. Can we change the pointer cursor only while it's dragging (similar to how we do it for channels in the sidebar)?
2. Clicking on a server in the list doesn't always switch. If I click (without dragging) on the left side where the label and icon are, it doesn't change servers. But, if I click in the middle where there is no label or icon, it switches. The whole row should be clickable. The drag function should only be engaged when I've held down the mouse and start dragging (similar to how we do in the channel sidebar).

matthewbirtch (approved):
I think we're good to go with this now @devinbinnie!

PR #1655, opened 2021-07-14
[MM-36431] Logic to support multiple configurable tabs per server

devinbinnie:
#### Summary
This PR contains just the logic to control multiple tabs based on a single server. The relationship is now setup such that we have any number of servers with any number of tabs, and each of those tabs maps to a view. Currently we have 3 different tab types, for each of the new products, but they all just map back to the root server type. Follow-up PRs will expand upon this logic and add behaviours that the app will follow with respect to the different tabs.

This should just be a logic PR so only Dev Review is needed at this time.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36431

#### Release Note
```release-note
Converted the tabs interface to support multiple configurable tabs based on the server added.
```


src/main/Validator.ts
```diff
@@ -92,6 +92,35 @@ const configDataSchemaV2 = Joi.object<ConfigV2>({
...
 
+const configDataSchemaV3 = Joi.object<ConfigV3>({
+    version: Joi.number().min(2).default(2),
+    teams: Joi.array().items(Joi.object({
+        name: Joi.string().required(),
+        url: Joi.string().required(),
+        order: Joi.number().integer().min(0),
+        lastActiveTab: Joi.number().integer().min(0),
+        tabs: Joi.array().items(Joi.object({
+            name: Joi.string().required(),
```
Willyfrog, line 103: nit:
```suggestion
            type: Joi.string().required(),
```

src/main/Validator.ts
```diff
@@ -92,6 +92,35 @@ const configDataSchemaV2 = Joi.object<ConfigV2>({
     downloadLocation: Joi.string(),
 });
 
+const configDataSchemaV3 = Joi.object<ConfigV3>({
+    version: Joi.number().min(2).default(2),
+    teams: Joi.array().items(Joi.object({
+        name: Joi.string().required(),
+        url: Joi.string().required(),
+        order: Joi.number().integer().min(0),
+        lastActiveTab: Joi.number().integer().min(0),
```
Willyfrog, line 101: 0/5 on this one
```suggestion
        lastActiveTab: Joi.number().integer().min(0).defaut(0),
```

src/main/Validator.ts
```diff
@@ -92,6 +92,35 @@ const configDataSchemaV2 = Joi.object<ConfigV2>({
...
+        tabs: Joi.array().items(Joi.object({
+            name: Joi.string().required(),
+            order: Joi.number().integer().min(0),
+        })).default([]),
+    })).default([]),
+    showTrayIcon: Joi.boolean().default(false),
+    trayIconTheme: Joi.any().allow('').valid('light', 'dark').default('light'),
```
Willyfrog, line 108:
```suggestion
    trayIconTheme: Joi.string().allow('').valid('light', 'dark').default('light'),
```

src/main/Validator.ts
```diff
@@ -92,6 +92,35 @@ const configDataSchemaV2 = Joi.object<ConfigV2>({
...
+    minimizeToTray: Joi.boolean().default(false),
+    notifications: Joi.object({
+        flashWindow: Joi.any().valid(0, 2).default(0),
+        bounceIcon: Joi.boolean().default(false),
+        bounceIconType: Joi.any().allow('').valid('informational', 'critical').default('informational'),
```
Willyfrog, line 113:
```suggestion
        bounceIconType: Joi.string().allow('').valid('informational', 'critical').default('informational'),
```

src/main/windows/windowManager.ts
```diff
@@ -437,12 +452,52 @@ export function handleHistory(event: IpcMainEvent, offset: number) {
...
+    const currentTeamTabs = status.config?.teams.find((team) => team.name === currentView.tab.server.name)?.tabs;
+    const currentTab = currentTeamTabs?.find((tab) => tab.name === currentView.tab.type);
+    if (!currentTeamTabs || !currentTab) {
+        return;
+    }
+
+    const currentOrder = currentTab.order;
+
+    // js modulo operator returns a negative number if result is negative, so we have to ensure it's positive
+    const nextOrder = ((currentTeamTabs.length + (currentOrder - 1)) % currentTeamTabs.length);
```
Willyfrog, line 495: isn't this assuming that all tabs are open? although maybe you were planning on adding support for opening/closing tabs in a different PR

src/renderer/components/MainPage.tsx
```diff
@@ -274,12 +238,17 @@ export default class MainPage extends React.PureComponent<Props, State> {
         if (addedIndex === undefined || removedIndex === addedIndex) {
             return;
         }
-        const teamIndex = this.props.moveTabs(removedIndex, addedIndex < this.props.teams.length ? addedIndex : this.props.teams.length - 1);
+        const currentTabs = this.props.teams.find((team) => team.name === this.state.activeServerName)?.tabs;
+        if (!currentTabs) {
+            // TODO: figure out something here
```
Willyfrog, line 243: isn't messaging always displayed?

Willyfrog (approved):
i've added a few comments here and there that even might be addressed in subsequent PRs, but overall it is a good looking PR

src/common/tabs/TabView.ts
```diff
@@ -0,0 +1,61 @@
...
+export function getServerView(srv: MattermostServer, tab: Tab) {
+    switch (tab.name) {
+    case TAB_MESSAGING:
+        return new MessagingTabView(srv);
+    case TAB_FOCALBOARD:
+        return new FocalboardTabView(srv);
+    case TAB_PLAYBOOKS:
+        return new PlaybooksTabView(srv);
+    default:
+        throw new Error('Not implemeneted');
```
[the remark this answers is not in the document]
devinbinnie: The base case should actually be an abstract class, which is something I didn't know existed in the TypeScript world. I will change it to that.

src/common/tabs/TabView.ts
```diff
@@ -0,0 +1,61 @@
...
+    get type(): TabType;
+    get url(): URL;
+}
+
+export function getDefaultTeamWithTabsFromTeam(team: Team) {
+    return {
+        ...team,
+        tabs: [
+            {
+                name: TAB_MESSAGING,
```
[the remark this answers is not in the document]
devinbinnie: The actual name would be a display name that we'll add to the class. I don't assume we're ever going to allow the user to change the name of their tabs. But if we do, I assume we'd call it `custom_name` or something. It's kinda both a `name` and a `type`, so I'm not sure it matters. 0/5

src/common/tabs/BaseTabView.ts
```diff
@@ -0,0 +1,23 @@
...
+    server: MattermostServer;
+
+    constructor(server: MattermostServer) {
+        this.server = server;
+    }
+    get name(): string {
+        return getTabViewName(this.server.name, this.type);
+    }
+    get url(): URL {
+        throw new Error('Not implemented');
```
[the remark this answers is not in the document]
devinbinnie: See below. This should be an abstract class only meant to be inherited. I'm going to make that change so it's clear.

src/common/tabs/TabView.ts
```diff
@@ -0,0 +1,61 @@
...
+    case TAB_FOCALBOARD:
+        return new FocalboardTabView(srv);
+    case TAB_PLAYBOOKS:
+        return new PlaybooksTabView(srv);
+    default:
+        throw new Error('Not implemeneted');
+    }
+}
+
+export function getTabViewName(serverName: string, tabType: string) {
```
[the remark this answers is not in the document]
devinbinnie: It's not a member of a class though so it won't be used in this way. The classes itself as implemented will have their names printed when you call the `name` property.

I've added this method so that we can construct the name based on config items vs. having to create a `TabView` object everytime. Ideally we should have kind of a config shim of sorts that stores all these objects at a higher level so that we're not relying on the base config everytime, but that's a refactor that I didn't think I had time for right now.

src/main/Validator.ts
devinbinnie: Was `any` maybe used here to define a constant?

src/main/windows/windowManager.ts
devinbinnie: Yes, that's a different PR

src/renderer/components/MainPage.tsx
devinbinnie: This is just a null check in case a team isn't found for some reason.

src/common/tabs/TabView.ts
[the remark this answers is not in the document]
Willyfrog: yeah, I got the idea, I was questioning if having a "degraded mode tab"would be more useful. or maybe keep the abstract but have a separate 'this-is-some-sort-of-tab-but-we-don-t-know-what-it-is' 😅

src/main/Validator.ts
Willyfrog: looks like it, but why accept a number and then check that the only valid options `'' | 'light' | 'dark'` are actually strings?

src/common/tabs/TabView.ts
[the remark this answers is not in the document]
devinbinnie: I don't think we would be using it for anything, plus it means that other parts of the app might be to handle a "we don't know what it is" tab. The case for using `google.com` would be akin to the legacy Mattermost server case where we just check if the `/channels` route exists and if it doesn't then we redirect to the base url.

All of that logic is coming in a future PR, this one was just to make sure we can actually handle mapping tabs to browserviews and allowing servers to have an arbitrary number of tabs.

src/main/Validator.ts
devinbinnie: Idk, I didn't write this code, this is from the v2 schema :P

src/common/utils/url.ts
```diff
@@ -156,32 +159,35 @@ function isManagedResource(serverUrl: URL | string, inputURL: URL | string) {
...
-    return secondOption;
+    teams.forEach((team) => {
+        const srv = new MattermostServer(team.name, team.url);
+        team.tabs.forEach((tab) => {
+            const tabView = getServerView(srv, tab);
+            parsedServerUrl = parseURL(tabView.url);
+            if (parsedServerUrl) {
+                // check server and subpath matches (without subpath pathname is \ so it always matches)
+                if (equalUrlsWithSubpath(parsedServerUrl, parsedURL, ignoreScheme)) {
+                    firstOption = {name: tabView.name, url: parsedServerUrl};
```
hmhealey, line 178: Is it a problem that we might overwrite firstOption and secondOption since we aren't breaking out of this loop immediately? Like if multiple tabs match the URL, then the second one will overwrite the `firstOption` set for the first tab.

hmhealey (approved):
Looks good, but I found one place there might be a minor functionality change.

src/common/utils/url.ts
devinbinnie: Yeah I refactored this because you can't break the loop of a `forEach` and I was asked to change it above. I'm not sure if it makes sense to do this or just go back to the `for` loop.

@Willyfrog I can't think of a case where we would match more than one URL unless we had 2 of the same server configured, which I'm pretty we don't allow. Does this seem fine or should we go back to what we had?

Willyfrog: I don't think it would be a problem, if we ever get into that situation (unlikely, but sometimes the unexpected happens) it means that both are valid options so I'm not sure which of those is the good one, so one or the other is fine to me.

Anyway it was a suggested change, we can revert back if you foresee any problems, there was nothing wrong with the previous code other than being less legible.

devinbinnie: Let's leave as-is for now. We have lots of time for this to bake in and be tested.

PR #1656, opened 2021-07-15
add a reminder on how to proceed with releases

Willyfrog:
#### Summary

Whenever we do a release we need to update the links that point to the latests mattermost release

this should live in a release playbook, but since we don't have one yet, this would serve as a reminder of the process.

#### Release Note

```release-note
NONE
```


metanerd (approved):

metanerd:
@devinbinnie please can you log into the GitLab instance so I can give you permissions to run this project?

amyblais:
> this should live in a release playbook, but since we don't have one yet, this would serve as a reminder of the process.

If you're referring to Incident Collaboration, I've added you to the current desktop app playbook so you can edit it as well. @Willyfrog

devinbinnie (approved):

PR #1660, opened 2021-07-16
[MM-36998] Ensure server key must be at least 0

devinbinnie:
#### Summary
In certain circumstances, for example, if there are no current servers, the server key in the renderer process was calculating to -1 or NaN. This PR ensures that it's always a number and must be at least 0 (ie. a valid array index)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36998

### DIFF
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -103,7 +103,7 @@ export default class MainPage extends React.PureComponent<Props, State> {
         this.threeDotMenu = React.createRef();
 
         this.state = {
-            key: this.props.teams.findIndex((team) => team.order === 0),
+            key: this.props.teams.length ? this.props.teams.findIndex((team) => team.order === 0) : 0,
             sessionsExpired: {},
             unreadCounts: {},
             mentionCounts: {},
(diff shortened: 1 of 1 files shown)


src/renderer/components/MainPage.tsx
```diff
@@ -103,7 +103,7 @@ export default class MainPage extends React.PureComponent<Props, State> {
         this.threeDotMenu = React.createRef();
 
         this.state = {
-            key: this.props.teams.findIndex((team) => team.order === 0),
+            key: this.props.teams.length ? this.props.teams.findIndex((team) => team.order === 0) : 0,
```
BenCookie95, line 106: Can we add a comment here to notify people in future that key must be at least 0 (a valid array index), just so people know why it's like this

BenCookie95 (approved):

src/renderer/components/MainPage.tsx
Willyfrog: actually, there is one option where it could be non-0 with this one, unlikely to happen but we can easily prevent it.
```js
const found = this.props.teams.findIndex((team) => team.order === 0);
const key = found >= 0 ? found : 0;
```

regarding the comment 0/5, it might help someone to read it, so it is fine :)

Willyfrog (changes requested):

src/renderer/components/MainPage.tsx
devinbinnie: If that issue happens, I'd like the app to crash because that means there is something else wrong with the configuration or our validation. If we check for it, it will fail silently and we may not notice the erroneous behaviour.

Willyfrog (approved):

PR #1653, opened 2021-07-12
[MM-36938] Disable gpu before reading the registry

Willyfrog:
#### Summary

Disabling the gpu once the app is ready had some unexpected results (see #1645 ) this will trigger reading the config file and applying the change before the app is ready and before it finishes reading the registry (which is really slow in windows)

After merging this, we'll need to reimplement it in master as TS, so don't cherry pick.

#### Ticket Link

Fixes #1645 
Jira: [MM-36938](https://mattermost.atlassian.net/browse/MM-36938)

#### Device Information
This PR was tested on: windows 10 with and without GPO policies

#### Release Note

```release-note
Fixed GUI missbehaviour by disabling GPU after app initialization.
```

PR #1665, opened 2021-07-19
[MM-36753] Revert to Electron v12.0.1

devinbinnie:
#### Summary
A breaking change was introduced into Electron v12.0.2 that causes the top part of our BrowserViews for Mattermost to become draggable, matching the draggable area added to the main window. This PR reverts us back to v12.0.1 where this issue does not exist.

I will be filing a bug with Electron to point out the issue.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36753

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "cross-env": "^5.2.0",
         "css-loader": "^1.0.1",
         "devtron": "^1.4.0",
-        "electron": "^12.0.10",
+        "electron": "12.0.1",
         "electron-builder": "^22.10.5",
         "electron-connect": "^0.6.3",
         "electron-mocha": "^10.0.0",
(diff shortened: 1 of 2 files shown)


Willyfrog (approved):
the change obviously looks good, but I'd like to point a few things:
- on https://github.com/mattermost/desktop/pull/1653 had the same effect and the actual cause was the renderer process crashing. So we might want to check if there is any kind of error anywhere that might be obscured but actually crashing.
- 12.0.14 seems to fix a couple of bugs that might be related. Although it might not affect at all as the relation is a bit far fetched, but it doesn't hurt to try!
- 12.0.12 has a [fix to something that seems related])https://github.com/electron/electron/pull/29733/files), might that be the reason for it failing?

Anyways, I'll approve as we can always bump later before v6.0 :)

devinbinnie:
> the change obviously looks good, but I'd like to point a few things:
> 
> * on [[MM-36938] Disable gpu before reading the registry #1653](https://github.com/mattermost/desktop/pull/1653) had the same effect and the actual cause was the renderer process crashing. So we might want to check if there is any kind of error anywhere that might be obscured but actually crashing.
> * 12.0.14 seems to fix a couple of bugs that might be related
> * 12..0.12 has a [fix to something that seems related])https://github.com/electron/electron/pull/29733/files), it might be the reason for it failing?
> 
> anyways approving as we can always bump later before v6.0

I don't think the renderer process is crashing as there's no error anywhere and the app is behaving normally. I feel like we would see some other evidence if that was happening.
v12.0.14 unfortunately doesn't fix the issue, there's a specific change that I'm pretty sure is causing this, as it is reproducing across v11 and v12 (and presumably higher). This was the PR: https://github.com/electron/electron/pull/28268

I'm going to be filing a bug with Electron citing this change and hopefully it can be fixed.

jwilander (approved):

amyblais:
@Willyfrog Was there a discussion if this is needed for v4.7.1? I think the bug itself is not urgent, but I wasn't sure what the priority is on the Electron revert.

Willyfrog:
I think it is fine for 4.7, we didn't have a discussion about it but we might run into other bugs that were already fixed.

PR #1667, opened 2021-07-22
[Mm 37198] enable global sandboxing to increase security

Willyfrog:
#### Summary

Global sandboxing is enabled.

once this is merged, I'll do a separate PR for 4.7

#### Ticket Link
[MM-37198](https://mattermost.atlassian.net/browse/MM-37198)

#### Device Information

This PR was tested on: windows 10 & macos big sur

#### Release Note

```release-note
global sandbox enabled
```

### DIFF
--- a/src/common/config/index.ts
+++ b/src/common/config/index.ts
@@ -50,7 +50,11 @@ export default class Config extends EventEmitter {
         super();
         this.configFilePath = configFilePath;
         this.registryConfig = new RegistryConfig();
-        this.useNativeWindow = os.platform() === 'win32' && os.release().startsWith('10');
+        try {
+            this.useNativeWindow = os.platform() === 'win32' && (parseInt(os.release().split('.')[0], 10) < 10);
+        } catch {
+            this.useNativeWindow = false;
+        }
     }
 
     // separating constructor from init so main can setup event listeners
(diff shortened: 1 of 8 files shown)


jupenur (approved):
LGTM. Thanks @Willyfrog!

src/common/config/index.ts
```diff
@@ -315,7 +322,7 @@ export default class Config extends EventEmitter {
      */
     regenerateCombinedConfigData = () => {
         // combine all config data in the correct order
-        this.combinedData = Object.assign({}, this.defaultConfigData, this.localConfigData, this.buildConfigData, this.registryConfigData);
+        this.combinedData = Object.assign({}, this.defaultConfigData, this.localConfigData, this.buildConfigData, this.registryConfigData, {useNativeWindow: this.useNativeWindow});
```
devinbinnie, line 325: Just a suggestion, and we don't have to do this here, but I wonder if it makes sense for us to define a type for more of these little items that should be passed around with the config. One thing that comes to mind was what I mentioned before, passing around the tab objects used by the main process that contain info on how the tabs are constructed wrt. products and the relative URLs.

devinbinnie (approved):

srkgupta (approved):
Tested the security fix and the issue is working fine now. Thanks @Willyfrog. LGTM.

PR #1692, opened 2021-08-10
fix package lock

devinbinnie:
#### Summary
non-macOS clients were throwing an error on `npm i` because `package-lock` was expecting them to download a mac-only version of `fsevents`. I ran `npm dedupe` and it seems to have solved the issue.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1845,9 +1845,9 @@
       }
     },
     "node_modules/@babel/preset-env/node_modules/babel-plugin-polyfill-corejs3": {
-      "version": "0.2.3",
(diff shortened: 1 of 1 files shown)


jwilander (approved):
Has package-lock.json ever solved any problems for us or just caused them? :P

devinbinnie:
> Has package-lock.json ever solved any problems for us or just cause them? :P

Well it's always been there so it may have solved some problems, but we'd never know cause they never happened :P

hmhealey:
I think it's finally at a point where it's solved the dependency-of-a-dependency problem for the most part. The only time that still seems to be an issue is with libraries that do weird things like installing native code or dependencies mucking with the package-lock themselves.

hmhealey (approved):

PR #1696, opened 2021-08-12
[MM-37797] Re-added the check for flashing the window

devinbinnie:
#### Summary
The setting determining whether the window should flash or not on Windows/Linux when a new mention arrives was not being checked before flashing the window, so we've re-added the check.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37797

#### Release Note
```release-note
Fixed an issue where the window would flash on Windows/Linux when a new mention arrives regardless of the setting to turn it on/off.
```


devinbinnie:
@amyblais Let me know if you'd like to cherry-pick this to v4.7

jwilander (approved):

hmhealey (approved):

PR #1697, opened 2021-08-16
[MM-36923] Wait to return to the application before turning off the custom login setting

devinbinnie:
#### Summary
Some login providers like GitLab require a CloudFlare check before hitting the login page, which will trigger a redirect. This PR allows us to stay in 'custom login' mode until we return back to the application.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36923

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/views/webContentEvents.ts
+++ b/src/main/views/webContentEvents.ts
@@ -76,7 +76,7 @@ const generateDidStartNavigation = (getServersFunction: () => TeamWithTabs[]) =>
 
         if (server && urlUtils.isCustomLoginURL(parsedURL, server, serverList)) {
             customLogins[contentID].inProgress = true;
-        } else if (customLogins[contentID].inProgress) {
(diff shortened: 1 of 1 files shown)


devinbinnie:
@Willyfrog Opened as a draft because I'm not sure if this poses a security risk, as we basically allow the application to stay in custom login mode as long as it's not on the Mattermost server.

I'm not sure exactly why this broke in v5.0, because the reason this happens is that `didStartNavigation` fires before `willNavigate` and cuts us off after we navigate a second time.

Willyfrog:
I've been thinking this, and I think it is fine! the real way to fix this is to have custom logins to happen on the browser and then redirect to the app, but we currently don't have the time to do the propper fix, so for now this should be ok.

Willyfrog (approved):

BenCookie95 (approved):
Just make sure to add a ticket that details Guillermo's point about redirecting to the browser for custom logins

PR #1707, opened 2021-08-18
[MM-36793][MM-37954] Upgrade to Electron v13.2.0

devinbinnie:
#### Summary
The dragging bug that overlapped the top of the Mattermost window has miraculously disappeared (on my machine, macOS Big Sur v11.5.2). Corey found another bug that had similar behaviour, but only reproduced on the old version of Electron (v12.0.1). So we're upgrading to Electron v13.2.0 because we now can and the bugs are fixed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36793
https://mattermost.atlassian.net/browse/MM-37954

#### Release Note
```release-note
Upgrade to Electron v13.2.0
```


devinbinnie:
@amyblais Should we backport this to v4.7 as well since we're doing a dot release anyways?

amyblais:
> @amyblais Should we backport this to v4.7 as well since we're doing a dot release anyways?

If it's urgent. We don't normally backport Electron upgrades to dot releases and in my understanding Electron updates need additional regression testing.

Willyfrog:
@devinbinnie upgrading to the latest v12 doesn't fix that, right? if we just upgrade electron within the same major version we won't need to do an exahustive regression testing.

Willyfrog (approved):
LGTM

devinbinnie:
> @devinbinnie upgrading to the latest v12 doesn't fix that, right? if we just upgrade electron within the same major version we won't need to do an exahustive regression testing.

It actually does, so we could upgrade to the latest v12 for the dot release and do v13 for 5.0 if that would work.

hmhealey (approved):

PR #1717, opened 2021-09-01
[MM-35483] Flip fuses on ELECTRON_RUN_AS_NODE

devinbinnie:
#### Summary
Flipping the fuse on build so that the `ELECTRON_RUN_AS_NODE` environment variable is disabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35483

#### Release Note
```release-note
NONE
```


Willyfrog (approved):

devinbinnie:
NOTE: I'm still having issues with this on Linux, but I'm wondering if the behaviour is expected.
With macOS and Windows, when the environment variable is enabled the app simply will not launch. With Linux, we get a Node.js command prompt, which feels wrong but might not be.

jupenur:
> With Linux, we get a Node.js command prompt, which feels wrong but might not be.

This is definitely wrong. Once the PR is building successfully I can do some testing, but I don't know enough about the packaging process to help otherwise. No idea where the issue could be.

jupenur:
@devinbinnie not working on Windows either:
image

devinbinnie:
@jupenur Should be working now, turns out that the documentation was a touch confusing for flipping the fuses, so nothing was actually working.

jupenur (approved):
LGTM. Thanks!

jupenur:
Random observation: earlier the build didn't fail even though the fuses weren't being flipped. I would've expected `flipFuses` to throw an error, but are uncaught errors just ignored in the `afterPack` hook?

Is there something we should do to improve on that?

devinbinnie:
> Random observation: earlier the build didn't fail even though the fuses weren't being flipped. I would've expected `flipFuses` to throw an error, but are uncaught errors just ignored in the `afterPack` hook?
> 
> Is there something we should do to improve on that?

Yeah that's worth looking into. I know `electron/fuses` throws errors when things don't work properly, so I guess `afterPack` is just ignoring them for some reason.

BenCookie95 (approved):

PR #1720, opened 2021-09-01
[MM-36347] Force min window width of 700px

devinbinnie:
#### Summary
This PR forces a minimum window width of 700px, and includes fixes to make sure that those minimum sizes work with all UI elements.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36437

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/common/utils/constants.ts
+++ b/src/common/utils/constants.ts
@@ -16,3 +16,8 @@ export const BACK_BAR_HEIGHT = 36;
 export const THREE_DOT_MENU_WIDTH = 40;
 export const THREE_DOT_MENU_WIDTH_MAC = 80;
 export const MENU_SHADOW_WIDTH = 24;
+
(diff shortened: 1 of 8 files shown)


Willyfrog (approved):
looks good, the only thing I'm not understanding is why use the update_teams_dropdown channel to send the bounds as well, but given all the info we send along I guess it is just not named properly.

devinbinnie:
> looks good, the only thing I'm not understanding is why use the update_teams_dropdown channel to send the bounds as well, but given all the info we send along I guess it is just not named properly.

It's just a general name for all the data we send down to the dropdown. I had it as a single channel for everything. It might be a bit heavy handed, but isn't impact performance as of yet so I think we're okay.

hmhealey (approved):

PR #1723, opened 2021-09-02
Show topbar regardless of whether there are servers or not

devinbinnie:
#### Summary
When no servers are configured, the Add Server modal pops up. However if you close that, on Windows, you will just end up with a blank screen that you can't interact with. This PR makes sure the top bar always shows regardless of how many servers you have.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -223,6 +223,16 @@ export default class MainPage extends React.PureComponent<Props, State> {
                 }
             });
         }
+
+        window.addEventListener('click', this.handleCloseTeamsDropdown);
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('click', this.handleCloseTeamsDropdown);
+    }
+
+    handleCloseTeamsDropdown = () => {
+        window.ipcRenderer.send(CLOSE_TEAMS_DROPDOWN);
     }
 
     handleMaximizeState = (_: IpcRendererEvent, maximized: boolean) => {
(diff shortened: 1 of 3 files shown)


src/renderer/components/MainPage.tsx
```diff
@@ -295,18 +305,11 @@ export default class MainPage extends React.PureComponent<Props, State> {
...
     render() {
-        if (!this.state.activeServerName || !this.state.activeTabName) {
-            return null;
-        }
-        const currentTabs = this.props.teams.find((team) => team.name === this.state.activeServerName)?.tabs;
-        if (!currentTabs) {
-            // TODO: figure out something here
-            return null;
-        }
+        const currentTabs = this.props.teams.find((team) => team.name === this.state.activeServerName)?.tabs || [];
```
hmhealey, line 312: I don't know if this'll be a problem, but it will cause the `TabBar` to re-render whenever this does when the empty array is passed.

hmhealey (approved):

src/renderer/components/MainPage.tsx
devinbinnie: The empty array should only be passed if there are no servers. Once a server is added that should go away.

src/renderer/components/TeamDropdownButton.tsx
```diff
@@ -65,7 +65,7 @@ const TeamDropdownButton: React.FC<Props> = (props: Props) => {
                 <i className='icon-server-variant'/>
                 {badgeDiv}
             </div>
-            <span>{activeServerName}</span>
+            <span>{activeServerName || 'No servers found'}</span>
```
Willyfrog, line 68: nit, maybe there is even better wording for that
```suggestion
            <span>{activeServerName || 'No servers configured'}</span>
```

Willyfrog (approved):
other than a small nit on wording, LGTM

PR #1726, opened 2021-09-08
[MM-38275] Use browser history push for deeplinking and other inter-tab navigation when using a v6.0 server

devinbinnie:
#### Summary
Since we now have functionality to have the desktop app update the webapp's current URL without reloading it, we can now leverage that functionality when deep-linking and clicking links in other tabs so that the page doesn't reload.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38275

```release-note
NONE
```

### DIFF
--- a/src/main/views/MattermostView.ts
+++ b/src/main/views/MattermostView.ts
@@ -24,6 +24,7 @@ import {
 
 import {TabView} from 'common/tabs/TabView';
 
+import {ServerInfo} from 'main/server/serverInfo';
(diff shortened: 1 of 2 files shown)


esethna (approved):
Yay, thanks @devinbinnie. I'll test once merged from the daily builds channel unless there is something you need my attention on prior to merging

Willyfrog (approved):

BenCookie95 (approved):

PR #1728, opened 2021-09-08
[MM-38433] Error checking when a bad server URL is entered

devinbinnie:
#### Summary
In some cases, if you entered a bad server URL and the returned data was not in JSON format, it would crash the app. I've added some error checking to fail gracefully and log it if this happens.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38433

```release-note
NONE
```

### DIFF
--- a/src/main/server/serverAPI.ts
+++ b/src/main/server/serverAPI.ts
@@ -37,8 +37,12 @@ export async function getServerAPI<T>(url: URL, isAuthenticated: boolean, onSucc
             if (response.statusCode === 200) {
                 response.on('data', (chunk: Buffer) => {
                     const raw = `${chunk}`;
-                    const data = JSON.parse(raw) as T;
-                    onSuccess(data);
+                    try {
+                        const data = JSON.parse(raw) as T;
+                        onSuccess(data);
+                    } catch (e) {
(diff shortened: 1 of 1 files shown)


src/main/server/serverAPI.ts
```diff
@@ -37,8 +37,12 @@ export async function getServerAPI<T>(url: URL, isAuthenticated: boolean, onSucc
             if (response.statusCode === 200) {
                 response.on('data', (chunk: Buffer) => {
                     const raw = `${chunk}`;
-                    const data = JSON.parse(raw) as T;
-                    onSuccess(data);
+                    try {
+                        const data = JSON.parse(raw) as T;
+                        onSuccess(data);
+                    } catch (e) {
+                        log.error('Error parsing server data from', url.toString());
```
Willyfrog, line 44: shouldn't we do a `onError?.(...)` here? so the requester know it failed?

Willyfrog (changes requested):

src/main/server/serverAPI.ts
devinbinnie: Seems like a good idea yeah :P

Willyfrog (approved):

hmhealey (approved):

PR #1735, opened 2021-09-14
[MM-38524] Rework the closing and opening tab logic, fixed login issue

devinbinnie:
#### Summary
I've changed around some of the closed tab logic to allow us to be explicit if the tab should be open or closed, and if there's no value we can open it. We will also open the Boards and Playbooks tabs for servers that support them instead of have them open by default and close them if not needed.

This should solve a couple issues with creating servers in which you would not see the loading screen for the new tab, and with removing servers in which you wouldn't be redirected to an active tab.

I've also added a tap into the logic functionality so that the desktop tabs will reload correctly when logging in on one of them.

```release-note
NONE
```

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -47,7 +47,7 @@ import {
 import Config from 'common/config';
 import {MattermostServer} from 'common/servers/MattermostServer';
 import {getDefaultTeamWithTabsFromTeam, TAB_FOCALBOARD, TAB_MESSAGING, TAB_PLAYBOOKS} from 'common/tabs/TabView';
-import Utils, {isServerVersionGreaterThanOrEqualTo} from 'common/utils/util';
+import Utils from 'common/utils/util';
 
 import urlUtils from 'common/utils/url';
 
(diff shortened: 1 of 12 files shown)


Willyfrog (approved):
minor nit, but LGTM

src/main/main.ts
```diff
@@ -799,28 +799,26 @@ function updateServerInfos(teams: TeamWithTabs[]) {
...
-            if (tab.name === TAB_FOCALBOARD && !remoteInfo.hasFocalboard) {
-                log.info(`closing ${team.name}___${tab.name} on !hasFocalboard`);
-                tab.isClosed = true;
-            }
-            if (tab.name !== TAB_MESSAGING && remoteInfo.serverVersion && !isServerVersionGreaterThanOrEqualTo(remoteInfo.serverVersion, '6.0.0')) {
-                log.info(`closing ${team.name}___${tab.name} on !serverVersion`);
-                tab.isClosed = true;
+            if (tab.name !== TAB_MESSAGING && remoteInfo.serverVersion && isServerVersionGreaterThanOrEqualTo(remoteInfo.serverVersion, '6.0.0')) {
+                if (tab.name === TAB_PLAYBOOKS && remoteInfo.hasPlaybooks && tab.isOpen !== false) {
+                    log.info(`opening ${team.name}___${tab.name} on hasPlaybooks`);
```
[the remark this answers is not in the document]
devinbinnie: Yeah, just in case we need to debug anything around why the tabs won't automatically open for someone when they add a new server.

devinbinnie:
@Willyfrog Added a fix for the login issue, have a look.

src/main/preload/mattermost.js
```diff
@@ -242,4 +243,10 @@ ipcRenderer.on(BROWSER_HISTORY_PUSH, (event, pathName) => {
     );
 });
 
+window.addEventListener('storage', (e) => {
```
Willyfrog, line 246: oh! I didn't knew about this! interesting!

src/main/views/viewManager.ts
```diff
@@ -89,6 +89,13 @@ export class ViewManager {
         view.once(LOAD_FAILED, this.failLoading);
     }
 
+    reloadViewIfNeeded = (viewName: string) => {
+        const view = this.views.get(viewName);
+        if (!view?.getWebContents()?.getURL().startsWith(view.tab.url.toString())) {
+            view?.load(view.tab.url);
```
Willyfrog, line 95: my only worry here is that we trigger a loop somehow,
go to base url, that triggers the `__login__` storage event, which reloads the base url, etc.
but we can't listen to the event only once, because someone might logout and login again.

so I'm not proposing to change it, but just raise a bit of awareness in case it is a real possibility

Willyfrog (approved):
added a comment, but LGTM

hmhealey (approved):

src/main/views/viewManager.ts
devinbinnie: I'm pretty sure the `__login__` event is only thrown once so this shouldn't be a concern

PR #1743, opened 2021-09-20
[MM-33112] Add support for multiple custom spellchecking languages

devinbinnie:
#### Summary
This PR allows the user to configure multiple languages to be used by the spellchecker so that multilingual users can enjoy spellchecking in the languages of their choice, regardless of system language.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33112

#### Release Note
```release-note
Add support for multiple languages to be used by the spellchecker.
```


devinbinnie:
@esethna Just for awareness, I have this merged but not to style spec. I'd like to skip UX review for now so that the feature can be merged before we cut an RC for v5.0

src/common/config/defaultPreferences.ts
```diff
@@ -31,7 +31,7 @@ const defaultPreferences: ConfigV3 = {
     useSpellChecker: true,
     enableHardwareAcceleration: true,
     autostart: true,
-    spellCheckerLocale: 'en-US',
+    spellCheckerLocales: [],
```
Willyfrog, line 34: should we default to english again?

src/main/main.ts
```diff
@@ -638,6 +641,12 @@ function handleRemoveServerModal(e: IpcMainEvent, name: string) {
     }
 }
 
+function updateSpellCheckerLocales() {
+    if (config.data?.spellCheckerLocales.length && app.isReady()) {
```
Willyfrog, line 645: if app is not ready, it will revert to system locale, right? wouldn't that feel weird?

src/renderer/components/SettingsPage.tsx
```diff
@@ -426,21 +424,36 @@ export default class SettingsPage extends React.PureComponent<Record<string, nev
...
+                        type='checkbox'
+                        key='inputSpellChecker'
+                        id='inputSpellChecker'
+                        ref={this.useSpellCheckerRef}
+                        checked={this.state.useSpellChecker}
+                        onChange={this.handleChangeUseSpellChecker}
+                    />
+                    {'Check spelling'}
+                    <FormText>
+                        {'Highlight misspelled words in your messages based on your system language configuration. '}
```
Willyfrog, line 439: I think this can be expressed better, but it is not the system language anymore
```suggestion
                        {'Highlight misspelled words in your messages based on your language selection. '}
```

Willyfrog (approved):

src/common/utils/constants.ts
```diff
@@ -21,3 +22,234 @@ export const DEFAULT_WINDOW_WIDTH = 1000;
 export const DEFAULT_WINDOW_HEIGHT = 700;
 export const MINIMUM_WINDOW_WIDTH = 700;
 export const MINIMUM_WINDOW_HEIGHT = 240;
+
+export const localeTranslations: Record<string, string> = {
```
hmhealey, line 26: Is it necessary to specify this as a `Record<string, string>` or can that be omitted? 0/5

hmhealey (approved):

src/common/config/defaultPreferences.ts
devinbinnie: The app already defaults to whatever language your system, so this is unnecessary.

src/main/main.ts
devinbinnie: Yes, but then it runs this again on `initializeAfterAppReady`. I had to check that because there was a case where it would run before the app was ready and cause a crash.

src/common/utils/constants.ts
devinbinnie: TS defaults it to `any` for some reason if I don't specify this.

src/renderer/components/SettingsPage.tsx
devinbinnie: Agreed on the change, but I think we should maybe 'language preference' instead, since the absence of a selection will default to a system language.

PR #1767, opened 2021-09-27
[MM-38695] Fixed minimize to tray setting on Linux

devinbinnie:
#### Summary
This PR fixes the 'minimize to tray' setting on Linux, as it was not changeable. This enables a user to uncheck the setting if they don't have a system tray to work with, thus allowing the window to minimize to the GNOME app bar if needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38695

#### Release Note
```release-note
NONE
```


svelle:
disabling the checkbox as in the screenshot brings back the old behavior which at least helps with not having to manually kill the mattermost app process so this is definitely a win for now. 
image

svelle (approved):
approved as per my last comment!

Willyfrog (approved):

Issue #1814, opened 2021-10-14
Desktop Client Crashes If HTTP Proxy Auth Is Required: net::ERR_TUNNEL_CONNECTION_FAILED


UberPinguin:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Mattermost Desktop Client for Windows crashes when using an HTTP proxy with authentication: net::ERR_TUNNEL_CONNECTION_FAILED

**Steps to reproduce**
Operating System: Windows 10.0.19042_x64
Mattermost Desktop App Version: 5.0.0

1. Configure an HTTP proxy that requires authentication
2. Configure the computer to use that proxy
3. Launch Mattermost client

**Expected behavior**
Mattermost Client should prompt for proxy credentials and authenticate to the proxy.

**Observed behavior**
Client immediately crashes, creates 10 crash dialogs stacked on top of each other.
![image](https://user-images.githubusercontent.com/638306/137352164-58cf5970-f868-434c-aafe-c48756a6b0cf.png)

```
Application: Mattermost 5.0.0
Platform: Windows_NT 10.0.19042 x64
Error: net::ERR_TUNNEL_CONNECTION_FAILED
    at SimpleURLLoaderWrapper.<anonymous> (electron/js2c/browser_init.js:105:7068)
    at SimpleURLLoaderWrapper.emit (events.js:376:20)
```

PR #1820, opened 2021-10-15
Fixed serverAPI throwing error on proxy server challenge, improved modals

devinbinnie:
#### Summary
When trying to use the Desktop App with a proxy server, the app would crash. This is due to a bug in the request logic here: https://github.com/electron/electron/issues/24948

I've implemented a workaround that catches the error. Also did some improvements on how the modals are handled for proxy servers, so now you should only get 1 modal instead of n where n is the number of views.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/1814

#### Release Note
```release-note
Fixed an issue where using a proxy server with the Desktop app causes the app to crash.
```

### DIFF
--- a/src/main/authManager.ts
+++ b/src/main/authManager.ts
@@ -62,7 +62,7 @@ export class AuthManager {
         if (!mainWindow) {
             return;
         }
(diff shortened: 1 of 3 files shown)


amyblais:
@Willyfrog @BenCookie95 This is planned for v5.0.1 dot release.

Issue #1840, opened 2021-10-27
Uncaught exception mattermost client 5.0 & above


zbeggs:
Hello,
I have encountered an issue when upgrading the Mattermost client from version 4.7.2 to version 5.0 and above. After uninstalling and then installing the new version of Mattermost, opening the app gives an uncaughtException error. The error states that the config.json located in /AppData/Roaming/Mattermost/ is busy or locked. This error is consistent with dozens of workstations that had Mattermost 4.7.2 installed before and then upgraded to 5.0. All workstations are running Windows 10 21H2. 

The server version is ubuntu 20.04, running Mattermost server version 6.

I have attached the pop message and uncaughtException error.  

[uncaughtException-1635349069239.txt](https://github.com/mattermost/desktop/files/7427551/uncaughtException-1635349069239.txt)
 
![Mattermost pop up error](https://user-images.githubusercontent.com/93212625/139105298-6bf12d1c-d3fb-42f6-8dfd-7115ce19f1b3.png)

Issue #1825, opened 2021-10-19
Windows: Crashing when changing between Channels and Boards


s1gh:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
When changing back and forth between Channels and Boards the desktop app sometimes crashes.

**Environment**
- Operating System: Windows 10 19042
- Mattermost Desktop App version: 5.0.0 
- Mattermost Server version: 6.0.0 

**Steps to reproduce**
Change back and forth between the two tabs "Channels" and "Boards".

**Expected behavior**
No crash. Channels and Boards loading as usual.

**Observed behavior**
Desktop app sometimes crashes with the following error message:
```
Application: Mattermost 5.0.0
Platform: Windows_NT 10.0.19042 x64
Error: EBUSY: resource busy or locked, open 'C:\Users\username\AppData\Roaming\Mattermost\config.json'
    at C:\Users\username\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:40992:19
    at fs.js:1504:7
    at FSReqCallback.oncomplete (fs.js:180:23)
```

**Possible fixes**
Not sure

devinbinnie:
@s1gh So just to understand, if you have the Boards tab closed, and then re-open it by clicking a link (or by using the Global Header) and then while the loading screen is running you switch back to the Channels tab, and that causes a crash?

devinbinnie:
Thanks for the info, create a ticket to track: https://mattermost.atlassian.net/browse/MM-39428

PR #1845, opened 2021-10-29
[MM-39428] Check for EBUSY, add retry logic and logging for writing to config

devinbinnie:
#### Summary
Some situations can cause the config file to be locked while trying to write to it. This PR adds retry logic to avoid this issue, and also adds logging if any other file related issues happen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39428
Closes https://github.com/mattermost/desktop/issues/1840
Closes https://github.com/mattermost/desktop/issues/1825

#### Release Note
```release-note
Fixed an issue where the app would crash trying to write to the config
```

### DIFF
--- a/src/common/config/index.ts
+++ b/src/common/config/index.ts
@@ -179,7 +179,11 @@ export default class Config extends EventEmitter {
         try {
             this.writeFile(this.configFilePath, this.localConfigData, (error: NodeJS.ErrnoException | null) => {
                 if (error) {
-                    throw new Error(error.message);
+                    if (error.code === 'EBUSY') {
+                        this.saveLocalConfigData();
+                    } else {
+                        this.emit('error', error);
+                    }
                 }
                 this.emit('update', this.combinedData);
                 this.emit('synchronize');
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -166,6 +166,9 @@ async function initializeConfig() {
             config.on('update', handleConfigUpdate);
             config.on('synchronize', handleConfigSynchronize);
             config.on('darkModeChange', handleDarkModeChange);
+            config.on('error', (error) => {
+                log.error(error);
+            });
             handleConfigUpdate(configData);
 
             // can only call this before the app is ready


hmhealey (approved):
Looks good, but do we need anything to potentially back off on this? If the file stays busy, then we might overflow the stack

src/common/config/index.ts
```diff
@@ -179,7 +179,11 @@ export default class Config extends EventEmitter {
         try {
             this.writeFile(this.configFilePath, this.localConfigData, (error: NodeJS.ErrnoException | null) => {
                 if (error) {
-                    throw new Error(error.message);
+                    if (error.code === 'EBUSY') {
+                        this.saveLocalConfigData();
```
Willyfrog, line 183: should it have a max number of retries in case something is wrong with the OS? Not sure how much that might happen so 0/5

Willyfrog (approved):

src/common/config/index.ts
devinbinnie: @hmhealey @Willyfrog I think we're okay to leave as is, since as Harrison said the stack would overflow trying to retry. If the file does stay locked, I would consider that an error and that the app probably should crash since there's no way to handle not having access to the configuration. And I would want it to fail silently since then the user would be unaware that anything is wrong and none of their configuration updates would work.

Willyfrog: maybe it shouldn't crash/fail silently, but gracefully exit with a message about the config not being possible to be saved? not sure if that would confuse the user, but at least he would know.

Willyfrog: also, that might as well come in a separate PR/ticket :)

hmhealey: I agree that we should probably handle it better than just letting the stack overflow, but I'm fine with that being a separate ticket. This is one of those things that's like handling errors when closing a file. There's only so much we can do to handle it, and there's only so much we want to do since it's an error within an edge case.

PR #1863, opened 2021-11-15
[MM-35441] Remove old cert, changes for building node modules

devinbinnie:
This PR removes the existing codesign cert for Windows (we have a new one in CircleCI), also made a change to upgrade `node-gyp` since the old version wasn't working on the Windows box.

```release-note
NONE
```


Willyfrog (commented):
Shouldn't we remove the certificate? I believe we get it from the environment and we shouldn't have it in the repo.

Willyfrog (approved):

jwilander (approved):

PR #1869, opened 2021-11-18
[MM-39491] Force Add Server modal to stay until server has been added

devinbinnie:
#### Summary
When the Add Server modal pops up for the first time when the app is launched, the user is no longer able to remove it until they add a server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39491

#### Release Note
```release-note
Stop the Add Server modal from being removed until the first server has been added.
```


Willyfrog (approved):

src/renderer/modals/newServer/newServer.tsx
```diff
@@ -25,14 +26,38 @@ const onSave = (data: TeamWithIndex) => {
...
+            setUnremovable(event.data.data);
+            break;
+        }
+        default:
+            break;
+        }
+    };
+
+    useEffect(() => {
+        window.addEventListener('message', handleNewServerMessage);
```
hmhealey, line 44: Don't we need to remove this event listener at some point?

src/main/preload/modalPreload.js
```diff
@@ -6,10 +6,29 @@
...
+    MODAL_SEND_IPC_MESSAGE,
+    GET_DARK_MODE,
+    DARK_MODE_CHANGE,
+    GET_MODAL_UNCLOSEABLE,
+    MODAL_UNCLOSEABLE,
+} from 'common/communication';
 
 console.log('preloaded for the modal!');
 
+let uncloseable = false;
```
hmhealey, line 23: Something about storing this state here feels weird to me, but I don't really have a better suggestion of how to do it.

I could see it potentially causing problems if we have multiple modals visible at once since the newest one will dictate whether or not any of the modals are closeable, but I don't think we have any case that this happens currently anyway.

hmhealey (changes requested):

src/main/preload/modalPreload.js
devinbinnie: This actually won't happen because each instance of this code is in its own tab (or BrowserView), so the instance of `uncloseable` is only related to the specific modal that was opened. It is a weird place to store the state, but I don't think I need to create a ModalWrapper component just for this bit of state at this time.

src/renderer/modals/newServer/newServer.tsx
devinbinnie: Based on the comment below, we don't NEED to, but it is a best practice so I'll remove it.

hmhealey (approved):

PR #1874, opened 2021-11-19
[MM-40146][MM-40147] Unit tests for authManager and certificateManager

devinbinnie:
#### Summary
Unit tests for `main/authManager` and `main/certificateManager`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40146
https://mattermost.atlassian.net/browse/MM-40147

```release-note
NONE
```

### DIFF
--- a/src/main/authManager.ts
+++ b/src/main/authManager.ts
@@ -8,13 +8,12 @@ import {PermissionType} from 'types/trustedOrigin';
 import {LoginModalData} from 'types/auth';
 
 import {BASIC_AUTH_PERMISSION} from 'common/permissions';
-
 import urlUtils from 'common/utils/url';
 
-import * as WindowManager from './windows/windowManager';
+import * as WindowManager from 'main/windows/windowManager';
+import {addModal} from 'main/views/modalManager';
+import {getLocalURLString, getLocalPreload} from 'main/utils';
 
-import {addModal} from './views/modalManager';
-import {getLocalURLString, getLocalPreload} from './utils';
 import TrustedOriginsStore from './trustedOrigins';
 
 const modalPreload = getLocalPreload('modalPreload.js');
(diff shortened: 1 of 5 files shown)


src/main/authManager.ts
```diff
@@ -43,7 +42,10 @@ export class AuthManager {
 
     handleAppLogin = (event: Event, webContents: WebContents, request: AuthenticationResponseDetails, authInfo: AuthInfo, callback?: (username?: string, password?: string) => void) => {
         event.preventDefault();
-        const parsedURL = new URL(request.url);
+        const parsedURL = urlUtils.parseURL(request.url);
+        if (!parsedURL) {
```
hmhealey, line 46: When is `parsedURL` falsey? It this changing any behaviour that the user would see?

src/main/certificateManager.ts
```diff
@@ -70,5 +70,6 @@ export class CertificateManager {
                 log.error(`There was a problem using the selected certificate: ${e}`);
             }
         }
+        this.certificateRequestCallbackMap.delete(server);
```
hmhealey, line 73: Similar question here. Was this just not needed before since we'd only ever call this once per server before?

src/main/authManager.ts
devinbinnie: `parseURL` ensures that the URL passed is a valid URL. I changed this for consistency with the rest of our app. Shouldn't change any user behaviour unless the URL passed is invalid which should generate an error anyways.

src/main/certificateManager.ts
devinbinnie: Yeah I think this was just missed, I don't see any reason to keep the request in the callback map after the cert has been approved or the request was cancelled. If we wanted to re-initiate the request we would just reload the page.

hmhealey (approved):

Willyfrog (approved):

PR #1876, opened 2021-11-23
[MM-40277][MM-40279][MM-40301][MM-40307][MM-40310][MM-40313] Unit tests and refactors for main module

devinbinnie:
#### Summary
This PR contains a bunch of small refactors and unit tests for files directly in the `src/main/` folder.
Will add comments below to explain any changes.

NOTE: We shouldn't squash these commits when merging, since it will be easier to rollback one commit if a change caused an issue.
In the future, I will be doing them by folder as well so I don't have 6 tickets per PR :P

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40277
https://mattermost.atlassian.net/browse/MM-40279
https://mattermost.atlassian.net/browse/MM-40301
https://mattermost.atlassian.net/browse/MM-40307
https://mattermost.atlassian.net/browse/MM-40310
https://mattermost.atlassian.net/browse/MM-40313

```release-note
NONE
```


src/main/AutoLauncher.ts
```diff
@@ -17,35 +17,44 @@ export default class AutoLauncher {
         });
     }
 
-    isEnabled() {
-        return this.appLauncher.isEnabled();
+    async upgradeAutoLaunch() {
```
devinbinnie, line 20: This was in a different file for some reason, it uses the same library it just looks like there was 2 implementations for some reason. So I've removed that file and put the code here.

src/main/CriticalErrorHandler.ts
```diff
@@ -57,7 +57,7 @@ export default class CriticalErrorHandler {
             defaultId: 0,
         }).then(({response}) => {
             if (response === 0) {
-                throw new Error('BrowserWindow \'unresponsive\' event has been emitted');
+                app.relaunch();
```
devinbinnie, line 60: This didn't work, it would just throw an error and fail silently. So now we do what I would expect which is to relaunch the app.

src/main/Validator.ts
```diff
@@ -197,22 +176,45 @@ function cleanURL(url: string): string {
     return updatedURL;
 }
 
-export function validateV2ConfigData(data: ConfigV2) {
-    if (Array.isArray(data.teams) && data.teams.length) {
-        // first replace possible backslashes with forward slashes
-        let teams = data.teams.map((team) => {
+function cleanTeam<T extends {name: string; url: string}>(team: T) {
```
devinbinnie, line 179: I consolidated all the functions to clean the URLs in teams and change around that logic. Made so that code coverage was increased without writing duplicate tests, and also reduces a bit of code bloat.

src/main/contextMenu.ts
```diff
@@ -51,11 +51,7 @@ export default class ContextMenu {
     reload = () => {
         this.dispose();
 
-        /**
-         * Work-around issue with passing `WebContents` to `electron-context-menu` in Electron 11
-         * @see https://github.com/sindresorhus/electron-context-menu/issues/123
-         */
-        const options = {window: {webContents: this.view.webContents, inspectElement: this.view.webContents.inspectElement.bind(this.view.webContents), isDestroyed: this.view.webContents.isDestroyed.bind(this.view.webContents), off: this.view.webContents.off.bind(this.view.webContents)} as unknown as WebContents, ...this.menuOptions};
+        const options = {window: this.view, ...this.menuOptions};
```
devinbinnie, line 54: This workaround is no longer needed, it was fixed with the patch I sent to `electron-context-menu` a few versions back, I just never noticed until now XD

Willyfrog (approved):

hmhealey (changes requested):

package.json
```diff
@@ -70,7 +71,10 @@
       "jsx",
       "json"
     ],
-    "testMatch": ["**/src/**/*.test.js"]
+    "testMatch": ["**/src/**/*.test.js"],
+    "globals": {
+      "__HASH_VERSION__": "5.0.0"
```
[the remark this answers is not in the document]
devinbinnie: We aren't using webpack to run the unit tests, they're run by Jest now meaning there's no compilation, this is the correct way to pass the globals to Jest.

src/main/Validator.ts
devinbinnie: I don't think it really matters one way or another, since generics allow us to just check for the fields we care about for this function. It's not like there's a large amount of in-line typing happening here, so I think it's fine doing it this way.

hmhealey (approved):

PR #1926, opened 2021-12-15
[MM-40009] Add servers dropdown shortcut to menu

devinbinnie:
#### Summary
We were missing a menu item for the Servers menu and instead used a globalShortcut. This PR adds the menu item and the correct keyboard shortcut, as well as a reference to the shortcut in the menu itself.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40009

#### Release Note
```release-note
Added menu item to open the dropdown menu called 'Show Servers' under 'Window'
```

### DIFF
--- a/src/main/menus/app.ts
+++ b/src/main/menus/app.ts
@@ -212,7 +212,7 @@ export function createTemplate(config: Config) {
             accelerator: 'CmdOrCtrl+W',
         }, separatorItem, {
             label: 'Show Servers',
-            accelerator: 'CmdOrCtrl+Shift+S',
+            accelerator: `${process.platform === 'darwin' ? 'Cmd+Ctrl' : 'Ctrl+Shift'}+S`,
             click() {
                 ipcMain.emit(OPEN_TEAMS_DROPDOWN);
             },
--- a/src/main/menus/app.ts
+++ b/src/main/menus/app.ts
@@ -5,7 +5,7 @@
 
 import {app, ipcMain, Menu, MenuItemConstructorOptions, MenuItem, session, shell, WebContents, webContents} from 'electron';
 
-import {SHOW_NEW_SERVER_MODAL} from 'common/communication';
+import {OPEN_TEAMS_DROPDOWN, SHOW_NEW_SERVER_MODAL} from 'common/communication';
 import {Config} from 'common/config';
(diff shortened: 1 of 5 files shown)


M-ZubairAhmed (approved):

Willyfrog (approved):

matthewbirtch (changes requested):
Thanks for this @devinbinnie.

Looks like we have the shortcut showing inconsistently for Mac (Windows seems to be ok).
- In the Servers Menu it is showing as `^⌘S`, but that shortcut doesn't actually work. It's showing as `⇧⌘S` in the Mac OS menu and that is the shortcut that actually works. I believe the correct shortcut was supposed to be `^⌘S` though
- Also, the `^` character displays weirdly with this font. I'm wondering if we change the font for the shortcut to `sans-serif` to just default to the OS font for these shortcut characters - that way they will display as they do in the OS. Thoughts?

image

image

devinbinnie:
@matthewbirtch Fixed the shortcut issue in the menu, it should now be Cmd+Ctrl+S (I always confuse the ^ for Shift)

As for the font on the menu, I think it looks fine, 0/5 on if we change it.

matthewbirtch:
> @matthewbirtch Fixed the shortcut issue in the menu, it should now be Cmd+Ctrl+S (I always confuse the ^ for Shift)

Thanks @devinbinnie

> As for the font on the menu, I think it looks fine, 0/5 on if we change it.

I worry that it would be confused with the `^` character (which would be `Shift 6`) though.

devinbinnie:
> > @matthewbirtch Fixed the shortcut issue in the menu, it should now be Cmd+Ctrl+S (I always confuse the ^ for Shift)
> 
> Thanks @devinbinnie
> 
> > As for the font on the menu, I think it looks fine, 0/5 on if we change it.
> 
> I worry that it would be confused with the `^` character (which would be `Shift 6`) though.

I didn't even realize it was a different character XD
I can make the change if needed.

matthewbirtch:
Thanks @devinbinnie is there a new build?

devinbinnie:
> Thanks @devinbinnie is there a new build?

@matthewbirtch Generating one now

matthewbirtch (approved):
Looks great @devinbinnie. Thanks!

PR #1946, opened 2022-01-10
[MM-39852] Setup docker image to run in CI for E2E

devinbinnie:
#### Summary
This PR adds configuration for the CI to run a version of Mattermost in Docker for the E2E tests to run against.
Also I've enabled a test to check that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39852

```release-note
NONE
```

### DIFF
--- a/scripts/setup_e2e_docker.js
+++ b/scripts/setup_e2e_docker.js
@@ -5,6 +5,7 @@ const {spawn, exec} = require('child_process');
 
 const axios = require('axios');
 
+console.log('path to mmctl', process.env.MMCTL_PATH);
 const mmctlPath = process.env.MMCTL_PATH | 'mmctl';
 
 const ping = setInterval(async () => {
--- a/scripts/setup_e2e_docker.js
+++ b/scripts/setup_e2e_docker.js
@@ -5,8 +5,7 @@ const {spawn, exec} = require('child_process');
 
 const axios = require('axios');
 
-console.log('path to mmctl', process.env.MMCTL_PATH);
-const mmctlPath = process.env.MMCTL_PATH | 'mmctl';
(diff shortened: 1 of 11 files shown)


hmhealey (approved):

scripts/setup_e2e_docker.js
```diff
@@ -0,0 +1,59 @@
...
+                            });
+                        });
+                    });
+                });
+            }
+        } else {
+            console.log(`ERROR: Trying to contact server, got ${pingRequest.status}`);
+        }
+    } catch {
+        console.log('waiting for server to respond...');
```
Willyfrog, line 57: we should have a max amount of retries, although most likely CI server will have a timeout eventually. 0/5 on this

scripts/setup_e2e_docker.js
```diff
@@ -0,0 +1,59 @@
...
+const ping = setInterval(async () => {
+    try {
+        const pingRequest = await axios.get('http://localhost:8065/api/v4/system/ping');
+        if (pingRequest.status === 200) {
+            const addUserRequest = await axios.post(
+                'http://localhost:8065/api/v4/users',
+                {
+                    email: 'test@test.com',
+                    username: 'admin1',
+                    password: 'Sys@dmin123',
```
Willyfrog, line 19: would it be possible for the CI to setup these as config variables instead of hardcoding them?

Willyfrog (approved):
Added a couple of comments, but lgtm

scripts/setup_e2e_docker.js
devinbinnie: Yeah I think it just makes sense to rely on the CI server here. For example, it could take longer than expected for the image to come up due to service issues at CircleCI and i'd like their timeout to be the determining factor if something has gone wrong.

PR #1948, opened 2022-01-12
change "tab" to "server" in developer tools menu

svelle:
Since we recently changed the layout of the desktop app and moved the server tabs into a drop down I think it makes more sense to change the wording for this to refer to the "Server" instead of "Tab". Another one I considered was "View" instead of "Tab". I'd be good with either one.

The changes refer to this menu: 
image

```release-note
Changed wording in "File -> View" Menu from "Tab" to "Server" to reflect recent changes in UI.
```

### DIFF
--- a/src/main/menus/app.ts
+++ b/src/main/menus/app.ts
@@ -152,7 +152,7 @@ export function createTemplate(config: Config) {
             }
         },
     }, {
-        label: 'Developer Tools for Current Tab',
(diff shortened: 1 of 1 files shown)


cwarnermm (approved):
Great catch, @svelle! I like your proposed terminology update.

devinbinnie (approved):
Thanks @svelle!

PR #1955, opened 2022-01-18
[MM-35666] Re-ordered the priority of badge sources to leave session expired to last

devinbinnie:
#### Summary
When the user has more than 1 server configured, where at least 1 server has mentions and another has a session expired, both the tray and the badge display the session expired message/badge before mentions. This PR flips that, such that session expired is shown last after mentions and unreads. You can still view session expired from the server menu.

Also I fixed some stuff that I broke. Oops.
And removed some logging that bothered me :P

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35666

```release-note
Mentions/unreads will now take precedence when setting the badge/tray icon over session expired
```

### DIFF
--- a/src/main/app/initialize.test.js
+++ b/src/main/app/initialize.test.js
@@ -202,11 +202,13 @@ describe('main/app/initialize', () => {
     });
 
     describe('initializeAfterAppReady', () => {
-        it('should set spell checker URL if applicable', async () => {
-            Config.spellCheckerURL = 'http://server-1.com';
-            await initialize();
-            expect(session.defaultSession.setSpellCheckerDictionaryDownloadURL).toHaveBeenCalledWith('http://server-1.com/');
-        });
+        if (process.platform !== 'darwin') {
+            it('should set spell checker URL if applicable', async () => {
+                Config.spellCheckerURL = 'http://server-1.com';
+                await initialize();
+                expect(session.defaultSession.setSpellCheckerDictionaryDownloadURL).toHaveBeenCalledWith('http://server-1.com/');
+            });
+        }
 
         it('should clear app cache if last version opened was older', async () => {
             wasUpdated.mockReturnValue(true);
--- a/src/main/preload/mattermost.js
+++ b/src/main/preload/mattermost.js
@@ -149,13 +149,6 @@ window.addEventListener('message', ({origin, data = {}} = {}) => {
         ipcRenderer.send(BROWSER_HISTORY_BUTTON, viewName);
         break;
     }
-    default:
-        if (typeof type === 'undefined') {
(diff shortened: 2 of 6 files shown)


src/main/preload/mattermost.js
```diff
@@ -149,13 +149,6 @@ window.addEventListener('message', ({origin, data = {}} = {}) => {
         ipcRenderer.send(BROWSER_HISTORY_BUTTON, viewName);
         break;
     }
-    default:
```
devinbinnie, line 152: These logs generate a LOT of spam when developing and made it hard to read the logs you want to read. So I've removed them.

src/main/views/MattermostView.ts
```diff
@@ -335,10 +335,6 @@ export class MattermostView extends EventEmitter {
     }
 
     updateMentionsFromTitle = (title: string) => {
-        if (this.serverInfo.remoteInfo.serverVersion && Util.isVersionGreaterThanOrEqualTo(this.serverInfo.remoteInfo.serverVersion, '5.29.0')) {
```
devinbinnie, line 338: Apparently I didn't read and thought that this function only needed to happen on v5.28 and below 😆

hmhealey (approved):

src/main/app/initialize.test.js
```diff
@@ -202,11 +202,13 @@ describe('main/app/initialize', () => {
...
-            await initialize();
-            expect(session.defaultSession.setSpellCheckerDictionaryDownloadURL).toHaveBeenCalledWith('http://server-1.com/');
-        });
+        if (process.platform !== 'darwin') {
+            it('should set spell checker URL if applicable', async () => {
+                Config.spellCheckerURL = 'http://server-1.com';
+                await initialize();
+                expect(session.defaultSession.setSpellCheckerDictionaryDownloadURL).toHaveBeenCalledWith('http://server-1.com/');
+            });
+        }
```
Willyfrog, line 211: is this change expected? seems to test a different thing. Looks ok, but not sure if you wanted to add that change as well.

src/main/preload/mattermost.js
Willyfrog: totally fine, those were to catch any unexpected message coming from the webapp. It's been a while since then so it shouldn't happen or it might be safely ignored.

src/main/views/MattermostView.ts
Willyfrog: only for the asterisk, at some point we should be able to remove that code as we don't expect people to run very old versions of the server.

Willyfrog (approved):
Looking good, added a couple of comments but LGTM to merge

PR #1957, opened 2022-01-19
Change zoom in shortcut to match expected behavior known from other apps

svelle:
This PR changes the zoom in shortcut to match the expected behavior known from other apps like Chrome, Firefox, etc.

The 'new' shortcut would be `ctrl/cmd`+`=` instead of `ctrl/cmd`+`shift`+`=`

```release-note
The app now uses 'ctrl+=' or 'cmd+=' to zoom in to match the behavior known from other apps like Chrome or Firefox.
```


devinbinnie (approved):
Thanks @svelle :)

M-ZubairAhmed (approved):

Issue #1298, opened 2020-05-19
[Windows]: Full Screen mode turns off on restart


PetbkA:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: http://www.mattermost.org/troubleshoot/](http://www.mattermost.org/troubleshoot/).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
If Full Screen mode was enabled, it turns off after restarting the app.

**Environment**
- Operating System: Microsoft Windows 10 LTSC (version 1809) x64, fully patched;
- Mattermost Desktop App version: 4.4.2;
- Mattermost Server version: 5.21.

**Steps to reproduce**
- Open the app;
- Enable Full Screen mode by pressing F11 or _View → Toggle Full Screen_;
- Restart the app.

**Expected behavior**
App restarts with Full Screen mode enabled.

**Observed behavior**
App restarts with Full Screen mode disabled

Willyfrog:
hi! thanks for the report.

I think that's the standard behaviour of windows. I've tested with chrome and edge and neither of them restart in full screen mode.

PetbkA:
> I think that's the standard behaviour of windows. I've tested with chrome and edge and neither of them restart in full screen mode.

Indeed, my mistake was to think that Chrome behaves differently. Although in Chrome I can force Full Screen by setting _--start-fullscreen_ option which doesn't work in the app. Is there analogue for this?

PetbkA:
Funny, I've found a ["reversed request"](https://github.com/mattermost/desktop/issues/538) which _required_ to restart without Full Screen.

This is very strange - usually the expected behavior is that the app should _preserve_ its settings on restart. May be this was done 'cause there were bugs with Full Screen that couldn't be fixed. For example, Full Screen mode is saved when restarting _Visual Studio Code_ which is also an Electron-based app.

If behavior can't be changed, it stilll would be nice to add some flag to start in Full Screen

Willyfrog:
Yeah, I think having a flag makes sense, so people can force it if needed.

I created [a ticket](https://mattermost.atlassian.net/browse/MM-25484) for it

devinbinnie:
Opening as HW.

[...]

PR #1968, opened 2022-01-24
Fullscreen mode

coltoneshaw:
#### Summary
- Added a settings box that allows the app to always be opened in full screen, if desired. 
- Added a CLI arg to force the app to be opened in full screen

If we don't like the specific config setting then we can just stick with the CLI arg.

CLI Example:

```bash
open release/mac/Mattermost.app --args --fullscreen true
open release/mac/Mattermost.app --args -f true
```

#### Ticket Link
  Fixes https://github.com/mattermost/desktop/issues/1298
https://mattermost.atlassian.net/browse/MM-25484

#### Checklist

- [X] Added or updated unit tests (required for all new features)
- [X] Has UI changes

#### Device Information
This PR was tested on: Mac Monterey 

#### Screenshots
Screen Shot 2022-01-24 at 2 48 46 PM

#### Release Note

```release-note
Added a new CLI parameter to open the application in full screen.
Additional setting/preference to always open the desktop application in full screen.
```


src/main/windows/mainWindow.ts
```diff
@@ -65,6 +65,16 @@ function createMainWindow(options: {linuxAppIcon: string}) {
...
     const spellcheck = (typeof Config.useSpellChecker === 'undefined' ? true : Config.useSpellChecker);
+    const isFullScreen = (savedWindowStateFullscreen: boolean | undefined) => {
+        if (global.args.fullscreen !== undefined) {
+            return global.args.fullscreen;
+        }
+
+        if (Config.startInFullscreen !== undefined) {
+            return Config.startInFullscreen;
+        }
+        return savedWindowStateFullscreen ?? false;
```
devinbinnie, line 76: Since the defaulted value of `Config.startInFullscreen` is always `false`, this will always return `false` since it will never be `undefined`. We should probably just check for `Config.startInFullscreen` and if it's falsy then we can default to the savedWindowState.

src/types/config.ts
```diff
@@ -40,6 +40,7 @@ export type ConfigV3 = {
     downloadLocation: string;
     spellCheckerURL?: string;
     lastActiveTeam?: number;
+    startInFullscreen: boolean;
```
devinbinnie, line 43: This should be optional.
```suggestion
    startInFullscreen?: boolean;
```

devinbinnie (changes requested):
Thanks @coltoneshaw! Just a few comments.

devinbinnie:
@esethna Pinging to check the wording on the settings page.

devinbinnie:
@coltoneshaw Also looks like a test is failing after merging in master.

src/types/config.ts
coltoneshaw: Thanks! What makes it require the optional? I notice most aren't optional params. I assume this has to do with params added after the V3 model was released?

devinbinnie: That, for one thing, and also it makes sense for stuff like booleans to be optional since we can just check for a falsy value and have the same behaviour happen.

coltoneshaw:
@devinbinnie - Okay, I fixed the original test. It looks like it's now throwing an error on the below which I'm a bit confused about. The only change I made to this test was to read the `Config.startInFullscreen`. Any idea what broke it? 

```bash
 FAIL  src/main/windows/mainWindow.test.js
  ● main/windows/mainWindow › createMainWindow › should set window size using bounds read from file

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: ObjectContaining {"fullscreen": undefined, "height": 700, "maximized": false, "width": 1280, "x": 400, "y": 300}
    Received: {"backgroundColor": "#fff", "frame": false, "fullscreen": false, "fullscreenable": true, "height": 700, "maximized": false, "minHeight": 240, "minWidth": 700, "paintWhenInitiallyHidden": true, "show": false, "title": undefined, "titleBarStyle": "hidden", "trafficLightPosition": {"x": 12, "y": 12}, "webPreferences": {"disableBlinkFeatures": "Auxclick", "nativeWindowOpen": true, "preload": undefined, "spellcheck": true}, "width": 1280, "x": 400, "y": 300}

    Number of calls: 1

      109 |         it('should set window size using bounds read from file', () => {
      110 |             createMainWindow({});
    > 111 |             expect(BrowserWindow).toHaveBeenCalledWith(expect.objectContaining({
          |                                   ^
      112 |                 x: 400,
      113 |                 y: 300,
      114 |                 width: 1280,

      at Object.<anonymous> (src/main/windows/mainWindow.test.js:111:35)
```

devinbinnie:
> @devinbinnie - Okay, I fixed the original test. It looks like it's now throwing an error on the below which I'm a bit confused about. The only change I made to this test was to read the `Config.startInFullscreen`. Any idea what broke it?
> 
> ```shell
>  FAIL  src/main/windows/mainWindow.test.js
>   ● main/windows/mainWindow › createMainWindow › should set window size using bounds read from file
> 
>     expect(jest.fn()).toHaveBeenCalledWith(...expected)
> 
>     Expected: ObjectContaining {"fullscreen": undefined, "height": 700, "maximized": false, "width": 1280, "x": 400, "y": 300}
>     Received: {"backgroundColor": "#fff", "frame": false, "fullscreen": false, "fullscreenable": true, "height": 700, "maximized": false, "minHeight": 240, "minWidth": 700, "paintWhenInitiallyHidden": true, "show": false, "title": undefined, "titleBarStyle": "hidden", "trafficLightPosition": {"x": 12, "y": 12}, "webPreferences": {"disableBlinkFeatures": "Auxclick", "nativeWindowOpen": true, "preload": undefined, "spellcheck": true}, "width": 1280, "x": 400, "y": 300}
> 
>     Number of calls: 1
> 
>       109 |         it('should set window size using bounds read from file', () => {
>       110 |             createMainWindow({});
>     > 111 |             expect(BrowserWindow).toHaveBeenCalledWith(expect.objectContaining({
>           |                                   ^
>       112 |                 x: 400,
>       113 |                 y: 300,
>       114 |                 width: 1280,
> 
>       at Object.<anonymous> (src/main/windows/mainWindow.test.js:111:35)
> ```

Your function `isFullScreen` defaults to false, the object is expecting `undefined`.

coltoneshaw:
@devinbinnie - Finally fixed this. Sorry for the delay, I haven't had a chance to look at this again until now.

src/main/windows/mainWindow.ts
```diff
@@ -74,7 +84,7 @@ function createMainWindow(options: {linuxAppIcon: string}) {
         minWidth: MINIMUM_WINDOW_WIDTH,
         minHeight: MINIMUM_WINDOW_HEIGHT,
         frame: !isFramelessWindow(),
-        fullscreen: savedWindowState.fullscreen,
+        fullscreen: isFullScreen(options.fullscreen ?? savedWindowState.fullscreen),
```
Willyfrog, line 87:
```suggestion
        fullscreen: isFullScreen(),
```
if we apply the changes above, we don't need any parameters here.

src/main/windows/mainWindow.ts
```diff
@@ -65,6 +65,16 @@ function createMainWindow(options: {linuxAppIcon: string}) {
...
+    const isFullScreen = (savedWindowStateFullscreen: boolean | undefined) => {
+        if (global?.args?.fullscreen !== undefined) {
+            return global.args.fullscreen;
+        }
+
+        if (Config.startInFullscreen) {
+            return Config.startInFullscreen;
+        }
+        return savedWindowStateFullscreen ?? undefined;
+    };
```
Willyfrog, line 77: I don't think you need any parameters for this function as they are not going to change from the definition.

```suggestion
    const isFullScreen = () => {
        if (global?.args?.fullscreen !== undefined) {
            return global.args.fullscreen;
        }

        if (Config.startInFullscreen) {
            return Config.startInFullscreen;
        }
        return options.fullscreen || savedWindowState.fullscreen || false
    };
```

Willyfrog (approved):
Added a comment, but it is not really blocking or needed so 0/5 on it

coltoneshaw:
@Willyfrog - Thanks for your changes! That makes total sense. Just commited them.

PR #1976, opened 2022-01-28
test:MM-T816 Toggle Full Screen in the Menu Bar

codedsun:
test:MM-T816 Toggle Full Screen in the Menu Bar

#### Ticket Link

  Fixes https://github.com/mattermost/desktop/issues/1900

#### Checklist

- [x] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
- [x] executed `npm run lint:js` for proper code formatting

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

PR #1975, opened 2022-01-28
test:MM-T824/MM-T825 Minimize/Close in the Menu Bar

srkgupta:
#### Summary
MM-T824/MM-T825 Minimize/Close in the Menu Bar

#### Ticket Link
Fixes https://github.com/mattermost/desktop/issues/1903

#### Checklist
- [x] Added or updated unit tests (required for all new features)
- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [x] completed [Mattermost Contributor Agreement](http://www.mattermost.org/mattermost-contributor-agreement/)
- [x] executed `npm run lint:js` for proper code formatting
- [ ] 

```release-note
NONE
```

PR #1983, opened 2022-02-01
[MM-T2637] Double-Clicking on the header should minimize/maximize the app

gohyinhao:
#### Summary

e2e test to check minimize/maximize of app when double clicking on header

#### Ticket Link

Fixes https://github.com/mattermost/desktop/issues/1914

#### Checklist

#### Device Information
This PR was tested on: Ubuntu 20.04.3

#### Release Note

```release-note
NONE
```

### DIFF
--- a/e2e/specs/server_management/header.test.js
+++ b/e2e/specs/server_management/header.test.js
@@ -1,0 +1,47 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+'use strict';
+
+const fs = require('fs');
+
+const env = require('../../modules/environment');
+const {asyncSleep} = require('../../modules/utils');
+
+describe('header', function desc() {
+    this.timeout(30000);
+
+    beforeEach(async () => {
+        env.createTestUserDataDir();
+        env.cleanTestConfig();
+    });
+
+    afterEach(async () => {
+        if (this.app) {
+            try {
+                await this.app.close();
+            // eslint-disable-next-line no-empty
+            } catch (err) {}
+        }
+    });
+
+    it('MM-T2637 Double-Clicking on the header should minimize/maximize the app', async () => {
+        const initialBounds = {x: 0, y: 0, width: 800, height: 400, maximized: false};
+        fs.writeFileSync(env.boundsInfoPath, JSON.stringify(initialBounds));
+        this.app = await env.getApp();
+        const mainWindow = await this.app.firstWindow();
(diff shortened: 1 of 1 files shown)


gohyinhao:
i been able to get the test to work as intended in my local env but it seems to always fail in circle ci. let me continue to take a look

devinbinnie:
@gohyinhao Did you manage to find out why the test wouldn't pass in CircleCI?

gohyinhao:
hi @devinbinnie originally, i thought there was something wrong with my implementation so i tried various implementations but they all had similar results (pass in local linux env, but fail in CircleCI). i noticed that other tests relating to minimizing/maximizing the app (#1976 #1975) face similar issues where the app does not seem to be minimized/maximized in CircleCI, so i tried looking for the problem outside of the test implementation.

so far, what I found out is that when i run the test using xvfb locally, the test fails. when i log the window bounds using the playwright methods, it seems like the app does not get maximized/minimized at all. however, i am still unable to ascertain whether the issue lies with xvfb itself, or if it is an issue specific to xvfb with electron apps, or possibly even xvfb with playwright with electron. as such, I'm sorry but I have yet to find a solution for this.

by any chance, are you familiar with such issues or have any insights pertaining to this issue?

devinbinnie:
> 



> hi @devinbinnie originally, i thought there was something wrong with my implementation so i tried various implementations but they all had similar results (pass in local linux env, but fail in CircleCI). i noticed that other tests relating to minimizing/maximizing the app (#1976 #1975) face similar issues where the app does not seem to be minimized/maximized in CircleCI, so i tried looking for the problem outside of the test implementation.
> 
> so far, what I found out is that when i run the test using xvfb locally, the test fails. when i log the window bounds using the playwright methods, it seems like the app does not get maximized/minimized at all. however, i am still unable to ascertain whether the issue lies with xvfb itself, or if it is an issue specific to xvfb with electron apps, or possibly even xvfb with playwright with electron. as such, I'm sorry but I have yet to find a solution for this.
> 
> by any chance, are you familiar with such issues or have any insights pertaining to this issue?

@gohyinhao I have noticed similar things happening with tests that attempt to minimize/maximize the Desktop App, it appears your test is also affected. I just ran it on my local machine and it passed just fine so I think we can say that it's good to merge. 

I'm going to look into removing the check for PRs since a lot of these tests are kind of flaky, but I will merge this one once reviews are complete. Thanks for the hard work :)

M-ZubairAhmed (approved):
🆗  It looks good to me. Thank you

gohyinhao:
thanks both, for taking the time to review

PR #1988, opened 2022-02-02
[MM-30713] Stop Linux and Windows app from minimizing/hiding without user warning

devinbinnie:
#### Summary
Many Linux users have reported that the app doesn't obey closing rules as they would normally expect, so we've refined the way the app closes on Linux and added some dialog boxes to inform the user of what happens when they close the main window:

- Upon pressing X, if the 'Leave app running in notification area when application window is closed' is checked, a dialog box will pop up telling the user that the app will still be running even if the window is closed. If 'Don't show again' is checked (it is by default) then the user will not see that message again.
image

- If the above setting is NOT checked, the user will see a dialog box asking them whether they really want to close the app permanently. If they reply 'Yes', the app is quit. If they reply 'No', nothing happens. If the 'Don't ask again' box is check and the user replies yes, they will not see the dialog again.
image

Additionally, the 'Leave app running in notification area when application window is closed' checkbox is now false by default on Linux, so it will need to be enabled before minimizing to tray will work correctly.

I've also mimicked the same behavior on Windows to give users the option to turn that on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30713

#### Release Note
```release-note
Updated Linux closing behaviour to allow the app to close complely when pressing X.
Changed the default setting for 'Leave app running in notification area when application window is closed' on Linux to 'false'
```

### DIFF
--- a/src/common/config/defaultPreferences.ts
+++ b/src/common/config/defaultPreferences.ts
@@ -21,7 +21,7 @@ const defaultPreferences: ConfigV3 = {
     teams: [],
     showTrayIcon: true,
     trayIconTheme: 'use_system',
-    minimizeToTray: true,
+    minimizeToTray: process.platform !== 'linux',
     notifications: {
         flashWindow: 2,
         bounceIcon: true,
--- a/src/common/config/index.ts
+++ b/src/common/config/index.ts
@@ -275,6 +275,12 @@ export class Config extends EventEmitter {
     get lastActiveTeam() {
         return this.combinedData?.lastActiveTeam;
     }
+    get alwaysClose() {
+        return this.combinedData?.alwaysClose;
+    }
+    get alwaysMinimize() {
+        return this.combinedData?.alwaysMinimize;
+    }
 
     // initialization/processing methods
 
--- a/src/main/Validator.ts
+++ b/src/main/Validator.ts
@@ -125,6 +125,8 @@ const configDataSchemaV3 = Joi.object<ConfigV3>({
     darkMode: Joi.boolean().default(false),
     downloadLocation: Joi.string(),
     lastActiveTeam: Joi.number().integer().min(0).default(0),
+    alwaysMinimize: Joi.boolean(),
+    alwaysClose: Joi.boolean(),
 });
 
 // eg. data['community.mattermost.com'] = { data: 'certificate data', issuerName: 'COMODO RSA Domain Validation Secure Server CA'};
--- a/src/main/windows/mainWindow.test.js
+++ b/src/main/windows/mainWindow.test.js
@@ -5,7 +5,7 @@ import fs from 'fs';
 
 import path from 'path';
 
-import {BrowserWindow, screen, app, globalShortcut} from 'electron';
+import {BrowserWindow, screen, app, globalShortcut, dialog} from 'electron';
 
 import {SELECT_NEXT_TAB, SELECT_PREVIOUS_TAB} from 'common/communication';
 import Config from 'common/config';
@@ -24,6 +24,10 @@ jest.mock('electron', () => ({
     app: {
         getPath: jest.fn(),
         hide: jest.fn(),
+        quit: jest.fn(),
+    },
+    dialog: {
+        showMessageBox: jest.fn(),
     },
     BrowserWindow: jest.fn(),
     ipcMain: {
@@ -38,7 +42,9 @@ jest.mock('electron', () => ({
     },
 }));
 
-jest.mock('common/config', () => ({}));
+jest.mock('common/config', () => ({
+    set: jest.fn(),
+}));
 jest.mock('common/utils/util', () => ({
     isVersionGreaterThanOrEqualTo: jest.fn(),
 }));
@@ -236,15 +242,17 @@ describe('main/windows/mainWindow', () => {
             };
             BrowserWindow.mockImplementation(() => window);
             Config.minimizeToTray = true;
+            Config.alwaysMinimize = true;
             createMainWindow({});
             Config.minimizeToTray = false;
+            Config.alwaysMinimize = false;
             Object.defineProperty(process, 'platform', {
                 value: originalPlatform,
             });
             expect(window.hide).toHaveBeenCalled();
         });
 
-        it('should minimize window on close for Linux if app wont quit and config item is not set', () => {
+        it('should close app on close window for Linux if app wont quit and config item is not set', () => {
             const originalPlatform = process.platform;
             Object.defineProperty(process, 'platform', {
                 value: 'linux',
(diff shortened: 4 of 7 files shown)


esethna:
@devinbinnie is there a need for the dialog/user prompt at all? Can't we just default the "Leave app running in notification area when application window is closed" to false on Linux so the app quits with the "x", and let the user change that setting if they so desire? We don't have this prompt on windows so I'm not sure why it might be needed here? 

Also, wondering if the other scenarios of quitting the app were fixed in this PR?
image

devinbinnie:
@esethna I've added the dialog box just as a safeguard for anyone who might not be aware of the settings. It appears that some users were not aware that the app was still running in the background so I wanted to make sure that users were aware almost 100% of the time. It used to be a pretty common practice to show a box like that even on Windows, so I don't really think it's too invasive.

All the cases above are captured either by the dialog box, or will ensure the app is quit.

esethna:
Thanks @devinbinnie, glad we captured all those cases. My opinion would be that we probably don't need the dialog box, or at least in the case of the new default setting (app quits) we don't need the dialog box. @matthewbirtch any thoughts from your end?

devinbinnie:
> Thanks @devinbinnie, glad we captured all those cases. My opinion would be that we probably don't need the dialog boxes, or at least in the case of the new default setting (app quits) we don't need the dialog box. @matthewbirtch any thoughts from your end?

I think that's a good reason for needing the dialog box, since users may not immediately expect the new behaviour. 2/5 personally.

matthewbirtch:
> Thanks @devinbinnie, glad we captured all those cases. My opinion would be that we probably don't need the dialog box, or at least in the case of the new default setting (app quits) we don't need the dialog box. @matthewbirtch any thoughts from your end?

I don't have a strong opinion either way. It's definitely not something as common anymore. Having said that Adobe does do something similar on Mac with their creative cloud since it just sits in the menubar (not in the doc).

image

Out of curiosity, is it only Linux users who have brought this up? Just wondering why we're proposing it only on Linux.

devinbinnie:
> > Thanks @devinbinnie, glad we captured all those cases. My opinion would be that we probably don't need the dialog box, or at least in the case of the new default setting (app quits) we don't need the dialog box. @matthewbirtch any thoughts from your end?
> 
> I don't have a strong opinion either way. It's definitely not something as common anymore. Having said that Adobe does do something similar on Mac with their creative cloud since it just sits in the menubar (not in the doc).
> 
> 
> 
> Out of curiosity, is it only Linux users who have brought this up? Just wondering why we're proposing it only on Linux. Is this because it is a deviation from the default Linux behavior for closing apps?

It's on Windows as well.

M-ZubairAhmed (approved):

devinbinnie:
@esethna @matthewbirtch Are we good to merge or did you want additional changes?

matthewbirtch:
@devinbinnie is there anything we can do to clean the dialogs up a bit? They look a bit rough to me. The message content in the dialog looks to be too close to the top and needs some padding. Not sure if this is just using native OS dialogs though.

Also, comparing the two provided screenshots, one of them has a bold 4-line title with a centered line of text under it. Is it possible to avoid the bold title?

image

image

Maybe we can also adjust the messages so we don't need a title and a detail:

`Upon closing the Mattermost window, it will continue to run in the system tray. This can be disabled in the Settings.`

`Are you sure you would like to close Mattermost? You will no longer receive notifications for messages. If you wish to minimize to the system tray instead, you can enable this in the Settings.`

devinbinnie:
> @devinbinnie is there anything we can do to clean the dialogs up a bit? They look a bit rough to me. The message content in the dialog looks to be too close to the top and needs some padding. Not sure if this is just using native OS dialogs though.
> 
> Also, comparing the two provided screenshots, one of them has a bold 4-line title with a centered line of text under it. Is it possible to avoid the bold title?
> 
> image
> 
> image
> 
> Maybe we can also adjust the messages so we don't need a title and a detail:
> 
> `Upon closing the Mattermost window, it will continue to run in the system tray. This can be disabled in the Settings.`
> 
> `Are you sure you would like to close Mattermost? You will no longer receive notifications for messages. If you wish to minimize to the system tray instead, you can enable this in the Settings.`

@matthewbirtch The look and feel of the dialogs I can't do anything about, those are native OS dialogs for Kubuntu, so they all look like that :P

I can, however, update the messages.

esethna:
@devinbinnie just to confirm, what OS's will see this?

+1 to avoiding the bold title or if we are stuck with native styling, can we rejigging the text in the various fields/leave some fields blank so the bolding doesn't occur? 

For the first one, let's update text to: "Mattermost will continue to run in the system tray. This can be disabled in the Settings."

For the second one: "Are you sure you want to quit? You will no longer receive notifications for messages. If you want to leave Mattermost running system tray, you can enable this in the Settings." and then get rid of the "Are you sure you would like to close Mattermost" line

devinbinnie:
> > Thing is that if we change it on one OS, another one could have a different styling. I think we should keep the positioning of message and description as they are to be in line with what Electron expect. Each Linux WM is going to look different, so I don't want to change things to satisfy one of them. 3/5 on this.
> 
> That's fair @devinbinnie. Can you share an updated screenshot of Windows for UX and PM review?

I'll generate some builds so you can play with the functionality and see how it looks on your local machine.

esethna (changes requested):
Thanks Devin. 

1. It looks like the default on Windows is to quit on the close button? I don't think we should change that default on Windows, just on Linux.

image

2. The checkbox above ^ is greyed out (in a disabled state) so I can't click it

3. Also notice some UI oddities:
- check mark is squiggly
- "Close Application" header is not aligned (could just say "Mattermost")
- Don't ask again text is not aligned with checkbox

image

devinbinnie:
> 1. It looks like the default on Windows is to quit on the close button? I don't think we should change that default on Windows, just on Linux.
> 2. The checkbox above ^ is greyed out (in a disabled state) so I can't click it

This is definitely not the default, I just blew out my configuration and it correctly made sure the checkbox wasn't checked.

> 3. Also notice some UI oddities:
> 
> * check mark is squiggly
> * "Close Application" header is not aligned (could just say "Mattermost")
> * Don't ask again text is not aligned with checkbox
> 
> image

Unfortunately these are all Electron issues. I can change the header to Mattermost but everything else is managed using Electron's `dialog` library, which I guess they haven't tested thoroughly on Windows 11 yet.

esethna:
@devinbinnie I'm still seeing the issue where the app doesn't seem to be running in the taskbar if I close it with the "x" (even though it's running in the tray), see Slack as an example: 
image

Did we have a separate ticket for this? ^

Looks good otherwise ✅

PR #2023, opened 2022-03-23
[MM-42538] Submit nightly builds to TestFlight for macOS

devinbinnie:
#### Summary
This PR contains changes in order to have the macOS app build for the App Store.
Currently the workflow will upload builds to TestFlight on a nightly basis. Additional changes will be required in order to get RC and Release builds made.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42538

```release-note
NONE
```


jupenur:
@devinbinnie I've added @spirosoik and @pfltdv as reviewers. We talked about this a little bit with Spiros and this seems like another thing that should go straight into GitLab instead of CircleCI. Now that we have Mustafa (👋), the team should have enough bandwidth to help you make that happen.

CircleCI is going away so we don't want to keep adding new things there 🙂

scripts/notarize.js
```diff
@@ -9,7 +9,8 @@ const config = require('../electron-builder.json');
 
 exports.default = async function notarizing(context) {
     const {electronPlatformName, appOutDir} = context;
-    if (electronPlatformName !== 'darwin' || process.platform !== 'darwin') {
+
+    if (!(['darwin', 'mas'].includes(electronPlatformName) && process.platform === 'darwin')) {
```
jupenur, line 13: Is this right? Does mas need notarization? Is there a benefit to notarizing the mas build?

electron-builder.json
```diff
@@ -93,6 +101,19 @@
       "LSFileQuarantineEnabled": true
   }
   },
+  "mas": {
+    "hardenedRuntime": false,
```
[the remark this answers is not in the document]
devinbinnie: The app crashes on opening when this is enabled. I think it has to do with accessing other files or something?
It might just not work for Electron at all right now.

scripts/notarize.js
devinbinnie: We originally thought it did but it actually doesn't. I'll revert this change.

hmhealey (approved):
Looks good to me from what I can tell, but I'm definitely not super familiar with this area

electron-builder.json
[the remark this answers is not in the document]
Willyfrog: that's weird, as it works with self-publishing. I'd suggest to at least add a ticket to investigate further at some point.

Willyfrog (approved):

fastlane/Fastfile
```diff
@@ -0,0 +1,35 @@
+fastlane_version '2.71.0'
+fastlane_require 'aws-sdk-s3'
+fastlane_require 'erb'
+fastlane_require 'json'
+fastlane_require 'pathname'
+
+lane :publish_test do |options|
+  api_key = ''
+  unless ENV['IOS_API_KEY_ID'].nil? || ENV['IOS_API_KEY_ID'].empty? ||
+    ENV['IOS_API_ISSUER_ID'].nil? || ENV['IOS_API_ISSUER_ID'].empty? ||
```
pfltdv, line 10: Is it possible to create releases with those environment variables ? 

By changing steps of circle-ci pipeline with those environment variables, it is possible to create release.

electron-builder.json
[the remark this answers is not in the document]
jupenur: I agree with Guillermo. We *should* be able to enable this and investigating why that's not the case is important, but I'd be fine with shipping with it disabled for now.

src/main/app/initialize.ts
```diff
@@ -198,10 +206,15 @@ function initializeBeforeAppReady() {
...
 
     // If there is already an instance, quit this one
-    const gotTheLock = app.requestSingleInstanceLock();
-    if (!gotTheLock) {
-        app.exit();
-        global.willAppQuit = true;
+    // eslint-disable-next-line no-undef
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    if (!__IS_MAC_APP_STORE__) {
```
jupenur, line 212: I assume this is because of https://github.com/electron/electron/issues/15958 but I don't see `LSMultipleInstancesProhibited` anywhere, so how do we handle multiple instances?

devinbinnie: Ahh, that's probably something I need to add. I will add that to `electron-builder.json`

fastlane/Fastfile
devinbinnie: I'm not sure what you mean, this is a custom script for Fastlane (was provided by @enahum) that uses those variables to upload to the App Store.

What needs to be changed here?

src/main/app/initialize.ts
devinbinnie: So I added that line and tested, but it doesn't seem to work as we want it to. You can still execute `./Mattermost.app/Contents/MacOS/Mattermost` from the command line and you'll get a second instance.

I'm not sure what a good way to handle this is.

fastlane/Fastfile
enahum: You can create releases all you want, but to publish on TestFlight and subsequently on the Appstore you'll need to set the API_KEYS to be able to login to appleconnect.

What I would say here is that probably the name of the variables should be something like `MACOS_API_KEY` or `APPLE_API_KEY` instead of `IOS...`

src/main/app/initialize.ts
jupenur: Oh yeah definitely you'll still get another instance if you launch the Electron binary from the command line. That's because it'll be a child process of the Terminal instance. But going through launchd using `open ./Mattermost.app` should behave more nicely.

devinbinnie: Looks like it doesn't actually matter if the line is there or not, launchd will always check if an instance is there first. I think that's something handled by the Mac App Store APIs, I feel like I read that somewhere at some point.

If that case is covered, are we okay to proceed? Normally the `singleInstanceLock` call would stop the command line binary from generating a second instance as well, but not sure if there's any concerns around that.

jupenur: Whoops actually `LSMultipleInstancesProhibited` is entirely different. From Apple docs: "A Boolean value indicating whether more than one user can launch the app simultaneously." We don't want that! I just saw it mentioned as a workaround in the Electron ticket, but didn't look into it further, sorry! 🙄

jupenur (approved):

pfltdv (approved):

PR #2043, opened 2022-04-07
[MM-39793][MM-T2633] E2E Test: Use back button to return to the desktop app login - and fixes

devinbinnie:
#### Summary
E2E test for the Back Button. Also some fixes since the Back button wasn't working as expected by the test case.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39793
Closes https://github.com/mattermost/desktop/issues/1912

```release-note
NONE
```

### DIFF
--- a/e2e/specs/mattermost/back_button.test.js
+++ b/e2e/specs/mattermost/back_button.test.js
@@ -1,0 +1,96 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+'use strict';
+
+const fs = require('fs');
+
+const robot = require('robotjs');
+
+const env = require('../../modules/environment');
+const {asyncSleep} = require('../../modules/utils');
+
+describe('Menu/window_menu', function desc() {
+    this.timeout(30000);
+
+    const config = {
+        ...env.demoConfig,
+        teams: [
+            {
+                name: 'community',
+                url: 'https://community.mattermost.com',
+                order: 0,
+                tabs: [
+                    {
+                        name: 'TAB_MESSAGING',
+                        order: 0,
+                        isOpen: true,
+                    },
+                    {
+                        name: 'TAB_FOCALBOARD',
+                        order: 1,
+                        isOpen: true,
+                    },
+                    {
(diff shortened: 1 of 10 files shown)


hmhealey (approved):

e2e/specs/mattermost/back_button.test.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('MM-T2633 Back button should behave as expected', async () => {
+        const mainWindow = this.app.windows().find((window) => window.url().includes('index'));
+        const loadingScreen = this.app.windows().find((window) => window.url().includes('loadingScreen'));
+        await loadingScreen.waitForSelector('.LoadingScreen', {state: 'hidden'});
+        const firstServer = this.serverMap[`${config.teams[0].name}___TAB_MESSAGING`].win;
+
+        await firstServer.click('a:has-text("OneLogin")');
+        let backButton = await mainWindow.waitForSelector('button:has-text("Back")');
```
agarciamontoro, line 65: Nit: should all the `let` in this file be `const`?

e2e/specs/mattermost/back_button.test.js
```diff
@@ -0,0 +1,94 @@
...
+        this.serverMap = await env.getServerMap(this.app);
+    });
+
+    afterEach(async () => {
+        if (this.app) {
+            await this.app.close();
+        }
+    });
+
+    it('MM-T2633 Back button should behave as expected', async () => {
```
agarciamontoro, line 58: Nit: this title is a bit vague, can we clarify it a bit more?

agarciamontoro (approved):
Looks good to me!

Disclaimer: this is the first time I read a Playwright test (it's neat, by the way!) :)

e2e/specs/mattermost/back_button.test.js
devinbinnie: Most of the variables are reused, so that's why we use `let`

e2e/specs/mattermost/back_button.test.js
devinbinnie: It kinda covers a large number of cases, which is why it's purposely vague.

PR #2049, opened 2022-04-12
[MM-31547] Stop users from being able to enter the same server name or URL twice

devinbinnie:
#### Summary
Previously, users were able to add duplicate servers, containing either the same name or URL as another server. This causes a few issues when a user does it, for example, strange behaviours in the dropdown menu.

This PR restricts users from being able to add the same exact server name or URL twice, thus avoiding the duplication issues.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31547

```release-note
Fixed an issue where users could add the same server name or URL twice
```


hmhealey:
Do we have to do anything to account for users that might already be in this situation, or is just asking them to remove the duplicate server enough?

Willyfrog (approved):

PR #2053, opened 2022-04-14
Fix issue where opening a new view causes the original view still to push

devinbinnie:
#### Summary
When using the Global Header (or some other non `target=_blank` link) to open a new tab, the original tab would also navigate to the new location. This was due to the check for `isLoggedIn` on the new view, which of course won't be set until the view has finished loading.

This PR stops the rest of the `handleBrowserHistoryPush` flow from happening after the tab is opened, as when we open a tab we always go directly to where the push link told us to go.

```release-note
Fixed an issue where opening a new view causes the original view to go to the requested link as well
```


harshilsharma63:
@devinbinnie where can I find the build for this branch? I checked CircleCI build artifacts but those are empty. How can I test this one?

devinbinnie:
> @devinbinnie where can I find the build for this branch? I checked CircleCI build artifacts but those are empty. How can I test this one?

@harshilsharma63 I just added the 'Build Apps for PR' label, it should be built within a half hour.

hmhealey (approved):

PR #2060, opened 2022-04-19
Update NOTICE.txt

amyblais:
Based on https://github.com/mattermost/desktop/blob/master/package.json.

This can be included in the next RC later when it's cut (we don't need to cut an RC just for this PR).

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -8,6 +8,19 @@ This document includes a list of open source components used in Mattermost Deskt
 
 --------
 
+## @mattermost/compass-icons
+
+This product contains '@mattermost/compass-icons' by Mattermost.
+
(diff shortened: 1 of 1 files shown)


NOTICE.txt
```diff
@@ -551,46 +658,6 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 
 ---
 
-## wix
```
devinbinnie, line 554: We still use WiX for creating the MSI build. Can't remove this.

devinbinnie (changes requested):

devinbinnie (approved):

NOTICE.txt
amyblais: I added it back. It's not listed here though: https://github.com/mattermost/desktop/blob/master/package.json#L173.

PR #2070, opened 2022-04-22
MAS: remove default download location

devinbinnie:
#### Summary
As per MAS guidelines, we're not allowed to prescribe the default download location for users, so we've removed it for that build.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43657

```release-note
NONE
```

### DIFF
--- a/src/common/config/defaultPreferences.ts
+++ b/src/common/config/defaultPreferences.ts
@@ -12,12 +12,14 @@ import os from 'os';
 
 import {ConfigV3} from 'types/config';
 
-export const getDefaultDownloadLocation = (): string => {
+export const getDefaultDownloadLocation = (): string | undefined => {
     // eslint-disable-next-line no-undef
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    const homePath = __IS_MAC_APP_STORE__ ? path.join(os.homedir(), '../../../../') : os.homedir();
-    return path.join(homePath, 'Downloads');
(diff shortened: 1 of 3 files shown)


src/common/config/defaultPreferences.ts
```diff
@@ -12,12 +12,14 @@ import os from 'os';
 
 import {ConfigV3} from 'types/config';
 
-export const getDefaultDownloadLocation = (): string => {
+export const getDefaultDownloadLocation = (): string | undefined => {
     // eslint-disable-next-line no-undef
```
hmhealey, line 16: Are all these comments still relevant? This one in particular doesn't look needed

src/common/config/defaultPreferences.ts
```diff
@@ -12,12 +12,14 @@ import os from 'os';
...
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    const homePath = __IS_MAC_APP_STORE__ ? path.join(os.homedir(), '../../../../') : os.homedir();
-    return path.join(homePath, 'Downloads');
+    if (__IS_MAC_APP_STORE__) {
+        return undefined;
```
hmhealey, line 20: 0/5, but would it help anything if we used the empty string instead of undefined here? I'm still on a mission to get JS to work like Go where the zero value for a string is the empty string instead of undefined 😛

hmhealey (approved):

PR #2075, opened 2022-04-25
Fix release markdown

devinbinnie:
#### Summary
@JulienTant pointed out a couple issues with the release markdown, so I've made the changes here
- Fix the renaming of `arm64` to `m1` for macOS builds
- Removed Windows ZIP files from release (since we don't build them anymore)
- Renamed RPM build for `i386` to `i686` (those poor 486 users :P)

```release-note
NONE
```


JulienTant (approved):
Looks good!

PR #2080, opened 2022-04-26
[MM-43601] Switch upgrade button to an actual button and moved onClick event

devinbinnie:
#### Summary
The upgrade button was only clickable from the icon and not the entire highlighted area. We had the click event on a `div` inside of a `span`. This PR changes to use a `button` with an `onClick` event.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43601

```release-note
NONE
```


M-ZubairAhmed (approved):

agarciamontoro (approved):
Good ol' semantic HTML :) Thanks, @devinbinnie!

PR #2094, opened 2022-05-10
only refresh view when the server URL changes (MM-34565) (#2082)

vaaas:
#### Summary

Create views of current and incoming tabs indexed by unique [URL, TABTYPE] tuples, and diffing them. Tuples that are identical are recycled, merely porting the new server name over.

#### Notes

The root cause is that the tabs are indexed by a key in the format:

```
SERVERNAME___TABTYPE
```

Obviously, when the server's name changes, it necessarily results in different keys.

The solution is indexing by the server's HREF and the tab type, which should be unique. A proper solution would involve a refactor of ViewManager that is out of scope for fixing this bug, so while I explored that solution, I decided on a more minimal and specific fix, because this pull request is already large enough.

Immutable tuples are introduced as a dependency. Bloomberg's implementation is provided, which is better than a naive implementation using nested maps, as the combination of WeakRef and FinalizationRegistry prevents unnecessary memory use. This bumps the node requirement to 14.6.0, the oldest version supporting these APIs. We seem to ship node 16+ based on our electron dependency. I couldn't find Typescript definition files for tuples and records, so I provided my own.

Some old tests have been modified, mostly to reflect the addition of the `urlTypeTuple` getters in `MattermostView` and `TabView`.

I suspect simplification of ViewManager is possible by leveraging a reactivity library, possibly React's, if it works similarly to Vue's. Touching up WindowManager, TabView, MattermostView, etc. is also a good idea, as they like to fiddle with their dependencies' internal state, which breaks encapsulation.

#### Ticket Link

Fixes https://github.com/mattermost/desktop/issues/2082

JIRA https://mattermost.atlassian.net/browse/MM-34565

#### Checklist

#### Device Information

This PR was tested on: Arch Linux, 5.15.37-1-lts

#### Release Note

```release-note
Do not reload tab when a server's name changes, only when the URL changes.
```


vaaas:
I had some time to clean up this pull request. Original description changed to match its current state.

src/common/tabs/BaseTabView.ts
```diff
@@ -14,6 +16,9 @@ export default abstract class BaseTabView implements TabView {
     get name(): string {
         return getTabViewName(this.server.name, this.type);
     }
+    get tuple(): TabTuple {
```
devinbinnie, line 19: Can we update the field name here? The name `tuple` doesn't necessarily describe what it is. Maybe something like `urlTypeTuple`?

src/main/views/viewManager.ts
```diff
@@ -112,44 +122,73 @@ export class ViewManager {
...
+
+        const current: Map<TabTuple, MattermostView> = new Map();
+        for (const x of this.views.values()) {
+            current.set(x.tuple, x);
+        }
+
+        const views: Map<TabTuple, MattermostView> = new Map();
+        const closed: Map<TabTuple, {srv: MattermostServer; tab: Tab; name: string}> = new Map();
+
+        const sortedTabs = configServers.flatMap((x) => x.tabs.sort(by((x) => x.order)).map((t) => duad(x, t)));
```
devinbinnie, line 140: We can simplify and standardize this without needing these functional helper functions, see my suggestion:
```suggestion
        const sortedTabs = configServers.flatMap((x) => x.tabs.sort((a, b) => a.order - b.order).map((t) => [x, t]);
```

devinbinnie (changes requested):
Thanks for the PR @vaaas, I've added some comments for you to review.

devinbinnie (approved):
Thanks @vaaas this is looking good now :)

package.json
```diff
@@ -115,6 +115,7 @@
     "@babel/preset-env": "7.16.11",
     "@babel/preset-react": "7.16.7",
     "@babel/register": "7.17.7",
+    "@bloomberg/record-tuple-polyfill": "^0.0.4",
```
hmhealey, line 118: Do we really need a polyfill to add this behaviour? If we want to use tuple types here, I think I'd rather just use a plain array and trust that we don't modify the result rather than introduce an experimental polyfill for it. We're also having to work around the fact that it doesn't natively support TypeScript in quite a few places, and it seems like TypeScript may not natively support it for some time even once the proposal reaches stage 3.

All-in-all, I'd rather we not use this until it becomes more standardized. The rest of the logic of returning `[href, type]` and passing that around still looks fine to me though as long as we're careful to use a shallow equality check instead of `===`

src/main/views/viewManager.ts
```diff
@@ -112,44 +121,76 @@ export class ViewManager {
...
+     * Servers or tabs have been added or edited. We need to
+     * close, open, or reload tabs, taking care to reuse tabs and
+     * preserve focus on the currently selected tab. */
     reloadConfiguration = (configServers: TeamWithTabs[]) => {
-        this.configServers = configServers.concat();
-        const oldviews = this.views;
+        const focusedTuple: TabTuple | undefined = this.views.get(this.currentView as string)?.urlTypeTuple;
+
+        const current: Map<TabTuple, MattermostView> = new Map();
+        for (const x of this.views.values()) {
```
hmhealey, line 132: Could use more self-explanatory names for loop variables like this?

src/main/views/viewManager.ts
```diff
@@ -112,44 +121,76 @@ export class ViewManager {
...
+            current.set(x.urlTypeTuple, x);
+        }
+
+        const views: Map<TabTuple, MattermostView> = new Map();
+        const closed: Map<TabTuple, {srv: MattermostServer; tab: Tab; name: string}> = new Map();
+
+        const sortedTabs = configServers.flatMap((x) =>
+            x.tabs.
+            sort((a, b) => a.order - b.order).
```
hmhealey, line 141: We need to copy this array since JS's `sort` method mutates the array it's working on. Just doing `[...x.tabs].sort(...)` should be enough to fix that though

src/main/views/viewManager.ts
```diff
@@ -112,44 +121,76 @@ export class ViewManager {
...
+
+        const current: Map<TabTuple, MattermostView> = new Map();
+        for (const x of this.views.values()) {
+            current.set(x.urlTypeTuple, x);
+        }
+
+        const views: Map<TabTuple, MattermostView> = new Map();
+        const closed: Map<TabTuple, {srv: MattermostServer; tab: Tab; name: string}> = new Map();
+
+        const sortedTabs = configServers.flatMap((x) =>
```
hmhealey, line 139: I'm having a tough time figuring out what we're trying to do here. This is making an array of tuples for each server's tabs, is that right? It seems a bit odd that we'd be both combining them into a single array (presumably for easier iteration) but then sorting them when they won't be displayed that way.

src/main/views/viewManager.ts
```diff
@@ -80,25 +81,33 @@ export class ViewManager {
...
     }
 
-    loadView = (srv: MattermostServer, serverInfo: ServerInfo, tab: Tab, url?: string) => {
+    makeView = (srv: MattermostServer, serverInfo: ServerInfo, tab: Tab, url?: string): MattermostView => {
         const tabView = getServerView(srv, tab);
-        if (!tab.isOpen) {
-            this.closedViews.set(tabView.name, {srv, tab});
-            return;
-        }
         if (this.closedViews.has(tabView.name)) {
```
hmhealey, line 86: Should this still be happening in `makeView` when it's a destructive action? Since you split up the old `loadView`, shouldn't this still happen in it or in `addView`? The lifecycle of a `MattermostView` is a bit confusing to me

hmhealey (changes requested):

src/main/views/viewManager.ts
vaaas: Yes, that's right. I'm sorting them because the original implementation did as well, at line 119. I am not sure if it serves any purpose, so I elected to trust the original implementation had something in mind that I'm not aware of.

src/main/views/viewManager.ts
vaaas: Agreed, both in the solution and in the confusion. I have elected to add this behaviour into `addView`, where it is most relevant.

package.json
hmhealey: Oh, I see. I didn't realize that Maps needed the keys to be checkable by `===` to work. I knew one of the benefits of Map is that it supports non-string keys, but I don't think I've ever ended up actually using that.

In that case, I think these are fine even if they don't have proper support yet. I've definitely wanted something like this in the past and had to settle for encoding things as strings, but this is definitely nicer than that.

PR #2096, opened 2022-05-10
[MM-44066] Truncate server names in the system tray to 50 characters

devinbinnie:
#### Summary
The server name was overflowing on the tray menu, causing bad UX (specifically on Ubuntu). This PR truncates the server name to 50 characters so that it doesn't cause any issues.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44066

```release-note
Fixed bad UX for the system tray menu with long server names
```


tweekie2 (approved):

hmhealey (approved):
Is there not a better way to truncate this? It might not cover all cases like if someone names a team `wwwwwwwwwwwwwwww` etc. It looks fine though even if it doesn't cover everything

PR #2106, opened 2022-05-14
automatically add protocol to Server URL (#2083)

### LINKED ISSUE #2083: Automatically add https:// or http:// to Server URL
It would help UX a lot when mattermost-desktop would probe the domain automatically if it supports https:// and adds it to the URL and only use http:// if it fails.
image

vaaas:
#### Summary

When adding and editing a server, automatically complete the protocol (http, https) if the user only types a domain name (www.example.com).

Wait for the user to stop typing, then ping the server with both protocols. If https resolves, use https. If only http resolves, use http. If neither responds, return an error.

#### Notes

I am unsure how or if we want to display the error, so I don't do anything with it besides setting it.

#### Ticket Link

Fixes #2083 

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Debian sid, 5.17.0-1-amd64

#### Release Note

```release-note
automatically prefix the server URL with http or https when the user enters it
```


src/common/utils/requests.ts
```diff
@@ -0,0 +1,17 @@
...
+// See LICENSE.txt for license information.
+
+import * as http from 'http';
+import * as https from 'https';
+
+type Method = 'GET' | 'POST' | 'DELETE' | 'PUT' | 'OPTIONS' | 'HEAD';
+
+export function ping(x: URL, method: Method = 'GET'): Promise<http.ServerResponse> {
+    const f = x.protocol === 'http:' ? http.request : https.request;
+    return new Promise((yes, no) => {
```
devinbinnie, line 11: Instead of returning a Promise, we could mark this function as `async`. It's a bit cleaner.

src/main/preload/modalPreload.js
```diff
@@ -4,7 +4,11 @@
 
 'use strict';
 
-import {ipcRenderer} from 'electron';
+import {ipcRenderer, contextBridge} from 'electron';
+
+contextBridge.exposeInMainWorld('ipcRenderer', {
```
devinbinnie, line 9: We don't need to expose this in the main world, see the code below. We can have specific messages sent by the front end that will do specific functions. This way we don't expose the `ipcRenderer` module to the `window` where it could be accessed via the Developer Console for example, making the approach more secure.

src/renderer/components/NewTeamModal.tsx
```diff
@@ -115,9 +123,17 @@ export default class NewTeamModal extends React.PureComponent<Props, State> {
     }
 
     handleTeamUrlChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-        this.setState({
-            teamUrl: e.target.value,
-        });
+        const teamUrl = e.target.value;
+        this.setState({teamUrl});
+        if (!teamUrl.startsWith('http') && teamUrl.length >= 4) {
```
devinbinnie, line 128: Technically `a.b` could be a valid domain, so we may not want to bother enforcing the length requirement.

src/renderer/components/NewTeamModal.tsx
```diff
@@ -115,9 +123,17 @@ export default class NewTeamModal extends React.PureComponent<Props, State> {
...
+            this.addProtocolToUrlThrottled(teamUrl);
+        }
+    }
+
+    addProtocolToUrl = (teamUrl: string): void => {
```
devinbinnie, line 133: I think we can avoid having this displayed to the user. I understand trying to 'teach' the user how to correctly form a URL, but this isn't something we have to be responsible for. For example, most browsers don't do this.

I'm thinking instead we can, on save, ping the URLs as we're doing here and adding the `http` after the fact. It will appear in the config and when you go to edit the URL later, but I don't think adding it while the user is typing is the correct solution here.

devinbinnie (changes requested):
Comment to request changes noted above :P

src/common/utils/requests.ts
vaaas: Is there any point in that? Since node only provides event handlers for the request and the response objects, we'd have to do manual Promise resolution anyway, and end up generating one more promise than necessary from V8's point of view.

vaaas: I don't expect this function to be called to the extent that the performance gains matters, so if you think it looks cleaner, the few bytes of extra memory are worth the tradeoff.

src/renderer/components/NewTeamModal.tsx
Willyfrog: Agree on this one, also if the user types it without protocol, it then adds the protocol and the user goes back and changes the domain, the result might no longer be valid.

This is meant as a helper for users that are not super technical and might not know about http/s, so it should happen at the end of the process without them having to care.

src/renderer/components/NewTeamModal.tsx
```diff
@@ -27,11 +29,13 @@ type State = {
     teamIndex?: number;
     teamOrder: number;
     saveStarted: boolean;
+    errorMessage: string;
```
Willyfrog, line 32: is this being used somewhere? or are we only storing the value?

Willyfrog (changes requested):
Hello and thanks for the PR!

In general looks good, but from the user point of view, I'd not fix the url while typing as that might be confusing and error prone.

There are more comments but they are mostly nitpicking and curiosity, the only blocking one is the above.

src/common/utils/requests.ts
vaaas: I opted for adding async.

src/renderer/components/NewTeamModal.tsx
```diff
@@ -27,11 +29,13 @@ type State = {
...
 }
 
 export default class NewTeamModal extends React.PureComponent<Props, State> {
     wasShown?: boolean;
     teamNameInputRef?: HTMLInputElement;
+    addProtocolToUrlThrottled: (x: any) => void;
```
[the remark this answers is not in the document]
vaaas: This method no longer exists after refactoring.

src/renderer/components/NewTeamModal.tsx
vaaas: I decided to do the ping when `save()` is called. Is that okay, or is there something better?

src/renderer/components/NewTeamModal.tsx
vaaas: In the previous implementation, I considered perhaps displaying some alert about the server not existing. Since now I only ping when the form is saved, this property is pointless, so I removed it. I simply log the message for debugging purposes.

src/common/utils/timing.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+const NoValue = Symbol('represents absence of any value');
+
+/** throttles a function, so that it is called at most every t milliseconds */
+export function throttle<T>(f: (x: T) => any, t: number): (x: T) => any {
```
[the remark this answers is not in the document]
vaaas: These no longer exist in the current implementation, but to answer the question, I prefer implementing function if they are simple enough over introducing external dependencies.

Willyfrog (approved):

PR #2111, opened 2022-05-17
Update Display Name for MSI

JtheBAB:
#### Summary
Update Display Name to Mattermost Desktop to reflect the real name. Sorry don't know where to change the exe builder :/

```release-note
NONE
```

### DIFF
--- a/scripts/msi_installer.wxs
+++ b/scripts/msi_installer.wxs
@@ -35,7 +35,7 @@
   - there are any changes in the feature hierarchy (child
     feature moving out of a parent, or a parent feature getting a new child).
   -->
-  <Product Id="*" UpgradeCode="8523DAF0-699D-4CC7-9A65-C5E696A9DE6D" Name="Mattermost" Manufacturer="Mattermost, Inc." Language="1033" Codepage="1252" Version="0.0.0">
+  <Product Id="*" UpgradeCode="8523DAF0-699D-4CC7-9A65-C5E696A9DE6D" Name="Mattermost Desktop" Manufacturer="Mattermost, Inc." Language="1033" Codepage="1252" Version="0.0.0">
     <!--
     perMachine installs the app system wide (ALLUSERS is set to 1)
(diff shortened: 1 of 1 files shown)


devinbinnie:
Thanks @JtheBAB, I'm not sure this one is necessary though. I think the app name (at least on Windows) is still Mattermost.

JtheBAB:
I find it just strange as everywhere it is Mattermost Desktop. But there not. But up to you guys :)

wget (approved):
I'm upvoting this I may work on a server msi installer later on an this would help avoid confusion between the server and desktop parts in display names from the app uninstall windows feature. I'm also preparing the work towards WiX v4.

wget:
@devinbinnie Feel free to give approval so that I can merge :)

devinbinnie:
@wget Before I approve, and forgive my ignorance because I'm not super familiar with MSI building myself other than what I've had to work with, but is there any risk in changing the name here? Does that change the name of any shortcuts? Would notifications be affected (since Windows bases their existence off of some ID)?

wget:
@devinbinnie I don't pretend to know everything about msi myself either, but what matters most (:yum: ) is the `Product Id` and `UpgradeCode` attribute, not the `name` :) In the past, that name was usually changed depending on the version used to make it look different in the Add/Remove program control panel element.
So yes, the Name doesn't matter here. That name was important for Chocolatey for example when we hadn't any msi installer yet, chcolatey relied on it to grab the Uninstall string from the registry. Now with msi installer, they use the ID.

I checked by myself and an upgrade process is working flawlessly between a 5.2.2 and a 5.3.0-develop (with the new name).

Before:
Screenshot_20230120_204333
After:
Screenshot_20230120_212156

In the installer GUI (the name is also changed everywhere else, but I'm not gonna do screenshots for all of them :P ):
Screenshot_20230120_212126

devinbinnie (approved):

devinbinnie:
Seems like this is something everyone wants, so I shall approve and merge. Thanks @JtheBAB and @wget :)

Issue #2113, opened 2022-05-17
Assets missing on GitHub Release post


bbodenmiller:
https://github.com/mattermost/desktop/releases/tag/v5.1.0 is missing release assets whereas https://github.com/mattermost/desktop/releases/tag/v5.0.4 had them. This is required because this is the only approved download source for some users for the desktop client.

bbodenmiller:
Might be related to https://github.com/mattermost/desktop/pull/1810 or https://github.com/mattermost/desktop/pull/2059

@elyscape @Willyfrog @devinbinnie

devinbinnie:
@bbodenmiller Can you provide reasoning why users are not able to download from our website? I would like to understand the requirements better. Thanks!

elyscape:
This happens because `ghr`, the tool used to [create the GitHub release][upload], only looks at [the immediate descendents of provided paths][ghr]. The contents of `ghr-dist/` are assembled with this in mind.

The various build jobs place the release assets into the build workspace under the paths [`linux/`][linux], [`macos-release/`][macos-release], and [`win-release/`][win-release]. In the release job, the build workspace is attached to `build/`, after which `build/{macos-release,win-release,linux}/*` [is copied into `ghr-dist/`][copy].

The problem is that, as part of #1714, the build jobs were modified to place the release assets into versioned subdirectories of `macos-release/`, `win-release/`, and `linux/`. As a result, when `ghr` begins execution, the contents of `ghr-dist/` look something like this:
```
ghr-dist/
├── 5.1.0
│   ├── mattermost-desktop-5.1.0-linux-i386.AppImage
│   ├── mattermost-desktop-5.1.0-linux-i686.rpm
│   ├── mattermost-desktop-5.1.0-linux-ia32.tar.gz
│   ├── mattermost-desktop-5.1.0-linux-x64.tar.gz
│   ├── mattermost-desktop-5.1.0-linux-x86_64.AppImage
│   ├── mattermost-desktop-5.1.0-linux-x86_64.rpm
│   ├── mattermost-desktop-5.1.0-mac-m1.dmg
│   ├── mattermost-desktop-5.1.0-mac-universal.dmg
│   ├── mattermost-desktop-5.1.0-mac-x64.dmg
│   ├── mattermost-desktop-5.1.0-x64.msi
│   ├── mattermost-desktop-5.1.0-x86.msi
│   ├── mattermost-desktop-setup-5.1.0-win.exe
[...]

PR #2114, opened 2022-05-18
Fix GitHub release asset upload

elyscape:
`ghr` only looks at immediate descendents of the path it is provided, so we need to ensure that `ghr-dist/` is flattened.

Fixes: #2113

```release-note
NONE
```


devinbinnie:
@amyblais Will cherry-pick this over to v5.1 for any future dot releases.

PR #2126, opened 2022-05-25
[MM-44541] Fix issue with changing servers and tabs

devinbinnie:
#### Summary
A recent PR caused an issue with changing servers and tabs, effectively making it impossible to do so. This PR makes a few fixes to ensure that we can still do that.

Reported here: https://community-daily.mattermost.com/core/pl/98do1n9ey38k9cp83owxqspbwe

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44541

```release-note
NONE
```

### DIFF
--- a/src/main/views/viewManager.ts
+++ b/src/main/views/viewManager.ts
@@ -150,8 +150,7 @@ export class ViewManager {
             if (!tab.isOpen) {
                 closed.set(tabTuple, {srv, tab, name: view.name});
             } else if (recycle) {
-                recycle.serverInfo = info;
-                recycle.tab.server = srv;
+                recycle.updateServerInfo(srv);
                 views.set(tabTuple, recycle);
             } else {
                 views.set(tabTuple, this.makeView(srv, info, tab, tabTuple[0]));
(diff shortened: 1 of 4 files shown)


M-ZubairAhmed (approved):

src/main/views/viewManager.ts
```diff
@@ -189,8 +188,11 @@ export class ViewManager {
         // show the focused tab (or initial)
         if (focusedTuple && views.has(focusedTuple)) {
             const view = views.get(focusedTuple);
-            this.currentView = view!.name;
-            this.showByName(view!.name);
+            if (view) {
+                this.currentView = view.name;
+                this.showByName(view.name);
+                this.mainWindow.webContents.send(SET_ACTIVE_VIEW, view.tab.server.name, view.tab.type);
```
hmhealey, line 194: Shouldn't this be sent as part of `showByName`? Why's it being sent separately

devinbinnie: `showByName` does not send this if the view is already - active, which is the case when the server name changes but not the URL. This is to ensure that it is sent no matter what. It doesn't do any harm to send it twice.

hmhealey (approved):

Issue #2125, opened 2022-05-25
5.1.0 client can't switch to channels resembling subpath.


sphr:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
If mattermost server is installed in a subpath, like /app then creating a channel that starts with "app" and attempting to switch to it in the client will cause the client to refuse switching. Web and mobile clients can still swap.

**Environment**
- Operating System: Windows 10, MacOS
- Mattermost Desktop App version: 5.1.0 
- Mattermost Server version: 6.4.2 

**Steps to reproduce**
1. Install mattermost in subpath, like /app or /chat
2. create a channel starting with the subpath name, like "app" or "chat", like application or chatting
3. in desktop client, attempt to join the channel.
4. can not join

**Expected behavior**
Client switches to desired channel.

**Observed behavior**
Current "view" flashes and reloads. Log file if set to debug shows url to the channel as missing the /subdir mattermost was installed in as can be noted in [2022-05-25 16:19:55.742] where we expect url to be: 
  url: 'https://example.tld/app/myteam/channels/applikationsstod' 
but it is
  url: 'https://example.tld/app/myteam/channelslikationsstod'

Log:

[2022-05-25 16:19:55.738] [debug] TeamDropdownView.handleClose
[2022-05-25 16:19:55.739] [debug] WwindowManager.handleBrowserHistoryPush {
  viewName: 'MYTEAM___TAB_MESSAGING',
  pathName: '/myteam/channels/applikationsstod'
}
[2022-05-25 16:19:55.740] [debug] EindowManager.handleBrowserHistoryButton MYTEAM___TAB_MESSAGING
[2022-05-25 16:19:55.742] [debug] webContentEvents.did-start-navigation {
  webContentsId: 3,
  url: 'https://example.tld/app/myteam/channelslikationsstod'
}
[2022-05-25 16:19:55.788] [debug] webContentEvents.did-start-navigation {
  webContentsId: 3,
[...]

PR #2129, opened 2022-05-25
[MM-44607] Only cleanse subpath if it occurs at the beginning

devinbinnie:
#### Summary
When switching to a channel that matches (or starts with) the name of the server subpath (eg. server url `http://server.com/mattermost` and channel name `mattermosty`), navigation would not happen. This was caused by an issue with the replacement function that removes the subpath when switching tabs.

This PR fixes the function to only remove the subpath if its at the beginning of the path that's being cleansed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44607
Closes https://github.com/mattermost/desktop/issues/2125

```release-note
Fixed an issue where a channel name matching the server subpath would not be navigable
```

### DIFF
--- a/src/common/utils/url.test.js
+++ b/src/common/utils/url.test.js
@@ -303,6 +303,20 @@ describe('common/utils/url', () => {
         });
     });
 
+    describe('cleanPathName', () => {
+        it('should not clean path name if it occurs other than the beginning', () => {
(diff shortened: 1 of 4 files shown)


agarciamontoro (approved):
Looks good :tada:

hmhealey (approved):

PR #2149, opened 2022-06-07
Create configuration folder for notice.txt generation

pfltdv:
#### Summary
We are working to automate Notice.txt file generation and some of the properties need to be managed by developers.

To support that feature, we created a configuration file in the repo under build folder. 

#### Ticket Link
https://mattermost.atlassian.net/browse/DOPS-448

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.config/notice-file/Readme.md
+++ b/.config/notice-file/Readme.md
@@ -1,0 +1,34 @@
+# Notice.txt File Configuration
+
+We are automatically generating Notice.txt by using first-level dependencies of the project. The related pipeline uses `config.yaml` stored in this folder.
+
+
+## Configuration
+
+Sample:
+
+```
+title: "Mattermost Desktop"
+copyright: "© 2016-2017 Mattermost, Inc.  All Rights Reserved.  See LICENSE.txt for license information."
+description: "This document includes a list of open source components used in Mattermost Desktop, including those that have been modified."
+reviewers: 
+  - mattermost/release-managers
+  - mattermost/team-desktop
+search:
+  - "package.json"
+dependencies:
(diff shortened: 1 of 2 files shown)


amyblais (approved):

spirosoik (approved):

devinbinnie (commented):
Does this need to be in the `build` directory?
CircleCI uses the `build` directory to store built artifacts, just not sure if there might be any conflict.

pfltdv:
> Does this need to be in the `build` directory? CircleCI uses the `build` directory to store built artifacts, just not sure if there might be any conflict.

@devinbinnie it is configurable, we can use any folder in the repo, how about using `config` folder or any other folder suitable for the project ?

devinbinnie:
> > Does this need to be in the `build` directory? CircleCI uses the `build` directory to store built artifacts, just not sure if there might be any conflict.
> 
> @devinbinnie it is configurable, we can use any folder in the repo, how about using `config` folder or any other folder suitable for the project ?

That would be great, any folder that isn't currently used would work.

pfltdv:
> > > > Does this need to be in the `build` directory? CircleCI uses the `build` directory to store built artifacts, just not sure if there might be any conflict.
> > > 
> > > 
> > > @devinbinnie it is configurable, we can use any folder in the repo, how about using `config` folder or any other folder suitable for the project ?
> > 
> > 
> > That would be great, any folder that isn't currently used would work.
> 
> Awesome, i will change PR to use `config` folder which is not used atm.

Changed to use `.config` folder and PR is updated.

PR #2151, opened 2022-06-13
Fix MAS/TestFlight nightly build

devinbinnie:
#### Summary
Once a version is released, MAS doesn't allow subsequent releases of the same version to be uploaded to TestFlight, so I've updated the version in package.json to v5.2

```release-note
NONE
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "mattermost-desktop",
   "productName": "Mattermost",
-  "version": "5.1.0-develop.1",
+  "version": "5.2.0-develop.1",
   "description": "Mattermost",
   "main": "dist/index.js",
   "author": "Mattermost, Inc. <feedback@mattermost.com>",


hmhealey:
Since this is still a 5.1 nightly release, should we maybe be updating the build number after the `-` instead? I wouldn't expect us to go to 5.2 until after 5.1 is cut

devinbinnie:
> Since this is still a 5.1 nightly release, should we maybe be updating the build number after the `-` instead? I wouldn't expect us to go to 5.2 until after 5.1 is cut

This is how we've done it in the past/how most other products do things. Yes we haven't officially changed the version number, but technically the next version won't be v5.1 and it's confusing to be shipping out more builds after the "final" version v5.1 release.

On top of all this, as per the description, MAS won't allow a repeat of an older version number, and the build number has no relevance to them so it wouldn't solve the problem.

hmhealey:
Oh, I didn't realize this was a nightly version of what'd eventually become 5.2. I thought we were still waiting to release 5.1. Sorry about delaying this then

hmhealey (approved):

PR #2152, opened 2022-06-13
[MM-44644] Update configured server URL from SiteURL if available

devinbinnie:
#### Summary
Whenever the server is configured not using the `SiteURL` that is configured on the server, a lot of unexpected issues can occur. This most recently manifested itself when we started redirecting `community-daily` to `community`. 

This PR uses the same API the app uses to check for installed plugins and server version to grab the SiteURL and update the configured server URL with it.

This does appear to work with GPO configurations as well!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44644

```release-note
Force servers to update their configuration using the configured SiteURL if available
```

### DIFF
--- a/src/main/app/utils.ts
+++ b/src/main/app/utils.ts
@@ -55,11 +55,21 @@ export function updateServerInfos(teams: TeamWithTabs[]) {
     });
     Promise.all(serverInfos).then((data: Array<RemoteInfo | string | undefined>) => {
         const teams = Config.teams;
-        teams.forEach((team) => openExtraTabs(data, team));
+        teams.forEach((team) => {
+            updateServerURL(data, team);
+            openExtraTabs(data, team);
+        });
         Config.set('teams', teams);
     }).catch((reason: any) => {
         log.error('Error getting server infos', reason);
     });
+}
+
+function updateServerURL(data: Array<RemoteInfo | string | undefined>, team: TeamWithTabs) {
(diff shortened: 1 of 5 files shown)


src/main/app/utils.ts
```diff
@@ -55,13 +55,23 @@ export function updateServerInfos(teams: TeamWithTabs[]) {
...
         Config.set('teams', teams);
     }).catch((reason: any) => {
         log.error('Error getting server infos', reason);
     });
 }
 
+function updateServerURL(data: Array<RemoteInfo | string | undefined>, team: TeamWithTabs) {
+    const remoteInfo = data.find((info) => info && typeof info !== 'string' && info.name === team.name) as RemoteInfo;
+    if (remoteInfo && remoteInfo.siteURL) {
+        team.url = remoteInfo.siteURL;
```
Willyfrog, line 71: if we do this and then I find a community-daily url, will it be treated as a non-server?

maybe we could add a siteURL to the team structure and check that as well in openExtraTabs.

Willyfrog (changes requested):
Maybe I'm missing something but I think we should keep both the user typed url and the final siteurl, as both point to the same server.

Willyfrog (approved):

src/main/app/utils.ts
devinbinnie: Spoke with Guillermo and ironed this out

nevyangelova (approved):
LGTM

PR #2190, opened 2022-07-07
[MM-14058] Add support for i18n

devinbinnie:
#### Summary
Adding support for translating strings in the Desktop App.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-14058

```release-note
Added support for i18n
```


devinbinnie:
> Code looks good to me but I wonder if we would benefit from in depth manual QA ?

Yeah I think that might be worth doing, @jgilliam17 Did you want to QA beforehand or afterwards?

jgilliam17:
Thanks @devinbinnie 
I can test after merge on desktop daily builds.
Will there be a new settings/preferences menu option to change desktop app language? How soon can I verify app is localized (I assume I have to wait for contributors to translate the strings 🤔 )

devinbinnie:
> Thanks @devinbinnie I can test after merge on desktop daily builds. Will there be a new settings/preferences menu option to change desktop app language? How soon can I verify app is localized (I assume I have to wait for contributors to translate the strings 🤔 )

Yes, there is an option in the Settings dropdown to override the language. Unfortunately you probably will have to wait for translation strings to be done to see any noticeable difference.

jgilliam17:
I was trying to add `QA deferred` label, but I guess we don't have it - thanks for checking, ok to defer

devinbinnie:
> I was trying to add `QA deferred` label, but I guess we don't have it - thanks for checking, ok to defer

Just added it :)

src/main/i18nManager.ts
```diff
@@ -0,0 +1,60 @@
...
+// See LICENSE.txt for license information.
+
+import {ipcMain} from 'electron';
+import log from 'electron-log';
+
+import {GET_AVAILABLE_LANGUAGES, GET_LANGUAGE_INFORMATION} from 'common/communication';
+
+import {Language, languages} from '../../i18n/i18n';
+
+export function t(s: string, defaultString = '', values: any = {}) {
```
hmhealey, line 11: I didn't realize when we were discussing this before that you meant `t` was the function that was actually doing the translation. While I don't like `localizeMessage` since we use it all over the place in the web app where `react-intl` is actually available, I guess that'd be a good name for it, and it should be picked up by mmjstool properly. In mobile and web, `t` is just a marker function that doesn't actually do anything, so I don't think we should have it work differently here.

Also, I'm now realizing that more of this translation stuff would probably make a good shared library between the web app and other repos...

src/renderer/components/AutoSaveIndicator.tsx
```diff
@@ -33,12 +34,13 @@ function getClassNameAndMessage(savingState: SavingState, errorMessage?: string)
...
-    errorMessage?: string;
+    errorMessage?: React.ReactNode;
+    intl: IntlShape;
 };
 
-export default function AutoSaveIndicator(props: Props) {
+const AutoSaveIndicator: React.FC<Props> = (props: Props) => {
     const {savingState, errorMessage, ...rest} = props;
-    const {className, message} = getClassNameAndMessage(savingState, errorMessage);
+    const {className, message} = getClassNameAndMessage(props.intl, savingState, errorMessage);
```
hmhealey, line 43: Can we get `intl` from `useIntl` instead of using `injectIntl`? As far as I know, `injectIntl` doesn't really have a purpose in functional components. I'd just use it for class components at the moment

src/renderer/components/ErrorView.tsx
```diff
@@ -41,31 +42,74 @@ export default function ErrorView(props: Props) {
...
                                         href='#'
                                     >
                                         {props.url}
-                                    </a>{' is correct.'}</li>
-                                <li>{'You can reach '}
+                                    </a>
+                                    <FormattedMessage
+                                        id='renderer.components.errorView.troubleshooting.urlIsCorrect.isCorrect'
+                                        defaultMessage=' is correct'
+                                    />
```
hmhealey, line 94: We can't append translated strings like this sentence and the one below (see [my blog post](https://developers.mattermost.com/blog/common-i18n-mistakes/#mistake-4-concatenating-translated-strings)). We should use react-intl's [rich text formatting thing](https://developers.mattermost.com/blog/common-i18n-mistakes/#solution-4-rich-text-formatting-in-react-intl) to do this instead

src/renderer/components/RemoveServerModal.tsx
```diff
@@ -13,27 +14,36 @@ type Props = {
...
+                            values={{serverName}}
+                        />
                     </p>
                 </Modal.Body>
             )}
         />
     );
 }
+
+export default injectIntl(RemoveServerModal);
```
hmhealey, line 49: This can also use `useIntl` instead of `injectIntl`

src/renderer/components/SettingsPage.tsx
```diff
@@ -852,16 +1087,35 @@ export default class SettingsPage extends React.PureComponent<Record<string, nev
...
                     onChange={this.handleChangeLogLevel}
                 >
-                    <option value='error'>{'Errors (error)'}</option>
-                    <option value='warn'>{'Errors and Warnings (warn)'}</option>
-                    <option value='info'>{'Info (info)'}</option>
-                    <option value='verbose'>{'Verbose (verbose)'}</option>
-                    <option value='debug'>{'Debug (debug)'}</option>
-                    <option value='silly'>{'Finest (silly)'}</option>
+                    <option value='error'>
+                        {intl.formatMessage({id: 'renderer.components.settingsPage.loggingLevel.level.error', defaultMessage: 'Errors (error)'})}
```
hmhealey, line 1091: Does FormattedMessage not work here? As far as I know, it just renders plain text without any tags around it

src/renderer/intl_provider.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export default class IntlProvider extends React.PureComponent<any, State> {
+    constructor(props: any) {
+        super(props);
+        this.state = {};
+    }
+
+    componentDidMount() {
+        window.addEventListener('message', this.handleMessageEvent);
```
hmhealey, line 26: This should be removed on unmount since it seems like we're going to have multiple IntlProviders mounted at different points. Otherwise, we're leaking these

src/renderer/intl_provider.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import {GET_LANGUAGE_INFORMATION, RETRIEVED_LANGUAGE_INFORMATION} from 'common/communication';
+
+import {Language} from '../../i18n/i18n';
+
+type State = {
+    language?: Language;
+}
+
+export function t(id: string) {
```
hmhealey, line 15: Is this still needed with the other version of `t` around?

src/renderer/modals/login/loginModal.tsx
```diff
@@ -87,82 +91,125 @@ export default class LoginModal extends React.PureComponent<Props, State> {
...
 
     render() {
+        const {intl} = this.props;
+
         let theServer = '';
         if (!(this.state.request && this.state.authInfo)) {
             theServer = '';
         } else if (this.state.authInfo.isProxy) {
-            theServer = `The proxy ${this.state.authInfo.host}:${this.state.authInfo.port}`;
+            theServer = intl.formatMessage({
```
hmhealey, line 100: This also has the concatenation issue. In this case, I think it would be easiest to just include the "requires a username and password" in each of these strings.

We could also use FormattedMessage here then instead of injecting the intl context

hmhealey (changes requested):

i18n/en.json
```diff
@@ -0,0 +1,238 @@
...
+  "renderer.modals.login.loginModal.password": "Password",
+  "renderer.modals.login.loginModal.proxy": "The proxy {host}:{port}",
+  "renderer.modals.login.loginModal.server": "The server {url}",
+  "renderer.modals.login.loginModal.title": "Authentication Required",
+  "renderer.modals.login.loginModal.username": "User Name",
+  "renderer.modals.permission.permissionModal.body": "A site that's not included in your Mattermost server configuration requires access for {permission}.",
+  "renderer.modals.permission.permissionModal.requestOriginatedFrom": "This request originated from ",
+  "renderer.modals.permission.permissionModal.title": "{permission} Required",
+  "renderer.modals.permission.permissionModal.unknownOrigin": "unknown origin",
+  "undefined": ""
```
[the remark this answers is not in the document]
devinbinnie: I couldn't find it, so I just left it since it doesn't really cause any problems. I can try and dig a bit deeper.

src/renderer/components/SettingsPage.tsx
devinbinnie: It actually renders as a `span` by default, which I think is defined [here](https://github.com/mattermost/desktop/pull/2190/files#diff-b44a65967cbe699fe5946cb191a63b9d228d678e3a42fa4a305b5975d2363c42R48)

We do the same in the webapp, so I copied the behaviour here.

src/renderer/intl_provider.tsx
devinbinnie: After the changes I made above it won't be, but I wanted to use `t()` purely as a marker function in the renderer process and not in the main. It will change once other changes are made.

i18n/en.json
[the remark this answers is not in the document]
devinbinnie: Seems like it cleared itself up :P

src/renderer/components/SettingsPage.tsx
hmhealey: Oh interesting. I thought it didn't render a tag because it doesn't let you apply styles or a className to FormattedMessage directly.

devinbinnie: I think you can override it if you choose, but by default you'll get a `span`. Which was very convenient for me since most of my text was wrapped in `span` anyway.

src/renderer/modals/permission/permissionModal.tsx
```diff
@@ -87,32 +101,46 @@ export default class PermissionModal extends React.PureComponent<Props, State> {
...
+                            </Button>
+                        </div>
+                    </Modal.Footer>
+                </Modal>
+            </IntlProvider>
         );
     }
 }
+
+t('common.permissions.canBasicAuth');
```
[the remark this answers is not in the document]
hmhealey: Usually it's better to keep the `t` function with the code that uses it so that the translation string doesn't get forgotten about or out of sync. We've had cases before where we removed some text from the web app, but the `t` function still existed so it was still showing up as something that needed to be translated

PR #2203, opened 2022-07-19
[MM-45646] Change download notification text to display saved filename instead of original

devinbinnie:
#### Summary
When downloading a file, the user is allowed to save the file as any name they want. However, when the notification displays that the file is done, it would always show the original name, which can be confusing.

This PR makes a change to always show the saved file name instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45646

```release-note
Fixed an issue where the download notification would show the wrong file name
```


agarciamontoro (approved):
:tada:

PR #2207, opened 2022-07-20
[MM-45448] Force electron processes to close after every test on Windows

devinbinnie:
#### Summary
When Playwright runs tests under Windows, it seems to leave behind the unclosed renderer processes. These processes then seems to consume as much CPU as they can, leading to a system that could become unresponsive or very slow after the test suite is run.

This PR introduces a library that checks for the orphaned processes and closes them after every test.
Tested on Windows 11 and no processes were left behind after the suite completed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45448

```release-note
NONE
```


M-ZubairAhmed (approved):

M-ZubairAhmed:
is this just windows issue? doesnt happen in mac

devinbinnie:
> is this just windows issue? doesnt happen in mac

Yeah it's a weird one, only affects Windows apparently. Funny because Playwright is a Microsoft library :D

nevyangelova (approved):
LGTM :)

PR #2209, opened 2022-07-25
Fix some issues caused by i18n addition

devinbinnie:
#### Summary
A couple of the modals were not working correctly with the i18n changes, so I've made a couple fixes.

```release-note
NONE
```

PR #2210, opened 2022-07-25
[MM-45740] Allow users to authenticate using a server that isn't their current server

devinbinnie:
#### Summary
NOTE: Depends on this first: https://github.com/mattermost/desktop/pull/2209

When a user tries to use basic authentication where the login request hits a different server (ie. same as custom login flow), it will be rejected since the server URLs don't match.

This PR fixes the auth manager so that it will first check if the server has entered a custom login flow and will allow the user to confirm the login via a permission modal that will show the login URL.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45740

```release-note
Fixed an issue where users using basic authentication against a different URL than their server URL would be silently rejected.
```

### DIFF
--- a/src/main/authManager.test.js
+++ b/src/main/authManager.test.js
@@ -5,6 +5,7 @@
 import {AuthManager} from 'main/authManager';
 import WindowManager from 'main/windows/windowManager';
 import ModalManager from 'main/views/modalManager';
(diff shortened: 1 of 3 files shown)


devinbinnie:
I'm going to be doing some work to move external authentication to the browser. Closing this for now.

Issue #2199, opened 2022-07-15
Client does not overwrite pre-existing file on download from a chat


flxmrk:
I confirm (by marking "x" in the [ ] below: [x]):

- [x ] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x ] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [ x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Downloading files from a chat will not overwrite preexisting files on save, but acts like it does.

**Environment**
- Operating System: OSX 12.4
- Mattermost Desktop App version: 5.1.1
- Mattermost Server version: 7.1.1

**Steps to reproduce**
1. put two different images in a chat
2. hover the first image, click the download icon, save the first image as a new file test.png
3. verify the right image is downloaded
4.  hover the second image, click the download icon, save the second image over the previous file named test.png
5. check the file, it is still the first image

**Expected behavior**
File becomes overwritten

**Observed behavior**
File stays the same as it was

devinbinnie:
There’s an Electron issue with MAS entitlements causing this. I’ve submitted a bug report: https://github.com/electron/electron/issues/35104

As per the bug, I can at least set exceptions for the Downloads folder (and a couple other select folders) so I will start with that and finish up if/when the bug is resolved.

PR #2213, opened 2022-07-27
[MM-45980] Add entitlements for read/write to a few key user directories

devinbinnie:
#### Summary
There's an issue with Electron currently blocking users from replacing files if there are using the MAS build. This PR mitigates the issue by explicitly adding the entitlements for available folders (Downloads, Pictures, Movies, Music) that will allow file replacement.

#### Ticket Link
Partial resolution:
https://mattermost.atlassian.net/browse/MM-45980
https://github.com/mattermost/desktop/issues/2199

```release-note
Fixed an issue where the user could not replace files in the Downloads folder
```

### DIFF
--- a/entitlements.mas.plist
+++ b/entitlements.mas.plist
@@ -18,9 +18,15 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
-	<key>com.apple.security.files.user-selected.read-only</key>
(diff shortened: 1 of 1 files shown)


nevyangelova (approved):
LG!

M-ZubairAhmed (approved):

PR #2222, opened 2022-08-10
[MM-46305] Add reporting code for E2E tests

devinbinnie:
#### Summary
Adds the code that allows us to upload E2E test reports to AWS. Largely based on the code currently used by Cypress in the webapp.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46305

```release-note
NONE
```


e2e/save_report.js
```diff
@@ -0,0 +1,117 @@
...
+require('dotenv').config();
+
+const saveReport = async () => {
+    const {
+        BRANCH,
+        BUILD_ID,
+        BUILD_TAG,
+        // DIAGNOSTIC_WEBHOOK_URL,
+        // DIAGNOSTIC_USER_ID,
+        // DIAGNOSTIC_TEAM_ID,
```
saturninoabril, line 57: I agree. These can be deleted and the corresponding commented code blocks in lines 102-107.

e2e/save_report.js
```diff
@@ -0,0 +1,117 @@
...
+        FAILURE_MESSAGE,
+        TM4J_ENABLE,
+        TM4J_CYCLE_KEY,
```
saturninoabril, line 60: Please change `TM4J` to `ZEPHYR` being the new name. We'll update the ones in mattermost-webapp, but [other repo started using Zephyr already](https://github.com/mattermost/mattermost-mobile/pull/6560/files#diff-4feac1abcee84ad8446fc56c50ddebd56a090571d90a26718c0f917291052c2fR56-R57).
```suggestion
        ZEPHYR_ENABLE,
        ZEPHYR_CYCLE_KEY,
```

saturninoabril (approved):
Thank you Devin! Let us know in QA where we could add values for those env variables. 

(I'm adding Mario here as well)

mvitale1989 (approved):
lgtm!

One small note regarding the AWS credentials source: the [aws-sdk-js v2 docs](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-environment.html) mention that `AWS_*` env vars should be picked up automatically by the library, so there's no need to explicitly read and pass them to the `AWS.S3` constructor.

Iiuc a nice side effect of relying on the library's internal lookup mechanism is that [the `AWS_PROFILE` env var](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-shared.html) can be used to get credentials from the awscli config, instead of the secret key's env vars. This is convenient but not at all required, so i'm completely fine with this implementation as well.

saturninoabril:
@mvitale1989 Good point and thanks for bringing it up! On top of that, maybe let's start using https://github.com/aws/aws-sdk-js-v3. Would you be open updaating separately, starting from mattermost-webapp, then all other repos? We can put in our queue to further discuss and plan.

PR #2227, opened 2022-08-16
[MM-22239] Downloads dropdown

tboulis:
#### Summary
Add downloads dropdown menu with recently downloaded items.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22239

#### Checklist
- [x] Added or updated E2E tests
- [x] Has UI changes

#### Tested on
Macbook Pro, 12.5
Windows 10 Pro, 21H2
Ubuntu 22

#### Screen Recording
https://recordit.co/33LNBhwCI6

#### Release Note
```release-note
Added Downloads Dropdown list
```

### DIFF
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -529,7 +529,6 @@ class MainPage extends React.PureComponent<Props, State> {
                     )}
                     {tabsRow}
                     {downloadsDropdownButton}
-                    {/* {upgradeIcon} */}
                     {titleBarButtons}
                 </div>
             </Row>


src/common/config/defaultPreferences.ts
```diff
@@ -43,6 +43,7 @@ const defaultPreferences: ConfigV3 = {
     lastActiveTeam: 0,
     downloadLocation: getDefaultDownloadLocation(),
     startInFullscreen: false,
+    downloads: [],
```
devinbinnie, line 46: Just curious, are these downloads that will be resumed once the application closes?

If not, I feel like we can just either store them in a variable somewhere, or if it needs to be a file, then maybe a different one.

The `config.json` file is mostly for persistent user settings.

devinbinnie: Well I'm wondering if they need to be stored at all. If a download is finished or cancelled, I don't think we need to store that information after the app has closed. If it's in progress, then I could see it, but that assumes that we're able to resume the download afterwards.

srkgupta:
Hi @tboulis 
I have analyzed and dismissed the CodeQL alerts in this PR. Please feel free to add me or someone from security team to this PR to re-review once the PR is complete and dev reviewed. 
Thanks

src/main/Validator.ts
```diff
@@ -96,7 +96,46 @@ const configDataSchemaV2 = Joi.object<ConfigV2>({
 });
 
 const configDataSchemaV3 = Joi.object<ConfigV3>({
-    version: Joi.number().min(2).default(2),
+    version: Joi.number().min(3).default(3),
```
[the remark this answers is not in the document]
devinbinnie: Yeah this should be fine. I think it was a typo.

matthewbirtch:
@tboulis I see the builds were started, but did they finish. Usually I see links to the builds posted back to the PR. Let me know!

devinbinnie:
@matthewbirtch There's an issue with the Mattermod hook. Here's the links:
https://output.circle-artifacts.com/output/job/6fbea896-b0a2-49d5-87d0-e4965570457e/artifacts/0/tmp/artifacts/mattermost-desktop-5.2.0-develop.1-linux-x64.tar.gz
https://output.circle-artifacts.com/output/job/92230e44-445d-4916-8ad2-4f9854f92692/artifacts/0/tmp/artifacts/mattermost-desktop-5.2.0-develop.1-mac-m1.dmg
https://output.circle-artifacts.com/output/job/92230e44-445d-4916-8ad2-4f9854f92692/artifacts/0/tmp/artifacts/mattermost-desktop-5.2.0-develop.1-mac-x64.dmg
https://output.circle-artifacts.com/output/job/2bdbbd0a-c487-4198-b40e-19609fdf8eef/artifacts/0/tmp/artifacts/mattermost-desktop-5.2.0-develop.1-win-ia32.zip
https://output.circle-artifacts.com/output/job/2bdbbd0a-c487-4198-b40e-19609fdf8eef/artifacts/0/tmp/artifacts/mattermost-desktop-5.2.0-develop.1-win-x64.zip

src/renderer/components/DownloadsDropdown/DownloadsDropdownButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+        if (!isDownloadsDropdownOpen) {
+            buttonRef.current?.blur();
+        }
+    }, [isDownloadsDropdownOpen, buttonRef]);
+
+    const handleToggleButton = (event: React.MouseEvent<HTMLButtonElement>) => {
+        event.preventDefault();
+        event.stopPropagation();
+        if (isDownloadsDropdownOpen) {
+            window.ipcRenderer.send(CLOSE_DOWNLOADS_DROPDOWN);
```
devinbinnie, line 34: Nit: Keep the `ipcRenderer` calls in `MainPage` so that we only have one point where this is used. I'd rather not actually have `ipcRenderer` exposed and eventually I'll want to remove it so it would be easier if we didn't have references everywhere.

src/renderer/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    // use generic icon
+    return 'generic';
+};
+
+const isImageFile = (file: DownloadedItem): boolean => {
+    return file.mimeType?.toLowerCase().startsWith('image/') ?? false;
+};
+
+const prettyETA = (ms = 0, intl: IntlShape) => {
```
devinbinnie, line 71: Do we have a library for this? I'm not opposed to doing this ourselves, just so we don't have to manage this.

src/main/downloadsManager.ts
```diff
@@ -0,0 +1,463 @@
...
+        this.saveAll({});
+        this.fileSizes = new Map();
+        this.closeDownloadsDropdown();
+        this.toggleAppMenuDownloadsEnabled(false);
+    }
+
+    showFileInFolder = (item?: DownloadedItem) => {
+        log.debug('DownloadsDropdownView.showFileInFolder', {item});
+
+        if (item && fs.existsSync(item.location)) {
```
devinbinnie, line 145: Do we have a mechanism to mark the file deleted if it no longer exists? Might be something we want to do.

src/main/downloadsManager.test.js
```diff
@@ -0,0 +1,168 @@
...
+        ...original,
+        resolve: jest.fn(),
+        parse: jest.fn(),
+    };
+});
+jest.mock('main/windows/windowManager', () => ({
+    sendToRenderer: jest.fn(),
+}));
+jest.mock('common/JsonFileManager', () => {
+    class JsonFileManagerMock {
```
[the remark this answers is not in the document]
devinbinnie: So as a general rule, `fs` should ALWAYS be mocked out in unit tests and if we need it to return something, we can mock that implementation as well.

I think that's what needs to happen here, not mocking out the class that it's inheriting. And if we do that, we can actually just mock the methods, not the whole class.

src/main/app/initialize.test.js
```diff
@@ -143,9 +143,34 @@ jest.mock('main/windows/windowManager', () => ({
     getMainWindow: jest.fn(),
     showMainWindow: jest.fn(),
     sendToMattermostViews: jest.fn(),
+    sendToRenderer: jest.fn(),
     getServerNameByWebContentsId: jest.fn(),
 }));
-
+jest.mock('common/JsonFileManager', () => {
```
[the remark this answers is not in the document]
devinbinnie: See my below reply.

e2e/modules/environment.js
```diff
@@ -242,6 +262,27 @@ module.exports = {
         await window.click('#saveSetting');
     },
 
+    async openDownloadsDropdown(app) {
```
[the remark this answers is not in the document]
devinbinnie: For the future we'll make one but for now this is fine. Was more of a question of what we should do going forward.

I'm 100% on board of sharing the `cy.addCommand` paradigm where we would define these somewhere.

src/main/utils.test.js
```diff
@@ -28,6 +27,17 @@ jest.mock('path', () => {
         resolve: (basePath, ...restOfPath) => original.join('/path/to/app/src/main', ...restOfPath),
     };
 });
+jest.mock('fs', () => {
+    const original = jest.requireActual('fs');
```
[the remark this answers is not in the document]
devinbinnie: We probably need to mock out additional methods in `fs` then.

src/renderer/utils.ts
devinbinnie: Oh yeah, let's avoid moment if we can :P

src/main/downloadsManager.ts
devinbinnie: What about after the app has started?

devinbinnie: If there's an update when it is clicked so that the user is shown that immediately, then we're okay. If the dropdown is opened, it's feasible that the user could open it, then go delete a file and then try and click on it afterwards.

e2e/modules/environment.js
```diff
@@ -100,10 +100,14 @@ const demoConfig = {
...
     autostart: true,
+    hideOnStart: false,
+    spellCheckerLocales: [],
     darkMode: false,
     lastActiveTeam: 0,
-    spellCheckerLocales: [],
+    startInFullscreen: false,
+    autoCheckForUpdates: false,
     appLanguage: 'en',
+    logLevel: 'silly',
```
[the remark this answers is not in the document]
devinbinnie: I didn't think we could see any logs. If that turns it on so we can then I'm all for it.

matthewbirtch (changes requested):
Hey @tboulis nice work. I'm still seeing a few minor things. These are non-blocking in my opinion, so we can fix them up in a follow-up PR if you want:

1. The popover has extra space on the bottom that is unnecessary.  Can we reduce this bottom padding? It should only be 8px.
image

2. It looks like the the ••• icon here is not vertically aligned within the button container.
image

Other than that, I think we're in great shape

matthewbirtch (approved):
Thanks @tboulis. Both those issues look resolved now. Awesome work!

src/renderer/components/MainPage.tsx
```diff
@@ -161,7 +177,21 @@ class MainPage extends React.PureComponent<Props, State> {
...
     }
 
+    async requestDownloadsLength() {
+        try {
+            const count = await window.ipcRenderer.invoke(REQUEST_HAS_DOWNLOADS);
+            this.setState({
+                hasDownloads: count,
+            });
+        } catch (error) {
+            console.error(error);
```
devinbinnie, line 187: Is there any case that this will popup on load? We've had that happen in the past and I'm trying to avoid pointless error messages cause they confuse people.

devinbinnie (approved):
Looks good to me! Added a few non-blocking comments.

One UX thing I noticed: If you click on the downloads dropdown itself while the sub-menu is open, it doesn't close the sub-menu. I would expect it to close. @matthewbirtch Not sure if you have a differing opinion. We can address it in a follow-up ticket.

EDIT: Also the same thing happens if you click on the top bar, it doesn't close the menu.

matthewbirtch:
> One UX thing I noticed: If you click on the downloads dropdown itself while the sub-menu is open, it doesn't close the sub-menu. I would expect it to close. @matthewbirtch Not sure if you have a differing opinion. We can address it in a follow-up ticket.

Agree with this - not a blocker, but we should fix it

devinbinnie (approved):

M-ZubairAhmed (approved):

PR #2242, opened 2022-08-26
[MM-46424] Fix 3-dot menu not losing focus after clicking outside (Windows)

tboulis:
#### Summary
On Windows the 3-dot button(that opens the app menu) would not lose focus after closing the menu or clicking outside of it. 
The problem was that the BrowserView of the top bar(3-dot button, server dropdown, tabs) could not fire the `focusout` event on the 3-dot because it can't "see" outside of its bounds (when a user would click outside of it).
To address that I used a state variable inside the MainApp component that changes the style of the button when it is hovered, or when the menu opens/closes. 

I also tested that pressing the Alt button on windows focuses the 3-dot button

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46424

#### Device Information
This PR was tested on:
- Windows 10 Pro, version 21H2

#### Screenshots
https://recordit.co/huulXU3KQV

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -49,6 +49,7 @@ import {
     UPDATE_DOWNLOADS_DROPDOWN,
     REQUEST_HAS_DOWNLOADS,
     CLOSE_DOWNLOADS_DROPDOWN_MENU,
+    APP_MENU_WILL_CLOSE,
 } from 'common/communication';
 
 import restoreButton from '../../assets/titlebar/chrome-restore.svg';
@@ -101,6 +102,7 @@ type State = {
     isDownloadsDropdownOpen: boolean;
     showDownloadsBadge: boolean;
     hasDownloads: boolean;
+    threeDotsIsFocused: boolean;
 };
 
 type TabViewStatus = {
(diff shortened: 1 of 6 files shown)


src/renderer/components/MainPage.tsx
```diff
@@ -529,11 +537,16 @@ class MainPage extends React.PureComponent<Props, State> {
                     <button
                         className='three-dot-menu'
                         onClick={this.openMenu}
+                        onMouseOver={this.focusThreeDotsButton}
```
hmhealey, line 540: Does this properly handle when focusing an element with the keyboard? Perhaps we could make it so that instead of adding a separate state to track focusing, we instead just blur the menu button on `APP_MENU_WILL_CLOSE`

hmhealey (changes requested):

src/renderer/components/MainPage.tsx
hmhealey: 0/5 on that as long as it still works with the keyboard. I wouldn't be surprised if there was some weird issue with blurring it where maybe it works differently in a BrowserView than I'm used to in an actual browser, but I don't have enough Electron experience to know for sure

hmhealey (approved):

devinbinnie (approved):
LGTM! Just to double-check, did we test this with keyboard navigation? I don't think the mouse events will cause any issues, but just wanted to make sure we can still tab around and not run into issues.

src/renderer/components/MainPage.tsx
```diff
@@ -268,12 +269,10 @@ class MainPage extends React.PureComponent<Props, State> {
             });
         });
 
+        window.ipcRenderer.on(APP_MENU_WILL_CLOSE, this.unFocusThreeDotsButton);
```
devinbinnie, line 272: Actually, to my previous comment, this probably will cause issues for keyboard issues, because once they close the menu for whatever reason they will use their focus position. It might not be the worst thing, but I think we should have a quick look to make sure it's not a huge problem for those users.

devinbinnie: I think in that case if they've pressed `ESC` they would likely lose focus. But any other interaction shouldn't necessarily do that. If that's the only one that's a concern then we're probably okay.

devinbinnie (approved):

devinbinnie:
cc @amyblais We cherry-picked this to v5.2 since it was needed for some other changes.

PR #2243, opened 2022-08-26
MM-45981_Desktop: Add Server Screen: Improve Onboarding screens

julmondragon:
#### Summary
When the user has no servers configured in the desktop app, a friendlier ‘add server’ screen will be displayed. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45981

#### UX Design
https://www.figma.com/file/7BM8S6BL597zBE83U6H8CK/MM-29159-Webapp-Login-Update?node-id=1023%3A92011

#### Checklist
- [x] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [x] completed [Mattermost Contributor Agreement](https://mattermost.com/contribute/)
- [x] executed `npm run lint:js` for proper code formatting

#### Device Information
This PR was tested on: macOS 11.6.8

##### Light mode
Screen Shot 2022-08-30 at 9 19 08 AM

##### Dark mode
Screen Shot 2022-08-30 at 9 19 32 AM

##### Mobile view (< 1200px)
Screen Shot 2022-08-30 at 9 19 56 AM

#### Release Note
```release-note
Adding Configure Server Screen when starting the app without servers created.
```

PR #2247, opened 2022-08-31
Add webContents listeners to modals

devinbinnie:
#### Summary
I noticed that our modals don't handle external links at all, since we've never had them before. However, now that we might want to do that, it would make sense to have similar handling around our links.

```release-note
NONE
```


julmondragon (approved):
LGTM! Thanks @devinbinnie , will update https://github.com/mattermost/desktop/pull/2243 once this is merged

PR #2260, opened 2022-09-19
[MM-46857] Remove misleading console error about tab status when servers don't exist

devinbinnie:
#### Summary
There was a misleading error about tab status that was coming up when no servers were configured. This was a relic of older BrowserView/WebView architecture stuff where it was possible that we were missing tab statuses for certain tabs, which doesn't happen anymore. So we've removed that log entry.

Also fixed a couple lint headaches.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46857

```release-note
NONE
```

### DIFF
--- a/e2e/specs/settings.test.js
+++ b/e2e/specs/settings.test.js
@@ -5,7 +5,6 @@
 'use strict';
 
 const fs = require('fs');
-const path = require('path');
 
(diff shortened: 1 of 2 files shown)


nevyangelova (approved):
lgtm!

PR #2265, opened 2022-09-28
[MM-46993] Implement CallsWidgetWindow

streamer45:
#### Summary

PR adds the initial implementation of a floating desktop widget for Calls. Technically speaking this is a frameless and transparent window that loads a standalone react app that purely renders the widget component (option 2 of the [design proposal](https://docs.google.com/document/d/1axYeHZeGbQWwsQPKHGYbdoyEkRTx0nbwxUEHWfk0FMA/edit)).

While ready for initial review I expect a few changes to get added on top as we start testing this a bit. I am using the [MM-46971-global-widget](https://github.com/mattermost/desktop/tree/MM-46971-global-widget) feature branch as base for this and any subsequent PRs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46993

#### Related PRs

https://github.com/mattermost/mattermost-plugin-calls/pull/213

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -1,55 +1,180 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {BrowserWindow} from 'electron';
+import {EventEmitter} from 'events';
+import {BrowserWindow, Rectangle, ipcMain, IpcMainEvent} from 'electron';
 import log from 'electron-log';
 
-import {CallsWidgetWindowConfig} from 'types/calls';
+import {CallsWidgetWindowConfig, CallsWidgetResizeMessage} from 'types/calls';
 
 import {getLocalPreload} from 'main/utils';
 
-export default function createCallsWidgetWindow(mainWindow: BrowserWindow, config: CallsWidgetWindowConfig) {
-    const preload = getLocalPreload('callsWidget.js');
-    const win = new BrowserWindow({
-        parent: mainWindow,
-        width: 274,
-        height: 82,
-        title: 'Calls Widget', // TODO: possibly add channel name?
-        fullscreen: false,
-        resizable: false,
-        frame: false,
-        transparent: true,
-        show: false,
-        alwaysOnTop: true,
-        webPreferences: {
-            preload,
+import {PRODUCTION} from 'common/utils/constants';
+import Utils from 'common/utils/util';
+import {
+    CALLS_WIDGET_RESIZE,
+} from 'common/communication';
 
-            // Workaround for this issue: https://github.com/electron/electron/issues/30993
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-            // @ts-ignore
+type LoadURLOpts = {
+    extraHeaders: string;
+}
+
+function boundsDiff(base: Rectangle, actual: Rectangle) {
+    return {
+        x: base.x - actual.x,
+        y: base.y - actual.y,
+        width: base.width - actual.width,
+        height: base.height - actual.height,
+    };
+}
+
+export default class CallsWidgetWindow extends EventEmitter {
+    private win: BrowserWindow;
+    private main: BrowserWindow;
+    private config: CallsWidgetWindowConfig;
+    private minWidth = 280;
+    private minHeight = 86;
+    private pluginID = 'com.mattermost.calls';
+    private boundsErr: Rectangle = {
+        x: 0,
+        y: 0,
+        width: 0,
+        height: 0,
+    };
+    private offsetsMap = {
+        'calls-widget-menu': {
+            height: 0,
+        },
+    };
+
+    constructor(mainWindow: BrowserWindow, config: CallsWidgetWindowConfig) {
+        super();
+
+        this.config = config;
+        this.main = mainWindow;
+        this.win = new BrowserWindow({
+            width: this.minWidth,
+            height: this.minHeight,
+            minWidth: this.minWidth,
+            minHeight: this.minHeight,
+            title: 'Calls Widget',
+            fullscreen: false,
+            resizable: false,
+            frame: false,
             transparent: true,
-        }});
-
-    win.once('ready-to-show', () => {
-        win.show();
-    });
-
-    const size = win.getSize();
-    const mainPos = mainWindow.getPosition();
-    const mainSize = mainWindow.getSize();
-    win.setPosition(mainPos[0] + 12, (mainPos[1] + mainSize[1]) - size[1] - 12);
-    win.setBackgroundColor('#00ffffff');
-    win.setMenuBarVisibility(false);
-
-    win.webContents.openDevTools({mode: 'detach'});
-
-    const pluginID = 'com.mattermost.calls';
-    const widgetURL = `${config.siteURL}/static/plugins/${pluginID}/widget/widget.html?call_id=${config.channelID}`;
-    win.loadURL(widgetURL).catch(
-        (reason) => {
-            log.error(`Calls window failed to load: ${reason}`);
+            show: false,
+            alwaysOnTop: true,
+            webPreferences: {
+                preload: getLocalPreload('callsWidget.js'),
+            },
         });
 
-    return win;
+        this.win.once('ready-to-show', () => this.win.show());
+        this.win.once('show', this.onShow);
+        this.win.on('closed', this.onClosed);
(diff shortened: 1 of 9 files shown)


src/main/windows/windowManager.ts
```diff
@@ -724,19 +782,26 @@ export class WindowManager {
     handleGetDesktopSources = async (event: IpcMainEvent, viewName: string, opts: Electron.SourcesOptions) => {
         log.debug('WindowManager.handleGetDesktopSources', {viewName, opts});
 
+        const globalWidget = viewName === 'widget' && this.callsWidgetWindow;
         const view = this.viewManager?.views.get(viewName);
-        if (!view) {
+        if (!view && !globalWidget) {
```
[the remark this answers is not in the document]
streamer45: I think `&&` is the correct one as we are looking to support both the existing way (in app widget) and the future one (global widget) so either of those conditions being true is okay to proceed. I guess I can turn it into a `||` and avoid the negations to make it a bit simpler to parse.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -0,0 +1,195 @@
...
+import {
+    CALLS_WIDGET_RESIZE,
+    CALLS_WIDGET_SHARE_SCREEN,
+} from 'common/communication';
+
+type LoadURLOpts = {
+    extraHeaders: string;
+}
+
+function boundsDiff(base: Rectangle, actual: Rectangle) {
```
devinbinnie, line 27: Could this be a helper function (it could go into `utils`?)

src/main/windows/callsWidgetWindow.ts
```diff
@@ -0,0 +1,195 @@
...
+        width: base.width - actual.width,
+        height: base.height - actual.height,
+    };
+}
+
+export default class CallsWidgetWindow extends EventEmitter {
+    public win: BrowserWindow;
+    private main: BrowserWindow;
+    private config: CallsWidgetWindowConfig;
+    private minWidth = 280;
```
devinbinnie, line 40: These belongs in `common/utils/constants`

src/main/windows/callsWidgetWindow.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        const bounds = this.win.getBounds();
+        const mainBounds = this.main.getBounds();
+        const initialBounds = {
+            x: mainBounds.x + 12,
+            y: (mainBounds.y + mainBounds.height) - bounds.height - 12,
+            width: this.minWidth,
+            height: this.minHeight,
+        };
+        this.win.setBackgroundColor('#00ffffff');
```
devinbinnie, line 185: You should be able to just define this when you create the `BrowserWindow`

devinbinnie (commented):
Overall design looks good, I ran into these issues while I was playing around with it:
- menu bar in the pop out window should not be accessible
	- maybe we make another "custom" popup window for that?
- closing the pop out window seems to freeze the app (but the call continues)
	- you have to force it to reopen but then it doesn't work until you leave/rejoin again cause it think you're already in the call
- there's not much indication that you can drag the call outside of the main window cause it looks identical (and is placed identically) to the webapp one.

I'd also like to ask if this is something we think we could build directly into the app top bar (for example) instead of having another window. With the upcoming changes to the Desktop App, we will be able to have a more unified Web App/Desktop App experience, so we might not need an extra window. Just something to think about.

streamer45:
Thanks @devinbinnie, answering to your main comments below:

> * menu bar in the pop out window should not be accessible
>   
>   * maybe we make another "custom" popup window for that?

This should be easily fixable from the plugin side without adding more logic here but just passing a flag. I'll send an update.

> * closing the pop out window seems to freeze the app (but the call continues)
>   
>   * you have to force it to reopen but then it doesn't work until you leave/rejoin again cause it think you're already in the call
> * there's not much indication that you can drag the call outside of the main window cause it looks identical (and is placed identically) to the webapp one.

If it's what I think, it's some sort of bug with chrome and something we use for dev builds cause I was never able to reproduce it in production but only happened on local dev instances. Same thing should happen if you close the pop out window using the web version. Not sure there's much we can do at this point but can try to look deeper.

> I'd also like to ask if this is something we think we could build directly into the app top bar (for example) instead of having another window. With the upcoming changes to the Desktop App, we will be able to have a more unified Web App/Desktop App experience, so we might not need an extra window. Just something to think about.

@cpoile raised a similar feedback during the design phase (full context in [here](https://docs.google.com/document/d/1axYeHZeGbQWwsQPKHGYbdoyEkRTx0nbwxUEHWfk0FMA/edit?disco=AAAAcIwGXL4)). We discussed this internally and mostly agreed that having the separate window actually helped in making the feature feel more native and also would help decoupling Mattermost from the call in such a way that the users are not necessarily forced to have Mattermost in the foreground but still be able to follow and control the call. 

By the way we are also planning to soon remove the pop out window completely from the desktop experience and converge its functionality into the widget window instead by dynamically mutating the view to account for things like screen sharing and host controls. 

@tao pinging you for visibility in case you have any comments on the above.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -0,0 +1,195 @@
...
+        this.win.close();
+    }
+
+    public getServerName() {
+        return this.config.serverName;
+    }
+
+    private load() {
+        const opts = {} as LoadURLOpts;
+        if (Utils.runMode() !== PRODUCTION) {
```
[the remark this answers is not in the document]
streamer45: Sure. I am actually still figuring out if we need it in the first place. The problem I was facing was that the window would load stale (client cached) data and I had to hard refresh every time so using that header did that automatically.

devinbinnie:
Thanks for the comments @streamer45. Overall architecture looks good to me, so we can keep going with this PR. Once we've got everything sorted out or if you have any questions feel free to ping me and I'll give the final approval.

src/main/windows/callsWidgetWindow.ts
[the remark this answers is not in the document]
streamer45: Okay so the issue turned out to be even worse with a proxy in between as it would cache stuff almost permanently even when updating the plugin since the `widget.html` file was considered static (and served as such). I've now changed the way we are serving the widget app entirely, instead of relying on the static folder I'll be using a custom plugin API endpoint `/widget` to do it so that we get more control on how assets are served and also the caching now works as expected, only updating when the files are modified.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -0,0 +1,197 @@
...
+        log.debug('CallsWidgetWindow.onClosed');
+        this.emit('closed');
+        this.removeAllListeners('closed');
+        ipcMain.off(CALLS_WIDGET_RESIZE, this.onResize);
+        ipcMain.off(CALLS_WIDGET_SHARE_SCREEN, this.onShareScreen);
+    }
+
+    private getWidgetURL() {
+        const u = new url.URL(this.config.siteURL);
+        u.pathname += `/plugins/${CALLS_PLUGIN_ID}/widget/widget.html`;
```
streamer45, line 111: I am still considering whether we should be passing this path from the plugin instead of forming it here. I think it should be stable but the flexibility of having it on the server side may help in making future changes backwards compatible.

streamer45:
@devinbinnie This should be ready for a second pass. If testing, just make sure to update the plugin as I had to make changes on that side as well.

> there's not much indication that you can drag the call outside of the main window cause it looks identical (and is placed identically) to the webapp one.

Forgot to answer this before. I think that was almost intentional but I see it can get confusing knowing how it's operating now. 
@itao @tanmay-des Is there anything you can think of we could do to make it more explicit that the window is decoupled from the app window and draggable anywere?

src/main/windows/callsWidgetWindow.ts
devinbinnie: I think this is fine for now. Long term, we will eventually move this widget (or at least the setup for it) into the Desktop App so we won't need this reference. I'd also like to avoid additional message passing stuff if we can.

devinbinnie (approved):
Code looks good, I'll spend a bit of time playing with it, but I won't hold up the review otherwise.
Thanks for the work @streamer45 :)

devinbinnie:
@streamer45 Feel free to merge whenever this is ready.

streamer45:
> @streamer45 Feel free to merge whenever this is ready.

Thanks, I am mostly waiting to send a couple of minor changes first which I'll create a dedicated PR for pointing to this one, after which we'll be ready to merge to master branch.

PR #2270, opened 2022-10-04
[MM-47279] Upgrade to Electron v21

devinbinnie:
#### Summary
Upgrading to Electron v21

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47279

```release-note
Upgrade to Electron v21.0.1
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "build": "npm-run-all build:*",
     "build:main": "webpack-cli --config webpack.config.main.js",
     "build:renderer": "webpack-cli --config webpack.config.renderer.js",
-    "build-robotjs": "electron-rebuild -v 21.1.0 -m ./node_modules/robotjs",
+    "build-robotjs": "cross-env CL='/std:c++17' electron-rebuild -v 21.1.0 -m ./node_modules/robotjs",
     "start": "electron dist/ --disable-dev-mode",
     "restart": "npm run build && npm run start",
     "storybook": "start-storybook -p 9001 -c src/.storybook",
--- a/package.json
+++ b/package.json
@@ -9,14 +9,14 @@
   "desktopName": "Mattermost.Desktop",
   "homepage": "https://mattermost.com",
   "engines": {
-    "node": ">=14.6.0"
+    "node": ">=16.16.0"
   },
   "repository": {
     "type": "git",
     "url": "git://github.com/mattermost/desktop.git"
   },
   "config": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -173,6 +173,7 @@
     "mmjstool": "github:mattermost/mattermost-utilities#3b4506b0f6b14fbb402f9f8ef932370e459e3773",
     "mocha-circleci-reporter": "0.0.3",
     "mochawesome": "7.1.3",
+    "node-abi": "3.25.0",
```
m1lt0n, line 176: (question) Is the explicit dependency needed here? All the packages that need it have it in their dependencies (e.g. electron-rebuild etc). Is this needed to pin it to a specific version?

m1lt0n (approved):
Left a minor non-blocking comment about dependencies.

package.json
devinbinnie: We should be able to remove this, but for now it's needed because of how new Electron v21 is. Some of the downstream dependencies are using older versions of `node-abi` so we need to make sure we have the latest.

devinbinnie:
So unfortunately this is now blocked due to issues with the NodeJS Native Abstractions, there's a PR waiting to be merged to fix it here: https://github.com/nodejs/nan/pull/943

devinbinnie:
> Good to know @devinbinnie , so I guess this one should be closed and the one that upgrades electron to the latest 19.1.1 will be the relevant one.

@m1lt0n We can leave this one open for the time being, since I'd still like to upgrade to v21 once we're unblocked. It just won't be merged for a while.

package.json
```diff
@@ -174,6 +174,8 @@
     "mmjstool": "github:mattermost/mattermost-utilities#3b4506b0f6b14fbb402f9f8ef932370e459e3773",
     "mocha-circleci-reporter": "0.0.3",
     "mochawesome": "7.1.3",
+    "nan": "2.17.0",
```
devinbinnie, line 177: These were added to ensure compatibility with native code.

devinbinnie:
@amyblais @tboulis @m1lt0n Potential known issue I'd like to call out: I had to blow out my `C:\Users\<username>\AppData\Roaming\Electron` folder to get things working on Windows. It could have been a fluke with my system, but just in case.

m1lt0n (approved):

devinbinnie:
There's an issue with the loading spinner I'm currently sorting out:

PR #2273, opened 2022-10-07
Fix issue with downloads position after window (un)maximize

tboulis:
```release-note
NONE
```


m1lt0n (approved):
Looks good! Tested locally and the downloads list appears properly in maximized and unmaximized position

devinbinnie (approved):

PR #2274, opened 2022-10-07
Downloads - Remove duplicate handler for showing file in folder

tboulis:
I noticed this on ubuntu.
If you clicked a file from the downloads list it would open the downloads folder twice. On mac it would not open a second Finder window, that's why I missed it

```release-note
NONE
```


m1lt0n (approved):
Nice catch!

devinbinnie (approved):

PR #2290, opened 2022-10-18
[MM-47256] Notifications native module

tboulis:
#### Summary
Use [node-notifier](https://www.npmjs.com/package/node-notifier) for displaying native notifications.
Add support for sending test notification under Desktop Settings page

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-47617
- https://mattermost.atlassian.net/browse/MM-47256

#### Release Note

```release-note
Use native module for notifications
```

### DIFF
--- a/src/main/notifications/index.ts
+++ b/src/main/notifications/index.ts
@@ -118,4 +118,3 @@ export function sendTestNotification() {
         soundName: 'Bing',
     });
 }
-
--- a/src/main/preload/mainWindow.js
+++ b/src/main/preload/mainWindow.js
@@ -56,6 +56,7 @@ import {
     UPDATE_DOWNLOADS_DROPDOWN,
     APP_MENU_WILL_CLOSE,
     FOCUS_THREE_DOT_MENU,
+    SEND_TEST_NOTIFICATION,
 } from 'common/communication';
 
 console.log('Preload initialized');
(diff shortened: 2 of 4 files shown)


src/main/preload/mattermost.js
```diff
@@ -140,7 +140,7 @@ window.addEventListener('message', ({origin, data = {}} = {}) => {
     case 'dispatch-notification': {
         if (shouldSendNotifications) {
             const {title, body, channel, teamId, url, silent, data: messageData} = message;
-            ipcRenderer.send(NOTIFY_MENTION, title, body, channel, teamId, url, silent, messageData);
+            ipcRenderer.send(NOTIFY_MENTION, title, body, channel, teamId, url, silent, messageData?.soundName);
```
devinbinnie, line 143: Is there anything else that we could end up putting in `messageData` that might be useful here?

I actually forget what's in there besides the sound name :P

src/main/notifications/index.test.js
```diff
@@ -16,9 +16,10 @@ import {localizeMessage} from 'main/i18nManager';
```
devinbinnie, line 16: It might be worth adding some additional tests to check for the behaviour of the notification changes. One example could be making sure it calls the right stuff per OS.

devinbinnie (approved):
Overall looks good! Added a couple non-blocking comments.

src/types/notification.ts
```diff
@@ -1,12 +1,75 @@
...
+import WindowsBalloon from 'node-notifier/notifiers/balloon';
+import Growl from 'node-notifier/notifiers/growl';
+import NotifySend from 'node-notifier/notifiers/notifysend';
+import WindowsToaster from 'node-notifier/notifiers/toaster';
 
-export type MentionData = {
+export type NotificationOptions = WindowsToaster.Notification & WindowsBalloon.Notification & NotifySend.Notification & Growl.Notification & Notification;
+
+type MentionPayload = {
+    channel: {
```
m1lt0n, line 13: (nitpick) perhaps channel can be its own type, since it's used a couple of times (here and in `DisplayMentionArguments`)

m1lt0n (approved):

jgilliam17:
Thanks @tboulis 
Seeing one issue on Windows 10, I can hear the notification sound, but no notification popup for messages or when I test "send notification"  in settings. 
I did get "session expired" notifications, both sound and popup.
On macOS and Ubuntu this works as expected (sound and popup) for both new messages and testing the setting.

PR #2294, opened 2022-10-20
Disabled "Clear All" button when downloads include only the update

tboulis:
#### Summary
- Disable "Clear All" button in Downloads when only the application update is included
- When Downloads include files and application update, the "Clear All" button will only remove the files, not the update

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47784

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/renderer/downloadsDropdown.tsx
+++ b/src/renderer/downloadsDropdown.tsx
@@ -74,7 +74,13 @@ class DownloadsDropdown extends React.PureComponent<Record<string, never>, State
     }
 
     clearAll = () => {
-        window.postMessage({type: REQUEST_CLEAR_DOWNLOADS_DROPDOWN}, window.location.href);
+        if (!this.clearAllButtonDisabled()) {
+            window.postMessage({type: REQUEST_CLEAR_DOWNLOADS_DROPDOWN}, window.location.href);
+        }
+    }
+
+    clearAllButtonDisabled = () => {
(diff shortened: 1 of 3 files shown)


src/renderer/downloadsDropdown.tsx
```diff
@@ -74,7 +74,13 @@ class DownloadsDropdown extends React.PureComponent<Record<string, never>, State
...
 
     clearAll = () => {
-        window.postMessage({type: REQUEST_CLEAR_DOWNLOADS_DROPDOWN}, window.location.href);
+        if (!this.clearAllButtonDisabled()) {
+            window.postMessage({type: REQUEST_CLEAR_DOWNLOADS_DROPDOWN}, window.location.href);
+        }
+    }
+
+    clearAllButtonDisabled = () => {
+        return this.state.downloads?.length === 1 && this.state.downloads[0]?.type === 'update';
```
m1lt0n, line 83: (clarification) So, based on the behaviour of the code above (clearDownloadsDropdown function), clearing when there is an update and other downloads, it will remove everything and push into the downloads the update again right?

Also, is it guaranteed that the update will always be the first of the downloads or something like `this.state.downloads.filter(d => d.type === 'update'). length > 0` or `this.state.downloads.map(d => d.type).includes('update')` is safer?

m1lt0n (approved):

devinbinnie (approved):

PR #2299, opened 2022-10-20
[MM-47754] Remove update from downloads when "Restart and Upgrade" is clicked

tboulis:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47754

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/autoUpdater.test.js
+++ b/src/main/autoUpdater.test.js
@@ -17,7 +17,6 @@ jest.mock('electron', () => ({
         createFromPath: jest.fn(),
     },
     ipcMain: {
-        handle: jest.fn(),
         on: jest.fn(),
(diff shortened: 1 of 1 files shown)


src/common/JsonFileManager.ts
```diff
@@ -31,9 +31,9 @@ export default class JsonFileManager<T> {
         });
     }
 
-    setJson(json: T): void {
+    async setJson(json: T): Promise<void> {
         this.json = json;
-        this.writeToFile();
+        await this.writeToFile();
```
devinbinnie, line 36: Just to make sure, did we test other uses of this function? I don't assume it'll break anything but worth having a quick look.

m1lt0n (approved):

PR #2308, opened 2022-10-21
[MM-47682] Implement WebRequestManager

devinbinnie:
#### Summary
In order to support URL rewriting from many different listeners, we need a manager that will allow us to bind as many listeners to the web requests as possible. This PR adds a manage for the WebRequest handlers and allows the app to bind as many listeners as we want.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47682

```release-note
NONE
```

### DIFF
--- a/src/main/webRequest/webRequestHandler.test.js
+++ b/src/main/webRequest/webRequestHandler.test.js
@@ -8,27 +8,6 @@ jest.mock('electron-log', () => ({
 }));
 
 describe('main/webRequest/webRequestHandler', () => {
-    it('should not iterate through listeners when size is < 1', () => {
-        const handler = new WebRequestHandler();
-        handler.listeners = {
-            size: 0,
-            keys: jest.fn(),
-        };
-
-        const callback = jest.fn();
-        handler.handle({}, callback);
-        expect(handler.listeners.keys).not.toBeCalled();
-        expect(callback).toBeCalledWith({});
-
-        handler.listeners = {
-            size: 1,
-            keys: jest.fn().mockReturnValue([]),
-        };
-
-        handler.handle();
-        expect(handler.listeners.keys).toBeCalled();
-    });
-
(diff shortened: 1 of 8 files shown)


devinbinnie:
> I really like what you did hear @devinbinnie. I only have one question: Why didn't you extend the `EventEmitter` for webRequestHandler? I think it will make it easier to:
> 
> * Errors will be thrown inside the consumer and not the WebRequestManager
> * No need to keep track of the listeners

Yeah this seems like a good idea :P I'll make the change.

src/main/webRequest/webRequestHandler.test.js
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {WebRequestHandler} from './webRequestHandler';
+
+jest.mock('electron-log', () => ({
+    silly: jest.fn(),
+}));
+
+describe('main/webRequest/webRequestHandler', () => {
```
m1lt0n, line 10: (nitpick/non-blocking) Nice tests! Since the details are passed to the listeners would it make sense for the mock implementations to have a signature that uses details (e.g. `(details) => result += details.some + '1, 2 or 3'`?

m1lt0n (approved):
Nice one @devinbinnie ! I left only a small comment (non-blocking) on making the tests of the handler a bit more robust.

src/main/webRequest/webRequestHandler.test.js
devinbinnie: The handler class itself doesn't make use of the `details`, so it's a bit irrelevant there actually. It's mostly just a passthrough, so I test it in the manager instead. Bit of a break of the unit test paradigm, but at least this way all of the tests have a bit more meaning.

PR #2310, opened 2022-10-21
Fix macos-notification-state build issue

devinbinnie:
#### Summary
There's been a long running issue with the `macos-notification-state` node module wouldn't compile correctly under `npm run postinstall`, due to it missing an `install` script.

This PR adds that bit to the module as part of a postinstall script, and switches us over to using `electron-builder` to run our app dependencies.

```release-note
NONE
```


devinbinnie:
cc @amyblais This is a build process fix I'd like to include in v5.2, low risk.

devinbinnie:
@tboulis The good ol' `node-gyp` fails on Windows error is back. It could just be the Node version causing this. I'm gonna mess with it a bit and see if I can get it working. This always happens in CI...

devinbinnie:
> @devinbinnie Do we actually need to build `macos-notification-state` on Windows? 🤔

If I can find a way around it, no. But that's not the issue actually, it's the `windows-focus-assist` module that's causing the error, or rather just native code building in general. It always throws this weird error where `node-gyp` can't find the node header files, I still don't exactly know why it happens. We fixed it before by making sure they were pre-downloaded, but that's not working correctly right now. Will have to investigate.

PR #2311, opened 2022-10-24
[MM-47842] Make "save as" in context menu show the save dialog

tboulis:
#### Summary
The save as dialog does not open and the file would be downloaded right away in the default download location. 

The fix is to detect if the download was initiated by user gesture or not.

This happens because we use `electron-context-menu` to show the right click menu, which when you click "Save as" uses `electron-dl` to trigger the download.

I would recommend we remove the `electron-context-menu` and use electron's Menu on the renderer process instead, which will give us the option to use our `downloadManager` when clicking "Save as" instead of the hard-coded `electron-dl`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47842

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/downloadsManager.test.js
+++ b/src/main/downloadsManager.test.js
@@ -117,6 +117,7 @@ const item = {
     getStartTime: () => nowSeconds,
     getTotalBytes: () => 4242,
     getSavePath: () => locationMock,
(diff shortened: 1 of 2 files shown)


m1lt0n (approved):
Looks good!

PR #2315, opened 2022-10-24
[MM-47801][MM-45980] Added support for security-scoped bookmarks to allow the MAS build to save files wherever needed

devinbinnie:
#### Summary
Because of Apple's App Sandbox, saving files in any location other than the Downloads directory causes a number of issues. In order to resolve these, Apple provides an API called "security-scoped bookmarks" that allows an app to temporarily gain access to a file in the same way that the user would normally pick one using the dialog box and do some work on it.

This PR converts the downloads code to allow for these bookmarks to be created and stored in a persistent manner so that MAS build users can access and save files as normal. This approach conveniently also fixes the issue where users couldn't replace files.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47801
https://mattermost.atlassian.net/browse/MM-45980
Closes https://github.com/mattermost/desktop/issues/2199

```release-note
Fixed an issue where MAS users couldn't easily replace files.
```

### DIFF
--- a/entitlements.mas.plist
+++ b/entitlements.mas.plist
@@ -38,5 +38,9 @@
 	<string>UQ8HT4Q2XM.Mattermost.Desktop</string>
 	<key>com.apple.developer.team-identifier</key>
 	<string>UQ8HT4Q2XM</string>
+	<key>com.apple.security.files.bookmarks.app-scope</key>
+	<true/>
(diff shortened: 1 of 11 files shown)


m1lt0n (approved):
Awesome PR!

PR #2323, opened 2022-10-26
Stop unnecessary TestFlight builds

devinbinnie:
We've been pushing to TestFlight on every commit for the release branches, this generates a lot of noise and doesn't help us version well. This PR should stop that from happening.


zefhemel:
Thank you 😊

PR #2357, opened 2022-11-04
[MM-48167] Only grab the last extension from a filename during Save As

devinbinnie:
#### Summary
When right-clicking a file and choosing Save As, if the filename had more than one `.`, the dialog would pick up each string seperated by a `.` and add it to the file extension list. Commonly reproducible when sharing macOS screenshots.

This PR ensures that we always get the actual file extension (ie. the last one) and use it for the dialog.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48167

```release-note
NONE
```


nevyangelova (approved):
LGTM!

PR #2362, opened 2022-11-07
[MM-48083] Add websocket overrides

devinbinnie:
#### Summary
This implements an override for the `getSiteURL` module, and overrides the origin when connecting to the websocket. These two changes allow the websocket to connect and pass information as normal without any issue.

Relevant webapp change: https://github.com/mattermost/mattermost-webapp/pull/11383/commits/d2699127df051cc22f680e81c4c996bf26acd415

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48083

```release-note
NONE
```


hmhealey (approved):
Looks good to me

There is also a setting on the server to override the websocket URL, but I don't believe that'll be affected by this because it's not relative to the root URL for the page.

PR #2365, opened 2022-11-08
[MM-47707] Update styles to allow for general functionality

devinbinnie:
#### Summary
This is just a small styles update to make sure that most of the webapp and desktop app styles don't conflict. There will likely be some more work to be done as we merge more of the views together, and there may still be some less-than-ideal UX issues, but overall the app should now be mostly usable with this change.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47707

```release-note
NONE
```


src/renderer/css/components/TabBar.scss
```diff
@@ -0,0 +1,202 @@
+#main {
```
nevyangelova, line 1: Is there a reason to need to add another root div here? Potentially the wrapper .Mainpage should make the styles specific enough. IMO we should specify styled in #root, #app or #main only if they are really global, ex. heading fonts or colours.

src/renderer/css/components/TabBar.scss
```diff
@@ -0,0 +1,202 @@
...
+  .TabBar {
+    border: none;
+    max-height: 40px;
+    flex: 1 1 auto;
+    display: flex !important;
```
nevyangelova, line 6: Why is this important needed here? I can't see this being overwritten in another place. If its not being applied then probably there's a clash somewhere.

src/renderer/css/components/TabBar.scss
```diff
@@ -0,0 +1,202 @@
...
+  .TabBar.darkMode>li.teamTabItem>a>div.TabBar-tabSeperator>.teamTabItem__close {
+      color: rgba(221,221,221,0.32);
+  }
+
+  .TabBar.darkMode>li>a:focus {
+    background-color: #1f1f1f;
+    color: #ddd;
+  }
+
+  .TabBar>li>a>div.TabBar-tabSeperator {
```
nevyangelova, line 98: not sure if this is in the scope of this PR but nesting like this is probably what causes the style overrides. We might need to clean up desktop app styles. Regardless i would add a selector to either one of those elements, and remove two out the three.

src/renderer/css/components/TabBar.scss
```diff
@@ -0,0 +1,202 @@
...
+    background-color: rgba(31, 31, 31, 0.56);
+  }
+
+  .TabBar>li>a:focus {
+    background-color: #fff;
+    color: #3d3c40;
+    text-decoration: none;
+  }
+
+  .TabBar>li.teamTabItem>a>div.TabBar-tabSeperator>.teamTabItem__close {
```
nevyangelova, line 69: > not sure if this is in the scope of this PR but nesting like this is probably what causes the style overrides. We might need to clean up desktop app styles. Regardless i would add a selector to either one of those elements, and remove two out the three.

same here and in other places

src/renderer/css/components/TabBar.scss
```diff
@@ -0,0 +1,202 @@
...
+    border-left: none;
+    margin-left: 0px;
+  }
+
+  .TabBar>li.teamTabItem:not(.active):not(.disabled):hover+li.teamTabItem:not(.active)>a>div.TabBar-tabSeperator {
+    border-left: none;
+    margin-left: 0px;
+  }
+
+  .TabBar>li.teamTabItem:not(.active):not(.disabled)+li.teamTabItem:not(.active)>a:hover>div.TabBar-tabSeperator {
```
nevyangelova, line 136: how about applying only one class in the case where this condition is met?

src/renderer/css/index.scss
```diff
@@ -0,0 +1,219 @@
...
+#main {
+
+  //height: 100%;
+  //font-size: 14px;
+  //min-height: 100%;
+
+  position: absolute;
+  width: 100%;
+
+  .hovering-enter {
```
nevyangelova, line 17: I would avoid specifying action styles in a root level element, it might cause issues in the future as these styles are often mutable.

src/renderer/css/index.scss
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  .topBar .title-bar-btns>.close-button:hover>img {
+    filter: invert(100%);
+    -webkit-filter: invert(100%);
+    opacity: 1;
+  }
+
+  .topBar .title-bar-btns>.close-button:active {
+    background: #f1707a !important;
```
nevyangelova, line 172: I would investigate why these importants are needed. IMO the styles need some refactoring

nevyangelova (commented):
I won't block the PR but wanted to add some suggestions regarding style improvements, in order to avoid UI bugs in the future

src/renderer/index.html
```diff
@@ -5,6 +5,6 @@
         <title><%= htmlWebpackPlugin.options.title %></title>
     </head>
     <body>
-        <div id="root" />
+        <div id="app" />
```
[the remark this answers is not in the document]
devinbinnie: We'd have to change a lot of different code to change that id. Since it doesn't really make a huge difference can we keep it the same for now? We could refactor this later.

src/renderer/css/components/TabBar.scss
devinbinnie: It's mostly for specificity so that it doesn't interfere with any of the styles provided by the web app.

src/renderer/css/components/TabBar.scss
devinbinnie: Not sure, it's been around for a while. All of this CSS is old I just had to wrap it in a new file and a new tag.

src/renderer/css/components/TabBar.scss
devinbinnie: Yes these need a cleanup. Worth doing in a future PR.

src/renderer/css/components/TabBar.scss
devinbinnie: Same here, can address in a future cleanup.

nevyangelova (approved):

Issue #2373, opened 2022-11-10
5.2.0 - Windows 8 - no notification popup


licaon-kter:
Works fine on 5.1.1

amyblais:
Are you on Windows 8.1?

licaon-kter:
Ah indeed, there a difference. Yes, latest 8.1.

jgilliam17:
@amyblais I checked our QA device list and don't see anyone listed in possession of Win8. I can try to create a VM and see if notifications can be tested that way.

jgilliam17:
I created a vm with Win8 and I am not getting any notifications. Tested using Desktop v5.2.1. I'll chat with Tasos about this and post an update.

PR #2385, opened 2022-11-15
Fix issue with windows-focus-assist returning truthy value on older windows versions

tboulis:
#### Summary
`windows-focus-assist` is intended to be used on win 10,11. On windows 8.1 it returns `-1` which is a truthy value in javascript.

5E86FFAA-E22E-47AC-B09C-1436BC2B0755

This causes the application to not display notifications by erroneously receiving `true` for the do-not-disturb mode.

More info on the return value here: https://github.com/bitdisaster/windows-focus-assist/blob/master/lib/focus-assist.cc#L5-L9

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48397
Closes https://github.com/mattermost/desktop/issues/2373

#### Device Information
This PR was tested on: VirtualBox, windows 8.1

#### Release Note
```release-note
Fix issue with notifications not displayed on windows 8
```

### DIFF
--- a/src/main/notifications/dnd-linux.ts
+++ b/src/main/notifications/dnd-linux.ts
@@ -9,6 +9,9 @@ const GNOME_READ_DND = 'gsettings get org.gnome.desktop.notifications show-banne
 
 function getLinuxDoNotDisturb() {
     try {
+        if (process.platform !== 'linux') {
+            return false;
+        }
         const showNotifications = execSync(GNOME_READ_DND).toString().replace('\n', '');
         log.debug('getLinuxDoNotDisturb', {showNotifications});
 
--- a/src/main/notifications/dnd-windows.ts
+++ b/src/main/notifications/dnd-windows.ts
@@ -1,0 +1,33 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {getFocusAssist, isPriority} from 'windows-focus-assist';
+
+/**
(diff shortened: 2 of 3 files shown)


src/main/notifications/dnd-linux.ts
```diff
@@ -9,6 +9,9 @@ const GNOME_READ_DND = 'gsettings get org.gnome.desktop.notifications show-banne
 
 function getLinuxDoNotDisturb() {
     try {
+        if (process.platform !== 'linux') {
```
m1lt0n, line 12: Why is this check needed? `getDoNotDisturb` which is the "public" API that runs internally this code means `getLinuxDoNotDisturb` is only ran when `process.platform` is `linux`.

m1lt0n (changes requested):

src/main/notifications/dnd-linux.ts
devinbinnie: 0/5 on keeping this or not. It does act as a guard, but really only if someone manually overrides their `process.platform` somehow.

devinbinnie (approved):

src/main/notifications/dnd-linux.ts
m1lt0n: yep, 0/5 on that too, so resolving the convo.

amyblais:
@tboulis @devinbinnie Could we also cherry-pick this bug fix to 5.2.2 dot release?

PR #2390, opened 2022-11-16
[MM-48144] Diagnostics Step - Configure logger

tboulis:
#### Summary
Configure logger in Step 1 to print the logs in a new file each time the diagnostics run

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48144

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/diagnostics/steps/step1.logConfig.ts
+++ b/src/main/diagnostics/steps/step1.logConfig.ts
@@ -18,8 +18,8 @@ const stepDescriptiveName = 'logConfig';
 const run = async (logger: ElectronLog): Promise<DiagnosticStepResponse> => {
     try {
         const now = new Date();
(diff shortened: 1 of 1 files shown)


src/main/diagnostics/steps/internal/loggerHooks.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const maskMessageDataHook: ElectronLogHookCreator = (logger) => (message, transport) => {
+    if (transport !== logger.transports.file) {
+        return message;
+    }
+
+    for (let i = 0; i < message.data.length; i++) {
+        // Specific keywords
+        if (message.data[i]?.toLowerCase?.().includes('password')) {
+            return false;
```
m1lt0n, line 25: (question) What's the purpose of returning? to avoid logging altogether in case password is in the log? Why not filtering out the password?

m1lt0n (approved):
LGTM! I left only one question.

devinbinnie (approved):

PR #2399, opened 2022-11-17
Add better obfuscator functions that mask all types of data

tboulis:
#### Summary
Added support for obfuscating data inside `string`, `Array` and `Object`

#### Ticket Link
Related to https://mattermost.atlassian.net/browse/MM-48155

#### Release Note
```release-note
NONE
```


m1lt0n (approved):
Nice abstraction! Question: did it came from an actual issue with logs (ie different shapes of data from the message coming from electron-log? I mean between arrays+strings+object)

devinbinnie (approved):

Issue #2335, opened 2022-10-31
Onboarding screen blocked at the step of adding a new server despite the DefaultServerList GPO


cyberm8:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

Despite the configuration of Mattermost servers by GPO, the onboarding screen in version 5.2.0 requires the manual configuration of a server to access the application.

**Environment**
- Operating System: Windows 10 + Windows server 2019 with Mattermost 5.2.0 admx pack
- Mattermost Desktop App version: 5.2.0 (via msi installer)
- Mattermost Server version: 7.2.0

**Steps to reproduce**

- Configure group policy "DefaultServerList" ;
- Install Mattermost Desktop 5.2.0 ;
- Open Mattermost Desktop.

![image](https://user-images.githubusercontent.com/117109145/199072636-8754b662-8233-417e-957d-84d5cac757f5.png)

**Expected behavior**

- The user should not need to configure any servers as these are managed by a GPO ;
- The user should land directly on the discussion channels.

**Observed behavior**

- The user remains stuck on the "Let's connect to a server" onboarding page ;
- It seems to be mandatory to fill in the server information ;
- The server pushed by GPO does appear at the top left, but the button is greyed out.

![image](https://user-images.githubusercontent.com/117109145/199071869-8cb30fcb-a9f8-48b8-8982-5ad022f04db6.png)

**Possible fixes**

N/A.

markatk:
Same issue on our installation. We get quite a lot of complains from our users, I hope this is fixed quite fast.

slate17:
Same Issue here. Hope this gets fixed fast, very annoying for the users.

devinbinnie:
Fixed by https://github.com/mattermost/desktop/pull/2346

kairauer:
I've updated to 5.2.1. Doesn't seem to be fixed yet.

markatk:
For our installation it seems to be fixed with version 5.2.1.

[...]

PR #2408, opened 2022-11-21
[MM-48079] Fix race condition related to registry loading vs. the welcome screen popup

devinbinnie:
#### Summary
On some system, the registry has a tendency to slow a little slowly and may not have its data be fully available until after we've displayed the welcome screen. This can cause the uncloseable popup to be stuck there on systems with only GPO configured servers.

This PR ensures that the app on Windows waits for the registry data to be populated before trying to show the welcome screen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48079
https://github.com/mattermost/desktop/issues/2335

```release-note
Fixed a potential race condition where the welcome screen would get stuck on Windows systems with only GPO-configured servers
```

### DIFF
--- a/src/main/app/intercom.ts
+++ b/src/main/app/intercom.ts
@@ -4,10 +4,11 @@
 import {app, dialog, IpcMainEvent, IpcMainInvokeEvent, Menu} from 'electron';
 import log from 'electron-log';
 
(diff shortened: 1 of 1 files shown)


m1lt0n (approved):
Nice fix!

PR #2409, opened 2022-11-22
[MM-48151] Diagnostics step 6 - Permissions

tboulis:
#### Summary
Add diagnostics step 6 - Permissions checks:
- Notification is supported
- Write access to downloads location
- Write access to log file
- Microphone and Screen permissions check for Mac, Win
- Accessibility check for Mac
- TBA other checks (Suggestions are welcome)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48151

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/diagnostics/index.test.js
+++ b/src/main/diagnostics/index.test.js
@@ -20,6 +20,6 @@ describe('main/diagnostics/index', () => {
 
     it('should count the steps correctly', () => {
         const d = Diagnostics;
-        expect(d.getStepCount()).toBe(6);
+        expect(d.getStepCount()).toBe(7);
     });
 });


m1lt0n (approved):
LGTM! I can't think of other permissions checks (we can always add more if we come up with more ideas in subsequent iterations though :) )

devinbinnie:
A couple that came to mind from reading this: https://www.electronjs.org/docs/latest/api/session#sessetpermissionrequesthandlerhandler

- clipboard-read - Request access to read from the clipboard.
- fullscreen - Request for the app to enter fullscreen mode.
- openExternal - Request to open links in external applications.

Not sure which might be relevant but those seem important to us.

devinbinnie:
> > A couple that came to mind from reading this: https://www.electronjs.org/docs/latest/api/session#sessetpermissionrequesthandlerhandler
> > 
> > * clipboard-read - Request access to read from the clipboard.
> > * fullscreen - Request for the app to enter fullscreen mode.
> > * openExternal - Request to open links in external applications.
> > 
> > Not sure which might be relevant but those seem important to us.
> 
> @devinbinnie this seems to be a permission granting method from the main to the renderer process. This step is mostly related to checking if a permission is granted and not to manage new permission requests. Do I get this wrong maybe?

Gotcha, I didn't fully understand what you were going for. The reason I highlighted these was that I was wondering if it made sense to check if these were already granted as well.

If that goes outside your scope, then don't worry about it.

devinbinnie (approved):

PR #2428, opened 2022-11-30
[MM-48380] Reworked Linux build process to correctly support ARM64

devinbinnie:
#### Summary
For building ARM64 native modules on Linux, you need to manually specify that you're using the cross-compiler or else it won't work correctly. This PR does that specification and reworks the build process to avoid repeating too many of the steps.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48380

```release-note
Fixed an issue with the ARM64 Linux build process.
```


m1lt0n (approved):

Strit:
This does not seem to enable native arm64 compilation.

It will always look for the aarch64 cross-compiler, even though it's not needed when building natively.

So this only enables cross-compilation for arm64.

PR #2441, opened 2022-12-01
[MM-48574] Fixed issue on windows where in some cases window resize does not work

tboulis:
#### Summary
Some users reported that they could not resize the mattermost app. In same cases it was related to the snap feature being on or off (Settings => System => Multitasking). In my case when snap was off the window could not be resized. 

After doing some digging, it seems like the initial `will-resize` event would emit with the existing bounds, causing the `resize` event to not fire which would cause the `throttleResize` function to not reset the `isResizing` boolean.

I did some testing and the fix is to ignore any resize event that fires with the same bounds inside the `will-resize` listener

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48574

#### Device Information
This PR was tested on: PC, Windows 11 Pro `22H2`

#### Release Note
```release-note
Fixed window resize not working on some windows machines
```

### DIFF
--- a/src/main/windows/windowManager.ts
+++ b/src/main/windows/windowManager.ts
@@ -258,7 +258,7 @@ export class WindowManager {
     isResizing = false;
 
     handleWillResizeMainWindow = (event: Event, newBounds: Electron.Rectangle) => {
-        log.debug('WindowManager.handleWillResizeMainWindow');
+        log.silly('WindowManager.handleWillResizeMainWindow');
 
         if (!(this.viewManager && this.mainWindow)) {
             return;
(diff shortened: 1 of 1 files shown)


m1lt0n (approved):
Great catch! Seeing the milestone, does it mean we need to cut an RC3?

src/main/windows/windowManager.ts
```diff
@@ -258,14 +258,23 @@ export class WindowManager {
     isResizing = false;
 
     handleWillResizeMainWindow = (event: Event, newBounds: Electron.Rectangle) => {
-        log.silly('WindowManager.handleWillResizeMainWindow');
+        log.debug('WindowManager.handleWillResizeMainWindow');
```
devinbinnie, line 261: Can we switch this back to `silly`? It's going to generate a lot of noise for people who want just debug logging of events.

devinbinnie (approved):

Issue #2415, opened 2022-11-23
Empty white bar when switching between plugins


catz:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

**Environment**
- Operating System: macOS 13.0.1
- Mattermost Desktop App version: 5.2.1 (25105)
- Mattermost Server version: 7.3.0

**Steps to reproduce**

1. Clone repository  - https://github.com/catz/mattermost-plugin-test (forked from https://github.com/mattermost/mattermost-plugin-test)
2. Plugin source code - see [index.js](https://github.com/catz/mattermost-plugin-test/tree/master/webapp/src/index.js)
3. Deploy plugin to your server - https://developers.mattermost.com/integrate/plugins/developer-workflow/ (my steps: npm install inside webapp, make dist, upload gz manually)
4. Follow steps

![white-bar](https://user-images.githubusercontent.com/472690/203508305-073dd73c-c258-4fd6-ad15-42e1225b22f4.gif)

**Expected behavior**
No blank bar

**Observed behavior**
White blank bar appears

**Possible fixes**

Maybe this issue is related https://github.com/mattermost/desktop/issues/2388

catz:
Hello @tboulis 

No, nothing similar on the web, only desktop app. The strange moment here is I couldn't find the reason neither via application wrapper dev tools nor current server tools.

catz:
As I wrote above I couldn't figure it out. Seems that it's something that is not shown in the debugger.
Can I provide any additional info?

![img1 app wrapper](https://user-images.githubusercontent.com/472690/203567814-f17d66eb-8ebd-48c0-b908-85d09b16271e.jpg)

![img2 app wrapper](https://user-images.githubusercontent.com/472690/203567835-7c1a405c-4f9c-4ae1-bd75-209323689336.jpg)

![img3 current server](https://user-images.githubusercontent.com/472690/203567865-371aca3f-bc0d-47b9-b2d6-c569a5fc5909.jpg)

[...]

PR #2446, opened 2022-12-02
[MM-48766] - Check if url is a plugin url when calculating the window bounds

tboulis:
#### Summary
When the desktop app would navigate into a plugins UI and/or when resizing the window while in that UI, it would display a white bar between the UI and the Desktop application bar which is intended to be used with the "Back" bar but, it had no elements inside it.

The solution is to also check if the url of the UI is not a plugin url when calculating the window height. 

Issue for reference: https://github.com/mattermost/desktop/issues/2415 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48766

#### Device Information
This PR was tested on: PC, Ubuntu 22

#### Screenshots

Before:
[Screencast from 02-12-2022 06:25:30 ΜΜ.webm](https://user-images.githubusercontent.com/24255041/205339739-ab97319e-96a6-4502-ad88-0b02b471cfba.webm)

After:
[Screencast from 02-12-2022 06:26:46 ΜΜ.webm](https://user-images.githubusercontent.com/24255041/205339759-f8213c26-5b4b-40a3-8168-5eec1d6292f0.webm)

#### Release Note
```release-note
Fixed issue with plugin navigation displaying white empty bar
```

### DIFF
--- a/src/main/utils.ts
+++ b/src/main/utils.ts
@@ -58,7 +58,7 @@ export function shouldHaveBackBar(serverUrl: URL | string, inputURL: URL | strin
 
         return false;
     }
-    return !UrlUtils.isTeamUrl(serverUrl, inputURL) && !UrlUtils.isAdminUrl(serverUrl, inputURL);
+    return !UrlUtils.isTeamUrl(serverUrl, inputURL) && !UrlUtils.isAdminUrl(serverUrl, inputURL) && !UrlUtils.isPluginUrl(serverUrl, inputURL);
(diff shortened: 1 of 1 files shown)


Willyfrog:
I'm pretty sure it was at some point, but I think we fixed it at some point, not 100% sure as URLs were always problematic

Willyfrog (approved):

devinbinnie (approved):

Issue #2388, opened 2022-11-15
Title bar menus disappear after server switching


MFHava:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**

**Environment**
- Operating System: Windows 10 22H2 (19045.2130)
- Mattermost Desktop App version: 5.2.1

**Steps to reproduce**
1) Switch between server instances in the desktop client

**Expected behavior**
The title bar menus stay.

**Observed behavior**
The title bar menus disappear, and the title bar becomes inactive.

Before:
![image](https://user-images.githubusercontent.com/17790925/202042364-e71ab9c8-0323-43f3-a90c-fc70ced0c237.png)
After:
![image](https://user-images.githubusercontent.com/17790925/202042422-bf9f7f08-e5d6-4e52-b8c0-44484c867fb2.png)

**Possible fixes**

devinbinnie:
@MFHava None of our team is able to reproduce this. Is this consistently reproducing for you?

If so, before the white bar shows up can you go to View > Developer Tools for Application Wrapper, then reproduce the issue and post the log for us? Thanks!

MFHava:
It's consistent - and happens on two independent systems for me.
Console log right after switching:

`
Starting Mattermost v5.2.1
index_bundle.js:5515 SyntaxError: Invalid regular expression: /ISO C++___TAB_[A-Z]+/: Nothing to repeat (at index_bundle.js:8910:46)
    at String.match (<anonymous>)
    at index_bundle.js:8910:46
    at Array.reduce (<anonymous>)
    at MainPage.render (index_bundle.js:8908:69)
    at gi (index_bundle.js:4822:76)
    at fi (index_bundle.js:4813:10)
    at Rj (index_bundle.js:7129:86)
    at Qj (index_bundle.js:6593:11)
    at Kj (index_bundle.js:6585:27)
    at yj (index_bundle.js:6328:7)
Ci @ index_bundle.js:5515
Xi.c.callback @ index_bundle.js:6019
[...]

PR #2449, opened 2022-12-05
Escape server name when used inside regex

tboulis:
#### Summary
If the name included any special character, it could cause a regex validation to fail. Escaping the `teamName` fixes the issue. Perhaps we could allow only specific characters inside the team name when entered, but it's not necessary at this moment in my opinion. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48463
https://github.com/mattermost/desktop/issues/2388

#### Screenshots
https://recordit.co/u1Pp8fTAxj

#### Release Note
```release-note
Fixed issue with special characters in server name causing the top bar to disappear
```

### DIFF
--- a/src/common/utils/util.test.js
+++ b/src/common/utils/util.test.js
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import Utils from 'common/utils/util';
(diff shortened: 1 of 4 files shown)


devinbinnie (approved):

m1lt0n (commented):
LGTM! Perhaps having a unit test on the util with some input that verifies `escapeRegex` has the intented behaviour would be great.

PR #2464, opened 2022-12-08
[MM-47929] Switch to using mm-desktop:// protocol for navigating through the app

devinbinnie:
#### Summary
In order to effectively separate any origin-based storage system with the integration, I've added a special new protocol to the app called "mm-desktop" that it used as a substitute for the local `file:///` URL we've been using.

It will prefix the files with `mm-desktop://<server-host-name>`, so that Chrome knows that anything loaded under the URL should use it as its origin.

In order to facilitate this, we've had to do some CORS magic in the main process as well, effectively adding a proxy that the renderer process will hit and verify on behalf of the server that any cross-origin requests between the `mm-desktop` origin and the server are okay.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47929

```release-note
NONE
```

### DIFF
--- a/src/main/app/initialize.ts
+++ b/src/main/app/initialize.ts
@@ -3,7 +3,7 @@
 
 import path from 'path';
 
-import {app, ipcMain, session} from 'electron';
+import {app, ipcMain, protocol, session} from 'electron';
 import installExtension, {REACT_DEVELOPER_TOOLS} from 'electron-devtools-installer';
 import isDev from 'electron-is-dev';
 import log from 'electron-log';
@@ -227,6 +227,16 @@ function initializeBeforeAppReady() {
     } else if (mainProtocol) {
         app.setAsDefaultProtocolClient(mainProtocol);
     }
+
+    protocol.registerSchemesAsPrivileged([
+        {
+            scheme: 'mm-desktop',
+            privileges: {
+                standard: true,
+                supportFetchAPI: true,
+            },
+        },
+    ]);
 }
 
 function initializeInterCommunicationEventListeners() {
@@ -270,6 +280,10 @@ function initializeAfterAppReady() {
     updateServerInfos(Config.teams);
     app.setAppUserModelId('Mattermost.Desktop'); // Use explicit AppUserModelID
     const defaultSession = session.defaultSession;
+
+    defaultSession.protocol.registerFileProtocol('mm-desktop', (request, callback) => {
+        callback(request.url.replace(/mm-desktop:\/\/([A-Za-z0-9.]+)\//, '').replace(/#(.+)/, ''));
+    });
 
     WebRequestManager.initialize();
 
--- a/src/main/views/MattermostView.ts
+++ b/src/main/views/MattermostView.ts
@@ -10,6 +10,7 @@ import {
     BrowserViewConstructorOptions,
     BrowserWindow,
     CookiesSetDetails,
+    HeadersReceivedResponse,
     ipcMain,
     IpcMainEvent,
     OnBeforeSendHeadersListenerDetails,
@@ -41,6 +42,8 @@ export class MattermostView extends EventEmitter {
     isLoggedIn: boolean;
 
     cookies: CookiesSetDetails[];
+    corsHeaders: string[];
+    corsMethods: string[];
 
     constructor(tab: TabView, serverInfo: ServerInfo, window: BrowserWindow, options: BrowserViewConstructorOptions) {
         super();
@@ -69,20 +72,14 @@ export class MattermostView extends EventEmitter {
         // URL handling
         ipcMain.handle(GET_CURRENT_SERVER_URL, () => `${this.tab.server.url}`);
         WebRequestManager.rewriteURL(
-            new RegExp(`file:///${path.resolve('/').replace('\\', '/').replace('/', '')}(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
+            new RegExp(`mm-desktop://${this.tab.server.url.host}(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
             `${this.tab.server.url}/$2/$3`,
             this.view.webContents.id,
         );
 
         WebRequestManager.rewriteURL(
-            new RegExp(`file://(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
-            `${this.tab.server.url}/$2/$3`,
-            this.view.webContents.id,
-        );
-
-        WebRequestManager.rewriteURL(
-            new RegExp(`file:///${path.resolve('/').replace('\\', '/')}(\\?.+)?$`, 'g'),
-            `${getLocalURLString('index.html')}$1`,
+            new RegExp(`mm-desktop://${this.tab.server.url.host}${path.resolve('/').replace('\\', '/')}(\\?.+)?$`, 'g'),
+            `${getLocalURLString('mattermost.html')}$1`,
             this.view.webContents.id,
         );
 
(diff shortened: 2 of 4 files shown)


src/main/views/MattermostView.ts
```diff
@@ -96,7 +93,64 @@ export class MattermostView extends EventEmitter {
...
+
+        // CORS
+        this.corsHeaders = [];
+        this.corsMethods = [];
+        WebRequestManager.onRequestHeaders(this.extractCORSHeaders, this.view.webContents.id);
+        WebRequestManager.onResponseHeaders(this.addCORSResponseHeader, this.view.webContents.id);
+    }
+
+    private extractCORSHeaders = (details: OnBeforeSendHeadersListenerDetails) => {
+        if (!details.url.match(new RegExp(`${this.tab.server.url.origin}/(.+)`))) {
```
jupenur, line 106: The RegExp needs to be anchored to the beginning of the string, otherwise a request to a crafted URL like `https://example.com/?https://community.mattermost.com/` matches. Dots and square brackets (valid in IPv6 URLs) in the origin also need to be escaped. This applies everywhere where this type of check is being used.

Also personally I would prefer `RegExp.prototype.test` over `String.prototype.match` where a boolean is expected.

Really this shouldn't be using RegExps at all, though.

src/main/app/initialize.ts
```diff
@@ -271,6 +281,10 @@ function initializeAfterAppReady() {
     app.setAppUserModelId('Mattermost.Desktop'); // Use explicit AppUserModelID
     const defaultSession = session.defaultSession;
 
+    defaultSession.protocol.registerFileProtocol('mm-desktop', (request, callback) => {
+        callback(request.url.replace(/mm-desktop:\/\/([A-Za-z0-9.]+)\//, '').replace(/#(.+)/, ''));
```
jupenur, line 285: Missing anchoring here as well.

jupenur (changes requested):
Most of the regexps seem to be missing anchoring, and really you should be avoiding regexps in general.

For a proper review I need to test a bunch of things, so can you make sure this builds properly on CI first? My local dev environment is WSL and it can't build the desktop app.

devinbinnie:
> Most of the regexps seem to be missing anchoring, and really you should be avoiding regexps in general.
> 
> For a proper review I need to test a bunch of things, so can you make sure this builds properly on CI first? My local dev environment is WSL and it can't build the desktop app.

Thanks for the review @jupenur. I've updated the regex's with anchors and fixed the tests.

As for not using regexes at all, can you think of another way to match the URL without them? A simple `startsWith` would work in a few cases but we'd be adding a lot more cases and doesn't really solve the issue. That's all I can think of at the moment.

jupenur:
> As for not using regexes at all, can you think of another way to match the URL without them?

Most of these checks will be looking at two things: the origin and some arbitrary features of the path. Using a regex is a neat way of tackling both at once, but really they're two different things. The origin actually has relevance for security, because if we mix up two different origins, headers get attached to the wrong request, cookies end up in the wrong place, or the wrong URL gets rewritten. If we mix up the path, nothing dramatic will happen; the app might not behave entirely as intended, but it also won't be sending anyone's credentials to an attacker's server.

For that reason my recommendation would be to actually check the two separately. For the origin you should ideally be using a URL parser, or if for some reason one isn't available, the simplest possible `indexOf` / `startsWith` style check. For the path (and any other stuff) using a regexp is fine, although I think in most cases something else will be more efficient: for these types of blocking network event handlers, performance can actually be fairly important and compiling and evaluating multiple regexps every time can add up.

m1lt0n (approved):
Looks good to me (code-wise), the important review here is the one from Juho, to see if there are security concerns with the new solution as well.

devinbinnie:
@tboulis @m1lt0n @jupenur Did a bit of a re-work here, now we check for the host and protocol separately from the rest of the URL before parsing and rewriting. Hopefully that makes sense (would have used `origin` but apparently for non-standard URL schemes it's always null 🙄)

Asking for re-reviews across the board to make sure everyone's on board here.

src/main/webRequest/webRequestManager.ts
```diff
@@ -45,12 +47,25 @@ export class WebRequestManager {
...
+
+            if (parsedURL.host !== host) {
+                return {};
+            }
+
+            if (parsedURL.protocol !== protocol) {
+                return {};
+            }
+
+            if (!parsedURL.pathname.match(pathnameRegex)) {
```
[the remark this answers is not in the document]
devinbinnie: I just separated them for clarity's sake

src/main/views/MattermostView.ts
```diff
@@ -183,28 +177,65 @@ export class MattermostView extends EventEmitter {
         });
     }
 
-    get serverUrl() {
-        let url = `${this.tab.server.url}`;
-        if (url.endsWith('/')) {
-            url = url.slice(0, url.length - 1);
+    private extractCORSHeaders = (details: OnBeforeSendHeadersListenerDetails) => {
+        if (!details.url.startsWith(this.tab.server.url.origin)) {
```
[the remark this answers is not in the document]
devinbinnie: Probably, but if the server URL is undefined there's likely a bigger problem going on.

m1lt0n (approved):

jupenur:
> would have used `origin` but apparently for non-standard URL schemes it's always null 🙄

This is is accurate, but you're explicitly declaring the scheme as standard so origin should work. I just tested it with the build from CircleCI: 

```
> new URL("mm-desktop://potato/path").origin
'mm-desktop://potato'
> new URL("xx-desktop://potato/path").origin
'null'
```

jupenur (commented):
None of the `mm-desktop:` URLs are resolving for me:

```
(node:17776) electron: Failed to load URL: mm-desktop://community.mattermost.com/playbooks/ with error: ERR_FILE_NOT_FOUND
```
The path passed to the `registerFileProtocol` callback looks correct though: `C:/Users/Juho%20Nurminen/Downloads/mattermost-desktop-5.3.0-develop.1-win64/resources/app.asar/renderer/mattermost.html`

devinbinnie:
> > would have used `origin` but apparently for non-standard URL schemes it's always null 🙄
> 
> This is is accurate, but you're explicitly declaring the scheme as standard so origin should work. I just tested it with the build from CircleCI:
> 
> ```
> > new URL("mm-desktop://potato/path").origin
> 'mm-desktop://potato'
> > new URL("xx-desktop://potato/path").origin
> 'null'
> ```
> 
> Edit: nvm, looks like it only works correctly in the renderer context ¯\_(ツ)_/¯

Oh, Electron you silly framework XD

devinbinnie:
> None of the `mm-desktop:` URLs are resolving for me:
> 
> ```
> (node:17776) electron: Failed to load URL: mm-desktop://community.mattermost.com/playbooks/ with error: ERR_FILE_NOT_FOUND
> ```
> 
> The path passed to the `registerFileProtocol` callback looks correct though: `C:/Users/Juho%20Nurminen/Downloads/mattermost-desktop-5.3.0-develop.1-win64/resources/app.asar/renderer/mattermost.html`

Ahh, must be something wrong with the release build version. I'll have a look.

src/main/views/MattermostView.ts
```diff
@@ -183,28 +177,65 @@ export class MattermostView extends EventEmitter {
...
+        });
+        if (!this.corsMethods.includes(details.requestHeaders['Access-Control-Request-Method'])) {
+            this.corsMethods.push(details.requestHeaders['Access-Control-Request-Method']);
+        }
+
+        return {};
+    }
+
+    private addCORSResponseHeader = (details: OnHeadersReceivedListenerDetails): HeadersReceivedResponse => {
+        if (!details.url.startsWith(this.tab.server.url.origin)) {
```
jupenur, line 207: This check is broken and can be bypassed with XHR and credentials in URLs. If a user adds two servers, `mattermost.evil.com` and `community.mattermost.com`, the former can then do this:

```js
let xhr = new XMLHttpRequest();
xhr.open("GET", "https://mattermost.evil.com@community.mattermost.com/any/path");
xhr.send(null);
```

CORS headers will get added and the response will come through to the attacker. This needs to be fixed before merging.

jupenur (changes requested):
SSO still allows evil tricks with navigation and that also needs to be fixed prior to release. And ideally we'd also fix the thing where after navigating to an origin that's not the current server URL, the origin then loads scripts from that URL.

But with the CORS thing it looks like this still deserves more attention before merging anyway. It's just too complex. I'll have to continue next week.

src/main/views/MattermostView.ts
devinbinnie: I'll replace this with an exact origin check, that should fix this bug.

jupenur (approved):
Spotted another major issue that was introduced in this PR: https://github.com/mattermost/desktop/pull/2343

Cookies are not tied in any way to a specific origin, so if a webapp instance loads an external image, any cookies set on the server will be sent to that external host as well, and any cookies the external host sets get defined on the Mattermost instance.

Out-of-band we discussed with @devinbinnie regarding a strategy for reviewing all of this and I'm going to approve this now still knowing there are outstanding security issues. Later we'll do a proper security review once the feature branch is ready to be merged to master.

src/main/views/MattermostView.ts
```diff
@@ -113,28 +114,15 @@ export class MattermostView extends EventEmitter {
...
-            new RegExp(`file:///${path.resolve('/').replace('\\', '/').replace('/', '')}(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
-            `${this.tab.server.url}/$2/$3`,
-            this.view.webContents.id,
-        );
-
-        WebRequestManager.rewriteURL(
-            new RegExp(`file://(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
+            this.tab.server.url.host,
+            'mm-desktop:',
+            new RegExp(`^(${this.tab.server.url.pathname})?/(api|static|plugins)/(.*)`, 'g'),
```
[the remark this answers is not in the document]
devinbinnie: Right, I was being overzealous here. Will remove the 'g'

devinbinnie:
> Spotted another major issue that was introduced in this PR: #2343
> 
> Cookies are not tied in any way to a specific origin, so if a webapp instance loads an external image, any cookies set on the server will be sent to that external host as well, and any cookies the external host sets get defined on the Mattermost instance.

Created a bug ticket here to deal with this issue after this PR is merged: https://mattermost.atlassian.net/browse/MM-49712

PR #2471, opened 2022-12-13
Replace deprecated "gpu-process-crashed" with "child-process-gone"

tboulis:
#### Summary
I also passed the details as an argument and not concat them in the string in order to be displayed in the logs

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49039

#### Screenshots
5497A918-364C-4750-8CC4-119D4567749F

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/app/initialize.ts
+++ b/src/main/app/initialize.ts
@@ -63,10 +63,10 @@ import {
     handleAppBeforeQuit,
     handleAppBrowserWindowCreated,
     handleAppCertificateError,
-    handleAppGPUProcessCrashed,
     handleAppSecondInstance,
     handleAppWillFinishLaunching,
     handleAppWindowAllClosed,
+    handleChildProcessGone,
 } from './app';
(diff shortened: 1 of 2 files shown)


src/main/app/initialize.ts
```diff
@@ -183,7 +183,7 @@ function initializeAppEventListeners() {
     app.on('before-quit', handleAppBeforeQuit);
     app.on('certificate-error', handleAppCertificateError);
     app.on('select-client-certificate', CertificateManager.handleSelectCertificate);
-    app.on('gpu-process-crashed', handleAppGPUProcessCrashed);
+    app.on('child-process-gone', handleChildProcessGone);
```
devinbinnie, line 186: Can we handle both events using this handler? Not sure if they differ at all but just to make sure we cover all bases.

devinbinnie (changes requested):
Looks good! Just one request :P

devinbinnie (approved):

m1lt0n (approved):

PR #2474, opened 2022-12-15
PoC performance tests in CI

m1lt0n:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


e2e/performance/settings.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+    describe('Options', () => {
+        describe('Time to open settings window', () => {
+            it('Should open settings window within expected duration', async () => {
+                const expected = (process.platform === 'win32' || process.platform === 'linux');
+                this.app.evaluate(({ipcMain}, showWindow) => {
+                    ipcMain.emit(showWindow);
+                }, SHOW_SETTINGS_WINDOW);
+
+                const t0 = performance.now();
```
m1lt0n, line 42: TODO: this could be abstracted into a helper function that measures and runs the logic below multiple times (with a reset state (in this case close the settings window) after each run" and make calculations of avg and perhaps deviation.

e2e/performance/settings.test.js
```diff
@@ -0,0 +1,59 @@
...
+                });
+                await settingsWindow.waitForSelector('.settingsPage.container');
+                await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
+
+                const t1 = performance.now();
+                const duration = t1 - t0;
+
+                console.log(`Duration: ${duration}ms`);
+
+                duration.should.lessThanOrEqual(500);
```
m1lt0n, line 55: This is hard-coded there (the baseline), but can have a +/- expected deviation to not be flaky or the baselines can be stored somewhere (repo or in a db/s3 etc)

e2e/performance/settings.test.js
```diff
@@ -0,0 +1,59 @@
...
+                const settingsWindow = await this.app.waitForEvent('window', {
+                    predicate: (window) => window.url().includes('settings'),
+                });
+                await settingsWindow.waitForSelector('.settingsPage.container');
+                await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
+
+                const t1 = performance.now();
+                const duration = t1 - t0;
+
+                console.log(`Duration: ${duration}ms`);
```
m1lt0n, line 53: TODO: the web platform team is displaying results in a table, perhaps something similar vs a random console log would make sense for having all results in a single place

e2e/performance/settings.test.js
```diff
@@ -0,0 +1,75 @@
...
+            await settingsWindow.waitForSelector('.settingsPage.container');
+            await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
+
+            const t1 = performance.now();
+            const duration = t1 - t0;
+
+            duration.should.lessThanOrEqual(500);
+        });
+
+        it('Should happen within expected duration failing example', async () => {
```
m1lt0n, line 55: Note: this is the same test as above, just with an unreasonable duration expectation, so that we have a failing test for sure to check the report

.circleci/config.yml
```diff
@@ -156,7 +156,11 @@ jobs:
           key: lint-{{ arch }}-{{ .Branch }}-{{ checksum "package-lock.json" }}
           paths:
             - "node_modules"
-
+  performance-tests:
+    executor: check-image
+    steps:
+      - checkout
+      - run: npm run test:e2e:performance || true
```
m1lt0n, line 163: the true is to not fail the pipeline (at least in this initial iteration) when performance tests fail

e2e/performance/settings.test.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    describe('Open settings window', () => {
+        it('Should happen within expected duration', async () => {
+            const expected = (process.platform === 'win32' || process.platform === 'linux');
+            this.app.evaluate(({ipcMain}, showWindow) => {
+                ipcMain.emit(showWindow);
+            }, SHOW_SETTINGS_WINDOW);
+
+            const t0 = performance.now();
```
m1lt0n, line 41: That's just a quick and dirty example. We can (as e.g. web did) extract a performance measurement helper that runs the test multiple times and calculates the average (and potentially deviations), which will be reusable for all tests.

e2e/performance/settings.test.js
```diff
@@ -0,0 +1,75 @@
...
+            const settingsWindow = await this.app.waitForEvent('window', {
+                predicate: (window) => window.url().includes('settings'),
+            });
+            await settingsWindow.waitForSelector('.settingsPage.container');
+            await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
+
+            const t1 = performance.now();
+            const duration = t1 - t0;
+
+            duration.should.lessThanOrEqual(500);
```
m1lt0n, line 52: For a first iteration this could be hard-coded, but potentially in the future we may want to have it be calculated based on an average e.g. of the runs of the latest days/weeks or something, so that it doesn't need to be updated manually and frequently (unless we consider it the maximum acceptable duration and we care only that there is no regression against this).

e2e/performance/app.test.js
```diff
@@ -0,0 +1,39 @@
...
+// See LICENSE.txt for license information.
+
+'use strict';
+
+const env = require('../modules/environment');
+const {asyncSleep} = require('../modules/utils');
+
+describe('startup/app', function desc() {
+    this.timeout(30000);
+    this.slow(3000);
```
m1lt0n, line 11: @tboulis regarding the use of `.slow`, I see the same issues I've mentioned before (in using the e2e tests report):

1. The test/assertion relies on one run of the test, so it can be flaky
2. If the body of the test includes some setup steps (e.g. logging in etc), it will take those into account in its durations, too (this could be avoided if people writing performance tests are aware of this fact and include only the core of the test/the performance part in the test and everything else in before/after hooks, but it has a higher chance for errors perhaps).

e2e/performance/app.test.js
```diff
@@ -0,0 +1,39 @@
...
+            await this.app.close();
+        }
+        await env.clearElectronInstances();
+        env.createTestUserDataDir();
+        env.cleanTestConfig();
+        this.app = await env.getApp();
+        await asyncSleep(500);
+        const welcomeScreenModal = this.app.windows().find((window) => window.url().includes('welcomeScreen'));
+        const modalButton = await welcomeScreenModal.innerText('.WelcomeScreen .WelcomeScreen__button');
+        modalButton.should.equal('Get Started');
```
m1lt0n, line 37: @tboulis 

An example of running the test multiple times and measure the results (based on the previous test that was in place in this draft, but could be applied to this example):

```js
    const measurePerformance = async (beforeRun, test, afterRun, numberOfRuns, durationThreshold) => {
        const testRuns = [];
        for (let i = 0; i < numberOfRuns; i++) {
            testRuns.push((async () => {
                await beforeRun();

                const t0 = performance.now();
                await test();
                const t1 = performance.now();

                await afterRun();

                return t1 - t0;
            })());
        }

        const durations = await Promise.all(testRuns);
        const avgDuration = durations.reduce((a, b) => a + b, 0) / durations.length;
        avgDuration.should.lessThanOrEqual(durationThreshold);
    };

    describe('Open settings window', () => {
        it('Should happen within expected duration', async () => {
            const expected = (process.platform === 'win32' || process.platform === 'linux');
            let settingsWindow;

            await measurePerformance(
                async () => {
                    this.app.evaluate(({ipcMain}, showWindow) => {
                        ipcMain.emit(showWindow);
                    }, SHOW_SETTINGS_WINDOW);
                },
                async () => {
                    settingsWindow = await this.app.waitForEvent('window', {
                        predicate: (window) => window.url().includes('settings'),
                    });
                    await settingsWindow.waitForSelector('.settingsPage.container');
                    await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
                },
                async () => {
                    await settingsWindow.close();
                },
                10,
                500,
            );
        });
    });
```

The above uses `performance.now` which is available and runs the test multiple times, calculating an average duration of each run. The logic in that sense can be easily extracted in a helper (similar to how web app has `measurePerformance`) that can be reused by anyone. The pros of the approach is that we have access to the individual runs (so in the future we can do more advanced stuff, like having a deviation/percentiles of the runs instead of just averages, etc).

e2e/performance/app.test.js
m1lt0n: Nice :) If there's a way to run the same test multiple times and compare to their average, I'm 0/5 in terms of approaches. In that case I guess it won't be with slow though I guess? as it will need to get the outcome of the report, calculate averages or something and compare to something (so it needs to happen in the `checkReport` file that you have created?

e2e/performance/app.test.js
m1lt0n: Yep, I ran it (I don't know if the logic is fully correct, it's a quick thing I wrote). It runs and calculates 10 durations and averages them out.

m1lt0n: I don't have a particular preference (I'm 0/5) for one over the other approach (as I mentioned in the other comment where you suggest the loop on `it`). If you believe looping the test and adjusting the `checkReport` is a better approach, it's perfectly fine by me :)

Regarding the code, I can commit it (if you prefer the example I mentioned over the approach you suggest with `slow`), otherwise feel free to continue with the `slow` and looping over the tests approach if you feel it's more robust. Both accomplish the same purpose, so again 0/5.

m1lt0n: @tboulis with the multiple tests (looping over it) that you mention, perhaps the results may be more reliable. I've played around with the idea running this:

```js
    describe('Open settings window', () => {
        const durations = [];

        for (let i = 0; i < 3; i++) {
            it('Should happen within expected duration', async () => {
                const expected = (process.platform === 'win32' || process.platform === 'linux');

                this.app.evaluate(({ipcMain}, showWindow) => {
                    ipcMain.emit(showWindow);
                }, SHOW_SETTINGS_WINDOW);

                const t0 = performance.now();
                const settingsWindow = await this.app.waitForEvent('window', {
                    predicate: (window) => window.url().includes('settings'),
                });
                await settingsWindow.waitForSelector('.settingsPage.container');
                await settingsWindow.waitForSelector('#inputAutoStart', {state: expected ? 'attached' : 'detached'});
                const t1 = performance.now();

                durations.push(t1 - t0);
                await settingsWindow.close();

                if (durations.length === 3) {
                    const avgDuration = durations.reduce((a, b) => a + b, 0) / durations.length;
                    avgDuration.should.lessThanOrEqual(300);
                }
            });
        }
    });
```

And it runs one test after the other etc, so it works quite well as well (I have no idea if this can be abstracted in a similar way to a measurePerformance thing for others to use. Leaving that here to continue the discussion on the different approaches :)

package.json
```diff
@@ -188,40 +203,28 @@
...
     "shebang-loader": "0.0.1",
     "style-loader": "3.3.1",
     "ts-prune": "0.10.3",
     "typescript": "4.6.3",
+    "valid-url": "1.0.9",
     "webpack": "5.71.0",
     "webpack-cli": "4.10.0",
     "webpack-dev-server": "4.8.0",
     "webpack-merge": "5.8.0",
-    "@mattermost/compass-icons": "0.1.32",
```
devinbinnie, line 203: Since this is a new feature (and could be reverted if something goes wrong) can we split this fix into a separate PR?

webpack.config.performance.test.js
```diff
@@ -0,0 +1,53 @@
...
+const glob = require('glob');
+const {merge} = require('webpack-merge');
+
+const base = require('./webpack.config.base');
+
+const WEBSERVER_PORT = process.env.WEBSERVER_PORT ?? 9001;
+
+module.exports = merge(base, {
+    entry: {
+        e2e: glob.sync('./e2e/performance/**/*.test.js'),
```
devinbinnie, line 19: This covers a folder already covered by `webpack.config.test.js`, so that file would also run everything in here. Should we exclude these tests from that file?

e2e/performance/report-check.js
```diff
@@ -0,0 +1,38 @@
...
+        });
+        process.exitCode = slowTests.length;
+        return;
+    }
+
+    console.log('No slow tests detected');
+    process.exitCode = 0;
+}
+
+checkReport();
```
devinbinnie, line 38: This runs right on import? Wouldn't think wants to run at the end of the tests?

devinbinnie (commented):
Looks like there's still some work to be done here, let me know if I can help at all :)

webpack.config.performance.test.js
devinbinnie: Ahh you're right, nevermind :)

package.json
devinbinnie: Oh I'm okay with this, I just wanted to separate it out into a separate PR, since it's more of a cleanup step that's not in scope.

devinbinnie (approved):

PR #2475, opened 2022-12-15
[MM-49082] Hide menu bar in calls popout window

streamer45:
#### Summary

PR brings back the title bar to the calls popout window (when global widget is enabled, Desktop >= 5.3) which was originally removed to avoid showing the menu bar. 

At the time I wasn't aware that there was a simple way to customize a window opened from the renderer side but using [`webContents.setWindowOpenHandler()`](https://www.electronjs.org/docs/latest/api/web-contents#contentssetwindowopenhandlerhandler) is working nicely and allows us to hide the menu bar while retaining the title bar which is useful for easily resizing, closing and dragging the window.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49082

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/282

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -82,6 +82,15 @@ export default class CallsWidgetWindow extends EventEmitter {
         ipcMain.on(CALLS_WIDGET_SHARE_SCREEN, this.onShareScreen);
         ipcMain.on(CALLS_JOINED_CALL, this.onJoinedCall);
 
+        this.win.webContents.setWindowOpenHandler(() => {
+            return {
+                action: 'allow',
+                overrideBrowserWindowOptions: {
+                    autoHideMenuBar: true,
+                },
+            };
+        });
+
         this.load();
     }
 
--- a/src/main/windows/callsWidgetWindow.test.js
+++ b/src/main/windows/callsWidgetWindow.test.js
@@ -51,6 +51,9 @@ describe('main/windows/callsWidgetWindow', () => {
         baseWindow.setBackgroundColor = jest.fn();
         baseWindow.setMenuBarVisibility = jest.fn();
         baseWindow.setBounds = jest.fn();
+        baseWindow.webContents = {
+            setWindowOpenHandler: jest.fn(),
+        };
 
         beforeEach(() => {
             mainWindow.getBounds.mockImplementation(() => {
@@ -138,6 +141,7 @@ describe('main/windows/callsWidgetWindow', () => {
             });
 
             baseWindow.webContents = {
+                ...baseWindow.webContents,
                 openDevTools: jest.fn(),
             };
 
@@ -258,6 +262,7 @@ describe('main/windows/callsWidgetWindow', () => {
 
         it('onShareScreen', () => {
             baseWindow.webContents = {
+                ...baseWindow.webContents,
                 send: jest.fn(),
             };
 
@@ -272,6 +277,7 @@ describe('main/windows/callsWidgetWindow', () => {
 
         it('onJoinedCall', () => {
             baseWindow.webContents = {
+                ...baseWindow.webContents,
                 send: jest.fn(),
             };
 


devinbinnie (approved):

src/main/windows/callsWidgetWindow.ts
```diff
@@ -82,6 +82,15 @@ export default class CallsWidgetWindow extends EventEmitter {
         ipcMain.on(CALLS_WIDGET_SHARE_SCREEN, this.onShareScreen);
         ipcMain.on(CALLS_JOINED_CALL, this.onJoinedCall);
 
+        this.win.webContents.setWindowOpenHandler(() => {
+            return {
+                action: 'allow',
+                overrideBrowserWindowOptions: {
+                    autoHideMenuBar: true,
```
cpoile, line 89: Does autoHide imply that you can trigger the menu somehow? If so, can we prevent that? Maybe Devin knows?

cpoile (approved):

src/main/windows/callsWidgetWindow.ts
cpoile: Sorry, forgot to tag you @devinbinnie

streamer45: According to the documentation it can be open by pressing the `Alt` key but I tested this to make sure it didn't interfere with keyboard shortcuts and couldn't get it to show, probably because we are overriding it already.

devinbinnie: I know on Linux sometimes the menu needs to be specifically removed. @streamer45 Did you test on Linux as well?

streamer45: @devinbinnie I did, although on Linux the single `ALT` key works (shows the menu which by default is hidden) but shortcuts work as well. The only OS I couldn't try would be Windows since I don't have a machine with that.

If we could specify `isMenuBarVisible` from within `overrideBrowserWindowOptions` it may be best but looking at documentation I am not sure it's possible. I guess as long as the menu hidden by default it shouldn't be a huge deal but let me know.

devinbinnie: @streamer45 This is the fix I did for Linux on the main window, you might want to do the same thing for the calls widget: https://github.com/mattermost/desktop/blob/master/src/main/windows/mainWindow.ts#L224

Windows I think should be fine, but I can quickly test it if you'd like.

streamer45: I am just not sure we can do it, we don't really have direct access to the window. I could do it on the parent but I am not sure that would transfer.

devinbinnie: @streamer45 Not sure what you mean by not having direct access to the window, you create the window here: https://github.com/mattermost/desktop/blob/47edeea60157b420da146af5ca49a79986bc5096/src/main/windows/callsWidgetWindow.ts#L60

Can we not add the same listener above?

streamer45: @devinbinnie That's the parent window though. The window we are targeting here is opened through JS code at https://github.com/mattermost/mattermost-plugin-calls/blob/2bea09af46ba1b370dbfa92106f969297abc0d50/webapp/src/components/call_widget/component.tsx#L1418-L1422

devinbinnie: Ahh gotcha, that's probably a bit of a problem then. If you plan on opening a window in Electron that isn't part of the popout stuff (shown here: https://github.com/mattermost/desktop/blob/47edeea60157b420da146af5ca49a79986bc5096/src/main/views/webContentEvents.ts#L181) then you might want to think about passing that control to Electron instead of trying to do it as part of the plugin itself.

That use case isn't really supported right now because it allows us to open arbitrary Electron windows which we want to avoid.

streamer45: Thanks @devinbinnie, yeah it's a bit tricky because this popout actually relies on the `window.opener` to point to the widget window since the call client and all connectivity live there. 

Given all this should be temporary as we are planning to get rid of this extra window on Desktop I am proposing we disable the `Alt+` bindings in the popup from the plugin side when running on desktop to avoid the menu to show up on Linux but otherwise proceed with the changes here. Would that work?

devinbinnie: Yeah I think that should be fine, go ahead and make those changes.

streamer45: @devinbinnie Changes on plugin side have been merged in https://github.com/mattermost/mattermost-plugin-calls/commit/982bb11f54ab5fa95093a6856c360caee24de4d4 so we should be good once this PR gets in.

streamer45:
@devinbinnie Let me know if this needs further additions.

devinbinnie (approved):
Sorry, lost track of this one.

PR #2476, opened 2022-12-15
Restore original MattermostView functionality

devinbinnie:
#### Summary
With the recent revelations that the BrowserView architecture will have to stay as is for now, I've elected to restore most of the functionality that was removed from MattermostView before further changes are made. This means we can make some more incremental changes as we continue to migrate things over to the integrated app, and we can preserve as much as possible to potentially allow for backwards compatibility.

```release-note
NONE
```


src/main/views/MattermostView.test.js
```diff
@@ -1,9 +1,374 @@
...
+        const window = {on: jest.fn()};
+        const mattermostView = new MattermostView(tabView, {}, window, {});
+
+        beforeEach(() => {
+            Utils.getLocalURLString.mockReturnValue('file://some/file/path/mattermost.html');
+            mattermostView.loadSuccess = jest.fn();
+            mattermostView.loadRetry = jest.fn();
+        });
+
+        // it('should load provided URL when provided', async () => {
```
m1lt0n, line 67: (question) do we intend to bring back the tests, so this is why they're commented (a temporary thing)? If not, let's remove the test cases altogether.

m1lt0n (approved):
👍 left only a comment (non-blocking) for the commented test cases.

src/main/views/MattermostView.test.js
devinbinnie: I need to re-evaluate how the `load()` function is going to work going forward, because we'll always be loading relative to the local app. I'm going to leave them commented for now, and if possible I'd like to bring them back or modify them.

PR #2477, opened 2022-12-15
Re-enable multiple tabs and allow for browser history push between them

devinbinnie:
#### Summary
This includes a few changes to move the browser history push module back into the Desktop App and not use the one provided by the webapp, which should allow navigation between tabs once again.

It also moves some of the functionality out of the MattermostView so that multiple views can exist without conflict.

```release-note
NONE
```


src/main/preload/mattermost.js
```diff
@@ -152,11 +152,11 @@ window.addEventListener('message', ({origin, data = {}} = {}) => {
         }
         break;
     }
-    case 'browser-history-push': {
-        const {path} = message;
-        ipcRenderer.send(BROWSER_HISTORY_PUSH, viewName, path);
-        break;
-    }
+    // case 'browser-history-push': {
```
m1lt0n, line 155: (nitpick) this(same for below (line 286+)) can be deleted instead of commented I guess (we still have the repo history if we want to bring it back).

m1lt0n (approved):
👍 Left only a comment about commented code vs deleted in 2-3 places (non-blocking)

PR #2493, opened 2023-01-09
Remove any existing IPC hanlders/listeners before assigning new ones inside the downloadsManager

tboulis:
#### Summary
When the `UPDATE_PATHS` event is emitted (because of config migration etc), the downloads manager needs to be re-initialized with the new downloads.json file. This is mainly used with MAS installations and it would throw an error because the IPC handlers were not removed before the downloadsManager is re-initialized. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49550

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/main/app/initialize.test.js
+++ b/src/main/app/initialize.test.js
@@ -50,6 +50,8 @@ jest.mock('electron', () => ({
         on: jest.fn(),
         handle: jest.fn(),
         emit: jest.fn(),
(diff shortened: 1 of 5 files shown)


m1lt0n (approved):

devinbinnie (approved):
Works for me!

PR #2505, opened 2023-01-18
[MM-45565] Check if View->Page is destroyed before sending ipc message

tboulis:
#### Summary
It seems that this is an [issue of electron](https://github.com/electron/electron/issues/15725). Despite that, I added check if the view is destroyed before sending any messages to it. 
I couldn't repro so I'm not so sure about the resolution but, [based on the logs](https://mattermost.atlassian.net/browse/MM-45565?focusedCommentId=131683) (main.log) this error seems to be triggered by a destroyed view. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45565

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,7 @@
         "chromedriver",
         "chromedriverlog",
         "deauthorize",
+        "deeplink",
         "deeplinking",
         "diskimage",
         "dont",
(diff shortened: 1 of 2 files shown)


devinbinnie (approved):
Ahh yes, this has been happening a bit more recently. Hopefully avoids an unnecessary crash message when things fail.

m1lt0n (approved):

PR #2530, opened 2023-01-30
[MM-49689] Force `/api/v4/files` links to not open a new window and download correctly

devinbinnie:
#### Summary
With the changing over to using the `mm-desktop` protocol, we end up a cross-origin request when trying to download files from the server. Thus, it will try to open all files in a new window, which isn't ideal.

This PR forces them through the `DownloadsManager` manually, by denying the `new-window` request and manually calling `downloadURL` on the corresponding `webContents`.

@tboulis I'm honestly not totally happy with the way I've done this, but it's the best I could come up with for now. Feel free to provide suggestions to improve.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49689

```release-note
NONE
```

### DIFF
--- a/src/main/downloadsManager.ts
+++ b/src/main/downloadsManager.ts
@@ -49,6 +49,7 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
     progressingItems: Map<string, DownloadItem>;
     downloads: DownloadedItems;
     willDownloadURLs: Map<string, {filePath: string; bookmark?: string}>;
+    overrideUserGesturesURLs: Set<string>;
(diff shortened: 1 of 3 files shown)


m1lt0n (approved):

devinbinnie:
> I see that you use `webContents.downloadURL`. Did you try with `session.downloadURL` and it didn't work?

If we don't specify a `webContents` the handler in the `DownloadsManager` will fail because it requires one to do its magic.

> I wonder if using this in a preload script would eliminate any extra steps in the download manager.

I'd like to avoid exposing this unnecessarily to the preload script, as I don't want to allow a remote server to arbitrarily download files without having any tracking around where it may have come from.

PR #2535, opened 2023-02-02
[MM-50042] Detect Boards using the BuildBoards flag in the client config

devinbinnie:
#### Summary
With the changes to allow Boards to function as a product, newer server versions won't have the Boards plugins installed anymore, thus when the server is added to the Desktop App the Boards tab wouldn't be automatically added.

This PR changes the logic to check for the `BuildBoards` flag in the `ClientConfig` object. If it set to `true` then we show the Boards tab.

Also, made a fix to the API to allow for chunked responses, since `community.mattermost.com` was sending those.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50042

```release-note
Fixed an issue where sometimes the Boards/Playbooks tabs wouldn't appear automatically when a server was added.
```

### DIFF
--- a/src/main/server/serverAPI.test.js
+++ b/src/main/server/serverAPI.test.js
@@ -23,6 +23,9 @@ jest.mock('electron', () => ({
                         if (event === 'data') {
                             responseCallback(url === badDataURL ? '98&H09986t&(*6BV789RhN^t97rb6Ev^*e5v89 re5bg^&' : JSON.stringify(testData));
                         }
+                        if (event === 'end') {
+                            responseCallback();
+                        }
                     }),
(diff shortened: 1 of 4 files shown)


devinbinnie:
> Would fix an issue where the boards is loaded even when it is disabled? https://community.mattermost.com/core/pl/z4erm7jjobbemk4i6y17noh8ga

No that's a separate one. Although, I'm wondering if it makes sense to fix that anyways, since presumably if Boards is installed it will be re-enabled at some point. Though I guess it's also pre-packaged so idk.

m1lt0n (approved):

PR #2536, opened 2023-02-02
[MM-49586] Use location as is when displaying thumbnails inside downloads

tboulis:
#### Summary
Using the `temp` folder of the sandboxed app would cause issues with MAS builds. I tested it by changing the downloads location, replacing files, restarting the app and it seems to work. 
We should test it again once released through Testflight

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49586

#### Release Note
```release-note
Fixed issue with image thumbnails not always displayed in downloads for MAS builds
```

### DIFF
--- a/src/main/views/downloadsDropdownView.ts
+++ b/src/main/views/downloadsDropdownView.ts
@@ -1,8 +1,7 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-import path from 'path';
 
-import {app, BrowserView, BrowserWindow, ipcMain, IpcMainEvent, IpcMainInvokeEvent} from 'electron';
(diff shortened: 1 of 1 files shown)


devinbinnie (approved):
LGTM! Going to add QA to review though since @jgilliam17 was the one who found the original bug working with files on the MAS build.

m1lt0n (approved):

jgilliam17:
@tboulis Can you please link the test builds?

devinbinnie:
> It was Matt afaik. But of course Jelena should also test this!

The original bug was one Jelena found that required us to use security-scoped bookmarks :P

> @tboulis Can you please link the test builds?

Actually this could be an issue, because the bug is only reproducible through MAS/TestFlight builds. I think it might be possible to have us build one locally and send it over for testing, but it could get complicated. We can merge as-is if everyone's okay with that and test it on the next nightly.

PR #2544, opened 2023-02-06
Remove use of `getView()` and replace with `getViewByURL()` where necessary.

devinbinnie:
#### Summary
This is part of a general rework of how we detect and verify server URLs. Currently we were using a rather complicated and inefficient way of trying to detect which server and tab to use from an arbitrary URL, which was exacerbated by the multi-server, multi-tab infrastructure. I'm looking to start finding ways to better verify our server URLs since we still have a few gaps (this bug for example: https://mattermost.atlassian.net/browse/MM-49072)

Luckily, most of the detection we need to do for URLs can be reworking by checking the `webContents` id of whatever renderer process was invoking an event. Examples include pretty much anything we do in the `webContentEvents` module.

For other cases (ie. checking which URL to forward to for multi-tab servers, and deeplinking), I have simplified the check for an arbitrary URL into a function called `getViewByURL()`, which should be a bit more performant and a bit simpler to understand and work with.

```release-note
Improved performance by reducing the number of calls for URL detection.
```


src/common/utils/url.ts
```diff
@@ -178,24 +136,26 @@ export function equalUrlsIgnoringSubpath(url1: URL, url2: URL, ignoreScheme?: bo
...
-function isTrustedURL(url: URL | string, teams: TeamWithTabs[]) {
+function isTrustedURL(url: URL | string, rootURL: URL | string) {
     const parsedURL = parseURL(url);
-    if (!parsedURL) {
+    const rootParsedURL = parseURL(rootURL);
+    if (!parsedURL || !rootParsedURL) {
         return false;
     }
-    return getView(parsedURL, teams) !== null;
+    return (getFormattedPathName(rootParsedURL.pathname) !== '/' && equalUrlsWithSubpath(rootParsedURL, parsedURL)) ||
```
devinbinnie, line 145: This check is here to ensure that whatever URL we are allowing is "trusted", meaning that it belongs to one of the configured servers. We were using `getView()` for this, but there was never a case where we needed to check across servers or tabs, so this is enough.

m1lt0n (approved):

src/common/utils/url.ts
devinbinnie: It's usually enough, but we want to make sure that the server URL matches exactly. It's unlikely, but I could see a case where an MM instance is hosted on a potentially untrusted domain that has a trusted subpath.

PR #2550, opened 2023-02-09
[MM-50411] Fix media permissions failure on Calls window

streamer45:
#### Summary

With changes in https://github.com/mattermost/desktop/pull/2544, `WindowManager.getServerURLFromWebContentsId()` wasn't taking into account the Calls widget window which doesn't really have a view so it would return undefined and eventually fail during a permission check for media in `setPermissionRequestHandler()`.

PR adds a check to make sure that we also match against the web contents id for the calls window, if present.

@cpoile Could you give this a quick try on your system as well?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50411

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.test.js
+++ b/src/main/windows/callsWidgetWindow.test.js
@@ -358,5 +358,26 @@ describe('main/windows/callsWidgetWindow', () => {
             expect(popOut.focus).toHaveBeenCalled();
             expect(popOut.restore).toHaveBeenCalled();
         });
+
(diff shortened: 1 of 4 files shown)


cpoile (approved):
It works, awesome, thank you Claudio!

devinbinnie (approved):
Ahh, didn't realize that the calls widget window would have those events attached, makes sense though. Thanks for fixing that @streamer45!

PR #2553, opened 2023-02-13
Merge and migrate locally-trusted preload scripts to `contextBridge` API

devinbinnie:
#### Summary
Before the inclusion of `contextBridge` we were making use of the `window.postMessage` API for running data back and forth between the preloaded context and the main context of a renderer process. However, it is now much cleaner to use `contextBridge` across the board.

This PR migrates all of our local trusted APIs used by local trusted windows (eg. main window, settings windows, local modals, dropdowns) into one preload script (now called `desktopAPI.js`) and moves all of the calls to the `contextBridge` API, so that now we can simply call a method like `window.desktop.someAPICall()` and it will directly (and safely) allow us to access the information we need.

I did this in anticipation of [MM-50352](https://mattermost.atlassian.net/browse/MM-50352) which will need to add several more calls to the main process to get information about the URLs we use.

```release-note
NONE
```

[MM-50352]: https://mattermost.atlassian.net/browse/MM-50352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


m1lt0n (approved):
Great cleanup!

M-ZubairAhmed (approved):

PR #2556, opened 2023-02-16
[MM-46980] Improve screensharing permissions flow

streamer45:
#### Summary

PR addresses a couple of outstanding issues with screen sharing:

1. After https://github.com/mattermost/desktop/pull/2554 there was still an issue preventing the error message from surfacing up again if dismissed and permissions were missing after the first attempt to share.
2. Since macOS 10.15 Catalina screen permissions need to be explicitly (and manually) granted to an app through the system settings. This has caused a lot of confusion for users on why screen sharing isn't working (see ticket below). On top of that the system isn't really helping since permissions require an app restart to take effect and the prompt to fix them is only showed once per app execution. To help with this we reset permissions on a share request if missing and also open the related system setting pane in an attempt to make it easier for the user (see video below).

#### Video

https://user-images.githubusercontent.com/1832946/219493880-7fc7f30f-faf9-46de-ac0f-8521b19fea99.mov

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46980

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/338

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/utils.ts
+++ b/src/main/utils.ts
@@ -5,6 +5,11 @@
 import path from 'path';
 import fs from 'fs';
 
+import {exec as execOriginal} from 'child_process';
+
+import {promisify} from 'util';
+const exec = promisify(execOriginal);
+
 import {app, BrowserWindow} from 'electron';
 
 import {Args} from 'types/args';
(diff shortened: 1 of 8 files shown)


devinbinnie (approved):

src/main/utils.ts
```diff
@@ -136,3 +141,19 @@ export function shouldIncrementFilename(filepath: string, increment = 0): string
...
+        {timeout: 1000});
+}
+
+export function openScreensharePermissionsSettingsMacOS() {
+    if (process.platform !== 'darwin') {
+        return Promise.resolve();
+    }
+    return exec('open "x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture"',
+        {timeout: 1000});
+}
```
cpoile, line 159: This is awesome...

cpoile (approved):
Wow, really nice.

src/main/utils.ts
streamer45: That's how much I care about you :)

PR #2565, opened 2023-02-27
[MM-50527] Explicitly allow the Mattermost views access to write to the clipboard using Copy Link

devinbinnie:
#### Summary
When upgrading to Electron v23, it seems that a new permission was added to check if we were allowed to write to the clipboard using the Clipboard API. I think the permission has existed for a while (since it didn't come up in the Breaking Changes) but it must have been recently enforced (or a bug was fixed)

This PR makes sure that we allowed that permission when coming from a trusted source. I've also added the calls widget window as a trusted source since it asks to play media sometimes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50527

```release-note
NONE
```

### DIFF
--- a/src/main/app/initialize.ts
+++ b/src/main/app/initialize.ts
@@ -407,7 +407,7 @@ function initializeAfterAppReady() {
                 callback(true);
                 return;
             }
-            if (callsWidgetWindow.popOut && webContents.id === callsWidgetWindow.win.webContents.id) {
+            if (callsWidgetWindow.popOut && webContents.id === callsWidgetWindow.popOut.webContents.id) {
                 callback(true);
                 return;
             }
--- a/src/main/app/initialize.ts
+++ b/src/main/app/initialize.ts
@@ -380,12 +380,15 @@ function initializeAfterAppReady() {
         'notifications',
         'fullscreen',
         'openExternal',
+        'clipboard-sanitized-write',
     ];
 
     // handle permission requests
     // - approve if a supported permission type and the request comes from the renderer or one of the defined servers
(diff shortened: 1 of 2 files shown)


hmhealey (approved):

cpoile (approved):
Thanks! We also need it for copying links from the call thread RHS in the calls expanded view popout, should that be included here as well? (it's `callsWidgetWindow.popout`)

devinbinnie:
> Thanks! We also need it for copying links from the call thread RHS in the calls expanded view popout, should that be included here as well? (it's `callsWidgetWindow.popout`)

Yes it should! I'll add that.

src/main/app/initialize.ts
```diff
@@ -402,9 +402,15 @@ function initializeAfterAppReady() {
...
         const callsWidgetWindow = WindowManager.callsWidgetWindow;
-        if (callsWidgetWindow && webContents.id === callsWidgetWindow.win.webContents.id) {
-            callback(true);
-            return;
+        if (callsWidgetWindow) {
+            if (webContents.id === callsWidgetWindow.win.webContents.id) {
+                callback(true);
+                return;
+            }
+            if (callsWidgetWindow.popOut && webContents.id === callsWidgetWindow.win.webContents.id) {
```
cpoile, line 410: Tested it, it looks like it needs to be:
```suggestion
            if (callsWidgetWindow.popOut && webContents.id === callsWidgetWindow.popOut.webContents.id) {
```

cpoile (approved):
Thanks Devin!

cpoile (approved):

PR #2567, opened 2023-02-27
[MM-50712] Fix local prototype pollution flaw

devinbinnie:
#### Summary
We had a function that would allow an untrusted input to override the type of the object that was inputted via local prototype pollution. This PR explicitly disallows for an object with the key `__proto__` to be included in the obfuscated output.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50712

```release-note
NONE
```

### DIFF
--- a/src/main/diagnostics/steps/internal/loggerHooks.test.js
+++ b/src/main/diagnostics/steps/internal/loggerHooks.test.js
@@ -4,7 +4,7 @@
 import {MASK_EMAIL, MASK_PATH} from 'common/constants';
 
 import {maskMessageDataHook} from './loggerHooks';
-import { obfuscateByType } from './obfuscators';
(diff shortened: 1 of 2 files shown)


esarafianou (approved):

src/main/diagnostics/steps/internal/loggerHooks.test.js
```diff
@@ -4,6 +4,7 @@
 import {MASK_EMAIL, MASK_PATH} from 'common/constants';
 
 import {maskMessageDataHook} from './loggerHooks';
+import { obfuscateByType } from './obfuscators';
```
larkox, line 7:
```suggestion
import {obfuscateByType} from './obfuscators';
```

larkox (approved):

PR #2571, opened 2023-03-01
[MM-50995] Harden Calls events handling

streamer45:
#### Summary

PR implements some stricter checks to make sure Calls events are coming from the expected sources (either server's main view or calls widget window).

@amyblais These fixes would probably need to go in 5.3

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50995
https://mattermost.atlassian.net/browse/MM-50975
https://mattermost.atlassian.net/browse/MM-50982
https://mattermost.atlassian.net/browse/MM-50994

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -92,7 +92,7 @@ export default class CallsWidgetWindow extends EventEmitter {
     }
 
     public getServerName() {
-        return this.config.serverName;
+        return this.mainView.serverInfo.server.name;
     }
 
     public getChannelURL() {
@@ -121,8 +121,8 @@ export default class CallsWidgetWindow extends EventEmitter {
     }
 
     private getWidgetURL() {
-        const u = new url.URL(this.config.siteURL);
-        u.pathname += `/plugins/${CALLS_PLUGIN_ID}/standalone/widget.html`;
+        const u = new url.URL(this.mainView.serverInfo.server.url.toString());
+        u.pathname += `plugins/${CALLS_PLUGIN_ID}/standalone/widget.html`;
         u.searchParams.append('call_id', this.config.callID);
         if (this.config.title) {
             u.searchParams.append('title', this.config.title);
@@ -131,8 +131,13 @@ export default class CallsWidgetWindow extends EventEmitter {
         return u.toString();
     }
 
-    private onResize = (event: IpcMainEvent, msg: CallsWidgetResizeMessage) => {
+    private onResize = (ev: IpcMainEvent, _: string, msg: CallsWidgetResizeMessage) => {
         log.debug('CallsWidgetWindow.onResize', msg);
+
+        if (!this.isAllowedEvent(ev)) {
+            log.warn('CallsWidgetWindow.onResize', 'Disallowed calls event');
+            return;
+        }
 
         const zoomFactor = this.win.webContents.getZoomFactor();
         const currBounds = this.win.getBounds();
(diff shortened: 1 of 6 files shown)


esarafianou (approved):

src/main/windows/callsWidgetWindow.ts
```diff
@@ -121,8 +121,8 @@ export default class CallsWidgetWindow extends EventEmitter {
     }
 
     private getWidgetURL() {
-        const u = new url.URL(this.config.siteURL);
-        u.pathname += `/plugins/${CALLS_PLUGIN_ID}/standalone/widget.html`;
+        const u = new url.URL(this.mainView.serverInfo.server.url.toString());
+        u.pathname += `plugins/${CALLS_PLUGIN_ID}/standalone/widget.html`;
```
devinbinnie, line 125: For server URLs with a pathname (eg. `https://subpath.test.mattermost.com/mattermost`) there will likely not be a trailing slash, so the widget URL here won't work. We have a function in `common/utils/url` called `getFormattedPathName` that should help make this more predictable.

Might also be worth adding a unit test for this case as well.

devinbinnie (changes requested):

devinbinnie (approved):

src/main/windows/windowManager.ts
```diff
@@ -146,35 +157,44 @@ export class WindowManager {
...
     }
 
-    handleCallsLinkClick = (_: IpcMainEvent, msg: CallsLinkClickMessage) => {
+    handleCallsLinkClick = (_: string, msg: CallsLinkClickMessage) => {
         log.debug('WindowManager.handleCallsLinkClick with linkURL', msg.link);
-        this.mainWindow?.focus();
-        const currentView = this.viewManager?.getCurrentView();
-        currentView?.view.webContents.send(BROWSER_HISTORY_PUSH, msg.link);
+
+        if (this.callsWidgetWindow) {
```
cpoile, line 187: This seems like double check, is that intentional?

src/main/windows/windowManager.ts
```diff
@@ -146,35 +157,44 @@ export class WindowManager {
...
+            this.switchServer(this.callsWidgetWindow.getServerName());
```
cpoile, line 188: Ah, thanks

cpoile (approved):
Thanks Claudio! One minor Q, non-blocking.

src/main/windows/windowManager.ts
streamer45: @cpoile What do you mean by double check? It's the usual "if widget window exists do something" pattern. We could use optional chaining but I feel this is clearer. 

Unless you mean if we get there it's likely there's a widget open but I am not sure we can guarantee it won't be closed once we reach that as there's probably some asynchronous logic involved.

cpoile: Yah, the second. Sure, that makes sense, and it doesn't hurt. 👍 
Although... if all of the handlers do this check, we can move it into the `genCallsEventHandler` HOF.

streamer45: I've thought about that but it would break older plugin versions where the global widget is not available so I left it where strictly necessary to keep backwards compatibility. For example, all the logic to get screen sharing sources should work regardless of the global widget. At some point hopefully we can deprecate all that and simplify.

cpoile: Ah, ok. Cool, thanks!

amyblais:
@streamer45 Quick note - automated cherry-picks are currently not working (a ticket is open for Spiros to investigate), so you can cherry-pick this manually to ``release-5.3``.

PR #2573, opened 2023-03-01
[MM-50973] Harden Calls windows URL navigation checks

streamer45:
#### Summary

PR adds some stricter checks on what can be opened/navigated from calls widget and popout windows.

- widget window should only navigate to the widget url and open the popout url.
- popout window should only navigate/open to allowed sources (using `webContentsEventManager.addWebContentsEventListeners`).

@amyblais Similarly to the previous one it would be good to include this in 5.3 as well.

@cpoile Could you give this a pass to make sure I didn't break anything link clicking related?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50973
https://mattermost.atlassian.net/browse/MM-50974

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/common/utils/url.ts
+++ b/src/common/utils/url.ts
@@ -4,7 +4,7 @@
 import {isHttpsUri, isHttpUri, isUri} from 'valid-url';
 
 import buildConfig from 'common/config/buildConfig';
-import {customLoginRegexPaths, nonTeamUrlPaths} from 'common/utils/constants';
+import {customLoginRegexPaths, nonTeamUrlPaths, CALLS_PLUGIN_ID} from 'common/utils/constants';
 
 function isValidURL(testURL: string) {
     return Boolean(isHttpUri(testURL) || isHttpsUri(testURL)) && Boolean(parseURL(testURL));
@@ -184,6 +184,22 @@ function cleanPathName(basePathName: string, pathName: string) {
     return pathName;
 }
 
+function isCallsPopOutURL(serverURL: URL | string, inputURL: URL | string, callID: string) {
+    if (!serverURL || !inputURL || !callID) {
+        return false;
+    }
+
+    const parsedURL = parseURL(inputURL);
+    const server = getServerInfo(serverURL);
+    if (!parsedURL || !server || (!equalUrlsIgnoringSubpath(server.url, parsedURL))) {
+        return false;
+    }
+
+    const regexRule = `^${server.subpath}[A-Za-z0-9-_]+/${CALLS_PLUGIN_ID}/expanded/${callID}$`;
+
+    return new RegExp(regexRule, 'i').test(parsedURL.pathname);
+}
+
 export default {
     isValidURL,
     isValidURI,
@@ -201,4 +217,5 @@ export default {
     isUrlType,
     cleanPathName,
     startsWithProtocol,
+    isCallsPopOutURL,
 };
--- a/src/common/utils/url.ts
+++ b/src/common/utils/url.ts
@@ -98,12 +98,6 @@ function isAdminUrl(serverUrl: URL | string, inputURL: URL | string) {
 }
 
 function isTeamUrl(serverUrl: URL | string, inputURL: URL | string, withApi?: boolean) {
-    const parsedURL = parseURL(inputURL);
-    const server = getServerInfo(serverUrl);
-    if (!parsedURL || !server || (!equalUrlsIgnoringSubpath(server.url, parsedURL))) {
-        return false;
-    }
-
     const paths = [...getManagedResources(), ...nonTeamUrlPaths];
 
     if (withApi) {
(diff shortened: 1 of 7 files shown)


src/common/utils/url.ts
```diff
@@ -184,6 +184,22 @@ function cleanPathName(basePathName: string, pathName: string) {
...
+        return false;
+    }
+
+    const parsedURL = parseURL(inputURL);
+    const server = getServerInfo(serverURL);
+    if (!parsedURL || !server || (!equalUrlsIgnoringSubpath(server.url, parsedURL))) {
+        return false;
+    }
+
+    const regexRule = `^${server.subpath}[A-Za-z0-9-_]+/${CALLS_PLUGIN_ID}/expanded/${callID}$`;
```
streamer45, line 198: Ideally I would have made this static but getting the team name is a bit tricky and if we accept it from the plugin side it would be a breaking change so using a regexp for now which I think it's fine but let me know if there are concerns.

cpoile (approved):
I tested it and it doesn't break the earlier linking changes. I didn't think it would -- from what I understand, the widgetWindow events (those `will-navigate`s, `did-start-navigation`s, `onPopoutOpen` etc) wouldn't have anything to do with the popout (please correct me if I'm wrong). 
Nice work!

streamer45:
> I tested it and it doesn't break the earlier linking changes. I didn't think it would -- from what I understand, the widgetWindow events (those `will-navigate`s, `did-start-navigation`s, `onPopoutOpen` etc) wouldn't have anything to do with the popout (please correct me if I'm wrong). Nice work!

Yes, the only related change was applying `addWebContentsEventListeners` to the popout's webcontents vs only using `generateNewWindowListener`.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -209,7 +220,12 @@ export default class CallsWidgetWindow extends EventEmitter {
...
-    private onPopOutOpen = () => {
+    private onPopOutOpen = ({url}: {url: string}) => {
+        if (!urlUtils.isCallsPopOutURL(this.mainView.serverInfo.server.url, url, this.config.callID)) {
+            log.warn(`CallsWidgetWindow.onPopOutOpen: prevented window open to ${url}`);
+            return {action: 'deny' as const};
+        }
+
         return {
             action: 'allow' as const,
             overrideBrowserWindowOptions: {
```
esarafianou, line 231: Can we swap the order? Deny by default and allow only when `urlUtils.isCallsPopOutURL` is `true` so that we follow secure by default principles.

```suggestion
    private onPopOutOpen = ({url}: {url: string}) => {
        if (urlUtils.isCallsPopOutURL(this.mainView.serverInfo.server.url, url, this.config.callID)) {
            return {
                action: 'allow' as const,
                overrideBrowserWindowOptions: {
                    autoHideMenuBar: true,
                },
            }
        }
        
        log.warn(`CallsWidgetWindow.onPopOutOpen: prevented window open to ${url}`);
        return {action: 'deny' as const};
```

streamer45: Sure. It actually simplifies the logic as we avoid the negation.

src/common/utils/url.ts
devinbinnie: Is there any way we can have this call `isUrlType` instead? I spent some time consolidating all of the URL checks to be based off of that one function, so if we could continue to do that that would be ideal.

streamer45: @devinbinnie PTAL

src/common/utils/url.ts
```diff
@@ -184,6 +184,28 @@ function cleanPathName(basePathName: string, pathName: string) {
     return pathName;
 }
 
+function isCallsPopOutURL(serverURL: URL | string, inputURL: URL | string, callID: string) {
+    if (!serverURL || !inputURL || !callID) {
+        return false;
+    }
+
+    const parsedURL = parseURL(inputURL);
```
devinbinnie, line 192: `isURLType` already does this check. I think we can try and reduce the amount of checks we're doing here to avoid some bloat.

streamer45: @devinbinnie  `isTeamUrl` does the same really. We need to parse to strip the server portion. We could modify `isUrlType` to avoid parsing the input if already an URL and potentially receive the serverInfo object directly. It feels a bit out of scope but happy to do it in here. Also happy to hear alternatives of course.

devinbinnie: I was thinking we don't have to do this check twice:
https://github.com/mattermost/desktop/blob/7cfeebf2a27abd5e1b068364d616e0cb75f187a2/src/common/utils/url.ts#L89
https://github.com/mattermost/desktop/blob/7cfeebf2a27abd5e1b068364d616e0cb75f187a2/src/common/utils/url.ts#L194

streamer45: I guess I was trying to fail early if the server is not matching, same as in `isTeamUrl` but yeah, we could also defer it. Not sure it will improve the bloating that match but it's one less duplicated check in the happy path. I'll have another look.

streamer45: No worries, should be done. I simplified `isTeamUrl` as well since I was there.

devinbinnie (approved):

streamer45:
Will merge and manual cherry-pick as soon as the pipeline is green.

PR #2580, opened 2023-03-03
[MM-50948] Prevent popup windows from redirecting to internal URLs outside of the plugin/managed resource URLs

devinbinnie:
#### Summary
There was a case where a user could click on a link that was able to trick the application into navigating to an arbitrary URL by having it prefixed with the path: `/plugins/..%2F`. From there any internal URL would open in the popup window which has some looser restrictions around navigating to external URLs. This was caused because our URL checker would note that the URL was a plugin URL, but Chrome would not evaluate it as such.

This PR adds a listener on the `will-redirect` event to ensure that those URLs cannot be exploited by making sure that any redirects to internal URLs need to go to the matching plugin/managed resource that is allowed by the application.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50948

```release-note
Fixed an issue where a certain URL form would allow the popup window to navigate to any arbitrary URL
```

### DIFF
--- a/src/main/views/webContentEvents.ts
+++ b/src/main/views/webContentEvents.ts
@@ -187,6 +187,17 @@ export class WebContentsEventManager {
                             spellcheck: (typeof spellcheck === 'undefined' ? true : spellcheck),
                         },
                     });
+                    this.popupWindow.webContents.on('will-redirect', (event, url) => {
+                        const parsedURL = urlUtils.parseURL(url);
+                        if (!parsedURL) {
+                            event.preventDefault();
+                            return;
+                        }
+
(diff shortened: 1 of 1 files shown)


larkox:
Not sure I understand what is happening here, but if I remember correctly, the original reason to do it this way is so plugins like github could go through the oauth flow. Is that still possible with these changes?

devinbinnie:
> Not sure I understand what is happening here, but if I remember correctly, the original reason to do it this way is so plugins like github could go through the oauth flow. Is that still possible with these changes?

Yeah sorry the title is misleading (I just fixed it). The case here is that a URL like `https://community.mattermost.com/plugins/..%2Flogin/sso/saml?redirect_to=https://www.google.com` would still open in the popup window but would redirect to the `/login` route. We don't want that in the popup window because it's allowed to go externally unlike the main app is.

Other redirects should be unaffected.

larkox (approved):

m1lt0n (approved):

jupenur (approved):

PR #2581, opened 2023-03-03
[GH-2534] Refocus the popup window if already exists

### LINKED ISSUE #2534: Plugin support for popup/dialog windows with external URL or content. Window focus issue.
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).

---

Show (focus on) already opened popup on url change. It's hard to understand what's going on when main app is in fullscreen and popup is in background.

**Environment**
- Operating System: macOS
- Mattermost Desktop App version: 5.2.2
- Mattermost Server version: 7.5.1

**Steps to reproduce**

1. Plugin (right panel in our case) opens some url. Popup gets focus.
2. Go back to main app. Popup is in background
3. Plugin opens another url. Some intentional user action like user clicked another button. 

**Expected behavior**

Popup gets focus.

**Observed behavior**

Popup is in background.

**Possible fixes**

Change line 
https://github.com/mattermost/desktop/blob/master/src/main/views/webContentEvents.ts#L195

From 

`this.popupWindow.once('ready-to-show', () => {`

to

`this.popupWindow.on('ready-to-show', () => {`

devinbinnie:
#### Summary
Popup windows do not refocus for the user when clicking on a URL in the app that would open an already existing popup. This PR adds a call to `show()` on the popup window when already open to make sure it's refocused.

#### Ticket Link
Closes #2534 

```release-note
Refocus the popup window if already exists
```


src/main/views/webContentEvents.ts
```diff
@@ -177,7 +177,11 @@ export class WebContentsEventManager {
 
             // TODO: move popups to its own and have more than one.
             if (urlUtils.isPluginUrl(serverURL, parsedURL) || urlUtils.isManagedResource(serverURL, parsedURL)) {
-                if (!this.popupWindow) {
+                if (this.popupWindow) {
+                    this.popupWindow.once('ready-to-show', () => {
+                        this.popupWindow!.show();
```
enahum, line 182: It should not need the bang

enahum (approved):

src/main/views/webContentEvents.ts
devinbinnie: Whoops 😓

devinbinnie: It does because when the callback is called it can't guarantee that the object exists (but it obviously does)

enahum: What about optional chaining instead? `this.popupWindow?.show()`

hmhealey: Since it's written as `this.popupWindow`, I think it might be possible for `this.popupWindow` to change between the the call to `once` and when this callback is actually used. If you were to capture the value like
```ts
const popupWindow = this.popupWindow;
popupWindow.once('ready-to-show', () => popupWindow.show());
```
I think it would be fine without the `!`

hmhealey (approved):

hmhealey (approved):

PR #2591, opened 2023-03-07
[MM-51222] Fix race condition on switching calls

streamer45:
#### Summary

PR fixes a race condition when switching from one call to another (or starting a new call when already in one). Having two Calls clients active at the same time even for just a few moments led to a permission error when accessing the microphone. I am not exactly sure why Electron would prevent that but it's a not a strong issue at the moment since we don't want to allow a user to be in more than one call at the same time. 

The fix here is to wait until the widget window is properly closed and only then proceed with creating a new one.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51222

#### Release Note

```release-note
Fixed a race condition leading to a permissions error when switching calls.
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.test.js
+++ b/src/main/windows/callsWidgetWindow.test.js
@@ -75,13 +75,15 @@ describe('main/windows/callsWidgetWindow', () => {
         baseWindow.setAlwaysOnTop = jest.fn();
         baseWindow.setBackgroundColor = jest.fn();
         baseWindow.setMenuBarVisibility = jest.fn();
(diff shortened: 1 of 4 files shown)


cpoile (approved):
Tested, fixes the problem, thanks Claudio!

devinbinnie (approved):
Does this need to be cherry-picked to the release?

devinbinnie:
> > Does this need to be cherry-picked to the release?
> 
> @devinbinnie It wouldn't hurt if we are still in time to do so. Let me know.

Yep, feel free to merge and cherry-pick over. cc @amyblais

PR #2621, opened 2023-03-15
Fix the issue where we sometimes need to wait for GPO teams properly

devinbinnie:
#### Summary
A while back, there was an issue in which the new server modal or the Welcome modal would show up when the user had GPO teams configured. At the time, it was fixed by checking for the teams to be added later on and not popping the modal until then. What was neglected was that we could just await the call before the rest of the app even loaded, which seems to be the right call and was removed at some point during v4.7.

This PR goes back to doing it the correct way and should be cleaner in general.

```release-note
Fix the issue where we sometimes need to wait for GPO teams properly.
```

### DIFF
--- a/src/common/config/index.ts
+++ b/src/common/config/index.ts
@@ -106,9 +106,17 @@ export class Config extends EventEmitter {
         if (process.platform === 'darwin' || process.platform === 'win32') {
             nativeTheme.on('updated', this.handleUpdateTheme);
         }
-        this.registryConfig = new RegistryConfig();
-        this.registryConfig.once(REGISTRY_READ_EVENT, this.loadRegistry);
-        this.registryConfig.init();
+    }
+
+    initRegistry = () => {
+        return new Promise<void>((resolve) => {
+            this.registryConfig = new RegistryConfig();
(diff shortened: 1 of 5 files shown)


hmhealey (approved):

src/common/config/index.ts
```diff
@@ -106,9 +106,17 @@ export class Config extends EventEmitter {
...
             nativeTheme.on('updated', this.handleUpdateTheme);
         }
-        this.registryConfig = new RegistryConfig();
-        this.registryConfig.once(REGISTRY_READ_EVENT, this.loadRegistry);
-        this.registryConfig.init();
+    }
+
+    initRegistry = () => {
+        return new Promise<void>((resolve) => {
+            this.registryConfig = new RegistryConfig();
```
enahum, line 113: Should we handle calling this function more than one time?

enahum (approved):
Small question

src/common/config/index.ts
devinbinnie: I don't think it's necessary, we don't read the registry more than once when the application is started and once it's started I don't believe there's a way for us to find out if the registry has changed and we need to reload the GPO teams.

So I think it's fine as is for now :)

PR #2629, opened 2023-03-20
[MM-51527] Escape regexp input in isCallsPopOutURL

streamer45:
#### Summary

Looks like regular expressions  in a URL are considered valid so the parsing would not complain and we could potentially execute something undesired. Adding a regexp escaping function, as recommended in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping should fix this.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51527

#### Release Notes

```release-note
NONE
```

### DIFF
--- a/src/common/utils/url.test.js
+++ b/src/common/utils/url.test.js
@@ -329,4 +329,18 @@ describe('common/utils/url', () => {
             expect(urlUtils.isCallsPopOutURL()).toBe(false);
         });
     });
+
+    describe('escapeRegExp', () => {
+        it('simple', () => {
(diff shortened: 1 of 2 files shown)


devinbinnie (approved):
LGTM, but is this something that might affect other RegExps we use for checking for URLs?

cc @esarafianou

devinbinnie:
> @devinbinnie Should we go ahead with merging this while we discuss of potential next steps?

Yep, go for it. We're still waiting on other items to finish QA anyways.

esarafianou:
@devinbinnie missed your comment. I'll take it as an action item

PR #2650, opened 2023-04-02
[MM-51871] Migrate mainWindow and settingsWindow to singletons

devinbinnie:
#### Summary
This PR moves the majority of the functionality of the Main Window, including event listeners and associated `BrowserWindow` to a singleton that can be independently called and imported outside of the `windowManager`. The same is done for the `settingsWindow`

This is a step towards removing most of the `windowManager` monolith and into more modular self-contained pieces.

Related PR: https://github.com/mattermost/desktop/pull/2628

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51871

```release-note
NONE
```


m1lt0n (approved):

src/main/windows/windowManager.ts
```diff
@@ -201,100 +196,49 @@ export class WindowManager {
...
-            this.teamDropdown = new TeamDropdownView(this.mainWindow, Config.teams, Config.darkMode, Config.enableServerManagement);
-            this.downloadsDropdown = new DownloadsDropdownView(this.mainWindow, downloadsManager.getDownloads(), Config.darkMode);
-            this.downloadsDropdownMenu = new DownloadsDropdownMenuView(this.mainWindow, Config.darkMode);
+            this.createMainWindow();
         }
-        this.initializeViewManager();
 
         if (deeplinkingURL) {
-            this.viewManager!.handleDeepLink(deeplinkingURL);
+            this.viewManager?.handleDeepLink(deeplinkingURL);
```
enahum, line 214: 👍

src/main/windows/windowManager.ts
```diff
@@ -464,111 +401,40 @@ export class WindowManager {
...
+            if (settingsWindow) {
+                settingsWindow.focus();
             } else {
-                this.mainWindow!.focus();
+                mainWindow.focus();
             }
-        } else if (this.settingsWindow) {
-            this.settingsWindow.focus();
+        } else if (SettingsWindow.get()) {
+            SettingsWindow.get()?.focus();
```
enahum, line 424: interesting that it asks you to use the optional chaining here, but hey it is what it is, I rather have it optional than with the unwrapping

src/main/app/initialize.test.js
```diff
@@ -157,13 +158,17 @@ jest.mock('main/UserActivityMonitor', () => ({
...
     sendToRenderer: jest.fn(),
     getServerNameByWebContentsId: jest.fn(),
     getServerURLFromWebContentsId: jest.fn(),
 }));
+jest.mock('main/windows/settingsWindow', () => ({
+    show: jest.fn(),
+}));
+jest.mock('main/windows/mainWindow', () => ({
+    get: jest.fn(),
+}));
```
enahum, line 171: I'm seeing this repeating itself.. can you add the mock in a test setup file of sort? we have one in the mobile app that works pretty well.

src/main/app/initialize.ts
```diff
@@ -344,7 +346,7 @@ function initializeAfterAppReady() {
 
     WindowManager.showMainWindow(deeplinkingURL);
 
-    CriticalErrorHandler.setMainWindow(WindowManager.getMainWindow()!);
+    CriticalErrorHandler.setMainWindow(MainWindow.get()!);
```
enahum, line 349: can we safely assume that a main window exists at this point?

src/main/badge.ts
```diff
@@ -2,20 +2,79 @@
...
 const MAX_WIN_COUNT = 99;
 
 let showUnreadBadgeSetting: boolean;
 
+/**
+     * Badge generation for Windows
+     */
+
+function drawBadge(text: string, small: boolean) {
+    const scale = 2; // should rely display dpi
```
enahum, line 24: is there a reason why the display dpi is not being used?

src/main/badge.ts
```diff
@@ -2,20 +2,79 @@
...
+    const size = (small ? 20 : 16) * scale;
+    const canvas = document.createElement('canvas');
+    canvas.setAttribute('width', `${size}`);
+    canvas.setAttribute('height', `${size}`);
+    const ctx = canvas.getContext('2d');
+
+    if (!ctx) {
+        log.error('Could not create canvas context');
```
enahum, line 32: should we log the reason for this or is there a particular reason that we already know this happens?

src/main/badge.ts
```diff
@@ -2,20 +2,79 @@
...
+    ctx.textBaseline = 'middle';
+    ctx.font = (11 * scale) + 'px sans-serif';
+    ctx.fillText(text, size / 2, size / 2, size);
+
+    return canvas.toDataURL();
+}
+
+function createDataURL(win: BrowserWindow, text: string, small: boolean) {
+    // since we don't have a document/canvas object in the main process, we use the webcontents from the window to draw.
+    const safeSmall = Boolean(small);
```
enahum, line 54: what is the reason for this "safety"?

src/main/badge.ts
```diff
@@ -2,20 +2,79 @@
...
+    if (mainWindow) {
+        if (badgeText) {
+            try {
+                const dataUrl = await createDataURL(mainWindow, badgeText, small);
+                overlay = nativeImage.createFromDataURL(dataUrl);
+            } catch (err) {
+                log.error('Could not generate a badge:', err);
+            }
+        }
+        mainWindow.setOverlayIcon(overlay, description);
```
enahum, line 74: so if the overlay fails to create in L#69, so we still want to apply it here?

src/main/diagnostics/steps/step5.browserWindows.ts
```diff
@@ -17,11 +17,11 @@ const stepDescriptiveName = 'BrowserWindowsChecks';
 const run = async (logger: ElectronLog): Promise<DiagnosticStepResponse> => {
     try {
         /** Main window check */
-        if (!windowManager.mainWindowReady) {
+        if (!MainWindow.isReady) {
```
enahum, line 20: is just logging the message that the main window is not ready enough at this point? don't we want to know why is not ready? I'm not sure at this point what is being diagnosed and what do we do with the results of this.

src/main/tray/tray.ts
```diff
@@ -85,9 +86,10 @@ export function setupTray(iconTheme: string) {
 
     trayIcon.setToolTip(app.name);
     trayIcon.on('click', () => {
-        if (WindowManager.mainWindow!.isVisible()) {
-            WindowManager.mainWindow!.blur(); // To move focus to the next top-level window in Windows
-            WindowManager.mainWindow!.hide();
+        const mainWindow = MainWindow.get(true)!;
+        if (mainWindow.isVisible()) {
+            mainWindow.blur(); // To move focus to the next top-level window in Windows
```
enahum, line 91: is this a windows only thing? are there any side effects in other OS' ? should this be wrapped under a platform only condition?

src/main/views/MattermostView.ts
```diff
@@ -118,7 +118,7 @@ export class MattermostView extends EventEmitter {
 
         this.altPressStatus = false;
 
-        this.window.on('blur', () => {
+        MainWindow.get()?.on('blur', () => {
```
enahum, line 121: main window at it's finest ;)

src/main/views/MattermostView.ts
```diff
@@ -235,21 +235,21 @@ export class MattermostView extends EventEmitter {
             this.status = Status.WAITING_MM;
             this.removeLoading = setTimeout(this.setInitialized, MAX_LOADING_SCREEN_SECONDS, true);
             this.emit(LOAD_SUCCESS, this.tab.name, loadURL);
-            this.setBounds(getWindowBoundaries(this.window, shouldHaveBackBar(this.tab.url || '', this.view.webContents.getURL())));
+            this.setBounds(getWindowBoundaries(MainWindow.get()!, shouldHaveBackBar(this.tab.url || '', this.view.webContents.getURL())));
```
enahum, line 238: unwrapping here, is this safe?

src/main/views/downloadsDropdownMenuView.ts
```diff
@@ -192,7 +194,7 @@ export default class DownloadsDropdownMenuView {
     }
 
     getX = () => {
-        const result = (this.windowBounds.width - DOWNLOADS_DROPDOWN_FULL_WIDTH - DOWNLOADS_DROPDOWN_MENU_FULL_WIDTH) + (this.coordinates?.x || 0) + (this.coordinates?.width || 0);
+        const result = (this.windowBounds!.width - DOWNLOADS_DROPDOWN_FULL_WIDTH - DOWNLOADS_DROPDOWN_MENU_FULL_WIDTH) + (this.coordinates?.x || 0) + (this.coordinates?.width || 0);
```
enahum, line 197: unwrapping 😨

src/main/views/downloadsDropdownView.ts
```diff
@@ -152,7 +154,7 @@ export default class DownloadsDropdownView {
     getBounds = (width: number, height: number) => {
         // Must always use integers
         return {
-            x: this.getX(this.windowBounds.width),
+            x: this.getX(this.windowBounds!.width),
```
enahum, line 157: 😨

src/main/views/viewManager.ts
```diff
@@ -422,7 +419,7 @@ export class ViewManager {
 
     setLoadingScreenBounds = () => {
         if (this.loadingScreen) {
-            this.loadingScreen.setBounds(getWindowBoundaries(this.mainWindow));
+            this.loadingScreen.setBounds(getWindowBoundaries(MainWindow.get()!));
```
enahum, line 422: you could add a cons for the main window and avoid unwrapping here adding the check in the if statement

enahum (approved):
Really cool, awesome work.. I've added some questions & comments but will leave the final decision to you.

src/main/windows/mainWindow.ts
```diff
@@ -3,146 +3,236 @@
...
-}
+        if (process.platform === 'linux') {
+            windowOptions.icon = path.join(path.resolve(app.getAppPath(), 'assets'), 'linux', 'app_icon.png');
+        }
+
+        this.win = new BrowserWindow(windowOptions);
+        this.win.setMenuBarVisibility(false);
+
+        if (!this.win) {
+            log.error('unable to create main window');
```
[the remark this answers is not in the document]
devinbinnie: I don't believe so, this should only fail if for some reason the app can't create a window. I'm not even sure where that case happens, maybe it's permissions?

However, you make a good point. Currently all we do is log the error and silently quit, but this should actually throw an error so that the user knows what happened.

src/main/windows/mainWindow.ts
```diff
@@ -3,146 +3,236 @@
...
+                    this.win.webContents.send(SELECT_PREVIOUS_TAB);
+                }
+            }
+        }
+    }
+
+    private onFocus = () => {
+        // Only add shortcuts when window is in focus
+        if (process.platform === 'linux') {
+            globalShortcut.registerAll(['Alt+F', 'Alt+E', 'Alt+V', 'Alt+H', 'Alt+W', 'Alt+P'], () => {
```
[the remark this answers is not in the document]
devinbinnie: I'm against using a constants file for anything not used outside of the file it's in, but I think seperating this into its own `const` makes some sense.

src/main/app/initialize.test.js
devinbinnie: I can, but it's really only used in a few places, and we're mocking different methods in different spots as well. Some files are adding additional methods, and others are mocking that `get` always just returns a blank object (useful for methods just checking if the MainWindow exists or not).

I think we'd just be moving the code bloat around, and I do like to be deliberate about what I mock where, especially when the mocked result has some consequence, so I'm going to leave this one as is for now. Might revisit later.

src/main/badge.ts
devinbinnie: Not sure, I'm guessing it's there since it's possible `canvas.getContext` can return null (probably if you provide the wrong string or something) and if it does we don't want to proceed. I think we log it in case someone asks why the badge isn't there and if so we can point to that code.

src/main/badge.ts
devinbinnie: Not sure, might have been a reason in the past. Will remove for now.

src/main/badge.ts
devinbinnie: Another one I'm not too sure of, I'll move it up for now.

src/main/diagnostics/steps/step5.browserWindows.ts
devinbinnie: I know it has to be visible for this to work, though I'm not sure in what case we're running diagnostics without the window visible. This stuff is a bit out of scope here (it was just updated) and I'm not super familiar with it so I'm going to leave it for now.

src/main/tray/tray.ts
devinbinnie: Another one that was just migrated, I don't remember the original context. Gonna leave it for now.

src/main/views/MattermostView.ts
```diff
@@ -235,21 +235,21 @@ export class MattermostView extends EventEmitter {
...
     }
 
     show = (requestedVisibility?: boolean) => {
         this.hasBeenShown = true;
         const request = typeof requestedVisibility === 'undefined' ? true : requestedVisibility;
         if (request && !this.isVisible) {
-            this.window.addBrowserView(this.view);
-            this.setBounds(getWindowBoundaries(this.window, shouldHaveBackBar(this.tab.url || '', this.view.webContents.getURL())));
+            MainWindow.get()?.addBrowserView(this.view);
+            this.setBounds(getWindowBoundaries(MainWindow.get()!, shouldHaveBackBar(this.tab.url || '', this.view.webContents.getURL())));
```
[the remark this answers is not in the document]
devinbinnie: Ahh thanks for reminding me I meant to check on all the unwrapping I did lol

enahum (approved):

PR #2652, opened 2023-04-02
[MM-51873] Create central logging module for adding prefixes to differentiate logs between modules

devinbinnie:
#### Summary
This PR adds a consistent logging module that decouples `electron-log` from the rest of the application, and gives up more control over how individual log entries are written.

The big takeaway here is to try and consistently have prefixes for log entries that specifically reference the module they're use, and also allows us to create hooks that we can use to add other custom prefixes, such as ones that automatically fill out server or view information (see here: https://github.com/mattermost/desktop/pull/2628/commits/e90c6f4c1ba514f1766bb7b0aadd518931033e17#diff-952df4a6d35a2fe5dac30403886c46bab3b29e9f344751a323ce5978917e472eR430)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51873

```release-note
NONE
```


src/common/log.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import log, {LevelOption, LogFunctions} from 'electron-log';
+
+import Util from 'common/utils/util';
+
+const logFunctions = {
```
hmhealey, line 8: 0/5 Maybe this could be a class instead of an object? It feels like that might make storing and applying the prefix a bit more straightforward, but that might not work with electron-log's interface

hmhealey (approved):
Looks good to me. I wonder if we should make something like this for the web app since I have missed having a tagged logger like mobile has

src/common/log.ts
devinbinnie: Yeah, I'm not sure why it's not a class. I will make it a class.

src/main/views/modalView.ts
```diff
@@ -48,13 +49,17 @@ export class ModalView<T, T2> {
...
-            log.error(e);
+            this.log.error('there was an error loading the modal:');
+            this.log.error(e);
         }
 
         this.contextMenu = new ContextMenu({}, this.view);
     }
 
+    private get log() {
+        return logger.withPrefix('ModalView', this.key);
```
larkox, line 60: Are we expecting the key to change? If not, we may want to assign this on the constructor, so we don't have to do all the `appendToLog` logic every time we need to log something. 0/5

src/main/views/webContentEvents.ts
```diff
@@ -106,17 +110,17 @@ export class WebContentsEventManager {
     };
 
     denyNewWindow = (details: Electron.HandlerDetails): {action: 'deny' | 'allow'} => {
-        log.warn(`Prevented popup window to open a new window to ${details.url}.`);
+        this.log().warn(`Prevented popup window to open a new window to ${details.url}.`);
```
larkox, line 113: If I read the code right, this will log as `[WebContentsEventManager] [undefined] Prevented popup...`. The undefined part is intended?

larkox (approved):
Nothing really blocking, but a few question and comments.

src/common/log.ts
```diff
@@ -0,0 +1,54 @@
...
+            funcs[key] = (...args: any[]) => func(...params, ...args);
+            return funcs;
+        }, {} as LogFunctions);
+    };
+
+    withPrefix = (...prefixes: string[]) => {
+        return this.appendToLog(...prefixes.map((prefix) => `[${Util.shorten(prefix)}]`));
+    };
+
+    setLoggingLevel = (level: string) => {
```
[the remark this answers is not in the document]
devinbinnie: You're right, it is weird. I've reworked it a little. Have a look :)

src/main/views/webContentEvents.ts
devinbinnie: Ahh you're right it will, I'll fix that.

devinbinnie: Yeah that should be fixed. I'll fix it.

src/main/views/modalView.ts
devinbinnie: Yeah you're right, it's in the constructor now.

src/common/log.ts
```diff
@@ -0,0 +1,52 @@
...
+        };
+    }
+
+    private prefixed = (func: (...args: any[]) => void, ...additionalPrefixes: string[]) => {
+        return (...args: any[]) => func(...this.shortenPrefixes(...this.prefixes, ...additionalPrefixes), ...args);
+    }
+
+    private shortenPrefixes = (...prefixes: string[]) => {
+        return prefixes.map((prefix) => `[${Util.shorten(prefix)}]`);
+    }
```
larkox, line 43: Why not do this on creation, so we don't have to repeat it on every log?

If we are going to keep the `withPrefix`, we can just add it to the list.

src/common/log.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    error = this.prefixed(log.error);
+    warn = this.prefixed(log.warn);
+    info = this.prefixed(log.info);
+    verbose = this.prefixed(log.verbose);
+    debug = this.prefixed(log.debug);
+    silly = this.prefixed(log.silly);
+    log = this.prefixed(log.log);
```
larkox, line 51: If I am reading this right, any change on `this.prefixes` won't be visible here, since `this.prefixed` is executed on object creation. If we wanted to make this work as is (that I don't recommend) we would need something like:
`error = (args) => this.prefixed(log.error)(args)`

If we don't want the prefixes to change after logger creation, we can always define all this on the object constructor, and not store the prefixes on `this.prefixes`.

src/common/log.ts
```diff
@@ -0,0 +1,52 @@
...
+        return {
+            error: this.prefixed(log.error, ...prefixes),
+            warn: this.prefixed(log.warn, ...prefixes),
+            info: this.prefixed(log.info, ...prefixes),
+            verbose: this.prefixed(log.verbose, ...prefixes),
+            debug: this.prefixed(log.debug, ...prefixes),
+            silly: this.prefixed(log.silly, ...prefixes),
+            log: this.prefixed(log.log, ...prefixes),
+        };
+    }
```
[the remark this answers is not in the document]
devinbinnie: So in 90% of cases, most modules will just call `new Logger(...some prefixes)` and then will always use the object returned, avoiding the need to use `withPrefix`. But I'm providing this here for the more dynamic cases like the `WebContentsEventManager` where we don't know the `webContentsId` immediately.

You may think, "well that's a code smell" and you're right. But it's coming from the WebContentsEventsManager, which also should be refactored because there is a way to know the id ahead of time and avoid using this.

tl;dr It's there for convenience right now, should probably be removed when we fix other things :P

src/common/log.ts
devinbinnie: I would assume `this.prefixes` is immutable, hence why it's private. Unfortunately JavaScript doesn't have an `ImmutableSet` so I think the array works fine.

larkox (approved):

PR #2654, opened 2023-04-02
[MM-50485] Implement ServerManager, introduce id concept for identifying servers/views

devinbinnie:
#### Summary
This PR contains the Server Manager module that allows for an easier and single source of truth for configured server and view information. All objects are keyed off of a specific `id` (generally a UUID of some kind) making accessing a specific server cleaner and more streamlined.

In this PR, it is not currently used by any other modules, as there are some very messy inter-dependencies that made pulling out the code for that very, very difficult. I will be attempting to break it down further over the coming week.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50485

```release-note
NONE
```


m1lt0n (approved):

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        log.debug('ServerManager.updateTabOrder', serverId, tabOrder);
+
+        this.tabOrder.set(serverId, tabOrder);
+        this.persistServers();
+    }
+
+    addServer = (server: Team) => {
+        const newServer = new MattermostServer(server, false);
+        this.servers.set(newServer.id, newServer);
+        this.serverOrder.push(newServer.id);
```
enahum, line 175: if the identifier on the map already exists, you may be replacing an existing server. but will have the same id as duplicate in the `serverOrder` array. can we be a little bit safer here?

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        const tabOrder: string[] = [];
+        getDefaultTabs().forEach((tab) => {
+            const newTab = this.getTabView(newServer, tab.name, tab.isOpen);
+            this.tabs.set(newTab.id, newTab);
+            tabOrder.push(newTab.id);
+        });
+        this.tabOrder.set(newServer.id, tabOrder);
+
+        this.emit(SERVERS_URL_MODIFIED, [newServer.id]);
```
enahum, line 184: not sure the event being emitted here is actually descriptive, very confusing for someone who hasn't been around the desktop app codebase

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        this.persistServers();
+        return newServer;
+    }
+
+    editServer = (serverId: string, server: Team) => {
+        const existingServer = this.servers.get(serverId);
+        if (!existingServer) {
+            return;
+        }
```
enahum, line 193: see, we are being safe here.. I like this.

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+                this.tabs.set(tabId, tab);
+            }
+        });
+
+        urlModified?.();
+        this.persistServers();
+    }
+
+    removeServer = (serverId: string) => {
+        this.tabOrder.get(serverId)?.forEach((tabId) => this.tabs.delete(tabId));
```
enahum, line 216: I think this line here is irrelevant, no need for it as we are deleting the key from the map in the next line

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+    updateLastActive = (tabId: string) => {
+        const tab = this.tabs.get(tabId);
+        if (!tab) {
+            return;
+        }
+        this.lastActiveTab.set(tab.server.id, tabId);
+
+        this.currentServerId = tab.server.id;
+
+        const serverOrder = this.serverOrder.findIndex((srv) => srv === tab.server.id);
```
enahum, line 247: remember that `findIndex` could return `-1`

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+            const id = this.initServer(team, true);
+            serverOrder.push(id);
+        });
+        if (Config.enableServerManagement) {
+            Config.localTeams.sort((a, b) => a.order - b.order).forEach((team) => {
+                const id = this.initServer(team, false);
+                serverOrder.push(id);
+            });
+        }
+        this.filterOutDuplicateTeams();
```
enahum, line 263: can you share a bit of context as to why would it be possible to have duplicate teams?

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        this.serverOrder = serverOrder;
+        if (Config.lastActiveTeam) {
+            this.currentServerId = this.serverOrder[Config.lastActiveTeam];
+        } else {
+            this.currentServerId = this.serverOrder[0];
```
enahum, line 268: we are saying this can be `undefined`, yes ?

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        }
+    }
+
+    private filterOutDuplicateTeams = () => {
```
enahum, line 272: for some reason this function seems to be related to Servers and not `Teams` which is throwing me off a bit

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        }
+        return firstTab;
+    }
+
+    private persistServers = async (lastActiveTeam?: number) => {
+        this.emit(SERVERS_UPDATE);
+
+        const localServers = [...this.servers.values()].
+            filter((server) => !server.isPredefined).
+            map((server) => this.toConfigTeam(server));
```
enahum, line 323: maybe you can use `reduce` instead, iterating over the array only once.

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        await Config.setServers(localServers, lastActiveTeam);
+    }
+
+    private getLastActiveTab = (serverId: string) => {
+        let lastActiveTab: number | undefined;
+        if (this.lastActiveTab.has(serverId)) {
+            const index = this.tabOrder.get(serverId)?.indexOf(this.lastActiveTab.get(serverId)!);
+            if (typeof index !== 'undefined' && index >= 0) {
```
enahum, line 331: a shorthand can be

```suggestion
            if (index != null && index >= 0) {
```

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+                lastActiveTab = index;
+            }
+        }
+        return lastActiveTab;
+    }
+
+    private toConfigTeam = (server: MattermostServer): ConfigTeam => {
```
enahum, line 338: see, I definitely do not understand the work Team here

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+                });
+                return tabs;
+            }, [] as ConfigTab[]) ?? [],
+        };
+    }
+
+    private getTabView = (srv: MattermostServer, tabName: string, isOpen?: boolean) => {
+        switch (tabName) {
+        case TAB_MESSAGING:
+            return new MessagingTabView(srv, isOpen);
```
enahum, line 362: ok, I need an explanation as to why we do create a new object every time we call `getTabView`

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+
+    private openExtraTabs = (serverId: string) => {
+        const server = this.servers.get(serverId);
+        const remoteInfo = this.remoteInfo.get(serverId);
+
+        if (!(server && remoteInfo)) {
+            return false;
+        }
+
+        if (!(remoteInfo.serverVersion && Utils.isVersionGreaterThanOrEqualTo(remoteInfo.serverVersion, '6.0.0'))) {
```
enahum, line 396: any reason why we are checking 6.0.0?

src/common/servers/serverManager.ts
enahum, line 1: I think that when adding or initializing a server could be a great moment to show users an alert, the same way we do on mobile, to ask their sys admins to update at least to the latest ESR.

cc: @svelle @itao @amyblais @m1lt0n in case we want to create a follow up ticket

src/common/servers/MattermostServer.ts
```diff
@@ -1,14 +1,30 @@
...
     name: string;
-    url: URL;
-    constructor(name: string, serverUrl: string) {
-        this.name = name;
-        this.url = urlUtils.parseURL(serverUrl)!;
+    url!: URL;
+    isPredefined: boolean;
+
+    constructor(server: Team, isPredefined = false) {
+        this.id = uuid();
```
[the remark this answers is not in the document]
devinbinnie: Well for one, I didn't know we had that :P.

In this case, the unique identifier is purely there to have a consistent and non-colliding ID so that we can ID a server from anywhere in the app, the actual string itself doesn't have any meaning otherwise.

I can see some usefulness in using the diagnostic ID, but when we create the server we don't have access to that yet, we'd have to ping the server first and see if we can get that. I think for the purposes I'm using the ID for right now the UUID is fine, but I could see a future change in which we store the diagnostic ID after the server has been added and switch to using that when we do logging.

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+        const tabs = this.getOrderedTabsForServer(server.id);
+
+        let selectedTab = tabs.find((tab) => tab && tab.name === TAB_MESSAGING);
+        tabs.
+            filter((tab) => tab && tab.name !== TAB_MESSAGING).
+            forEach((tab) => {
+                if (parsedURL.pathname.match(new RegExp(`^${tab.url.pathname}(/(.+))?`))) {
+                    selectedTab = tab;
+                }
+            });
```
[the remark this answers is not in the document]
devinbinnie: Yeah this logic is very confusing, it's been around a while in various forms. What this is supposed to do is find the exact view that best matches an arbitrary URL that you provide.

The reason we iterate so many times is that we basically assume that the view of type `TAB_MESSAGING` (which has a url of the form `http://<somehost>`) always is the default case (and it should be, it's the Channels tab), so we grab that first as the default case. Then we check the other tabs, which currently have a base URL of the form `http://<somehost>/boards` or `http://<somehost>/playbooks` to see if either of them are better matches.

We do the `find()` first to separate out the Channels case because it will always match the above regex, and since we don't guarantee any order, there are cases where you click on a Boards link and it would forward you to the Channels tab because that's the last one selected.

A lot of this logic was shoehorned in when we added the tabs and it's complicated and annoying, and one of the biggest reasons I want to move to arbitrary tabs is to get rid of this. We could as a short term fix patch the regex to only look for `boards` or `playbooks`, but I don't see a huge benefit in potentially causing some issue now when I fully plan on getting rid of this logic altogether.

src/common/servers/serverManager.ts
devinbinnie: If we ever get the same ID somehow, that would mean that our guaranteed UUIDs have collided. It's really unlikely and should cause errors to happen anyways, but I can add an explicit error check here if we think it's necessary.

src/common/servers/serverManager.ts
devinbinnie: This event is specific to when a server URL is changed, and basically is there to tell the app to contact the server and update the config information we get from it.

We don't encapsulate this particular logic in the server manager because it's coupled to Electron, and we don't use Electron libraries in the `common` module when we can avoid it.

I'll make a comment to clarify this.

src/common/servers/serverManager.ts
devinbinnie: We're deleting from a different map in this line (that's what the `forEach` is for).
I suppose we could leave them in there but then they're never explicitly dereferenced and they might get stuck in memory.

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,426 @@
...
+
+        this.persistServers();
+    }
+
+    toggleTab = (tabId: string, isOpen: boolean) => {
+        const tab = this.tabs.get(tabId);
+        if (!tab) {
+            return;
+        }
+        tab.isOpen = isOpen;
```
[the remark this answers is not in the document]
devinbinnie: Good catch, I'll change the function name.

src/common/servers/serverManager.ts
devinbinnie: If this ever happens we should be hard crashing, but I can be explicit about the error thrown.

src/common/servers/serverManager.ts
devinbinnie: We have 3 different sources of servers in the app: the build config (servers configured at build time), Group Policy for Windows (for customers mostly) and then the user configuration.

It's possible that a user could add a server that's already pre-configured, or could edit their local config file to do this, causing a duplicate. This can cause some weird behaviour because then the app doesn't know where to send deep links, and the check for which tab to switch to gets fuzzy as well, so we filter out ones that are clear duplicates.

src/common/servers/serverManager.ts
devinbinnie: Yes, in the case where there are no servers for example, we can't have a current server ID.

src/common/servers/serverManager.ts
devinbinnie: I don't think this works here, because `null` and `undefined` are different identifiers. `Map.get()` returns undefined if the element doesn't exist, so index would be `undefined`.

I think this might get around TypeScript, but isn't technically correct. 0/5 on whether we change or not.

src/common/servers/serverManager.ts
devinbinnie: Okay this one I should probably change the type for, because I created this type.
Oh the merge conflicts I'm about to create 😅

src/common/servers/serverManager.ts
devinbinnie: Before v6.0, the logic to send React Router pushes through the Desktop App (used to dynamically switch between tabs) didn't exist, so if you're using an older server we don't want to enable these extra tabs in the first place.

This will be going away as well when we move to arbitrary tabs.

src/common/servers/MattermostServer.ts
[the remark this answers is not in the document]
enahum: sounds good to me

src/common/servers/serverManager.ts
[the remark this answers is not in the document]
enahum: is fine like this.. is not like this is a massive perf issue, but could definitely benefit from some logic change.

enahum (approved):

src/common/servers/serverManager.ts
larkox: `!= null` works in its own way (JavaScript magic) and checks either for `undefined` or `null`.

Be aware of `!=` instead of `!==`.

larkox (approved):
Great work. A few non-blocking questions.

src/common/config/RegistryConfig.ts
```diff
@@ -7,7 +7,7 @@ import log from 'electron-log';
 import WindowsRegistry from 'winreg';
 import WindowsRegistryUTF8 from 'winreg-utf8';
 
-import {RegistryConfig as RegistryConfigType, Team} from 'types/config';
+import {RegistryConfig as RegistryConfigType, FullTeam} from 'types/config';
```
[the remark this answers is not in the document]
devinbinnie: They wee called teams by the original writer of the Desktop App, not sure why but I think it had to do with them using it initially to manage multiple teams before we had an LHS team sidebar.

Also see here: https://github.com/mattermost/desktop/pull/2654#discussion_r1157293268

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,442 @@
...
+        const tabOrder: string[] = [];
+        getDefaultTabs().forEach((tab) => {
+            const newTab = this.getTabView(newServer, tab.name, tab.isOpen);
+            this.tabs.set(newTab.id, newTab);
+            tabOrder.push(newTab.id);
+        });
+        this.tabOrder.set(newServer.id, tabOrder);
+
+        // Emit this event whenever we update a server URL to ensure remote info is fetched
+        this.emit(SERVERS_URL_MODIFIED, [newServer.id]);
```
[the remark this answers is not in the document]
devinbinnie: I suppose it could be, but users are already used to doing this. Doesn't make too much sense to remove the functionality to me, and I can see where people might want to change the URL (if something updates) so we can leave it.

src/common/servers/serverManager.ts
```diff
@@ -0,0 +1,442 @@
...
+            this.lastActiveTab.set(server.id, tabOrder[team.lastActiveTab]);
+        }
+        return server.id;
+    }
+
+    private getFirstOpenTabForServer = (serverId: string) => {
+        const tabOrder = this.tabOrder.get(serverId);
+        if (!tabOrder) {
+            throw new Error(`Cannot find tabs for server id ${serverId}`);
+        }
```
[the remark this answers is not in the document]
devinbinnie: Ahh yeah, that's generally going to be more performant to do. Good catch :)

src/common/servers/serverManager.ts
devinbinnie: Ugh, Javascript lol.
I think I'm going to leave it as is to avoid confusion, this is more explicit.

cwarnermm:
@devinbinnie - Are product or developer documentation updates needed for this PR?

PR #2655, opened 2023-04-02
[MM-51874] Migrate loading screen to singleton

devinbinnie:
#### Summary
This PR migrates the loading screen object to a singleton, meaning that objects will not need to go through the `windowManager` or `viewManager` to interact with it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51874

```release-note
NONE
```


m1lt0n (approved):
Nice refactoring!

src/main/views/loadingScreen.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        if (process.env.NODE_ENV === 'test') {
+            app.emit('e2e-app-loaded');
+        }
+    }
+}
+
+const loadingScreen = new LoadingScreen();
+export default loadingScreen;
```
streamer45, line 116: Purely out of curiosity, when does the loading screen object get initialized? I suspect it may be caused by an import and the runtime is smart enough to do it only once but not sure. 

Mostly I wonder if this is a commonly established pattern in the JS world since from my low-level tuned brain perspective it feels a bit obscure as I'd expect it to be initialized somewhere in the entry point for the app.

streamer45 (approved):
Nice! Only left one non-blocking question for my own understanding :)

src/main/views/loadingScreen.ts
devinbinnie: So I'm not sure if it's established in the JS world, but I'm aiming to work in the singleton pattern for most objects now since it ensures we encapsulate the logic most relevant to said modules and means we don't end up with as many behemoths like the `windowManager`.

As for how it's imported, I believe webpack is smart enough to handle that. Once the first call of `import` happens for the file it'll run this code and export the object. I haven't done a deep dive into the NodeJS engine to see how it actually works but that's my current understanding.

PR #2656, opened 2023-04-03
[MM-51875] Migrate viewManager to singleton

devinbinnie:
#### Summary
This PR contains the changes to migrate the `viewManager` module out of the `windowManager` into a singleton class, meaning that many associated objects can now directly access the remote views without needing to import the `windowManager`. It should encapsulate that logic better and require less helper functions and such.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51875

```release-note
NONE
```

### DIFF
--- a/src/main/views/viewManager.ts
+++ b/src/main/views/viewManager.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+import {BrowserView, BrowserWindow, dialog, ipcMain, IpcMainEvent, IpcMainInvokeEvent} from 'electron';
 import log from 'electron-log';
-import {BrowserView, BrowserWindow, dialog, ipcMain, IpcMainEvent} from 'electron';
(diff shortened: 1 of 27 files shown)


enahum:
@devinbinnie this PR now has a bunch of conflicts, perhaps best to resolve to avoid having similar comments as in the other PR

src/main/views/viewManager.ts
```diff
@@ -53,35 +60,204 @@ export enum LoadingScreenState {
...
+    sendToFind = () => {
+        this.getCurrentView()?.openFind();
+    }
+
+    /**
+     * Deep linking
+     */
+
+    handleDeepLink = (url: string | URL) => {
+        // TODO: fix for new tabs
```
enahum, line 185: do we have a ticket for this?

enahum (approved):

src/main/views/viewManager.ts
devinbinnie: Oh that's a super old TODO, I should just remove that cause I think it's working now anyways, and if it's not the new tabs are going to change it.

hmhealey (approved):

PR #2676, opened 2023-04-12
Clean up appState module, move to common, simplify events

devinbinnie:
#### Summary
This PR moves the `appState` (responsible for keeping track of mentions, unreads and expiry) to the `common` module, while also refactoring it into a class, and simplifying and genericizing the events emitted so that we aren't adding a lot of additional code for each widget that might need this information.

```release-note
NONE
```


src/common/appState.ts
```diff
@@ -0,0 +1,82 @@
...
+            this.unreads,
+        );
+        this.emit(UPDATE_APPSTATE_TOTALS,
+            [...this.expired.values()].some((value) => value),
+            [...this.mentions.values()].reduce((total, value) => total + value, 0),
+            [...this.unreads.values()].some((value) => value),
+        );
+    };
+
+    private emitMentions = (viewId: string) => {
```
larkox, line 69: Instead of emitMentions, shouldn't be still emitStatus? Since it is for just one viewId (even thought then it calls emitStatus either way) maybe `emitStatusForView`?

larkox (approved):
LGTM, just one small nitpick.

src/common/appState.ts
devinbinnie: Yeah that makes sense. Will fix.

src/common/appState.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        this.expired.delete(viewId);
+        this.mentions.delete(viewId);
+        this.unreads.delete(viewId);
+    }
+
+    emitStatus = () => {
+        log.silly('emitStatus');
+
+        this.emit(UPDATE_APPSTATE,
```
hmhealey, line 57: Some style nitpicking, but I'd expect all the arguments to this to be on new lines for these

src/common/communication.ts
```diff
@@ -167,3 +164,7 @@ export const UPDATE_TAB_ORDER = 'update-tab-order';
 export const GET_LAST_ACTIVE = 'get-last-active';
 export const GET_ORDERED_SERVERS = 'get-ordered-servers';
 export const GET_ORDERED_TABS_FOR_SERVER = 'get-ordered-tabs-for-server';
+
+export const UPDATE_APPSTATE = 'update-appstate';
+export const UPDATE_APPSTATE_TOTALS = 'update-appstate-totals';
+export const UPDATE_APPSTATE_FOR_VIEWID = 'update-appstate-for-viewid';
```
hmhealey, line 170: I think this should be `UPDATE_APPSTATE_FOR_VIEW_ID` since they're two separate words. I also misread it as "for viewed" at first

hmhealey (approved):

src/common/appState.ts
devinbinnie: I like the idea of having the event on the same line to denote that it's kind of a special argument. I'll leave it as is for now, but something to think about for the future for sure.

PR #2680, opened 2023-04-14
Migrate downloads dropdown menus to singletons

devinbinnie:
#### Summary
This PR moves the last two objects (the downloads dropdown menus) out of the `WindowManager`  and into their own singletons.

```release-note
NONE
```


hmhealey (approved):

larkox (approved):
LGTM. Nevertheless, both classes seem to share a lot of code. It may be an opportunity to simplify this even a bit more. But 0/5

devinbinnie:
> LGTM. Nevertheless, both classes seem to share a lot of code. It may be an opportunity to simplify this even a bit more. But 0/5

Yes I agree, I'm thinking of creating a generic `MenuView` class that contains the code responsible for opening, closing and sizing the menu. Then other implementations can define how they're updated.

This will be a follow up item since I want to first focus on breaking down the monoliths.

PR #2682, opened 2023-04-18
Remove WindowManager, separate functionality into smaller modules

devinbinnie:
#### Summary
This refactor removes the monolithic `WindowManager` object that contained a lot of different functionality pertaining to the application. It had handled main window functions, server switching, resizing and managing some other views.

Most of this functionality can be self-contained within individual modules to avoid any more of these big monoliths. This is the first step towards a new more modular singleton-based architecture, in which each object should be able to mostly govern itself.

A lot of the functionality has been moved into the `main/app` folder, which will be further broken down into classes that will be self-governing over time. These changes will happen in future PRs. The hope is that with this PR, we should be able to work on those smaller pieces without having them interdependent on other changes.

I've separated the move into a number of commits, the best strategy for review might be to go commit-by-commit to understand gradually what is being separated from the `WindowManager`, where it's going and how it affects the modules that used the functionality. Any questions I will be happy to answer :)

```release-note
NONE
```

### DIFF
--- a/src/main/app/servers.ts
+++ b/src/main/app/servers.ts
@@ -29,7 +29,7 @@ export const switchServer = (serverId: string, waitForViewToExist = false) => {
         const timeout = setInterval(() => {
             if (ViewManager.getView(nextTab.id)) {
                 ViewManager.showById(nextTab.id);
-                clearTimeout(timeout);
+                clearInterval(timeout);
             }
         }, 100);
     } else {


src/main/app/index.ts
```diff
@@ -5,6 +5,11 @@
 
 import {initialize} from './initialize';
 
+// TODO: Singletons, we need DI :D
```
devinbinnie, line 8: This is a temporary solution. The best way to solve this would be a DI framework, but that might be a bit overkill for our application.

A quicker solution may be to just have a file that imports all the singletons when the app starts instead of having them all here.

enahum: we could consider a DI framework if this grows like crazy, so far with just three I don't think is needed either

src/main/app/servers.ts
```diff
@@ -0,0 +1,134 @@
...
+        return;
+    }
+    const nextTab = ServerManager.getLastActiveTabForServer(serverId);
+    if (waitForViewToExist) {
+        const timeout = setInterval(() => {
+            if (ViewManager.getView(nextTab.id)) {
+                ViewManager.showById(nextTab.id);
+                clearTimeout(timeout);
+            }
+        }, 100);
```
enahum, line 34: can you explain this value of 100ms?

src/main/app/servers.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (modalPromise) {
+        modalPromise.then((data) => {
+            const newTeam = ServerManager.addServer(data);
+            switchServer(newTeam.id, true);
+        }).catch((e) => {
+            // e is undefined for user cancellation
+            if (e) {
+                log.error(`there was an error in the new server modal: ${e}`);
+            }
+        });
```
enahum, line 62: this is fine, but I wonder if it makes it easier to read by using async/await

src/main/app/tabs.ts
```diff
@@ -0,0 +1,73 @@
...
+    const currentTeamTabs = ServerManager.getOrderedTabsForServer(currentView.tab.server.id).map((tab, index) => ({tab, index}));
+    const filteredTabs = currentTeamTabs?.filter((tab) => tab.tab.isOpen);
+    const currentTab = currentTeamTabs?.find((tab) => tab.tab.type === currentView.tab.type);
+    if (!currentTeamTabs || !currentTab || !filteredTabs) {
+        return;
+    }
+
+    let currentOrder = currentTab.index;
+    let nextIndex = -1;
+    while (nextIndex === -1) {
```
enahum, line 65: I wonder if another stop condition should be that there are no more tabs to check

enahum (changes requested):

src/main/app/servers.ts
devinbinnie: There was an issue where sometimes the view wouldn't exist after a new server was added, due to a race condition. We picked an arbitrary value.

I'm not sure if the problem still exists but I wanted to keep the logic the way it is during the refactor.

src/main/app/servers.ts
devinbinnie: Yeah that's probably worth doing at some point. I think this is older code so that's why it wasn't written that way.
Will address in a follow-up PR.

src/main/app/servers.ts
enahum: I'm ok with the logic, just want to make sure this value is not an overkill. 0/5 if we change it now or later.

m1lt0n (approved):

PR #2693, opened 2023-04-27
Prevent calls widget window from showing native shadow

streamer45:
#### Summary

PR sets the `hasShadow` option to `false` for the calls widget window. The native macOS shadow is not playing nicely and causing some weird artifacts and inconsistencies in sizing as we already have some shadow in place at the CSS level.

The only concern with this change is that it makes it a little more difficult to figure out whether the widget window is focused but not sure how important that is. @tanmay-des Maybe we could work on improving that at the widget level through some clever styling

#### Screenshots

Before:

Screenshot 2023-04-27 at 14 26 12

After:

Screenshot 2023-04-27 at 14 27 10

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -62,6 +62,7 @@ export default class CallsWidgetWindow extends EventEmitter {
             transparent: true,
             show: false,
             alwaysOnTop: true,
+            hasShadow: false,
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Looks good, thanks 👍

devinbinnie (approved):

PR #2699, opened 2023-05-01
[MM-52139] Force user to reset default downloads directory when the app is not allowed to access it

devinbinnie:
#### Summary
When the application is upgraded, it loses access to the folders it was previously granted entitlements to, including whatever the configured default downloads directory is. Since we still keep the configuration between upgrades however, it confuses the app when it doesn't have access.

This PR will force the user to reset the default downloads directory when the folder becomes inaccessible, allowing the app to download normally to whatever folder was chosen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52139

```release-note
Fixed an issue for the MAS build where the default downloads directory would be invalid after upgrade.
```

### DIFF
--- a/src/main/downloadsManager.ts
+++ b/src/main/downloadsManager.ts
@@ -137,8 +137,8 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
                 this.willDownloadURLs.set(url, {filePath: saveDialogResult.filePath, bookmark: saveDialogResult.bookmark});
             } else {
                 const filename = this.createFilename(item);
-                const savePath = this.getSavePath(`${Config.downloadLocation}`, filename);
-                await this.verifyMacAppStoreDownloadFolder(savePath);
+                const downloadLocation = await this.verifyMacAppStoreDownloadFolder(filename);
+                const savePath = this.getSavePath(`${downloadLocation}`, filename);
                 this.willDownloadURLs.set(url, {filePath: savePath});
             }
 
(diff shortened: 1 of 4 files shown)


enahum (approved):
lgtm

src/main/downloadsManager.ts
```diff
@@ -374,6 +378,46 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+        return result.filePaths[0];
+    }
+
+    private verifyMacAppStoreDownloadFolder = async (fileName: string) => {
+        let downloadLocation = Config.downloadLocation;
+
+        // eslint-disable-next-line no-undef
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        if (__IS_MAC_APP_STORE__ && downloadLocation) {
```
hmhealey, line 404: If this is because of `__IS_MAC_APP_STORE__`, you might want to add a `.d.ts` somewhere that defines any of the globals you're using to avoid having to add these comments

src/main/downloadsManager.ts
```diff
@@ -374,6 +378,46 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+            try {
+                const savePath = this.getSavePath(downloadLocation, fileName);
+                fs.writeFileSync(savePath, '');
+                fs.unlinkSync(savePath);
+            } catch (e) {
+                downloadLocation = await this.selectDefaultDownloadDirectory(
```
hmhealey, line 410: Should this still save the file after the path has been selected? Or maybe I'm misunderstanding this and the `writeFileSync` in the `try` block doesn't actually save the file

hmhealey (approved):

src/main/downloadsManager.ts
devinbinnie: Yeah this is a weird case. Because of the async nature of this code, after the method finishes for some reason the `DownloadItem` is marked destroyed and we can't re use it. So for the cases where we need to pick a folder to save to, we have this `willDownloadUrls` queue that we add to to recreate downloads effectively.

src/main/downloadsManager.ts
devinbinnie: Yeah this seems like a good cleanup item to do in a future PR.

PR #2713, opened 2023-05-05
MM-51956 - Fix for: Calls popout generating duplicate notifications

cpoile:
#### Summary
- The popout didn't have the `Mattermost/version` userAgent. The only way to set this (that works) is through the `browserWindow.loadURL` fn, but we don't use that for the popout, we use the `window.open` from the calls side. It's not possible to set the userAgent directly with `window.open` or in the `windowOpenHandler` return options. The best way I could come up with is to just reload the calls popout using the `loadURL`. Through trial and error I found a place to do that, which doesn't cause the screen to flicker or load and reload, luckily. We can do it even earlier in the process (e.g., the `dom-ready` event) but that returns an error (but, oddly, everything works and the userAgent does get correctly set). The `did-frame-finish-load` event is the earliest that doesn't return an error.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51956

#### Checklist

- [ ] Has UI changes

#### Release Note
```release-note
Calls: Fix duplicate desktop notifications when calls popout is open
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -16,7 +16,12 @@ import {
 
 import {MattermostView} from 'main/views/MattermostView';
 
-import {getLocalPreload, openScreensharePermissionsSettingsMacOS, resetScreensharePermissionsMacOS} from 'main/utils';
+import {
+    composeUserAgent,
+    getLocalPreload,
+    openScreensharePermissionsSettingsMacOS,
+    resetScreensharePermissionsMacOS,
+} from 'main/utils';
 
 import {Logger} from 'common/log';
 import {CALLS_PLUGIN_ID, MINIMUM_CALLS_WIDGET_HEIGHT, MINIMUM_CALLS_WIDGET_WIDTH} from 'common/utils/constants';
@@ -261,7 +266,7 @@ export class CallsWidgetWindow {
         this.setBounds(initialBounds);
     }
 
-    private onPopOutOpen = ({url}: {url: string}) => {
+    private onPopOutOpen = ({url}: { url: string }) => {
         if (!(this.mainView && this.options)) {
             return {action: 'deny' as const};
         }
@@ -294,6 +299,23 @@ export class CallsWidgetWindow {
 
         this.popOut.on('closed', () => {
             delete this.popOut;
+        });
+
+        // Set the userAgent so that the widget's popout is considered a desktop window in the webapp code.
+        // 'did-frame-finish-load' is the earliest moment that allows us to call loadURL without throwing an error.
+        this.popOut.webContents.once('did-frame-finish-load', async () => {
+            const url = this.popOut?.webContents.getURL() || '';
+            if (!url) {
+                return;
+            }
+
+            try {
+                await this.popOut?.loadURL(url, {
+                    userAgent: composeUserAgent(),
+                });
+            } catch (e) {
+                log.error('did-frame-finish-load, failed to reload with correct userAgent', e);
+            }
         });
     }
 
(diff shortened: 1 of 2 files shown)


streamer45 (approved):
Thanks @cpoile, would it make sense to set the user agent on the widget window for consistency if anything?

cpoile:
> Thanks @cpoile, would it make sense to set the user agent on the widget window for consistency if anything?

Sure, I don't see why not (just in case, for the future). Done.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -295,6 +302,23 @@ export class CallsWidgetWindow {
         this.popOut.on('closed', () => {
             delete this.popOut;
         });
+
+        // Set the userAgent so that the widget's popout is considered a desktop window in the webapp code.
+        // 'did-frame-finish-load' is the earliest moment that allows us to call loadURL without throwing an error.
+        this.popOut.webContents.once('did-frame-finish-load', async () => {
```
devinbinnie, line 308: This feels like a hack instead of solving the real problem of actually having the calls widget call the desktop app and have it open its own window. I think we should take that approach instead of trying to intercept the loading of a different window.

devinbinnie (changes requested):

src/main/windows/callsWidgetWindow.ts
cpoile: From what I understand, the reason to keep the window creation in the calls code is so that we can move quickly. The desktop product is more stable and its release schedule is (understandably) slower. Calls needs faster iteration; we're fixing and improving things weekly as we figure out how to best do things.
@streamer45 -- your thoughts?

devinbinnie: I don't think that should affect the actual creation of the window at all, we'll still use the window to load a calls plugin URL, meaning that it's still possible to make feature/bug fix changes to the app. Just the governance of the window should be under the control of the Desktop App.

Additionally, I don't think it makes sense to start accumulating further technical debt in an effort to try and allow for quick changes.

devinbinnie: Another way to detect the desktop app would be to have the app check to see if a specific field on the window exists, that you could set using the preload script. We just generally make use of the user agent instead since it's never been an issue. I think the reason it's flaky has to do with using `window.open` instead of creating a BrowserWindow, but I could be wrong.

Understandable, in that case I think we should be realistic and accept that the popout window will be the "semi-permanent" solution and at least make sure that it's secure and designed correctly as best we can.

I think either way there's going to be a minimum Desktop App version that will be needed to support certain features, that's just the nature of releasing it as a standalone app. One thing I've been pushing for is having the web app be aware of which version of the Desktop App is running it so that it's able to render/not render certain features.

streamer45: @devinbinnie  Had a good discussion on this with @cpoile  who reminded me of one very important technical detail I completely overlooked. The way things work now is that the popout always keeps the handler of the parent window (`window.opener`) which is used to share all sorts of state, including video tracks in case of screen sharing since the calls client is created and always living in the widget window. 

I actually remember playing with this sort of case when initially implementing the widget and I don't think I ever found a way to retain this behaviour without disabling security features. I am happy to revisit if you have suggestions on this front but otherwise the proper solution here feels just a bit too complex to justify the effort necessary to fix this issue.

devinbinnie: @streamer45 Are we not able to enable screen sharing without popping the pop out window? Is there not a way we could communicate that same state through the Electron API so that it can be shared between both windows? I guess I'm failing to see where the gap is here by having the window handler in the other renderer process instead of in the Electron main process.

Issue #2054, opened 2022-04-14
Desktop application opens any provided site when adding a server


userbradley:
I confirm (by marking "x" in the [ ] below: [x]):

- [x] This is not a troubleshooting question. [Troubleshooting questions go here: https://docs.mattermost.com/install/troubleshooting.html](https://docs.mattermost.com/install/troubleshooting.html).
- [x] This doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/platform/issues).
- [x] I have read [contributing guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

---

**Summary**
Desktop application allows you to open any website by adding new server

**Environment**
- Operating System: OSX 12.3 (21E230)
- Mattermost Desktop App version: 5.0.4
- Mattermost Server version: 6.6

1. Open desktop application
2. Click add new server
3. Name it anything
4. Fill in any web address
5. Site loads

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/41597815/163494132-2533ca4a-2371-4b93-8018-36a3daa3a724.png">

**Expected behavior**

Desktop application should respond with `This is not a mattermost server Address`

**Observed behavior**

Web page loads and you are able to interact with it

**Possible fixes**

1. Check headers contain the word `mattermost` (Will require the mattermost server to send these)
2. Check HTML contains certain mattermost specific phrases

devinbinnie:
This will be worked on as part of next-gen changes to the Desktop App. Currently we don't require any specific server side validation (apart from a couple cosmetic checks) so there's no incentive to support this, but in the future we likely will.

devinbinnie:
Relevant JIRA issue: https://mattermost.atlassian.net/browse/MM-50352

PR #2720, opened 2023-05-15
[MM-50352] Improve URL validation and add/edit server experience

devinbinnie:
#### Summary
This changes adds a feature to the server add/edit modals that will do some validation for the user to assist in making sure that their server URL is accurate. This should help alleviate some troubleshooting issues like:
- Site URL not matching the entered server URL
- Using HTTP instead of HTTPS
- Outdated server versions causing problems
- URL is valid, but points to the wrong server

The UX is a bit of a WIP, as I kind of just threw something together, but I would appreciate some prompt feedback from @asaadmahmood or @matthewbirtch as I think this is a really valuable feature to go out with v5.4.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50352
https://mattermost.atlassian.net/browse/MM-49072
Closes https://github.com/mattermost/desktop/issues/2054

```release-note
Improved URL validation and add/edit server experience
```


harshilsharma63:
I tried this build and found some issues-

1. When I enter an non-existant Mattermost URL, for example `https://localhost:8065` (local MM wasn't running), I get the warning about invalid server URL but can still click "Connect" and continue forward, eventually leading to an error. 

https://github.com/mattermost/desktop/assets/18575143/7a86ea27-42e1-4cfc-aa5d-6d3af1b3b4ab

matthewbirtch:
@devinbinnie before I review, I have one question:
- Do we need to require the use of `http://` or `https://` at the start of the URL? On the mobile app, there is some trickery done that allows users to just enter the URL without the HTTP/HTTPS, so I'm wondering/hoping we can do the same here. 
- Why do we need to warn them that their URL doesn't precisely match the configured siteURL? I might be missing context here, but it feels quite unnecessary to block them if they add a trailing `/` or use `http` instead of `https`. As @harshilsharma63 suggests, it would be great to just normalize the URL for them.

src/renderer/components/ConfigureServer.tsx
```diff
@@ -64,28 +65,26 @@ function ConfigureServer({
...
-            then((result: string) => {
-                const newURL = `${result}://${checkURL}`;
-                setUrl(newURL);
-                return newURL;
-            }).
-            catch(() => {
-                console.error(`Could not ping url: ${checkURL}`);
-                return checkURL;
+    useEffect(() => {
+        clearTimeout(validationTimeout);
```
larkox, line 79: It is more idiomatic to clear in the effect return function instead of before doing anything in the effect. Also, it will make sure it gets cleared on unmount.

src/renderer/components/ConfigureServer.tsx
```diff
@@ -35,8 +35,9 @@ type ConfigureServerProps = {
     onConnect: (data: UniqueServer) => void;
 };
 
+let validationTimeout: NodeJS.Timeout;
```
larkox, line 38: Any reason to tie this timeout to the file instead of to the component (using a `useRef`)?

src/renderer/components/ConfigureServer.tsx
```diff
@@ -64,28 +65,26 @@ function ConfigureServer({
...
-                return newURL;
-            }).
-            catch(() => {
-                console.error(`Could not ping url: ${checkURL}`);
-                return checkURL;
+    useEffect(() => {
+        clearTimeout(validationTimeout);
+        validationTimeout = setTimeout(() => {
+            setValidating(true);
+            window.desktop.validateServerURL(url).then((validationResult) => {
```
larkox, line 82: `validateServerUrl` is async, so you never know how long will it take. A rare scenario would be the following (haven't seen the whole component, so may not be possible).

- User writes the url
- The first timeout (TA) ends, and this gets executed
- The user changes the url
- The new timeout (TB) ends, and this gets executed
- TB sets again validating to true
- TA finishes with validateServerURL, setting the validation result to whatever the URL was on TA
- TA sets validating to false, while TB is still running
- TB finishes with validateServerURL, and now sets the correct result to setValidationResult.

There is also the problem of trying to set the state when the component is unmounted if the component gets unmounted while `validateServerURL` is running.

src/renderer/components/NewServerModal.tsx
```diff
@@ -93,108 +78,171 @@ class NewServerModal extends React.PureComponent<Props, State> {
...
+        const serverUrl = e.target.value;
+        this.setState({serverUrl, validationResult: undefined});
+        this.validateServerURL(serverUrl);
+    }
+
+    validateServerURL = (serverUrl: string) => {
+        clearTimeout(this.validationTimeout as unknown as number);
+        this.validationTimeout = setTimeout(() => {
+            this.setState({validationStarted: true});
+            window.desktop.validateServerURL(serverUrl, this.props.server?.id).then((validationResult) => {
```
larkox, line 91: Similar issues to the ones raised in the functional version.

src/renderer/components/NewServerModal.tsx
```diff
@@ -93,108 +78,171 @@ class NewServerModal extends React.PureComponent<Props, State> {
...
         }
         return null;
     }
 
-    validateForm() {
-        return this.getServerNameValidationState() === null &&
-            this.getServerUrlValidationState() === null;
-    }
-
     save = async () => {
```
larkox, line 229: This doesn't seem to need to be async anymore.

larkox (changes requested):
In general looks great. Just some nitpicks.

src/renderer/components/ConfigureServer.tsx
```diff
@@ -151,8 +164,34 @@ function ConfigureServer({
...
+                value: formatMessage({id: 'renderer.components.configureServer.url.validating', defaultMessage: 'Validating...'}),
+            });
+            validateURL(value).then(({validatedURL, serverName, message}) => {
+                if (currentTimeout.current !== validationTimeout.current) {
+                    return;
+                }
+                if (validatedURL) {
+                    setUrl(validatedURL);
+                }
+                if (serverName && !name) {
```
larkox, line 185: The value of `name` here may be from a few renders before, so may not be up to date. 0/5 on why this check is here, so maybe it is not an issue, but if you foresee this being an issue, I would suggest using the "functional form" of the set state.

```
if (serverName) {
    setName((prev) => prev ? prev : serverName);
}
```

src/renderer/components/NewServerModal.tsx
larkox: Still potential issues around `setState` being called when the component is already unmounted.

larkox (changes requested):
Still some nits regarding the components and states.

matthewbirtch:
> I'll generate a build that you guys can test out and let me know what feels off.

@devinbinnie did the build finish? I don't see a link to it anywhere

devinbinnie:
> > I'll generate a build that you guys can test out and let me know what feels off.
> 
> @devinbinnie did the build finish? I don't see a link to it anywhere

@matthewbirtch Yeah it did, sorry I think the new CI build doesn't post it here.
Here's the link: https://github.com/mattermost/desktop/actions/runs/5008088608

You'll download the ZIP for your OS and it should be in there. Ping me if you have questions :)

abhijit-singh (commented):
Looks good @devinbinnie. A few thoughts:
1. If the user does add an invalid URL and clicks on `Connect anyway`, it looks like currently the app gets stuck on the spinner screen. The only way I was able to go back was quitting and reopening the app. Can we throw an error here if it doesn't connect/takes too much time or maybe just give them an option to go back?

https://github.com/mattermost/desktop/assets/14369607/f100cc9c-6869-478c-bec5-9f8689b293fe

Also, the behaviour is different when you already have a server setup and then try to add a second one. In that case if you enter an invalid URL, the modal just closes without any feedback to the user. We should probably show a similar error over here also. 

2. If the user adds the server display name first and then the URL, the `Connect` button becomes active even before the URL is validated. Ideally, it should stay disabled until the validation is complete and the warning message is shown so that people know there's something wrong before they click on `Connect`.  

https://github.com/mattermost/desktop/assets/14369607/2ac7f70e-8f5c-4e1f-aea7-ee06b6e07e6e

3. If you delete a server and then try adding a server, the app crashes. If you reopen the app though, the server you just added does show up. 

https://github.com/mattermost/desktop/assets/14369607/98a06a20-3aef-4ad5-9b49-bbc8ca3e7e09

src/renderer/components/ConfigureServer.tsx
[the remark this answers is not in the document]
devinbinnie: Yeah this component was put together by the Growth team at some point. I'm still trying to figure out who is going to own the front-end of the Desktop App going forward, but either way definitely something we can fix later on.

devinbinnie:
Feedback has been addressed.

@abhijit-singh For #3 on your points, I have addressed the crashing bug in this PR: https://github.com/mattermost/desktop/pull/2727

src/renderer/components/ConfigureServer.tsx
```diff
@@ -56,35 +53,68 @@ function ConfigureServer({
...
+            return;
+        }
+
+        setValidating(true);
+        setURLError({
+            type: STATUS.INFO,
+            value: formatMessage({id: 'renderer.components.configureServer.url.validating', defaultMessage: 'Validating...'}),
+        });
+        validateURL(url).then(({validatedURL, serverName, message}) => {
+            if (editing) {
```
larkox, line 99: Given the times we are working with, probably is not going to happen unless you are looking for it, but there is a small issue here. Here the example scenario:
- I write an url
- It starts validating, and for whatever reason, takes ages. Lets call that execution `vURL1`
- I change the url
- It starts validating the second one. `vURL2`.
- Editing now is false.
From this point, two scenarios:
Scenario1:
- vURL1 finishes. `editing` is false, so executes everything. Among other things, rewriting the url the user last posted.
- vURL2 finishes. `editing` is also false, so executes everything. The user is really confused because what he wrote last is back there.
Scenario2:
- vURL2 finishes. `editing` is false, so executes everything. The user is happy with the results.
- vURL1 finishes (it was a long timeout). `editing` is still false, so executes everything. Suddenly the user sees an error and the previous URL. It gets really confused.

To fix this, we can go back to the idea of the timestamp. Actually, we would need both:
- timestamp for several validation running at the same time
- editing so a finishing validation doesn't overwrite what the user is editing right now

0/5 on whether this case is worth handling.

larkox (changes requested):
Almost there! A blocking comment, and two nitpicks/proposals.

src/renderer/components/ConfigureServer.tsx
devinbinnie: Yeah this is why I was originally checking the URL to make sure it matched so that this wouldn't happen, but I understand that it wasn't really working (it seemed like a good idea at the time :P)

The funny thing is there is a case where this could happen, where the URL check is timing out and it's probably going to confuse some people, so maybe we should go back to the timestamp.

larkox (approved):
LGTM! :smile:

abhijit-singh (commented):
Changes based on earlier feedback look great, thanks @devinbinnie! 

Just one other thing that I noticed - adding a server (valid or invalid) seems to be adding it twice for some reason. Not sure if it's because of any changes here though. 

https://github.com/mattermost/desktop/assets/14369607/5adfe275-ece2-4e64-ac77-0647792ae36f



https://github.com/mattermost/desktop/assets/14369607/dccaf03b-70e8-45a1-b5e6-0e6ec684be40

abhijit-singh (commented):
Changes based on earlier feedback look great, thanks @devinbinnie! 

Just one other thing that I noticed - adding a server (valid or invalid) seems to be adding it twice for some reason. Not sure if it's because of any changes here though. 

https://github.com/mattermost/desktop/assets/14369607/5adfe275-ece2-4e64-ac77-0647792ae36f



https://github.com/mattermost/desktop/assets/14369607/dccaf03b-70e8-45a1-b5e6-0e6ec684be40

harshilsharma63 (approved):
LGTM 👍

PR #2725, opened 2023-05-18
[MM-52879] Fixed calls popout URL checking

devinbinnie:
#### Summary
Some navigation code changes, as well as https://github.com/mattermost/desktop/pull/2713 revealed an issue with how we check for the calls widget/popout. We should be getting the root server URL to check against instead of the calls widget URL, whatever that may be, so I've fixed that.

This was causing a white screen for users of nightly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52879

```release-note
NONE
```


hmhealey (approved):

cpoile (approved):
Thanks!

PR #2730, opened 2023-05-24
Fix window issues

devinbinnie:
#### Summary
This PR fixes two issues with window sizing:
- On Windows 10/11, changing servers would cause the window to unsnap if snapped, due to [this bug](https://github.com/electron/electron/issues/25359). I've made an exception for Windows to fix this for our purposes.
- Restoring the saved window bounds for the `MainWindow` was failing due to calling the screen size before the app was ready, which wasn't working, so I've moved it over to the `init()` call.

```release-note
NONE
```

### DIFF
--- a/src/main/windows/mainWindow.test.js
+++ b/src/main/windows/mainWindow.test.js
@@ -495,14 +495,35 @@ describe('main/windows/mainWindow', () => {
 
         afterEach(() => {
             mainWindow.ready = false;
+            mainWindow.win.visible = false;
             jest.resetAllMocks();
         });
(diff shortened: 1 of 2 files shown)


hmhealey (approved):

hmhealey:
That snapping bug is interesting since I'm pretty sure that's bugged me in both Discord and Spotify in the past. Hopefully it gets fixed eventually

harshilsharma63 (approved):
Looks good on high level but I don't have a windows machine to try out the change.

PR #2736, opened 2023-05-25
Fix release post script

devinbinnie:
#### Summary
For some reason, the script running in GitHub Actions grabbed the same tag that we just made. So I fixed up the script to explicitly exclude it.

```release-note
NONE
```


marianunez (approved):
Anyway you can test this to make sure it's working as expected now?

hmhealey (approved):

devinbinnie:
> Anyway you can test this to make sure it's working as expected now?

I'm going to test it on the `release` branch actually, that's probably a better way of determining whether it works or not.

PR #2751, opened 2023-06-14
MM-52857 - Add call-join-request to calls widget

cpoile:
#### Summary
- We need the ability to send a request to the webapp to indicate that we want to join another call. That's what this does.
- Will be used in https://github.com/mattermost/mattermost-plugin-calls/pull/445 shortly.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-52857

#### Checklist

- [ ] Has UI changes

#### Release Note

```release-note
NONE
```


src/types/calls.ts
```diff
@@ -34,4 +34,8 @@ export type CallsLinkClickMessage = {
     link: string | URL;
 }
 
+export type CallsJoinRequestMessage = {
+    targetID: string;
```
streamer45, line 38: What's this target exactly? Is it the call?

cpoile: Yep. For now it's the channelID, but could change. Same as we've used targetID elsewhere.

streamer45: Interesting, I don't recall using `targetID` ever in the context of calls. I was expecting `callID` I guess. Is there a future this won't be accurate?

streamer45: Yeah, I think it should be `callID` on both sides ideally. The existing one on the calls side is only for internal use I suppose but here we are making a contract so I'd like to give it more thought and it seems like what we really mean is the id of the call (which can coincide with the channel ID but that's another matter) so we should name it accordingly.

cpoile: Should be good to go now, and the calls PR is updated.

cpoile:
I'm not sure what's happening with the channel toast e2e diff -- it passes locally, and I can't see why it would be narrower in the trace.

streamer45 (approved):
:+1:

cpoile:
@devinbinnie Any ideas what might still be the problem?

devinbinnie:
@cpoile

@yasserfaraazkhan is looking into it with this PR: https://github.com/mattermost/desktop/pull/2770

Looks like we might need to update the Windows image to a newer one as this is repeatedly failing.

devinbinnie (approved):
Looks good. I'll force merge this now since the CI issues are unrelated

PR #2764, opened 2023-06-20
[MM-53199] Update to use `notarytool`

devinbinnie:
#### Summary
We've been using a legacy method to notarize our app, and apparently a newer and much faster one exists called `notarytool`. I've updated the dependency, and enabled `notarytool`, so now macOS builds should be a LOT faster :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53199

```release-note
NONE
```

### DIFF
--- a/scripts/notarize.js
+++ b/scripts/notarize.js
@@ -26,5 +26,7 @@ exports.default = async function notarizing(context) {
         appPath: `${appOutDir}/${appName}.app`,
         appleId: process.env.APPLEID,
         appleIdPassword: process.env.APPLEIDPASS,
+        teamId: 'UQ8HT4Q2XM',
+        tool: 'notarytool',
     });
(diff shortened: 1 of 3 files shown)


scripts/notarize.js
```diff
@@ -26,5 +26,7 @@ exports.default = async function notarizing(context) {
         appPath: `${appOutDir}/${appName}.app`,
         appleId: process.env.APPLEID,
         appleIdPassword: process.env.APPLEIDPASS,
+        teamId: 'UQ8HT4Q2XM',
```
enahum, line 29: This should be an env var

enahum (changes requested):

enahum (approved):

marianunez (approved):

Issue #2769, opened 2023-06-22
[Crash]: When running "Run Diagnostics" from the help menu, the app crashes


LukasFiniarel:
### Before you file a crash report

- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

Version 5.4.0 commit: 420c63d

### Operating System

Windows 10 Pro x64 22H2 Build: 19045.3086

### Mattermost Server Version

7.10.0

### What type of crash did you experience?

Uncaught exception (I saw a dialog pop up that said "The Mattermost app quit unexpectedly")

### Crash report details

```shell
Application: Mattermost 5.4.0 [commit: 420c63d
]
Platform: Windows_NT 10.0.19045 x64
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at IncomingMessage.<anonymous> (C:\Users\lt\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:73601:31)
    at IncomingMessage.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

### Log Output

```shell
[2023-06-22 15:49:33.927] [debug] Diagnostics.run
[2023-06-22 15:49:33.927] [debug] Diagnostics.initializeValues
[2023-06-22 15:49:33.928] [debug] Diagnostics.getStepCount { stepsCount: 12 }
[2023-06-22 15:49:33.928] [debug] Diagnostics sendNotification DiagnosticsStarted
[2023-06-22 15:49:33.928] [info]  Diagnostics.executeSteps Started
```

devinbinnie:
Thanks for the report, created a ticket: https://mattermost.atlassian.net/browse/MM-53295

PR #2771, opened 2023-06-22
[MM-53295] Fix crash in diagnostics when server is unreachable

devinbinnie:
#### Summary
When running diagnostics, we check the current configured servers to ensure reachability. Unfortunately, there was a bug introduced when refactoring for the `ServerManager` in which the wrong URL was provided and would always crash. This revealed a further bug where if the server was unreachable, diagnostics would always crash.

This PR fixes the issue with the wrong URL, as well as catches the potential crash that could happen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53295
Closes https://github.com/mattermost/desktop/issues/2769

```release-note
Fix crash in diagnostics when server is unreachable
```

### DIFF
--- a/src/main/diagnostics/steps/step3.serverConnectivity.ts
+++ b/src/main/diagnostics/steps/step3.serverConnectivity.ts
@@ -24,7 +24,7 @@ const run = async (logger: ElectronLog): Promise<DiagnosticStepResponse> => {
                 throw new Error(`Invalid server configuration. Server Url: ${server.url}, server name: ${server.name}`);
             }
 
-            const serverOnline = await isOnline(logger, `${server.url}/api/v4/system/ping`);
+            const serverOnline = await isOnline(logger, `${server.url}api/v4/system/ping`);
 
             if (!serverOnline) {
                 throw new Error(`Server appears to be offline. Server url: ${server.url}`);
--- a/src/main/diagnostics/steps/step3.serverConnectivity.ts
+++ b/src/main/diagnostics/steps/step3.serverConnectivity.ts
@@ -5,6 +5,7 @@ import {ElectronLog} from 'electron-log';
 import {DiagnosticStepResponse} from 'types/diagnostics';
 
 import ServerManager from 'common/servers/serverManager';
+import {parseURL} from 'common/utils/url';
 
 import DiagnosticsStep from '../DiagnosticStep';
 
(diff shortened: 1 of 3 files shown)


src/main/diagnostics/steps/step3.serverConnectivity.ts
```diff
@@ -24,7 +25,7 @@ const run = async (logger: ElectronLog): Promise<DiagnosticStepResponse> => {
                 throw new Error(`Invalid server configuration. Server Url: ${server.url}, server name: ${server.name}`);
             }
 
-            const serverOnline = await isOnline(logger, `${server.url}/api/v4/system/ping`);
+            const serverOnline = await isOnline(logger, parseURL(`${server.url}/api/v4/system/ping`)?.toString());
```
marianunez, line 28: If the `parseUrl` here returns `undefined`, will it get defaulted above in the `isOnline` function? Do we need to check for `undefined`?

devinbinnie: Not in this case no, it's very unlikely that this is `undefined` but if it ever is (would only happen if someone tried to break something) the default case would catch it from crashing.

marianunez: I'm not familiar what this is trying to do just want to make sure that is OK to default to `'https://community.mattermost.com/api/v4/system/ping' in the rare case it's an invalid url here

marianunez (approved):

src/main/diagnostics/steps/step3.serverConnectivity.ts
devinbinnie: Yeah that's fine, this is all part of a health check in case users experience issues. Worst case scenario it pings `community`, which is only an indication that their server URL is wrong/unavailable. It's actually potentially helpful for it to work this way :P

harshilsharma63: Trying to understand the difference here, the use of `parseURL` here is only to validate if its a current URL, right? If so, are we including all the basic stuff such as removing trailing slashes, whitespaces, missing protocols, etc?

harshilsharma63 (approved):

src/main/diagnostics/steps/step3.serverConnectivity.ts
devinbinnie: `parseURL` is here to remove extra slashes, since we assume that one is needed before the `api` route. But often the slash is already there so we run it through `parseURL` to make sure it's all correct.

The whitespace and missing protocols should already all be taken care of at this point before the server was even added.

Issue #2766, opened 2023-06-22
[Bug]: Mattermost not detected in login screen, when server has plugins disabled


TwizzyDizzy:
### Before you file a bug report

- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [ ] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.4.0

### Operating System

Linux, Arch, x64

### Mattermost Server Version

7.8.7

### Steps to reproduce

* Open desktop client
* Type the URL of a Mattermost server that has plugins disabled
* Read error message

### Expected behavior

Normal reply:

> Server URL is valid. Server version: 7.8.7

### Observed behavior

Error message

> The server URL provided does not appear to point to a valid Mattermost server. Please verify the URL and check your connection.

### Log Output

```shell
No corresponding log output in the server.

Apache logfile that is in front of Mattermost:


0.0.0.0 - - [22/Jun/2023:09:59:24 +0200] "GET //api/v4/plugins/webapp HTTP/1.1" 501 194 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Mattermost/5.4.0 Chrome/112.0.5615.183 Electron/24.3.1 Safari/537.36"
```

### Additional Information

Response to the request (see log output) when plugins are disabled:

```
{"id":"app.plugin.disabled.app_error","message":"Plugins have been disabled. Please check your logs for details.","detailed_error":"","request_id":"h4ydi85jmbrbuckcdu4kfgbw7o","status_code":501}
```

instead of 

```
[]
```

when plugins are enabled.

[...]

PR #2772, opened 2023-06-22
[MM-53294] Fix bad user feedback on server URL validation when plugins are disabled

devinbinnie:
#### Summary
When a user tries to enter a server URL that has plugins disabled, the app will tell the user that their server isn't valid, even though it is.

This PR changes the logic to not require the user of the plugins API to get the server status.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53294
Closes https://github.com/mattermost/desktop/issues/2766

```release-note
Fix bad user feedback on server URL validation when plugins are disabled
```

### DIFF
--- a/src/app/serverViewState.ts
+++ b/src/app/serverViewState.ts
@@ -343,7 +343,7 @@ export class ServerViewState {
         const server = new MattermostServer({name: 'temp', url: parsedURL.toString()}, false);
         const serverInfo = new ServerInfo(server);
         try {
-            const remoteInfo = await serverInfo.fetchRemoteInfo();
+            const remoteInfo = await serverInfo.fetchConfigData();
             return remoteInfo;
         } catch (error) {
             return undefined;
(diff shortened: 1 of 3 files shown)


M-ZubairAhmed (approved):

src/app/serverViewState.ts
```diff
@@ -343,7 +343,7 @@ export class ServerViewState {
         const server = new MattermostServer({name: 'temp', url: parsedURL.toString()}, false);
         const serverInfo = new ServerInfo(server);
         try {
-            const remoteInfo = await serverInfo.fetchRemoteInfo();
```
larkox, line 346: Nit: this variable is no longer storing `remoteInfo`, but `configData`, so it should be renamed.

Similar in the many other places this function is called on this file.

larkox (approved):
Only one nit. Approving to not block it.

src/app/serverViewState.ts
devinbinnie: It actually is still storing the `RemoteInfo` type, it's just not necessarily complete at this point. So I think the naming is okay.

PR #2783, opened 2023-07-10
Fix React Developer Tools for some users

devinbinnie:
#### Summary
Some errors are popping up causing issues with React Developer Tools. This PR switches over to a different package to load them that seems to solve the issue.

```release-note
NONE
```


marianunez (approved):
Not familiar with that library but dependency change LGTM 👍

Issue #2786, opened 2023-05-26
App can't be resized to 1/3 screen width


duncan-bayne:
#### Summary
On macOS, the Mattermost desktop application won't allow itself to be resized to 1/3 of the screen width; this breaks "thirds" tiling in e.g. Rectangle.

#### Steps to reproduce
1. Install Mattermost 5.3.1 on macOS 13.4.
2. Open Mattermost and log in.
3. Set screen resolution either to 1800x1169, or the default of 1512x982.
4. Drag the edge of the Mattermost window and attempt to resize to 1/3 of the screen width.

#### Expected behavior
I expect the window to resize as directed.

#### Observed behavior (that appears unintentional) 
The window refuses to resize below a certain point; the resize pointer remains, but the window just stops resizing.

devinbinnie:
@duncan-bayne I've transferred this issue over to the Desktop App repo since we're making changes here.

We did a quick UX review and were able to drop the minimum width down to 600px. This is the lowest we can currently go while still maintaining a good UX in the Desktop App, so I hope that helps a bit. We might be able to drop it down further in the future but this is what we can do for now.

PR #2787, opened 2023-07-11
[MM-53490] Reduce min window width to 600px

devinbinnie:
#### Summary
Dropping the minimum width of the Desktop App down to 600px to accommodate for users wanting to have the app in a smaller window.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/2786
https://mattermost.atlassian.net/browse/MM-53490

```release-note
Set minimum window width to 600px
```

### DIFF
--- a/e2e/specs/menu_bar/full_screen.test.js
+++ b/e2e/specs/menu_bar/full_screen.test.js
@@ -19,7 +19,7 @@ describe('menu/view', function desc() {
         env.createTestUserDataDir();
         env.cleanTestConfig();
         fs.writeFileSync(env.configFilePath, JSON.stringify(config));
-        fs.writeFileSync(env.boundsInfoPath, JSON.stringify({x: 0, y: 0, width: 700, height: 240, maximized: false, fullscreen: false}));
+        fs.writeFileSync(env.boundsInfoPath, JSON.stringify({x: 0, y: 0, width: 600, height: 240, maximized: false, fullscreen: false}));
         await asyncSleep(1000);
         this.app = await env.getApp();
         this.serverMap = await env.getServerMap(this.app);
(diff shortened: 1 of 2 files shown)


streamer45:
Out of curiosity, wouldn't Mattermost render in "mobile view" at that point?

devinbinnie:
> Out of curiosity, wouldn't Mattermost render in "mobile view" at that point?

Yes, but that still happens at 700px as well.

streamer45 (approved):

M-ZubairAhmed (approved):

duncan-bayne:
> > Out of curiosity, wouldn't Mattermost render in "mobile view" at that point?
>
> Yes, but that still happens at 700px as well.

As the originator of this request, that's fine with me :). I use a tiling window manager, so apps that set a minimum size (which seems to be most Electron apps?) break my desktop.

Very happy for the app to switch to mobile view if that makes sense, I just want to be able to resize it the way _I_ want :)

streamer45:
> > > Out of curiosity, wouldn't Mattermost render in "mobile view" at that point?
> > 
> > 
> > Yes, but that still happens at 700px as well.
> 
> As the originator of this request, that's fine with me :). I use a tiling window manager, so apps that set a minimum size (which seems to be most Electron apps?) break my desktop.
> 
> Very happy for the app to switch to mobile view if that makes sense, I just want to be able to resize it the way _I_ want :).
> 
> Thanks for looking at this issue so promptly.

Interesting, I am an i3wm user and can resize at will. But probably because the window manager ignores anything the app says :p

duncan-bayne:
@streamer45 I normally use StumpWM, and it's the same - StumpWM does what StumpWM wants :)

The minimum size thing hits me on macOS.  I'm having the same problem with Spotify, FWIW.  The OS won't let the "WM" (Rectangle) override the app preferences.

Personal opinion: apps shouldn't have opinions on their geometry limits, let alone be able to enforce them over the user's instructions.

streamer45:
That makes sense, thanks for the context!

PR #2800, opened 2023-07-27
[MM-53847] Expose isDev API to allow webapp to check if the app is in dev mode

devinbinnie:
#### Summary
This change is being done to allow the Desktop App to continue to authenticate properly against versions of the server/webapp that have the new Desktop Auth flow enabled (see here: https://github.com/mattermost/mattermost/pull/24140)

It exposes a boolean call to check if the app is in dev mode, and if so, the server/webapp will post a deep link to the `mattermost-dev:` protocol which should avoid interference with the locally installed production version.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53847

```release-note
NONE
```


src/main/app/initialize.ts
```diff
@@ -235,7 +235,7 @@ function initializeBeforeAppReady() {
     AllowProtocolDialog.init();
 
     if (isDev && process.env.NODE_ENV !== 'test') {
-        log.info('In development mode, deeplinking is disabled');
+        app.setAsDefaultProtocolClient('mattermost-dev');
```
larkox, line 238: Shouldn't we keep the log, or is it no longer valid?

larkox (approved):

src/main/app/initialize.ts
devinbinnie: It's no longer valid, deep linking will technically be enabled in development mode, just under `mattermost-dev:`

devinbinnie:
@calebroseland Can you please expedite review so we can cut v5.5 RC1? Thanks!

calebroseland (approved):
Sorry about the delay; thanks for the ping!

PR #2812, opened 2023-08-29
[MM-54279] Fix issues around working with saved downloads

devinbinnie:
#### Summary
There was a report of users having the app crash when restarting due to missing files that were downloaded, deleted from the file system, but not cleared from the Downloads dropdown.

I wasn't able to reproduce the exact issue, but I found and fixed 2 issues that were likely contributing:
- The initial crash report was due to an `ENOENT` error (file not found) which I've wrapped in a `try/catch` to make sure that the app won't crash in the same way again.
- The reason that this was hard to reproduce was being caused by the saved downloads list not saving correctly (it was being corrupted by multiple async saves), so I've added a queue of promises that need to happen subsequently in order to save correctly.

Between these two fixes this should help alleviate the linked ticket. I also noticed that in my case I had an issue where the window wouldn't load at all, which should now be fixed (might also fix some of the other reports of this kind)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54279

```release-note
NONE
```

### DIFF
--- a/src/common/JsonFileManager.ts
+++ b/src/common/JsonFileManager.ts
@@ -6,6 +6,7 @@ import fs from 'fs';
 export default class JsonFileManager<T> {
     jsonFile: string;
     json: T;
+    private saving?: Promise<void>;
 
     constructor(file: string) {
         this.jsonFile = file;
@@ -16,9 +17,9 @@ export default class JsonFileManager<T> {
         }
     }
 
-    writeToFile(): Promise<void> {
+    write(json: string): Promise<void> {
         return new Promise((resolve, reject) => {
-            fs.writeFile(this.jsonFile, JSON.stringify(this.json, undefined, 2), (err) => {
+            fs.writeFile(this.jsonFile, json, (err) => {
                 if (err) {
                     // No real point in bringing electron-log into this otherwise electron-free file
                     // eslint-disable-next-line no-console
@@ -29,6 +30,16 @@ export default class JsonFileManager<T> {
                 resolve();
             });
         });
+    }
+
+    writeToFile(): Promise<void> {
+        const json = JSON.stringify(this.json, undefined, 2);
+        if (this.saving) {
+            this.saving = this.saving.then(() => this.write(json));
+        } else {
+            this.saving = this.write(json);
+        }
+        return this.saving;
     }
 
     setJson(json: T): void {
--- a/src/main/downloadsManager.ts
+++ b/src/main/downloadsManager.ts
@@ -174,14 +174,19 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
         }
 
         for (const file of Object.values(this.downloads)) {
-            if (file.bookmark) {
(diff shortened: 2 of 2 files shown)


src/main/downloadsManager.ts
```diff
@@ -174,14 +174,19 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
         }
 
         for (const file of Object.values(this.downloads)) {
-            if (file.bookmark) {
-                this.bookmarks.set(this.getDownloadedFileId(file), {originalPath: file.location, bookmark: file.bookmark});
+            try {
```
hmhealey, line 177: Should we minimize the try/catch to only be around whatever we expect to throw the error? 0/5

src/main/downloadsManager.ts
```diff
@@ -174,14 +174,19 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+                if (file.bookmark) {
+                    this.bookmarks.set(this.getDownloadedFileId(file), {originalPath: file.location, bookmark: file.bookmark});
```
hmhealey, line 179: Does this need to be unset in the catch statement at all, or would `clearFile` do that?

hmhealey (approved):
A few questions, but otherwise looks good

src/common/JsonFileManager.ts
```diff
@@ -31,6 +32,16 @@ export default class JsonFileManager<T> {
         });
     }
 
+    writeToFile(): Promise<void> {
+        const json = JSON.stringify(this.json, undefined, 2);
+        if (this.saving) {
+            this.saving = this.saving.then(() => this.write(json));
```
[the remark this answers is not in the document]
devinbinnie: It should be saved as a resolved Promise so it should be called immediately on the next event loop. I haven't actually confirmed this is how it works, but testing seems to prove it does :)

src/main/downloadsManager.ts
devinbinnie: I want to make sure we also try/catch the bookmark call, it's another system call that has potential to fail. I think we're okay to do it all.

src/common/JsonFileManager.ts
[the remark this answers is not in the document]
hmhealey: Yep, that seems to work. This is a pretty nice way of writing an async queue that doesn't involve a `while(true)` somewhere

Screenshot 2023-09-01 at 11 06 22 AM

src/common/JsonFileManager.ts
```diff
@@ -31,6 +32,16 @@ export default class JsonFileManager<T> {
...
+        } else {
+            this.saving = this.write(json);
+        }
+        return this.saving;
```
marianunez, line 42: So we need this return?

marianunez (approved):

src/common/JsonFileManager.ts
devinbinnie: Yes, we still want the method to return a promise in case it wants to `await` it.

PR #2824, opened 2023-09-06
Fix app name test

devinbinnie:
#### Summary
One of the E2E tests broke when the app name change went in. This fixes it.

```release-note
NONE
```


e2e/specs/startup/app.test.js
```diff
@@ -59,7 +59,7 @@ describe('startup/app', function desc() {
         it('MM-T4985 should show app name in title bar when no servers exist', async () => {
             const mainWindow = this.app.windows().find((window) => window.url().includes('index'));
             const titleBarText = await mainWindow.innerText('.app-title');
-            titleBarText.should.equal('Mattermost');
+            titleBarText.should.equal('Electron');
```
marianunez, line 62: App name is Electron in tests?

devinbinnie: Yeah it's the default for dev mode.

marianunez (approved):

PR #2829, opened 2023-09-11
[MM-54463] Enforce secure keyboard entry on macOS when a password box is focused

devinbinnie:
#### Summary
When a user types a password, there is a risk that another process on the user's computer could read the keyboard input and retrieve sensitive information from the user. On macOS, there is an API designed to prevent this.

This PR adds a call to that API that fires whenever the user has a password box focused.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54463

```release-note
NONE
```

### DIFF
--- a/src/main/app/intercom.ts
+++ b/src/main/app/intercom.ts
@@ -147,3 +147,9 @@ export function handlePingDomain(event: IpcMainInvokeEvent, url: string): Promis
         throw new Error('Could not find server ' + url);
     });
 }
+
+export function handleToggleSecureInput(event: IpcMainEvent, secureInput: boolean) {
+    // Enforce macOS to restrict processes from reading the keyboard input when in a password field
+    log.debug('handleToggleSecureInput', secureInput);
+    app.setSecureKeyboardEntryEnabled(secureInput);
+}
--- a/src/main/preload/mattermost.js
+++ b/src/main/preload/mattermost.js
@@ -378,21 +378,12 @@ window.addEventListener('resize', () => {
 let isPasswordBox = false;
 
 window.addEventListener('focusin', (event) => {
-    if (isPasswordBox) {
-        return;
-    }
-
-    if (event.target.tagName === 'INPUT' && event.target.type === 'password') {
+    const targetIsPasswordBox = event.target.tagName === 'INPUT' && event.target.type === 'password';
+    if (targetIsPasswordBox && !isPasswordBox) {
         ipcRenderer.send(TOGGLE_SECURE_INPUT, true);
-        isPasswordBox = true;
-    }
-});
-
-window.addEventListener('focusout', () => {
-    if (!isPasswordBox) {
-        return;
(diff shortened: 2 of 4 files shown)


devinbinnie:
Must be merged after the final v5.5 release is cut on Friday. This will go into a patch release.

src/main/app/intercom.ts
```diff
@@ -147,3 +147,9 @@ export function handlePingDomain(event: IpcMainInvokeEvent, url: string): Promis
         throw new Error('Could not find server ' + url);
     });
 }
+
+export function handleToggleSecureInput(event: IpcMainEvent, secureInput: boolean) {
+    // Enforce macOS to restrict processes from reading the keyboard input when in a password field
+    log.debug('handleToggleSecureInput', secureInput);
```
larkox, line 153: Do we want to keep the log?

src/main/preload/mattermost.js
```diff
@@ -373,3 +374,25 @@ ipcRenderer.on(CALLS_JOIN_REQUEST, (event, message) => {
 window.addEventListener('resize', () => {
     ipcRenderer.send(VIEW_FINISHED_RESIZING);
 });
+
+let isPasswordBox = false;
```
larkox, line 378: Are we positive any focus out event always come before a new focusin?

larkox (approved):
LGTM, just two minor things to confirm.

src/main/preload/mattermost.js
devinbinnie: From testing it would appear so, I'm not sure if by definition it has to.

Either way it's low risk if not, in the worst case the secure input stays on a little longer.

src/main/preload/mattermost.js
```diff
@@ -373,3 +374,25 @@ ipcRenderer.on(CALLS_JOIN_REQUEST, (event, message) => {
...
+        return;
+    }
+
+    if (event.target.tagName === 'INPUT' && event.target.type === 'password') {
+        ipcRenderer.send(TOGGLE_SECURE_INPUT, true);
+        isPasswordBox = true;
+    }
+});
+
+window.addEventListener('focusout', () => {
```
hmhealey, line 391: This might not work if you switch between two password input boxes like we probably have for confirming your password in the settings modal.

Instead of clearing `isPasswordBox` on `focusout`, what if we use a single focus handler and write it something like
```js
// obviously, this needs a better name
const isNowPasswordBox = event.target.tagName === 'INPUT' && event.target.type === 'password'; 
if (isNowPasswordBox && !isPasswordBox) {
    // turn on secure
} else if (!isNowPasswordBox && isPasswordBox) {
    // turn off secure
}

isPasswordBox = isNowPasswordBox;
```
That way, it'll stay active when you move between password inputs, and it'll only need one handler so you don't need to worry about event order either

hmhealey (changes requested):

hmhealey (approved):

PR #2831, opened 2023-09-13
[MM-54465] Remove lowercase call when checking path names

devinbinnie:
#### Summary
When we were checking if two URLs were equal, we were assuming that like the URL's origin, the pathname was also not case-sensitive. This is not the case so it was possible that two URLs with different URLs could be considered equal.

This PR removed the call to `toLowerCase()` to make sure we respect case sensitivity for path names.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54465

```release-note
Fixed an issue where two different URLs could be considered the same.
```

### DIFF
--- a/src/common/Validator.ts
+++ b/src/common/Validator.ts
@@ -201,7 +201,7 @@ export function validateV0ConfigData(data: ConfigV0) {
 function cleanURL(url: string): string {
     let updatedURL = url;
     if (updatedURL.includes('\\')) {
-        updatedURL = updatedURL.toLowerCase().replace(/\\/gi, '/');
+        updatedURL = updatedURL.replace(/\\/gi, '/');
     }
     return updatedURL;
 }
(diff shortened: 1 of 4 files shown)


src/common/Validator.ts
```diff
@@ -201,7 +201,7 @@ export function validateV0ConfigData(data: ConfigV0) {
 function cleanURL(url: string): string {
     let updatedURL = url;
     if (updatedURL.includes('\\')) {
-        updatedURL = updatedURL.toLowerCase().replace(/\\/gi, '/');
```
marianunez, line 204: Are there any cases where later we compare the server url that it make break now if it's there any casing mistmatch?

devinbinnie: Shouldn't be. We were doing this to make sure the path name was lower case, and this function is mostly to get rid of double slashes. We do a lower case when we compared host and origin in other parts of the code.

marianunez (approved):
LGTM

src/common/Validator.ts
esarafianou: @devinbinnie isn't it this: https://github.com/mattermost/desktop/blob/master/src/common/utils/url.ts#L9 that we should be changing?

esarafianou (approved):

larkox (approved):

PR #2832, opened 2023-09-13
[MM-54465] Add Permissions Manager, require manual interaction to approve use of microphone and camera, to send notifications, and to use location per server

devinbinnie:
#### Summary
We've added a `PermissionsManager` class to take care of permissions requests. This is put in place to make sure that we validate permissions requests per server, instead for just one server.

This will check for `media` (microphone/camera), `notifications` and `geolocation` requests and pop a dialog box to the user asking them to verify the request's origin.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54465

```release-note
Added extra permissions check for microphone, camera, notifications and location on a per-server basis
```


src/main/notifications/index.ts
```diff
@@ -49,6 +50,10 @@ export function displayMention(title: string, body: string, channel: {id: string
         data,
     };
 
+    if (!await PermissionsManager.doPermissionRequest(webcontents.id, 'notifications', view.view.server.url.toString())) {
```
devinbinnie, line 53: We add a manual permission check here because notifications are sent manually by Electron.

M-ZubairAhmed (approved):

src/main/notifications/index.ts
hmhealey, line 53: 0/5 on if this is actually an issue, but it feels weird to see an await inside a negated if condition. I'd tend to split that onto its own line

src/main/permissionsManager.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    doPermissionRequest = async (
+        webContentsId: number,
+        permission: string,
+        requestingURL: string,
+    ) => {
+        log.debug('doPermissionRequest', requestingURL, permission);
+
+        // is the requested permission type supported?
+        if (!supportedPermissionTypes.includes(permission)) {
```
hmhealey, line 74: You may want to consider adding an explicit type for `permission`. If you do it as a union type or string enum, you shouldn't need to confirm if the permission is supported or not

src/main/permissionsManager.ts
```diff
@@ -0,0 +1,170 @@
...
+        if (!isTrustedURL(parsedURL, serverURL)) {
+            return false;
+        }
+
+        // For certain permission types, we need to confirm with the user
+        if (authorizablePermissionTypes.includes(permission)) {
+            const currentPermission = this.json[parsedURL.origin]?.[permission];
+
+            // If previously allowed, just allow
+            if (currentPermission?.allowed) {
```
hmhealey, line 110: Nothing to change here. I just wanted to mention that I appreciate the way these are broken up since it makes it very easy to understand the flow through this method

hmhealey (approved):

src/main/permissionsManager.ts
```diff
@@ -0,0 +1,170 @@
...
+            };
+            this.writeToFile();
+
+            if (response > 0) {
+                return false;
+            }
+        }
+
+        // We've checked everything so we're okay to grant the remaining cases
+        return true;
```
esarafianou, line 153: Which are the remaining cases?

src/main/permissionsManager.ts
```diff
@@ -0,0 +1,170 @@
...
+        // is the requested permission type supported?
+        if (!supportedPermissionTypes.includes(permission)) {
+            return false;
+        }
+
+        // allow if the request is coming from the local renderer process instead of the remote one
+        const mainWindow = MainWindow.get();
+        if (mainWindow && webContentsId === mainWindow.webContents.id) {
+            return true;
+        }
```
esarafianou, line 82: I see that this was in the old code too, but why don't we check the rest if the request is coming from the local renderer process?

devinbinnie: The local renderer process is trusted, and we control what's loaded into it. Unless the Mattermost Desktop App itself is malicious, we can assume that local code is trusted.

src/main/notifications/index.ts
devinbinnie: I've been using this more and more, it's a nice shorthand. Going to keep as-is.

src/main/permissionsManager.ts
devinbinnie: I tried working with that, the problem is that there is a set of constants Electron allowed, a set we allow, and a set we allow but also require user interaction for. It really wasn't providing any extra value, and we still have to check that the permission coming from the handler is one of our supported ones.

src/main/permissionsManager.ts
esarafianou: Thanks for the explanation!

src/main/permissionsManager.ts
esarafianou: Haven't we already returned `true` for all these 3 cases?

devinbinnie: The only two we return true for earlier are:
- if it's the main window
- if the permission was previously allowed by the user

esarafianou: Sounds good, thanks for explaining

esarafianou (approved):
LGTM!

PR #2807, opened 2023-08-17
Replaced hardcoded app name with parameter in localizable strings

Partizann:
#### Summary
Replaced hardcoded app name with parameter in localizable strings

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [x] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [x] completed [Mattermost Contributor Agreement](https://mattermost.com/contribute/)
- [x] executed `npm run lint:js` for proper code formatting

#### Device Information
This PR was tested on: mac OS 13.2.1

#### Release Note
```release-note
NONE
```

PR #2862, opened 2023-09-29
[MM-54591] Use getVersion endpoint to report app name

devinbinnie:
#### Summary
For some reason `app` wasn't initialized in the preload script at the right time, causing issues that were breaking the downloads dropdown. This PR pushes us back to using the old endpoint of `getAppVersion` that provides both the version and app name when the dropdown initializes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54591

```release-note
Fixed an issue where the downloads dropdown would not open on auto update notification.
```

### DIFF
--- a/src/renderer/components/DownloadsDropdown/Update/UpdateWrapper.tsx
+++ b/src/renderer/components/DownloadsDropdown/Update/UpdateWrapper.tsx
@@ -11,14 +11,25 @@ import 'renderer/css/components/Button.scss';
 
 type OwnProps = {
     item: DownloadedItem;
+    appName: string;
 }
 
-const UpdateWrapper = ({item}: OwnProps) => {
+const UpdateWrapper = ({item, appName}: OwnProps) => {
     if (item.state === 'available') {
-        return <UpdateAvailable item={item}/>;
+        return (
+            <UpdateAvailable
+                item={item}
+                appName={appName}
+            />
+        );
     }
     if (item.state === 'completed') {
-        return <UpdateDownloaded item={item}/>;
+        return (
+            <UpdateDownloaded
+                item={item}
+                appName={appName}
+            />
+        );
     }
     return null;
 };
--- a/src/renderer/downloadsDropdown.tsx
+++ b/src/renderer/downloadsDropdown.tsx
@@ -18,6 +18,7 @@ type State = {
     darkMode?: boolean;
     windowBounds?: Electron.Rectangle;
     item?: DownloadedItem;
+    appName?: string;
 }
 
 class DownloadsDropdown extends React.PureComponent<Record<string, never>, State> {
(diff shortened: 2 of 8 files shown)


marianunez (commented):
@devinbinnie just to understand, is this a regression from a change in 5.5? If so, can you point me to the PR so I can reference for this review?

devinbinnie:
> @devinbinnie just to understand, is this a regression from a change in 5.5? If so, can you point me to the PR so I can reference for this review?

Yep, was caused here: https://github.com/mattermost/desktop/pull/2807 and partially fixed here: https://github.com/mattermost/desktop/pull/2813

src/renderer/components/DownloadsDropdown/Update/UpdateWrapper.tsx
```diff
@@ -11,14 +11,25 @@ import 'renderer/css/components/Button.scss';
...
+        );
     }
     if (item.state === 'completed') {
-        return <UpdateDownloaded item={item}/>;
+        return (
+            <UpdateDownloaded
+                item={item}
+                appName={appName}
+            />
+        );
```
marianunez, line 32: Not introduced in this PR but wondering why are these two cases separate? Seems they return the same thing

marianunez (approved):
Thanks @devinbinnie! Any tests we can consider in the future for this area?

src/renderer/components/DownloadsDropdown/Update/UpdateWrapper.tsx
devinbinnie: One is `UpdateAvailable` the other is `UpdateDownloaded`, there are separate components.

devinbinnie:
> Thanks @devinbinnie! Any tests we can consider in the future for this area?

We don't currently have front-end testing set up for Desktop App, since it wasn't creating a lot of value.
E2E would be nice but we don't have a way to simulate auto-updates in the testing environment yet :(

src/renderer/downloadsDropdown.tsx
```diff
@@ -40,6 +41,9 @@ class DownloadsDropdown extends React.PureComponent<Record<string, never>, State
             window.desktop.downloadsDropdown.focus();
         });
 
+        window.desktop.getVersion().then(({name}) => {
```
hmhealey, line 44: I'm a bit surprised to see that this works when you're removing this field in the type definition below. I guess it gets its type from somewhere else?

hmhealey (approved):

src/renderer/downloadsDropdown.tsx
devinbinnie: It does! It's here: https://github.com/mattermost/desktop/blob/2a88175b09e46fdae6b680d251825a711a1941f5/src/types/window.ts#L65

PR #2864, opened 2023-10-03
integrating WIP languages

ctlaltdieliet:
#### Summary
Integrating the WIP-translations
#### Checklist
- [X] completed [Mattermost Contributor Agreement](https://mattermost.com/contribute/)
#### Release Note
```release-note
NONE
```


ctlaltdieliet:
cc @lieut-data

lieut-data (approved):
I need to double check that the desktop app only lists supported languages. @devinbinnie, do you know if this is the case already?

devinbinnie:
> I need to double check that the desktop app only lists supported languages. @devinbinnie, do you know if this is the case already?

Yes, we have a supported languages list here: https://github.com/mattermost/desktop/blob/master/i18n/i18n.ts
Do we want to add the new ones we've added to that list?

lieut-data:
> > I need to double check that the desktop app only lists supported languages. @devinbinnie, do you know if this is the case already?
> 
> Yes, we have a supported languages list here: https://github.com/mattermost/desktop/blob/master/i18n/i18n.ts Do we want to add the new ones we've added to that list?

No need to update: we're going to start committing the WIP languages to the desktop app, but we wanted to avoid these showing up in the UI. Thanks!

devinbinnie (approved):
Thanks for the work on this guys :)

PR #2868, opened 2023-10-03
[MM-54584] Fixed an issue where clicking on a link to an unregistered protocol on macOS would cause the app to crash

devinbinnie:
#### Summary
When calling `shell.openExternal` on macOS, it returns a `Promise`. If the protocol used in the URL provided is not registered, the Promise will reject.

We were not properly handling this rejection, and due to the changes in v5.4 that included unhandled promise rejections, the app was simply crashing even though it should just fail silently.

This PR adds the handler to catch the error, log in, and then fail silently like we would expect.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54584

```release-note
Fixed an issue where clicking on a link to an unregistered protocol on macOS would cause the app to crash.
```


hmhealey (approved):

marianunez (approved):
Thanks @devinbinnie

PR #2869, opened 2023-10-04
[MM-54742] Force secure keyboard entry off when window loses focus and when servers switch

devinbinnie:
#### Summary
We missed a couple cases in which it wouldn't make sense to have secure keyboard entry still enabled, so I've made sure to turn it off when the window is out of focus, and when servers switch. I've also added a check so that when the view is focused, if the password entry is still the active element, we turn the secure keyboard entry back on.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54742

```release-note
NONE
```

### DIFF
--- a/src/main/windows/mainWindow.ts
+++ b/src/main/windows/mainWindow.ts
@@ -27,6 +27,7 @@ import {
     MAIN_WINDOW_RESIZED,
     MAIN_WINDOW_FOCUSED,
     VIEW_FINISHED_RESIZING,
+    TOGGLE_SECURE_INPUT,
 } from 'common/communication';
 import Config from 'common/config';
 import {Logger} from 'common/log';
@@ -314,6 +315,7 @@ export class MainWindow extends EventEmitter {
         globalShortcut.unregisterAll();
 
         this.emit(MAIN_WINDOW_RESIZED, this.getBounds());
+        ipcMain.emit(TOGGLE_SECURE_INPUT, null, false);
 
         // App should save bounds when a window is closed.
         // However, 'close' is not fired in some situations(shutdown, ctrl+c)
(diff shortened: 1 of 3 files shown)


hmhealey (commented):
What is the `force` option for? It seems like every case here has it set to `false`

Edit: Never mind. I can't read. I blame the PR diff view having line wrapping for code

devinbinnie:
> What is the `force` option for? It seems like every case here has it set to `false`

See here: https://github.com/mattermost/desktop/pull/2869/files#diff-cf2a83e275be0d5e5a6d16e74136e69e59f363ad257978dd01bae9752d613d19R395

We use it when the view gains focus.

M-ZubairAhmed (approved):
Wasnt this nice, using mattermost was always secure keyword entry on :P Just kidding

src/main/windows/mainWindow.ts
```diff
@@ -314,6 +315,7 @@ export class MainWindow extends EventEmitter {
         globalShortcut.unregisterAll();
 
         this.emit(MAIN_WINDOW_RESIZED, this.getBounds());
+        ipcMain.emit(TOGGLE_SECURE_INPUT, null, false);
```
jupenur, line 318: This is the right idea because main process is trusted and so the event here can't be forged. The web context in the renderer can still forge focus events though, so the main process `onBlur` and `onFocus` handlers should be made "authoritative". Meaning, if the window is blurred according to trusted data from the main process, any focus events from the renderer should be ignored. Events from the renderer should only be acted on if the window is already in focus.

jupenur (changes requested):

jupenur (approved):

Issue #2881, opened 2023-10-11
[Bug]: Permission for audio not allowed


BKapelari:
### Before you file a bug report

- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.5.1

### Operating System

Multible macOS Versions (v13.5.1 and v14.0)

### Mattermost Server Version

9.0.1

### Steps to reproduce

We have on mutlible Macs the same problem with the new version of Mattermost Desctop on Mac. The Version 5.5.1 has an error with microphone and screen sharing. The permissions are given in the Mac system settings. It is, i have tested it a lot, only a problem with Mattermost Desktop 5.5.1 (universal, X64 and silicon). Version 5.5.0 works perfektly - So we use this now.

<img width="303" alt="Bildschirmfoto 2023-10-11 um 11 50 32" src="https://github.com/mattermost/desktop/assets/4718590/2b048a10-0d99-420e-848b-d768f60fc913">

### Expected behavior

sound should work

### Observed behavior

the permissions are given, but the programm reacts like it would not be

### Log Output

```shell
[2023-10-11 12:24:12.501] [error] Logger Log level set to: info
[2023-10-11 12:24:12.624] [warn]  [App.Initialize] Current working directory is /, changing into /Applications/Mattermost.app/Contents/MacOS
[2023-10-11 12:24:12.732] [info]  [App.Initialize] Autoupgrade disabled: false
[2023-10-11 12:24:12.820] [info]  [ServerDropdownView] init
[2023-10-11 12:24:12.842] [info]  [i18nManager] Set new language de
```

devinbinnie:
[...]

PR #2893, opened 2023-10-24
[MM-55054] Consider a matching origin for a media request as a trusted URL when checking permissions

devinbinnie:
#### Summary
With the new permission changes, we were stopping servers with subpaths from being able to grant the `media` permission, due to being too strict about which URLs we were considering trusted.

The `media` permission generally populates the `securityOrigin` field when doing a permission check, so when checking for that permission if the URL can be trusted, we just need to match the two origins.

This PR makes that exception in the permissions handler.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55054
Closes https://github.com/mattermost/desktop/issues/2881

```release-note
Fixed an issue where servers on a subpath could not grant the `media` permission
```

### DIFF
--- a/src/main/permissionsManager.test.js
+++ b/src/main/permissionsManager.test.js
@@ -65,7 +65,7 @@ describe('main/PermissionsManager', () => {
                 return null;
             }
         });
-        isTrustedURL.mockImplementation((url, baseURL) => baseURL.toString().startsWith(url.toString()));
+        isTrustedURL.mockImplementation((url, baseURL) => url.toString().startsWith(baseURL.toString()));
     });
 
     afterEach(() => {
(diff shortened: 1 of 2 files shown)


larkox (approved):

marianunez (approved):

marianunez:
@devinbinnie could we add steps in the tickets on how we will do QA validation on this ?

marianunez:
@devinbinnie could we add steps in the tickets on how we will do QA validation on this ?

esarafianou:
@devinbinnie if a domain e.g. example.com hosts 2 Mattermost servers: one with URL example.com/mm1 and one with URL example.com/mm2, will after this change these 2 Mattermost servers need to ask for media permission separately or if example.com/mm1 is granted `media` permission, example.com/mm2 will inherit that permission without asking?

devinbinnie:
> @devinbinnie if a domain e.g. example.com hosts 2 Mattermost servers: one with URL example.com/mm1 and one with URL example.com/mm2, will after this change these 2 Mattermost servers need to ask for media permission separately or if example.com/mm1 is granted `media` permission, example.com/mm2 will inherit that permission without asking?

Correct, because we are using the `securityOrigin` field which is what I assume is the correct field to use for the `media` permission. It's not provided for the others. Not sure if this is a problem or not.

esarafianou (approved):
LGTM!

PR #2905, opened 2023-11-08
[MM-55261] Allow embedded video links to go fullscreen after a permission check

devinbinnie:
#### Summary
When adding per-server permissions, I neglected to test for YouTube and other embedded videos that will want to fullscreen the app. Those requests always come from the video's source and would get auto-denied otherwise.

This PR adds an exception for the `fullscreen` permission to allow the user to decide whether to allow/deny the full screen permission, which allows these embedded videos to work again.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55261

```release-note
Fix an issue where user's could not fullscreen embedded videos.
```

### DIFF
--- a/src/main/permissionsManager.ts
+++ b/src/main/permissionsManager.ts
@@ -105,13 +105,17 @@ export class PermissionsManager extends JsonFileManager<Permissions> {
             return false;
         }
 
+        // Exception for embedded videos such as YouTube
+        // We still want to ask permission to do this though
+        const isExternalFullscreen = permission === 'fullscreen' && parsedURL.origin !== serverURL.origin;
+
         // is the requesting url trusted?
-        if (!(isTrustedURL(parsedURL, serverURL) || (permission === 'media' && parsedURL.origin === serverURL.origin))) {
+        if (!(isTrustedURL(parsedURL, serverURL) || (permission === 'media' && parsedURL.origin === serverURL.origin) || isExternalFullscreen)) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/permissionsManager.ts
```diff
@@ -105,13 +105,17 @@ export class PermissionsManager extends JsonFileManager<Permissions> {
             return false;
         }
 
+        // Exception for embedded videos such as YouTube
+        // We still want to ask permission to do this though
+        const isExternalFullscreen = permission === 'fullscreen' && parsedURL.origin !== serverURL.origin;
+
         // is the requesting url trusted?
-        if (!(isTrustedURL(parsedURL, serverURL) || (permission === 'media' && parsedURL.origin === serverURL.origin))) {
+        if (!(isTrustedURL(parsedURL, serverURL) || (permission === 'media' && parsedURL.origin === serverURL.origin) || isExternalFullscreen)) {
```
marianunez, line 113: Do we want to ask permission or not for external full screen? Is this supposed to be `!isExternalFullscreen`?

devinbinnie: We do, there's an additional `!` at the beginning of that statement.

marianunez: Ah gotcha, didn't see that it was added in the two parenthesis there

marianunez (approved):

larkox (approved):

esarafianou (approved):
LGTM!

PR #2930, opened 2023-12-15
fix: Use labeled event from pull request to trigger apps build

toninis:
#### Summary

Currently when we added the label `Build Apps for PR` mattermod picked this up and created the related branches .  
Related core [here](https://github.com/mattermost/mattermost-mattermod/blob/master/server/build_apps.go) . 

We will remove the related code which was used for mobile and desktop repositories in favor of the labeled pull request event. 

- Check status will be generated as before on the PR level

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-5751

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build-for-pr.yml
+++ b/.github/workflows/build-for-pr.yml
@@ -1,11 +1,9 @@
 name: build-for-pr
 
 on:
-  push:
-    # only for build-pr branches
-    # build-pr-*
-    branches:
-      - build-pr-*
+  pull_request:
+    types:
+      - labeled
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


toninis:
> Thanks @toninis!
> 
> @devinbinnie In majority of the PRs, do you find it necessary to have the builds in all platforms available for testing or are targeted to specific platform only? I'm asking for I'll propose to have some level of granularity instead of for all always, such as:
> 
> * Build Apps for PR
> * Build App for Linux
> * Build App for MacOS
> * Build App for Windows
> 
> Just thinking about the cost of running the pipeline especially with [MacOS](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers).

Thanks @saturninoabril for the comment . We are doing this for mobile . It's possible to add some more labels for granularity . cc: @devinbinnie for feedback . 
In fact I do not think we pay that much for the desktop repo . We can handle that

phoinixgrr (approved):

devinbinnie:
@saturninoabril @toninis I'm okay with adding the granularity if it saves us some time/money. Generally I add that label for UX/QA to review a Desktop App build so I can just target whatever system they're testing on.

That said, for QA we really should be testing across the board for some features, so having the ability to do all 3 is still helpful. But in those cases, I think it's fine to just add 3 labels.

devinbinnie (approved):

PR #2936, opened 2023-12-22
[MM-55720] Add menu item to open developer tools for Call Widget window

streamer45:
#### Summary

PR adds a dynamic menu item to open developers tools for the Call widget window (see screenshot).

UPDATE: after the below conversation we are adding a dedicated `Developer Tools` submenu to keep things tidy.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55720

#### Screenshots

image

#### Release Note

```release-note
Added a new View > Developer Tools submenu that contains items to access the developer tools for all desktop created windows.

Added a new menu item to open the developers tools for the Call widget window
```

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -84,6 +84,7 @@
   "main.menus.app.view.devToolsAppWrapper": "Developer Tools for Application Wrapper",
   "main.menus.app.view.devToolsCurrentCallWidget": "Developer Tools for Call Widget",
   "main.menus.app.view.devToolsCurrentServer": "Developer Tools for Current Server",
+  "main.menus.app.view.devToolsSubMenu": "Developer Tools",
   "main.menus.app.view.downloads": "Downloads",
   "main.menus.app.view.find": "Find..",
   "main.menus.app.view.fullscreen": "Toggle Full Screen",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -82,6 +82,7 @@
   "main.menus.app.view.actualSize": "Actual Size",
   "main.menus.app.view.clearCacheAndReload": "Clear Cache and Reload",
   "main.menus.app.view.devToolsAppWrapper": "Developer Tools for Application Wrapper",
+  "main.menus.app.view.devToolsCurrentCallWidget": "Developer Tools for Call Widget",
   "main.menus.app.view.devToolsCurrentServer": "Developer Tools for Current Server",
   "main.menus.app.view.downloads": "Downloads",
   "main.menus.app.view.find": "Find..",
--- a/src/main/menus/app.test.js
+++ b/src/main/menus/app.test.js
@@ -10,7 +10,26 @@ import ServerManager from 'common/servers/serverManager';
 
 import ServerViewState from 'app/serverViewState';
 
+import CallsWidgetWindow from 'main/windows/callsWidgetWindow';
+
 import {createTemplate} from './app';
+
+jest.mock('fs-extra', () => ({
+    readFileSync: jest.fn(),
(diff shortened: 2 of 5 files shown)


devinbinnie:
@streamer45 I actually don't think this change is necessary. We can already open the Dev Tools for the Calls Widget simply by clicking on the widget (so that it's focused) and then going to View > Developer Tools for Application Wrapper.

I can see this being a bit confusing, but since we might want to support additional widgets or windows down the line, we might actually want to not set the precedent to have to add a Dev Tools for each one. So maybe the best thing would be to fix the string so that it says something like "Developer Tools for Focused Application Window".

If we really want to have a separate menu item for each one, I think that would be okay, but we might want to rethink the menu formatting (ie. maybe have a separate Developer Tools menu so that we're not constantly expanding things?)

streamer45:
> @streamer45 I actually don't think this change is necessary. We can already open the Dev Tools for the Calls Widget simply by clicking on the widget (so that it's focused) and then going to View > Developer Tools for Application Wrapper.
> 

Thanks Devin. I didn't know that was possible, maybe because it doesn't seem to be working on my Linux system.
Maybe the window is getting out of focus the moment I click on the menu The keyboard shortcut works as expected though.

> I can see this being a bit confusing, but since we might want to support additional widgets or windows down the line, we might actually want to not set the precedent to have to add a Dev Tools for each one. So maybe the best thing would be to fix the string so that it says something like "Developer Tools for Focused Application Window".

It's not much the label that makes it confusing (easy to change) but the fact that it needs to be focused. The widget is a frameless window and depending on your operating system / window manager it may be very hard to tell if it's focused. 

> If we really want to have a separate menu item for each one, I think that would be okay, but we might want to rethink the menu formatting (ie. maybe have a separate Developer Tools menu so that we're not constantly expanding things?)

An explicit item still makes sense to me as it's consistent and easy to reproduce. I agree it won't scale too well but that sounds like a UX problem, maybe we should be looping in someone from that side.

streamer45:
> Yes this is true, but as a developer I think we can know since we'll see which Developer Tools pops up :P.

Unfortunately "as a developer" doesn't apply that often when looking to get calls debugging logs from end users, hence why I am stressing on making this dead simple :p

> Thinking on this more, I actually like the idea of a sub-menu to do this. On top of which we could iterate and provide explicit items for the servers as well. Would you be willing to add the sub-menu for this PR? Just to move the App Wrapper, Current Server and Calls Widgets items into one menu. I can work on refining it later.

Sure, if you are happy with the minor "breaking change" UX wise I can give it a go.

devinbinnie:
> > Yes this is true, but as a developer I think we can know since we'll see which Developer Tools pops up :P.
> 
> Unfortunately "as a developer" doesn't apply that often when looking to get calls debugging logs from end users, hence why I am stressing on making this dead simple :p
> 

Lol understandable :P.

> > Thinking on this more, I actually like the idea of a sub-menu to do this. On top of which we could iterate and provide explicit items for the servers as well. Would you be willing to add the sub-menu for this PR? Just to move the App Wrapper, Current Server and Calls Widgets items into one menu. I can work on refining it later.
> 
> Sure, if you are happy with the minor "breaking change" UX wise I can give it a go.

Yep go for it. I'll take a look again at this PR when that's done :)

streamer45:
@devinbinnie Updated, let me know:

image

devinbinnie (approved):
Thanks @streamer45!

streamer45:
I am adding the docs needed label since I'd expect this to require a quick pass to our documentation in case we are mentioning how to access these menus. /cc @cwarnermm

PR #2943, opened 2024-01-08
CLD-4665 Fix package versioning

mvitale1989:
#### Summary

TBD

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-4665

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/actions/patch-nightly-version/patch-version.go
+++ b/.github/actions/patch-nightly-version/patch-version.go
@@ -28,7 +28,7 @@ func main() {
 	json.Unmarshal(packageBytes, &packageInfo)
 
 	originalVersion := fmt.Sprintf("%s", packageInfo["version"])
-	nightlyVersion := fmt.Sprintf("%s~nightly.%s", strings.Split(originalVersion, "~")[0], time.Now().Format("20060102"))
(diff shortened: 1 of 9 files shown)


devinbinnie:
@mvitale1989 One thing we'll have to make sure of is that auto-updating works still for Windows, because I think they only support `semver` style versioning. Looks like right now it's failing because of the tilde.

mvitale1989:
@devinbinnie understood, i'll need to keep windows in mind as well.

Although the tilde seems to be a problem even for node itself, since the [version field](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version) gets validated by the [`semver` package](https://www.npmjs.com/package/semver#versions), which doesn't accept tildes:
```
$ ./node_modules/.bin/semver 1.2.3-1 || echo failure
1.2.3-1
$ ./node_modules/.bin/semver 1.2.3~1 || echo failure
failure
```
I think we need to look for a solution other than modifying the version field in `package.json`. The package ordering issue (and tilde as a solution) seems to be linux specific, hopefully electron-builder gives us some tunable for this.

I'll take a look tomorrow but if you already have an idea how this could be done please let me know

mvitale1989:
This needs to be fixed upsteram, on the `electron-builder` project. I will open a PR. See [this comment on the ticket](https://mattermost.atlassian.net/browse/CLD-4665?focusedCommentId=160413) for details.

PR #2958, opened 2024-02-09
[MM-54872] Ensure matched subpaths are exact and not substrings of each other

devinbinnie:
#### Summary
If you tried to add a server on a subpath, where the domains were the same and the second server's subpath was a substring of the first server's subpath, the app would erroneously claim it was the same server. This is because we weren't comparing the server subpaths as though they were the last in the chain.

This PR adds the `getFormattedPathName` call around the original subpath as well as the first to ensure we're comparing them correctly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54872
Closes https://github.com/mattermost/mattermost/issues/24891

```release-note
Fixed an issue where you couldn't add a second server with a similar subpath as a configured server
```

### DIFF
--- a/src/common/servers/serverManager.test.js
+++ b/src/common/servers/serverManager.test.js
@@ -13,7 +13,7 @@ jest.mock('common/config', () => ({
 jest.mock('common/utils/url', () => ({
     parseURL: jest.fn(),
     isInternalURL: jest.fn(),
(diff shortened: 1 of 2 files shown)


harshilsharma63 (approved):
Looks good!

larkox (approved):

Issue #2891, opened 2023-10-21
[Bug]: "Add Server" dialogue cannot be used to add server


bsl:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

Version 5.5.1 (63830002741)

### Operating System

macOS Ventura 13.4 (22F66)

### Mattermost Server Version

9.1.0

### Steps to reproduce

Click the server selection dropdown. Click "Add a server". Type or paste a Mattermost URL that the app does not already know about.

### Expected behavior

I expect that the app will go verify that the URL is actually for a Mattermost server, leave the entered URL in the text field, allow setting the display name of the server, and allow me to click Add.

### Observed behavior

When the URL is entered, the app goes and checks that the URL is for a Mattermost server. Having verified that it is, it clears the URL text field. It's still possible to click Add, but no server is actually added.

### Log Output

```shell
[2023-10-21 13:41:03.422] [error] Logger Log level set to: info
[2023-10-21 13:41:03.537] [warn]  [App.Initialize] Current working directory is /Users/b/Library/Containers/Mattermost.Desktop/Data, changing into /Applications/Mattermost.app/Contents/MacOS
[2023-10-21 13:41:03.654] [info]  [App.Utils] MAS: No files to migrate, skipping
[2023-10-21 13:41:03.666] [info]  [App.Initialize] Autoupgrade disabled: false
[2023-10-21 13:41:03.712] [info]  [ServerDropdownView] init
[2023-10-21 13:41:03.725] [warn]  [i18nManager] Failed to set new language en-US
[...]

PR #2959, opened 2024-02-09
[MM-55134] Don't blank the Site URL when it's not properly configured on an otherwise working server

devinbinnie:
#### Summary
As part of server validation, we are checking against the Site URL on the configured server. However, when the URL wasn't configured we weren't doing any further validation, but we were assuming that there was a Site URL there, which would blank out the URL for users.

This PR assumes that if the user's URL is okay and there's no Site URL configured, to use the user's provided URL.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/2891
https://mattermost.atlassian.net/browse/MM-55134

```release-note
Fixed an issue where the user's URL would be cleared after being entered in the Add Server modal
```

### DIFF
--- a/src/app/serverViewState.test.js
+++ b/src/app/serverViewState.test.js
@@ -446,6 +446,22 @@ describe('app/serverViewState', () => {
             expect(result.validatedURL).toBe('https://mainserver.com/');
         });
 
+        it('should not update the users URL when the Site URL is blank', async () => {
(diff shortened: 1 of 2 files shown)


M-ZubairAhmed (approved):

harshilsharma63 (approved):
Looks good!

PR #2966, opened 2024-02-27
MM-56996 Fix nan dependency

mvitale1989:
#### Summary

This PR adapts the [nan library](https://www.npmjs.com/package/nan) to work with V8 engine `v12.x` (which underwent a major version bump between [electron 27](https://github.com/electron/electron/releases/tag/v27.0.0) and [electron 28](https://github.com/electron/electron/releases/tag/v28.0.0)).

The root issue is that some function overloads for the `SetAccessor()` function were removed, but the `nan.h` header was not updated accordingly.
- SetAccessor in V8 `v11.x`: https://v8docs.nodesource.com/node-18.16/db/d5f/classv8_1_1_object_template.html#ad096a9b454b9ddc11d2d7124a03dc746
- SetAccessor in V8 `v12.x`: https://v8docs.nodesource.com/node-20.3/db/d5f/classv8_1_1_object_template.html#ab8b382adf193624286604916568b671c
  * Note that the two overloads that were removed were those with the `Local< AccessorSignature >` argument type

This issue arises in E2E test jobs, e.g.: https://github.com/mattermost/desktop/actions/runs/8014552242/job/21893346736
You can see that after the upgrades, there have been a bunch of compilation failures in `nan/nan.h` when trying to compile for the Electroni 29 runtime.

The `nan` repo now points to a fork under my account, until the fix is merged upsteram.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56996

#### Release Note
```release-note
NONE
```


package.json
```diff
@@ -235,6 +234,7 @@
   },
   "dependencies": {
     "macos-notification-state": "2.0.2",
+    "nan": "github:mvitale1989/nan#fix/electron-29",
```
saturninoabril, line 237: Proper way would be to install `nan@2.18.0` then patch using https://www.npmjs.com/package/patch-package so you don't need to maintain your own fork and eventually bump the version once available..

saturninoabril (changes requested):
Thanks @mvitale1989, nice find! Btw, looks like the Playwright tests were still failing.

package.json
mvitale1989: Interesting, I didn't know that existed, thanks Saturn! I'll re-implement it using patch-package later today

mvitale1989:
> Thanks @mvitale1989, nice find! Btw, looks like the Playwright tests were still failing.

Yep I could also reproduce it locally: the code currently freezes at [this line](https://github.com/mattermost/desktop/blob/master/e2e/modules/environment.js#L218) during the `beforeAll`, so all test cases time out because of that.
I'm not very familiar with this setup, so I have asked Devin if he could also take a look when possible.

devinbinnie:
@mvitale1989 Upgrading Playwright to 1.42.0 stopped the test hanging for me.

mvitale1989:
> @mvitale1989 Upgrading Playwright to 1.42.0 stopped the test hanging for me.

Super, that worked on my machine as well, thanks for checking Devin! :rocket: :tada: 
I'll implement the suggestion by Saturn and test in CI too

yasserfaraazkhan (approved):
thank you 🎉  @mvitale1989  @devinbinnie

package.json
mvitale1989: This is now implemented

saturninoabril (approved):
Awesome, thanks @mvitale1989!

PR #2970, opened 2024-03-07
[MM-52696] Upgrade and clean up Desktop App dev dependencies

devinbinnie:
This change includes upgrades to ESLint, React, TS and other dependencies used for Desktop App development. I've also removed a bunch of dependencies and unnecessary pieces of code from the app, including storybook.

Most of the files changed in this PR are just automatic fixes by ESLint, the rest are changes to the `package.json` file. You can separate them out by commit if need be to make the review easier.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52696

```release-note
NONE
```


api-types/package.json
```diff
@@ -18,7 +18,7 @@
     "directory": "api-types"
   },
   "peerDependencies": {
-    "typescript": "^4.3"
+    "typescript": "5.3.3"
```
hmhealey, line 21: Can this be made less restrictive so that this version doesn't need to be updated in lock step with the web app? I went with `^4.3.0 || ^5.0.0` for the client and types packages recently since I don't know of anything specific to 5.0 that is even needed for them, but we could potentially even go with `>=4.3.0`

hmhealey (changes requested):

hmhealey (approved):

PR #2972, opened 2024-03-08
[MM-45780] Update to macos-notification-state 3.0, update entitlements and provisioning profile to allow Communication Notification access

devinbinnie:
#### Summary
The DND status was being blocked for MAS users and users on Ventura (macOS 13) or higher. This was caused by Apple's change in the way they communicate their notification status to apps. There is now an API called Communication Notifications that controls this, and it needed to be enabled in Apple Developer to work.

This change does the following things:
- Upgrades us to `macos-notification-state` 3.0, which supports the new API
- Adds entitlements for both DMG and MAS builds to be able to request focus status.
- Adds a provisioning profile for the DMG build to allow it to access the focus status.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45780

```release-note
Fixed an issue for macOS 13+ users on the MAS build where the OS-level DND status was not respected
```

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -104,8 +104,10 @@
     "extendInfo": {
       "NSMicrophoneUsageDescription": "Microphone access may be used by Mattermost plugins, such as Jitsi video conferencing.",
       "NSCameraUsageDescription": "Camera access may be used by Mattermost plugins, such as Jitsi video conferencing.",
+      "NSFocusStatusUsageDescription": "Focus status is used by Mattermost to determine whether to send notifications or not.",
       "LSFileQuarantineEnabled": true
-  }
+    },
+    "provisioningProfile": "./mac.provisionprofile"
   },
   "mas": {
     "hardenedRuntime": false,
@@ -119,7 +121,7 @@
     "singleArchFiles": "*"
   },
   "masDev": {
-    "provisioningProfile": "./mas-dev.provisionprofile"
+    "provisioningProfile": "./dev.provisionprofile"
   },
   "dmg": {
     "background": "src/assets/osx/DMG_BG.png",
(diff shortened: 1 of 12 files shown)


src/main/notifications/index.test.ts
```diff
@@ -184,7 +184,7 @@ describe('main/notifications', () => {
                 value: 'darwin',
             });
 
-            getDarwinDoNotDisturb.mockReturnValue(true);
```
marianunez, line 187: Did this test pass before with the issue we had?

marianunez (approved):

src/main/notifications/index.test.ts
devinbinnie: Yes, because we were mocking the library. We unfortunately can't really test the behaviour of notifications since we don't have a library capable of verifying that the OS received them (that I know of).

marianunez: Ah ok, makes sense. There is so much that we are really testing with mocks 🤷‍♀️

toninis (approved):

electron-builder.json
```diff
@@ -100,12 +100,14 @@
     "hardenedRuntime": true,
     "gatekeeperAssess": true,
     "entitlements": "./entitlements.mac.plist",
-    "entitlementsInherit": "./entitlements.mac.plist",
+    "entitlementsInherit": "./entitlements.mac.inherit.plist",
     "extendInfo": {
       "NSMicrophoneUsageDescription": "Microphone access may be used by Mattermost plugins, such as Jitsi video conferencing.",
       "NSCameraUsageDescription": "Camera access may be used by Mattermost plugins, such as Jitsi video conferencing.",
+      "NSFocusStatusUsageDescription": "Focus status is used by Mattermost to determine whether to send notifications or not.",
```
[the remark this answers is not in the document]
devinbinnie: They are end-user yes, I believe I had this one reviewed a while back when I tried to fix this bug before, but I'm not 100%. You'll see a dialog pop up when the app tries to send a notification for the first time.

I'm not sure if we can translate these, they're not part of our JS app so we can't use our own translation code.

PR #2974, opened 2024-03-08
Remove build hash from release builds

devinbinnie:
#### Summary
When we migrated to GitHub Actions, the hash got stuck in release builds, because we were still checking for the old CircleCI env variables. This PR switches over to the GitHub Actions one.

```release-note
NONE
```

### DIFF
--- a/webpack.config.base.js
+++ b/webpack.config.base.js
@@ -14,7 +14,7 @@ const webpack = require('webpack');
 const VERSION = childProcess.execSync('git rev-parse --short HEAD').toString();
 const isProduction = process.env.NODE_ENV === 'production';
 const isTest = process.env.NODE_ENV === 'test';
-const isRelease = process.env.CIRCLE_BRANCH && process.env.CIRCLE_BRANCH.startsWith('release-');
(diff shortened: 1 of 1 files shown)


marianunez (approved):

webpack.config.base.js
```diff
@@ -14,7 +14,7 @@ const webpack = require('webpack');
 const VERSION = childProcess.execSync('git rev-parse --short HEAD').toString();
 const isProduction = process.env.NODE_ENV === 'production';
 const isTest = process.env.NODE_ENV === 'test';
-const isRelease = process.env.CIRCLE_BRANCH && process.env.CIRCLE_BRANCH.startsWith('release-');
+const isRelease = process.env.GITHUB_WORKFLOW && process.env.GITHUB_WORKFLOW.startsWith('release');
```
toninis, line 17: Just FYI this will pick up also `release-mas`

toninis (approved):
🚀

webpack.config.base.js
devinbinnie: Yep that's what we want. It should only show for nighties and dev builds.

PR #2975, opened 2024-03-08
Various build and developer fixes, some optimizations

devinbinnie:
#### Summary
This change contains a host of build and development changes, including, but not limited to the following:
- Move entitlements to `resources/mac`
- Move non-dev dependencies back to dependencies in `package.json`, exclude all `node_modules` except for native ones from final packaged builds
- Remove `node-jq`, `fs-extra`
- Change production builds to use `mode: production` for better minification
- Clean up the webpack configs and remove unnecessary config options
- Add specific preload webpack config to target `electron-preload`
- Added chunk splitting for the renderer process to avoid dependency duplication
- Use `patch-package` to fix the `macos-notification-state` build instead of the postinstall script
- Reworked the watch script to watch the renderer process properly
- Updated a few more modules
- Removed some unnecessary scripts
- Reworked some of the package.json actions and parallelized the builds
- Remove `scripts/notarize.js` in favour of just using `electron-builder`
- Remove `@babel/cli` and just use `tsc` to run `check-build-config`

There should be no behavioural change in the app as a result of these changes. But we will want to make sure to test this thoroughly to see if any app behaviour breaks as a result.

```release-note
NONE
```


harshilsharma63 (approved):

toninis (approved):

scripts/watch.js
```diff
@@ -0,0 +1,26 @@
...
+        compiler.watch({}, (err, stats) => {
+            if (err) {
+                console.error(err);
+            }
+            process.stdout.write(stats.toString({colors: true}));
+            process.stdout.write('\n');
+            if (!stats.hasErrors()) {
+                electron.restart();
+            }
+            resolve();
```
hmhealey, line 23: Not sure if this matters or is even correct since I've never used the Node.js API for Webpack, but could this resolve the promise multiple times, and if so, is that a problem at all?

webpack.config.preload.js
```diff
@@ -0,0 +1,28 @@
...
+        electronPreload: true,
+    },
+    node: {
+        __filename: true,
+        __dirname: true,
+    },
+    target: 'electron-preload',
+});
+
+/* eslint-enable import/no-commonjs */
```
hmhealey, line 28: Despite what we used to think, you don't actually need to re-enable ESLint rules at the end of a file

webpack.config.renderer.js
```diff
@@ -175,10 +173,7 @@ module.exports = merge(base, {
         __filename: false,
         __dirname: false,
     },
-    target: 'electron-renderer',
-    devServer: {
-        port: WEBSERVER_PORT,
-    },
+    target: 'web',
```
hmhealey, line 176: For my own curiosity, what does changing the target to web do? Is the old one removed or is there some other reason you changed it?

hmhealey (approved):

scripts/watch.js
devinbinnie: I don't believe you can resolve a Promise multiple times, so this should be fine.

webpack.config.preload.js
devinbinnie: Yeah wasn't sure about this. Next clean-up PR :+1:

webpack.config.renderer.js
devinbinnie: I believe if you use `electron-renderer` it assumes that `nodeIntegration` was enabled meaning that `require` should be exposed. This isn't usually a problem, but with chunking multiple files it is because it will use NodeJS `require` instead of the webpack baked-in function.

It's probably safer altogether to compile this for web, since we should assume anything run in the renderer process only has access to the same global fields that a Chrome process would.

footsim:
Hi, could you explain why you do that and how notarization works now?
> - Remove scripts/notarize.js in favour of just using electron-builder

PR #2980, opened 2024-03-12
Stop constant reloading of BrowserViews for servers that are not available

devinbinnie:
#### Summary
For servers that might be offline, we had some retry logic that would constantly reload the whole Chrome page to try and reconnect to the server when it came back.

This was a bit heavy handed, and with recent changes it's now gumming up the logs. So I've changed the retry logic to just ping the server using the backend Fetch API and when it returns a success we can then try and reload the server correctly.

```release-note
NONE
```

### DIFF
--- a/src/main/server/serverAPI.test.js
+++ b/src/main/server/serverAPI.test.js
@@ -8,7 +8,6 @@ import {net, session} from 'electron';
 import {getServerAPI} from './serverAPI';
 
 const validURL = 'http://server-1.com/api/endpoint';
-const badDataURL = 'http://server-1.com/api/bad/endpoint';
 const testData = {
     name: 'some data',
     value: 'some more data',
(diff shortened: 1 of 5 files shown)


harshilsharma63 (approved):
Looks good!

harshilsharma63:
Not a blocker for this PR, but can we optimise this further by just pinging the server and reloading the server only if ping is successful? That will save a reload.

M-ZubairAhmed (approved):

devinbinnie:
> Not a blocker for this PR, but can we optimise this further by just pinging the server and reloading the server only if ping is successful? That will save a reload.

I don't think it really saves much, I don't want to fail on the first connection nor should we be waiting for a ping to start trying to load the page, since that might make for a slower feeling UX. I think it's fine to just try to retry the load a few times in case network conditions are bad/someone just turned on their WiFi or something, this was just to stop it from happening indefinitely.

PR #2998, opened 2024-04-04
[MM-57348] Support notification metrics from the Desktop App client

devinbinnie:
#### Summary
This PR enables the Desktop App to return data around whether a notification was successful or not. It relays that data back to the web clients that support it, and the web client will push that data through the websocket as needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57348

```release-note
NONE
```

### DIFF
--- a/api-types/lib/index.d.ts
+++ b/api-types/lib/index.d.ts
@@ -1,6 +1,4 @@
-// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
-// See LICENSE.txt for license information.
-export declare type DesktopSourcesOptions = {
+export type DesktopSourcesOptions = {
     types: Array<'screen' | 'window'>;
     thumbnailSize?: {
         height: number;
@@ -8,12 +6,12 @@ export declare type DesktopSourcesOptions = {
     };
     fetchWindowIcons?: boolean;
 };
-export declare type DesktopCaptureSource = {
+export type DesktopCaptureSource = {
     id: string;
     name: string;
     thumbnailURL: string;
 };
-export declare type DesktopAPI = {
+export type DesktopAPI = {
     isDev: () => Promise<boolean>;
     getAppInfo: () => Promise<{
         name: string;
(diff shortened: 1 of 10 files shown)


api-types/lib/index.d.ts
```diff
@@ -1,19 +1,17 @@
-// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
-// See LICENSE.txt for license information.
```
larkox, line 2: Shouldn't the linter complain since we are removing this header?

src/main/notifications/index.ts
```diff
@@ -97,10 +78,33 @@ class NotificationManager {
             this.allActiveNotifications.delete(mention.uId);
         });
 
-        mention.on('failed', () => {
-            this.allActiveNotifications.delete(mention.uId);
+        return new Promise((resolve) => {
```
larkox, line 81: I wonder if there is any chance this promise never gets resolved. I mean... are we positive by calling `mention.show` either the show or the failed code is going to execute? Is there any chance they won't (or take a worrying amount of time)?

If that were the case, we could add some timeout to resolve this if it takes longer than X (30s?).

larkox (approved):
Looks great. Just two minor considerations.

api-types/lib/index.d.ts
devinbinnie: I turned it off for this - I don't really wanna fight with having it in the built types, it should just be in our source code.

src/main/notifications/index.ts
devinbinnie: I was thinking that, and that would indicate a clear failure, but wouldn't really add more information in terms of metrics. If we never get an ACK as a result of this Promise not returning, then we can count that as a missed notification since we can't guarantee that `show()` was called or finished correctly. I think the `failed()` callback handles most of that, but the lack of an ACK should be just as good of an indication that something went wrong, and the metrics will reflect that as part of notification health.

larkox: I am not so much worried about the metrics, but the memory leak of promises being created and never getting resolved.

devinbinnie: Ahh, that's a good point. In that case I think a timeout is warranted.

src/main/notifications/index.ts
```diff
@@ -32,19 +32,19 @@ class NotificationManager {
 
         if (!Notification.isSupported()) {
             log.error('notification not supported');
-            return;
+            return {result: 'error', reason: 'notification_api', data: 'notification not supported'};
```
streamer45, line 35: Should we consider "codifying" the reasons so that other sides (e.g. web client, server) can understand them? Maybe a common enum type or something could be an option.

src/main/notifications/index.ts
larkox: This is now a "purist nit", so feel free to ignore.

Resolve is supposed to be used when the promise ends correctly. In theory, when it fails, we should reject the promise.

Nevertheless, I was trying to look in the internet some good reason to do it one way or another, and I think in this case would only add unneeded boilerplate to it :P

devinbinnie: Yeah also `reject` commonly causes an error state which is sort of the case, but not what we're trying to achieve here.

streamer45 (approved):
Makes sense, nothing blocking :+1:

src/main/notifications/index.ts
devinbinnie: I'm thinking of doing that once I have a final final decision on how to track everything. I'll probably do that in the phase where I actually start adding metrics around this stuff :P

PR #2999, opened 2024-04-05
Various QoL fixes for Desktop App

devinbinnie:
#### Summary
A slew of fixes for the Desktop App:
- Fixed an issue where removing a server would not clear mentions: https://mattermost.atlassian.net/browse/MM-57690
- Added the login/logout signal back for the external API - to allow for servers to tell the app when login/logout occurs, better automating cookie storage and mention counts (NOTE: A small breakage may occur in older servers where cookies may not be flushed easily, but we want to push for this new method anyways)
- Fixed an issue where Save Image right-clicking would crash the app (bad package)
- Reduced the amount of noise the renderer process would dump into the logs, we didn't want to show every `console.log` outside of debugging
- Fixed an issue where typing in your local server followed by a port would trip up the URL validation. Also made it so that if your server has a default name, it will use the beginning of its hostname instead (good for if you need to add community/hub ;))
- Couple small ESLint fixes

```release-note
Fixed an issue where removing a server would not clear mentions: https://mattermost.atlassian.net/browse/MM-57690
Fixed an issue where Save Image right-clicking would crash the app (bad package)
Fixed an issue where typing in your local server followed by a port would trip up the URL validation
```

### DIFF
--- a/src/app/serverViewState.ts
+++ b/src/app/serverViewState.ts
@@ -294,15 +294,15 @@ export class ServerViewState {
                 // If we can't reach the remote Site URL, there's probably a configuration issue
                 const remoteSiteURLInfo = await this.testRemoteServer(parsedSiteURL);
                 if (!remoteSiteURLInfo) {
-                    return {status: URLValidationStatus.URLNotMatched, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName, validatedURL: remoteURL.toString()};
+                    return {status: URLValidationStatus.URLNotMatched, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName === 'Mattermost' ? remoteURL.host.split('.')[0] : remoteInfo.siteName, validatedURL: remoteURL.toString()};
                 }
             }
 
             // Otherwise fix it for them and return
-            return {status: URLValidationStatus.URLUpdated, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName, validatedURL: remoteInfo.siteURL};
+            return {status: URLValidationStatus.URLUpdated, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName === 'Mattermost' ? remoteURL.host.split('.')[0] : remoteInfo.siteName, validatedURL: remoteInfo.siteURL};
         }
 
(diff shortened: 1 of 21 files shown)


src/app/serverViewState.ts
```diff
@@ -292,15 +294,15 @@ export class ServerViewState {
                 // If we can't reach the remote Site URL, there's probably a configuration issue
                 const remoteSiteURLInfo = await this.testRemoteServer(parsedSiteURL);
                 if (!remoteSiteURLInfo) {
-                    return {status: URLValidationStatus.URLNotMatched, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName, validatedURL: remoteURL.toString()};
+                    return {status: URLValidationStatus.URLNotMatched, serverVersion: remoteInfo.serverVersion, serverName: remoteInfo.siteName === 'Mattermost' ? remoteURL.host.split('.')[0] : remoteInfo.siteName, validatedURL: remoteURL.toString()};
```
larkox, line 297: You are repeating `remoteInfo.siteName === 'Mattermost' ? remoteURL.host.split('.')[0] : remoteInfo.siteName` in three different places. Any chance we can extract this logic?

Either setting the `serverName` earlier, or creating a function `getServerName(remoteInfo, remoteUrl)`.

src/main/views/webContentEvents.ts
```diff
@@ -298,17 +298,14 @@ export class WebContentsEventManager {
...
         switch (level) {
         case ConsoleMessageLevel.Error:
             logFn = wcLog.error;
             break;
         case ConsoleMessageLevel.Warning:
             logFn = wcLog.warn;
             break;
-        case ConsoleMessageLevel.Info:
-            logFn = wcLog.info;
-            break;
```
larkox, line 311: Shouldn't we just use `console.debug` when we want debug logs?

larkox (approved):
LGTM, just a couple of nits.

harshilsharma63 (approved):
Lots of nice changes!!!

src/main/views/webContentEvents.ts
devinbinnie: If the front-end exclusively did it then yes, but we want to reduce the noise that the front-end is generating, and that includes general `console.log` statements which are interpreted as `info` level.

PR #3002, opened 2024-04-09
[MM-57736] Force maximize state when restoring the window

devinbinnie:
#### Summary
On Windows, when the window is maximized, and we restore the window from minimized using the `restore()` command, we often get into a case where the window will not return to its maximized state. This gets the app into a weird state where sometimes users will not be able to re-maximize or may see some whitespace where the app should be.

This PR forces the app to keep track of the maximized state of the main window, and re-enter the maximized state on restore if it was supposed to be there before.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57736

```release-note
Fixed an issue where restoring the window from the tray icon could cause a strange state if the window was previously maximized
```

### DIFF
--- a/src/main/windows/mainWindow.ts
+++ b/src/main/windows/mainWindow.ts
@@ -49,6 +49,7 @@ export class MainWindow extends EventEmitter {
     private ready: boolean;
     private isResizing: boolean;
     private lastEmittedBounds?: Electron.Rectangle;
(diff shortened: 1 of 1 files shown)


harshilsharma63 (approved):
👌👌👌

larkox (approved):

PR #3005, opened 2024-04-11
[MM-57348] Change `result` to `status` as a better descriptor of the notification status

devinbinnie:
#### Summary
For consistency, and overall better naming, I'm changing the field name for the notification status to `status` instead of result. This keeps it in line with the server descriptors/constants.

```release-note
NONE
```


hmhealey (approved):
The code changes look good, but for what it's worth, I feel like `result` is a bit less overloaded of a term for this

larkox (approved):

PR #3017, opened 2024-04-24
[MM-58001] Fix an issue where the window position has a decimal number

devinbinnie:
#### Summary
When I fixed up the window sizing code for the settings window, I introduced a bug where you can sometimes report a floating-point value for the window position. This PR fixes that by simply `floor()`-ing the result.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58001

```release-note
NONE
```

### DIFF
--- a/src/main/app/utils.ts
+++ b/src/main/app/utils.ts
@@ -155,8 +155,8 @@ function getNewWindowPosition(browserWindow: BrowserWindow) {
     const mainWindowPosition = mainWindow.getPosition();
 
     return [
-        mainWindowPosition[0] + ((mainWindowSize[0] - newWindowSize[0]) / 2),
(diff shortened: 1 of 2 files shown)


src/main/app/utils.ts
```diff
@@ -169,6 +169,7 @@ export function resizeScreen(browserWindow: BrowserWindow) {
         width: size[0],
         height: size[1],
     });
+    log.info(validPosition);
```
marianunez, line 172: Do we want to keep this log?

marianunez (approved):

src/main/app/utils.ts
devinbinnie: LOL nope :rofl:

streamer45 (approved):
Thanks @devinbinnie , appreciate the quick fix on this!

PR #3024, opened 2024-05-03
[MM-58089] Disallow redirects to untrusted URLs without a permission prompt

devinbinnie:
#### Summary
It's possible to allow a server to redirect to an external application without a permission prompt, so I've added the prompt to this PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58089

```release-note
Fixed an issue where a server could redirect to an external application without a prompt
```


marianunez (approved):
Thanks @devinbinnie! Should we include a test to cover that it would open the dialog for this new case we are covering here?

devinbinnie:
> Thanks @devinbinnie! Should we include a test to cover that it would open the dialog for this new case we are covering here?

The existing tests should actually cover this case now as well so I think we're okay :)

hmhealey (approved):

jupenur (approved):

PR #3028, opened 2024-05-07
[MM-55341][MM-56894] Use `electron-builder` to create the MSI installer, drop support for 32-bit Windows and the NSIS installer

devinbinnie:
#### Summary
This PR contains changes to revamp our Windows installer building process. It does away with the custom script we had to build the MSI installer, and elects to use `electron-builder` with a few modifications to the flow.

Some features provided by the new MSI installer that work out of the box:
- One-click per-user installation, similar to our original EXE/NSIS installer
- Auto-update functionality built in, so that if you choose, you can still get automatic updates to the newest version.
- A host of administrative functionality via the command line:
    - Per-machine installation by specifying `ALLUSERS=1`
    - Outright disabling of automatic updates by specifying `DISABLEAUTOUPDATE=1`
    - Ability to modify the installation directory by specifying `APPLICATIONFOLDER=<install directory>`
    
With this PR, this also marks the following changes in support for the Desktop App:
- The MSI installer is now officially out of beta :tada:
- 32-bit Windows will no longer be supported for newer versions, and we will be officially supporting ARM64 (in a beta capacity for the time being)
- The EXE/NSIS installer will no longer be distributed as the main way to setup the Desktop App, all users can either use the ZIP for portable installation, or the MSI.

I've added a number of people for review, I will make comments on the code to help people better understand some of the changes that are happening.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55341
https://mattermost.atlassian.net/browse/MM-56894

```release-note
Drop support for 32-bit Windows, add support for ARM64 (in beta)
Drop support for the EXE/NSIS installer, ship only the MSI
```


devinbinnie:
Link to built installer: https://github.com/mattermost/desktop/actions/runs/9020601613/artifacts/1488474351

.github/workflows/build-for-pr.yml
```diff
@@ -123,6 +122,10 @@ jobs:
       - name: ci/build
         env:
           MM_WIN_INSTALLERS: 1
+          PFX_KEY: ${{ secrets.MM_DESKTOP_MSI_INSTALLER_PFX_KEY }}
```
devinbinnie, line 125: Will likely remove this after this PR is done being tested - I think we're okay to still just ship the ZIP for PR builds.

nab-77 (approved):
Woot! Consolidation of asset, upgrade of installs for customers, and also more flexibity.

Wins!

electron-builder.json
```diff
@@ -156,21 +156,29 @@
...
     ],
     "extraFiles": [
       {
         "filter": [
           "LICENSE.txt",
           "NOTICE.txt"
         ]
+      },
+      {
+        "from": "resources/windows/gpo",
```
devinbinnie, line 170: All builds will now ship with the GPO artifacts, since it's only helpful to those that need them and doesn't matter for those that don't.

package.json
```diff
@@ -61,9 +61,10 @@
     "start": "npm run create-linux-dev-shortcut && electron dist/ --disable-dev-mode",
     "restart": "run-s build start",
     "watch": "npm run create-linux-dev-shortcut && node scripts/watch.js",
-    "package": "run-s package:windows package:windows-nsis package:mac-with-universal package:mas package:linux",
-    "package:windows": "npm run build-prod && electron-builder --win zip --x64 --ia32 --arm64 --publish=never",
-    "package:windows-nsis": "npm run build-prod-upgrade && electron-builder --win zip nsis --x64 --ia32 --arm64 --publish=never",
+    "package": "run-s package:windows package:mac-with-universal package:mas package:linux",
+    "package:windows": "run-s package:windows-*",
+    "package:windows-zip": "npm run build-prod && electron-builder --win zip --x64 --arm64 --publish=never",
+    "package:windows-installers": "npm run build-prod-upgrade && electron-builder --win nsis --x64 --arm64 --publish=never && node scripts/cp_msi_app_update.js && electron-builder --win msi --x64 --arm64 --publish=never",
```
devinbinnie, line 67: It is mandatory for the MSI installer that we build the NSIS first. This is so we can copy the `app-update.yml` file necessary to tell the app how to auto-update for the MSI build.

patches/app-builder-lib+24.13.3.patch
```diff
@@ -0,0 +1,43 @@
+diff --git a/node_modules/app-builder-lib/templates/msi/template.xml b/node_modules/app-builder-lib/templates/msi/template.xml
+index 2d5cd3c..04ed672 100644
+--- a/node_modules/app-builder-lib/templates/msi/template.xml
++++ b/node_modules/app-builder-lib/templates/msi/template.xml
+@@ -26,6 +26,11 @@
+     <Property Id="ARPPRODUCTICON" Value="${iconId}"/>
+     {{ } -}}
+ 
++    <CustomAction Id="removeExeInstaller" Directory="APPLICATIONFOLDER" ExeCommand="[APPLICATIONFOLDER]Uninstall ${productName}.exe /S" Execute="immediate" Impersonate="yes" Return="asyncWait"/>
```
devinbinnie, line 9: This patch adds a step to the MSI to check for per-user installation of the NSIS installer. If that installation exists, it will run the uninstall program. This covers the case where a user might have the NSIS build installed from a previous build, or from auto-update, to avoid duplicate installations on the system.

@jupenur I called this one out in our security meeting, because I'm not sure how comfortable we are executing an arbitrary EXE like this. Let me know your thoughts on this one.

patches/app-builder-lib+24.13.3.patch
```diff
@@ -0,0 +1,43 @@
...
+ 
++    <Directory Id="resourcesDir" Name="APPLICATIONFOLDER:\resources\"/>
+     <Directory Id="TARGETDIR" Name="SourceDir">
+       <Directory Id="${programFilesId}">
+         {{ if (menuCategory) { }}
+@@ -110,6 +117,10 @@
+     {{-dirs}}
+ 
+     <ComponentGroup Id="ProductComponents" Directory="APPLICATIONFOLDER">
++      <Component Directory="resourcesDir">
```
devinbinnie, line 37: This patch injects the auto-update YML file into the MSI, since it's not normally supported by `electron-builder`. It is done conditionally in case a system admin wants to disable it when installing per-machine.

scripts/Makefile.ps1
devinbinnie, line 1: This file is no longer needed at all, as well as any of its dependencies. `electron-builder` takes care of the majority of the process now.

scripts/generate_release_markdown.sh
```diff
@@ -21,15 +21,11 @@ Release notes can be found here: https://docs.mattermost.com/install/desktop-app
 
 The download links can be found below.
 
-#### Windows - msi files (beta)
-$(print_link "${BASE_URL}/mattermost-desktop-${VERSION}-x64.msi")
-$(print_link "${BASE_URL}/mattermost-desktop-${VERSION}-x86.msi")
-
-#### Windows - setup exe files
-$(print_link "${BASE_URL}/mattermost-desktop-setup-${VERSION}-win.exe")
```
devinbinnie, line 29: We still technically build the EXE for automatic updates, but it's no longer being advertised to download. Users should always download the MSI for their system architecture.

@nab-77 Can I get your help coordinating with whoever runs the `mattermost.com` site to update this link to use the MSI for their architecture? It will require some code changes.

scripts/installer.nsh
```diff
@@ -0,0 +1,8 @@
+# This macro fetches the currently install MM version via MSI (if any) and uninstalls it first
+!macro customInit
```
devinbinnie, line 2: This is an NSIS script macro for when auto-update occurs. The user will be switched over to the EXE installation method, so we need to remove the MSI.

This should only be for users who've installed on a machine they control, on a per-user basis. Auto-update is automatically **disabled** for per-machine installations.

scripts/cp_msi_app_update.js
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+const fs = require('fs');
+fs.copyFileSync('release/win-unpacked/resources/app-update.yml', 'release/msi-app-update.yml');
+fs.unlinkSync('release/win-unpacked/resources/app-update.yml');
```
larkox, line 6: Two nitpicks, probably unneeded.
- Does it make sense to put these routes in a constant? The first one is repeated twice in this file, and I guess the second one will be used somewhere else, but maybe not in a JS environment...
- Do we want to check any possible error on the copy before trying to unlink the original file?

larkox (approved):
No issues from my side, just a nitpick.

yasserfaraazkhan (approved):
- e2e test passed on windows
- able to install the `mattermost-desktop-5.9.0-develop.1-win-x64.msi` on machine.
- Verified that it picks up old servers configured.
image

- Had this 1 error while installing. Had to Uninstall the other MM desktop version and the reinstall using the msi installer. I think this is expected.

Screenshot 2024-05-13 025453

devinbinnie:
> Screenshot 2024-05-13 025453

@yasserfaraazkhan When did you encounter this error? Was it reproducible?

Also, did you happen to try any of the admin commands?

scripts/installer.nsh
```diff
@@ -0,0 +1,8 @@
...
+  nsExec::ExecToStack "$\"powershell.exe$\" -command $\"$$Installer = New-Object -ComObject WindowsInstaller.Installer; $$MMProduct = $$Installer.ProductsEx('', '', 7) | Where-Object -FilterScript {$$_.InstallProperty('ProductName') -eq 'Mattermost'}; if ($$MMProduct -ne $$null) {Write-Host -NoNewline $$MMProduct.ProductCode()}$\""
+  Pop $0
+  Pop $1
+  StrCmp $1 "" 0 +1
+  ExecWait '"msiexec.exe" /x $1 /qn'
```
jupenur, line 7: What's the working directory here and is it included in the search path when locating `msiexec.exe`? I.e. can we be sure we're actually executing the thing we're supposed to? The same question goes for the `powershell.exe` invocation above.

patches/app-builder-lib+24.13.3.patch
jupenur: This seems safe since it's an absolute path to a directory where you wouldn't be able to drop files e.g. from a web browser.

scripts/installer.nsh
devinbinnie: I would guess that it would be wherever the installer is running from, which for most people would be wherever the update is downloaded to.
All of those are likely being picked up from the local PATH variable.

I'm not sure if they can be overridden, Windows might have a thing to stop that, but if this is a concern I think it makes sense to just add the absolute paths.

scripts/installer.nsh
```diff
@@ -0,0 +1,8 @@
+# This macro fetches the currently install MM version via MSI (if any) and uninstalls it first
+!macro customInit
+  nsExec::ExecToStack "$\"C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe$\" -command $\"$$Installer = New-Object -ComObject WindowsInstaller.Installer; $$MMProduct = $$Installer.ProductsEx('', '', 7) | Where-Object -FilterScript {$$_.InstallProperty('ProductName') -eq 'Mattermost'}; if ($$MMProduct -ne $$null) {Write-Host -NoNewline $$MMProduct.ProductCode()}$\""
+  Pop $0
+  Pop $1
+  StrCmp $1 "" 0 +1
+  ExecWait '"C:\WINDOWS\system32\msiexec.exe" /x $1 /qn'
```
jupenur, line 7: It's at least theoretically possible that `C:` isn't the system drive, so these should use environment variables instead. `C:\Windows\` is `%SystemRoot%` or `%windir%`.

devinbinnie: Thank you for catching that :P

toninis (approved):
🚀

PR #3032, opened 2024-05-08
[MM-58158] Make sure closing the permissions dialog results in a Deny instead of an Allow

devinbinnie:
#### Summary
When the permissions dialog pops to ask permission, if you close the dialog using the X on Windows/Linux, it returns a response of `0` which was resulting in an Allow. This seems counterintuitive to the user experience and could result in users allowing permissions they didn't expect.

This PR changes the ordering so that Allow is now 2, and Deny is 0 such that closing the dialog results in a Deny.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58158

```release-note
Fixed the permission prompt to Deny on closing the dialog
```

### DIFF
--- a/src/main/permissionsManager.ts
+++ b/src/main/permissionsManager.ts
@@ -152,16 +152,16 @@ export class PermissionsManager extends JsonFileManager<Permissions> {
                 detail: localizeMessage(`main.permissionsManager.checkPermission.dialog.detail.${permission}`, 'Would you like to grant {appName} this permission?', {appName: app.name}),
                 type: 'question',
                 buttons: [
-                    localizeMessage('label.allow', 'Allow'),
                     localizeMessage('label.deny', 'Deny'),
                     localizeMessage('label.denyPermanently', 'Deny Permanently'),
+                    localizeMessage('label.allow', 'Allow'),
                 ],
             });
 
             // Save their response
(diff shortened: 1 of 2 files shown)


src/main/permissionsManager.ts
```diff
@@ -152,16 +152,16 @@ export class PermissionsManager extends JsonFileManager<Permissions> {
...
                     localizeMessage('label.deny', 'Deny'),
                     localizeMessage('label.denyPermanently', 'Deny Permanently'),
+                    localizeMessage('label.allow', 'Allow'),
                 ],
             });
 
             // Save their response
             const newPermission = {
-                allowed: response === 0,
-                alwaysDeny: (response === 2) ? true : undefined,
```
marianunez, line 164: Should we define these are const for readiability and also if we need to change something it will propagate to other places where we check like below?

devinbinnie: This is really the only place that does it, and it's based on the ordering of the options. So it's not repeatable really. I think it's okay as-is for now.

marianunez: FWIW - Below in line 174 it's also used and in tests

marianunez (approved):

hmhealey (approved):

PR #3040, opened 2024-05-23
Improve notification logging

devinbinnie:
#### Summary
The logging we have for each step of the notification process is insufficient, as we are missing every step where the app might trip up and decide not to show a notification.

This PR beefs up the amount of logging in debug mode (and removes the logging that showed message content).

```release-note
NONE
```

### DIFF
--- a/src/main/notifications/index.ts
+++ b/src/main/notifications/index.ts
@@ -28,7 +28,7 @@ class NotificationManager {
     private restartToUpgradeNotification?: UpgradeNotification;
 
     public async displayMention(title: string, body: string, channelId: string, teamId: string, url: string, silent: boolean, webcontents: Electron.WebContents, soundName: string) {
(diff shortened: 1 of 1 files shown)


hmhealey (approved):

src/main/notifications/index.ts
```diff
@@ -28,25 +28,28 @@ class NotificationManager {
...
         }
 
         if (await getDoNotDisturb()) {
+            log.debug('do not disturb is on, will not send');
             return {status: 'not_sent', reason: 'os_dnd'};
         }
 
         const view = ViewManager.getViewByWebContentsId(webcontents.id);
         if (!view) {
+            log.error('missing view', webcontents.id);
```
marianunez, line 45: 👀  Looking at you `missing_view`

marianunez (approved):

Issue #3039, opened 2024-05-21
[Bug]: Part of app window is showing on second monitor


Profesor08:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.8.0

### Operating System

Windows 11 x64

### Steps to reproduce

1. Run application
2. Align to same size where second monitor is located
3. Do some restarts, reopening of window, make pc to sleep/hybernation and back.

### Expected behavior

Application window to keep it position

### Observed behavior

It shifts a little to right (in my case) and appears on second screen.

![image](https://github.com/mattermost/desktop/assets/8218016/6800843a-b224-452d-bcff-ff35df73c464)

### Log Output

```shell
[2024-05-21 00:00:33.147] [info]  [UpdateManager] Update for version 5.8.0 is not available (latest version: 5.8.0, downgrade is disallowed).
[2024-05-21 01:00:32.565] [info]  [UpdateManager] Checking for update
[2024-05-21 01:00:33.146] [info]  [UpdateManager] Update for version 5.8.0 is not available (latest version: 5.8.0, downgrade is disallowed).
[2024-05-21 02:00:32.572] [info]  [UpdateManager] Checking for update
[2024-05-21 02:00:33.115] [info]  [UpdateManager] Update for version 5.8.0 is not available (latest version: 5.8.0, downgrade is disallowed).
[2024-05-21 02:13:37.054] [error] [WebContentsEventM...] [producktv] [TAB_MESSAGING] [renderer] TypeError: Failed to fetch
[...]

PR #3045, opened 2024-05-28
[MM-58357] Account for monitor scale factor when creating the window from saved bounds

devinbinnie:
#### Summary
When we create the main window of the app using saved bounds from the disk, if the screen that we are putting the window is scaled to a higher value than 100%, the window will try to render outside the boundaries of the screen in a multi-monitor environment.

This is due to a bug in Electron (https://github.com/electron/electron/issues/10862) where the `setBounds()` function of the `BrowserWindow` seems to ignore scale factor altogether. This PR contains a workaround that will manually resize the saved bounds using the scale factor of the target monitor to put the window back correctly, avoiding any multi-monitor overlap.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58357
Closes https://github.com/mattermost/desktop/issues/3039

```release-note
Fixed an issue where scaled monitors in a multi-monitor setup may cause the window to be opened across 2 screens.
```

### DIFF
--- a/src/common/Validator.ts
+++ b/src/common/Validator.ts
@@ -179,7 +179,7 @@ export function validateArgs(data: Args) {
 }
 
 // validate bounds_info.json
-export function validateBoundsInfo(data: SavedWindowState) {
+export function validateBoundsInfo(data: SavedWindowState | null) {
     return validateAgainstSchema(data, boundsInfoSchema);
 }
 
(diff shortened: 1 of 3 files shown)


M-ZubairAhmed (approved):
Those are too many logs 😛

cpoile (approved):
Nice!

PR #3051, opened 2024-06-03
[MM-46634] Force window to be focused after navigation on notification click

devinbinnie:
#### Summary
When the webapp is focused, we make a call to mark the current channel as read. When a notification is clicked, we make a call to both focus the window and navigate to the new route if needed. This introduces a race condition in which the channel the user was on before switching to the one where the notification was triggered from would be marked read instead, causing the perception that an unread was missed.

This doesn't really happen much on the web side, as the calls are all run as part of a single thread and the `push()` action will likely update the `currentChannel` value before the focus handle can get in the way, but on Desktop App the focus action is called much earlier and done as a result of Electron telling the renderer process to focus. Then afterwards an IPC messages takes care of the routing.

This PR waits for the navigation callback from the webapp before focusing the application ensuring that the webapp will have switched to the new location before the focus trigger happens.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46634

```release-note
Fixed an issue where clicking a notification would clear unreads for the wrong channel.
```

### DIFF
--- a/src/main/notifications/index.ts
+++ b/src/main/notifications/index.ts
@@ -1,11 +1,11 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {app, shell, Notification} from 'electron';
+import {app, shell, Notification, ipcMain} from 'electron';
 import isDev from 'electron-is-dev';
 import {getDoNotDisturb as getDarwinDoNotDisturb} from 'macos-notification-state';
 
-import {PLAY_SOUND, NOTIFICATION_CLICKED} from 'common/communication';
+import {PLAY_SOUND, NOTIFICATION_CLICKED, BROWSER_HISTORY_PUSH} from 'common/communication';
 import Config from 'common/config';
 import {Logger} from 'common/log';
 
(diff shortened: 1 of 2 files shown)


src/main/notifications/index.ts
```diff
@@ -70,11 +70,16 @@ class NotificationManager {
             log.debug('notification click', serverName, mention.uId);
 
             this.allActiveNotifications.delete(mention.uId);
-            MainWindow.show();
-            if (serverName) {
+
+            // Show the window after navigation has finished to avoid the focus handler
+            // being called before the current channel has updated
+            const focus = () => {
```
marianunez, line 76: Not sure if relevant but before we were checking on `serverName` before trying to move to that view, do we want to keep that check?

marianunez:
Daniel is out this week so you may want to get another reviewer

src/main/notifications/index.ts
devinbinnie: No it was just a null-check that didn't have any relevance, so I removed it.

marianunez (approved):
LGTM

hmhealey (approved):
Great idea for fixing this. That's a much more elegant and simple fix than I was expecting when you originally mentioned this

Issue #3052, opened 2024-06-06
[Bug]: Notification banners and sound are not played on macOS when FocusStatus permission is denied


blackout400:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.8.0

### Operating System

Mac OS Sonoma 14.5

### Mattermost Server Version

9.5.1

### Steps to reproduce

Install Mattermost 5.8 from AppStore or Github.
Receive notification.

### Expected behavior

Notifications must work as set in Mattermost and Mac OS notification settings.

### Observed behavior

No sound, banner or icon bounce. Only red badge on dock icon.

### Log Output

```shell
[2024-06-06 10:28:55.391] [error] Logger Log level set to: info
[2024-06-06 10:28:55.440] [warn]  [Config] Failed to load configuration file from the filesystem. Using defaults.
[2024-06-06 10:28:55.441] [info]  [Config] Migrating config items successfully.
[2024-06-06 10:28:55.449] [warn]  [App.Initialize] Current working directory is /, changing into /Applications/Mattermost.app/Contents/MacOS
[2024-06-06 10:28:55.514] [info]  [App.Initialize] Autoupgrade disabled: false
[2024-06-06 10:28:55.516] [error] [MainWindow] Error: ENOENT: no such file or directory, open '/Users/yaroslav/Library/Application Support/Mattermost/bounds-info.json'
    at Object.readFileUtf8 (node:internal/fs/sync:25:18)
    at Object.readFileSync (node:fs:441:19)
    at t.readFileSync (node:electron/js2c/node_init:2:9771)
[...]

PR #3053, opened 2024-06-06
[MM-58455] Add error handling when FocusStatus is not authorized on macOS

devinbinnie:
#### Summary
A recent change that allowed the macOS app to correctly use the `FocusStatus` to prevent notifications under DND was completely stopping notifications if permission wasn't given. This was caused by the app not performing any error handling from the DND check, which would throw an error if the app wasn't authorized to check the status.

This PR adds that error handling, and logs a warning in the logs to denote that the check failed.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/3052
https://mattermost.atlassian.net/browse/MM-58455

```release-note
Fixed an issue where notifications would not show on macOS in certain cases.
```

### DIFF
--- a/src/main/notifications/index.ts
+++ b/src/main/notifications/index.ts
@@ -27,6 +27,11 @@ class NotificationManager {
     private upgradeNotification?: NewVersionNotification;
     private restartToUpgradeNotification?: UpgradeNotification;
 
+    constructor() {
+        // Run this very early to perform an early permission check
+        getDoNotDisturb();
+    }
+
     public async displayMention(title: string, body: string, channelId: string, teamId: string, url: string, silent: boolean, webcontents: Electron.WebContents, soundName: string) {
         log.debug('displayMention', {title, channelId, teamId, url, silent, soundName});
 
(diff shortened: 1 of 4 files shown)


devinbinnie:
@amyblais I'm looking to do a dot release for this one if possible.

amyblais:
> @amyblais I'm looking to do a dot release for this one if possible.

Ok 👍

src/main/notifications/index.ts
```diff
@@ -27,6 +27,11 @@ class NotificationManager {
     private upgradeNotification?: NewVersionNotification;
     private restartToUpgradeNotification?: UpgradeNotification;
 
+    constructor() {
+        // Run this very early to perform an early permission check
+        getDoNotDisturb();
```
marianunez, line 32: Will this trigger a permission modal before getting a notification?

devinbinnie: Yes, it shows the macOS one that people were missing.

hmhealey, line 32: Should this happen as an explicit call from somewhere in the initialization code? It feels a bit weird to have this be a side effect of importing this file

devinbinnie: Yeah I guess that would be better, I wanted to keep everything centralized but the import case does make that a bit awkward.

marianunez (approved):

hmhealey (approved):

PR #3060, opened 2024-06-10
[MM-58693] Fix some issues with Windows notifications

devinbinnie:
#### Summary
Found a few issues with reporting notifications on Windows and a small bug that I wanted to fix:
- Sometimes Windows will fire both the `show` and `failed` events when a notification fails to show due to permissions not being enabled for Mattermost in Windows. This was then always being reported as a `success` when it should be `not_sent`, so I had to add a timeout to have the `show` callback wait for the `failed` one to potentially run (not an elegant solution)
- Windows would return `failed` when the notification fails the permission check. This should now be reported as `not_sent` as it is a user choice not to send the notification
- Our DND check for priority was returning the opposite of what it should, so I've swapped the flag.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58693

```release-note
Fixed an issue with DND mode on Windows
```

### DIFF
--- a/src/main/notifications/dnd-windows.ts
+++ b/src/main/notifications/dnd-windows.ts
@@ -20,13 +20,9 @@ function getWindowsDoNotDisturb() {
     case 2:
         return true;
     case 1:
-        return isPriority('Mattermost.Desktop');
-    case 0:
-    case -1:
-    case -2:
-        return false;
+        return Boolean(isPriority('Mattermost.Desktop').value);
     default:
-        return focusAssistValue;
+        return false;
     }
 }
 
--- a/src/main/notifications/dnd-windows.ts
+++ b/src/main/notifications/dnd-windows.ts
@@ -20,7 +20,7 @@ function getWindowsDoNotDisturb() {
     case 2:
         return true;
     case 1:
-        return Boolean(isPriority('Mattermost.Desktop').value);
+        return !(isPriority('Mattermost.Desktop').value);
     default:
         return false;
     }
--- a/src/main/notifications/dnd-windows.ts
+++ b/src/main/notifications/dnd-windows.ts
@@ -20,7 +20,7 @@ function getWindowsDoNotDisturb() {
     case 2:
         return true;
     case 1:
-        return !isPriority('Mattermost.Desktop');
+        return isPriority('Mattermost.Desktop');
     case 0:
     case -1:
     case -2:
--- a/src/main/notifications/index.ts
+++ b/src/main/notifications/index.ts
@@ -92,30 +92,37 @@ class NotificationManager {
                 log.debug('notification timeout', serverName, mention.uId);
                 resolve({status: 'error', reason: 'notification_timeout'});
             }, 10000);
+            let failed = false;
 
             mention.on('show', () => {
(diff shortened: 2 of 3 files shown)


marianunez (commented):
Is this a recent regression or something that we tracked down with recent metrics and added logging? Can we create a ticket for tracking purposes on notification fixes?

src/main/notifications/index.ts
```diff
@@ -92,34 +92,48 @@ class NotificationManager {
...
-                    this.mentionsPerChannel.set(mentionKey, mention);
-                }
-                const notificationSound = mention.getNotificationSound();
-                if (notificationSound) {
-                    MainWindow.sendToRenderer(PLAY_SOUND, notificationSound);
-                }
-                flashFrame(true);
-                clearTimeout(timeout);
-                resolve({status: 'success'});
+                }, 0);
```
marianunez, line 121: Is there any effect on adding this timeout if the on `show` is already running asynchronously ? I was expecting some min time set here to wait

devinbinnie:
> Is this a recent regression or something that we tracked down with recent metrics and added logging? Can we create a ticket for tracking purposes on notification fixes?

I just found this while testing the notifications manually, but I can add a ticket for it.

src/main/notifications/index.ts
devinbinnie: This is basically there to tell the JS engine to requeue this logic at the end of the event queue so that by then if the `failed` event had fired, the flag would be set and we wouldn't run both callbacks (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Event_loop#zero_delays)

I'm really not happy with this solution, but it's all I could think of to try and make sure we resolved the promise correctly, otherwise it's a race between the two callbacks.

marianunez (approved):
Could we add a test for the dnd for windows? Seems like we only have it for mac? https://github.com/mattermost/desktop/blob/release-5.5/src/main/notifications/index.test.js#L163

src/main/notifications/dnd-windows.ts
```diff
@@ -20,13 +20,9 @@ function getWindowsDoNotDisturb() {
     case 2:
         return true;
     case 1:
-        return !isPriority('Mattermost.Desktop');
-    case 0:
-    case -1:
-    case -2:
-        return false;
+        return !(isPriority('Mattermost.Desktop').value);
```
marianunez, line 23: So this turned out to be the right condition to check but we were not reading the right value?

devinbinnie: Correct

marianunez (approved):
Do we want to include this in the dot release as well?

devinbinnie:
> Do we want to include this in the dot release as well?

cc @amyblais

src/main/notifications/dnd-windows.ts
hmhealey: Truthy/falsy values strike again!

src/main/notifications/index.ts
hmhealey: I'm trying to think if there's any way to do this with something like a `Promise.race`. This just seems like a flaw in Electron though

hmhealey (approved):

Issue #3067, opened 2024-06-14
[Bug]: Mattermost window keeps resizing each time after relaunch


Izaya21:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.8.1

### Operating System

Windows 11

### Mattermost Server Version

9.8.1

### Steps to reproduce

Restart Mattermost

### Expected behavior

Maintains the resized window

### Observed behavior

Windows goes smaller each time the app is restarted

### Log Output

```shell
[2024-06-14 05:05:39.980] [error] Logger Log level set to: info
[2024-06-14 05:05:40.076] [warn]  [App.Initialize] Current working directory is C:\Windows\System32, changing into C:\Users\User\AppData\Local\Programs\mattermost-desktop
[2024-06-14 05:05:40.110] [info]  [App.Config] config.autostart has been configured: true
[2024-06-14 05:05:40.133] [info]  [App.Config] config.autostart has been configured: true
[2024-06-14 05:05:40.246] [info]  [ServerDropdownView] init
[2024-06-14 05:05:40.271] [info]  [i18nManager] Set new language pl
[2024-06-14 05:05:40.284] [info]  [App.Config] config.autostart has been configured: true
[2024-06-14 05:05:40.312] [info]  [App.Config] config.autostart has been configured: true
[2024-06-14 05:05:40.762] [info]  [App.Config] config.autostart has been configured: true
[2024-06-14 05:05:45.226] [info]  [UpdateManager] Checking for update
[...]

PR #3072, opened 2024-06-17
[MM-58780] Don't account for scaleFactor on Windows when resizing if the matching screen is the primary monitor

devinbinnie:
#### Summary
Further to the fix I did for the Electron bug here: https://github.com/electron/electron/issues/10862
The primary monitor on Windows doesn't need to care about the scale factor, and will respect it when resetting its window bounds. This PR checks to see if the primary monitor is the matching screen and won't use the `scaleFactor` if the windows bounds aren't set.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58780
Closes https://github.com/mattermost/desktop/issues/3067

```release-note
Fixed an issue where the window size will get smaller on Windows after a restart if the primary monitor is used and is scaled
```

### DIFF
--- a/src/main/windows/mainWindow.test.js
+++ b/src/main/windows/mainWindow.test.js
@@ -39,6 +39,7 @@ jest.mock('electron', () => ({
     },
     screen: {
         getDisplayMatching: jest.fn(),
+        getPrimaryDisplay: jest.fn(),
     },
     globalShortcut: {
         register: jest.fn(),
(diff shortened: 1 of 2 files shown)


marianunez (approved):

marianunez:
Harshil is out so you may want to get another reviewer

hmhealey (approved):

PR #3073, opened 2024-06-19
[MM-57964] Implement `openThread` and `openStopRecordingModal`

streamer45:
#### Summary

We are extending the API to support the functionality in https://github.com/mattermost/mattermost-plugin-calls/pull/779, namely:

- Ability to open the call thread from the global widget.
- Ability to show the stop recording confirmation modal when stopping a recording from the global widget.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57964

#### Release Note

```release-note
NONE
```


api-types/index.ts
```diff
@@ -70,6 +70,12 @@ export type DesktopAPI = {
 
     focusPopout: () => void;
 
+    openThread: (threadID: string) => void;
```
devinbinnie, line 73: Can we rename these so that we know they're specific to calls?
Something like `openThreadForCalls`.
I know it's a bit "smurf naming" but it's important in this context to make sure the API is only used for that.

src/main/windows/callsWidgetWindow.ts
```diff
@@ -504,6 +508,14 @@ export class CallsWidgetWindow {
         };
     };
 
+    private handleCallsOpenThread = (event: IpcMainEvent, threadID: string) => {
+        this.forwardToMainApp(CALLS_WIDGET_OPEN_THREAD)(event, threadID);
```
devinbinnie, line 512: Can you link me where in (I assume the calls plugin) code this is being received and what it's doing? We'll need to make sure that it's not something exploitable.

A quick look seems like nothing in the Desktop App should be an issue with this (since it's just forwarding through)

cc @enzowritescode

api-types/index.ts
streamer45: Sure. This functionality was generic enough that I thought it could be used by something else in the future, but we can rename it for now since Calls is currently the only user.

cpoile (approved):
Really slick, nicely done.

src/main/preload/externalAPI.ts
```diff
@@ -109,6 +111,12 @@ const desktopAPI: DesktopAPI = {
 
     focusPopout: () => ipcRenderer.send(CALLS_POPOUT_FOCUS),
 
+    openThread: (threadID) => ipcRenderer.send(CALLS_WIDGET_OPEN_THREAD, threadID),
+    onOpenThread: (listener) => createListener(CALLS_WIDGET_OPEN_THREAD, listener),
+
+    openStopRecordingModal: (channelID) => ipcRenderer.send(CALLS_WIDGET_OPEN_STOP_RECORDING_MODAL, channelID),
+    onOpenStopRecordingModal: (listener) => createListener(CALLS_WIDGET_OPEN_STOP_RECORDING_MODAL, listener),
```
streamer45, line 118: @devinbinnie @cpoile Since we are here, I was considering making this a bit more generic (e.g. `openCallsModal`) so we could pass a modal ID (string or enum) and at least the channel/call ID which is hopefully sufficient in most cases. That would avoid the need to make desktop changes every time we want to add a modal interaction from the widget.

Thoughts?

devinbinnie: I'd be okay with that, but I think we'd need to lock down the types of modals that could be opened somewhere, and this should never be used to open a modal that's native to the desktop app. If we're just using it as a relay to the calls plugin from the widget (or vice-versa) then I'm okay with it.

streamer45: Okay. Of course we'd still be handling this event only for Calls windows.

enzowritescode (approved):

src/main/preload/externalAPI.ts
streamer45: I'll be deferring this improvement to the next time we'll need this sort of functionality.

api-types/index.ts
streamer45: @devinbinnie Done

devinbinnie (approved):
LGTM!

streamer45:
@devinbinnie Would you be able to publish a new API types NPM version? Thanks!

devinbinnie:
> @devinbinnie Would you be able to publish a new API types NPM version? Thanks!

`+ @mattermost/desktop-api@5.9.0-1`
Done!

devinbinnie:
> > @devinbinnie Would you be able to publish a new API types NPM version? Thanks!
> 
> `+ @mattermost/desktop-api@5.9.0-1` Done!

@streamer45 https://github.com/mattermost/desktop/pull/3075
PR to bump the package numbers

PR #3093, opened 2024-07-15
[MM-59492] Fixed `isInsideRectangle`, and removed null check for matchingScreen since it will always return something

devinbinnie:
#### Summary
There was a case as described in the ticket below where a user could get the window to display off the screen. This was due to a flaw in `isInsideRectangle` where we weren't checking the bounds correctly (we were only checking that it was inside from the top-left corner, which is fine in single-monitor setups, but not multi-monitor). Additionally, I had expected `getMatchingScreen` to return `false` if it doesn't find one, but it always just gets the primary screen in the worse case.

This PR fixes and adds tests for `isInsideRectangle`, as well removes the null check for `getMatchingScreen`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59492
Closes https://github.com/mattermost/desktop/issues/3087

```release-note
Fixed an issue where the window might be rendered off-screen in multi-monitor setups
```

### DIFF
--- a/src/main/windows/mainWindow.ts
+++ b/src/main/windows/mainWindow.ts
@@ -254,8 +254,8 @@ export class MainWindow extends EventEmitter {
                 throw new Error('Provided bounds info file does not validate, using defaults instead.');
             }
             const matchingScreen = screen.getDisplayMatching(savedWindowState);
-            log.debug('matching screen for main window', matchingScreen);
-            if (!(matchingScreen && (isInsideRectangle(matchingScreen.bounds, savedWindowState) || savedWindowState.maximized))) {
+            log.debug('closest matching screen for main window', matchingScreen);
+            if (!(isInsideRectangle(matchingScreen.bounds, savedWindowState) || savedWindowState.maximized)) {
                 throw new Error('Provided bounds info are outside the bounds of your screen, using defaults instead.');
             }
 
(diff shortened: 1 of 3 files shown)


src/main/windows/mainWindow.ts
```diff
@@ -254,8 +254,8 @@ export class MainWindow extends EventEmitter {
                 throw new Error('Provided bounds info file does not validate, using defaults instead.');
             }
             const matchingScreen = screen.getDisplayMatching(savedWindowState);
-            log.debug('matching screen for main window', matchingScreen);
-            if (!(matchingScreen && (isInsideRectangle(matchingScreen.bounds, savedWindowState) || savedWindowState.maximized))) {
```
larkox, line 258: I wonder what happens if you disconnect all screens. I don't think anyone will ever face this, but someone with a desktop pc could be switching screens and suddenly mattermost crashes. 0/5 if we want to even test that.

larkox (approved):

src/main/windows/mainWindow.ts
devinbinnie: I think it's safe to say we don't support running Mattermost Desktop without any monitors connected :p

M-ZubairAhmed (approved):

PR #3095, opened 2024-07-17
[MM-59543] Disallow use of file: protocol in the app, remove all references to it, add mattermost-desktop: protocol to read local files

devinbinnie:
#### Summary
This PR is a companion to https://github.com/mattermost/desktop/pull/3094 where most of the security fuses were flipped. However this change flips only one, `GrantFileProtocolExtraPrivileges` to `false` and includes a number of changes to support that fuse being flipped:
- I have removed all use of the `file:` protocol in the app, replacing it with a custom protocol called `mattermost-desktop` as per the [recommendation by Electron](https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols). This protocol has a handler that allows only files from the `renderer` folder to be served up and prevents any path traversal using said protocol.
- As a result of the above I have also removed the method `getLocalURLString` since it's no longer required anywhere
- One of the files in the loading screen was being loaded directly as an image, so I've amended that and moved that file (`StippleMask.jpg`)
- The downloads dropdown was directly loading files from the disk for thumbnails, so I've replaced that with the `nativeImage` library that creates a base64 based thumbnail and sends that along with the other download data.

This change is being separated due to how much has changed to support this, so it should likely be given more time to bake than the other fuses flips.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59543

```release-note
Disallow use of file: protocol in the app
```


src/main/app/intercom.test.js
```diff
@@ -65,13 +63,12 @@ describe('main/app/intercom', () => {
             ModalManager.addModal.mockReturnValue(promise);
 
             handleWelcomeScreenModal();
-            expect(ModalManager.addModal).toHaveBeenCalledWith('welcomeScreen', '/some/index.html', '/some/preload.js', null, {}, true);
+            expect(ModalManager.addModal).toHaveBeenCalledWith('welcomeScreen', 'mattermost-desktop://renderer/welcomeScreen.html', '/some/preload.js', null, {}, true);
```
harshilsharma63, line 66: Worth making the string `mattermost-desktop://renderer` an exported constant as its being used in a lot of places and making it a constant will avoid possible typos.

harshilsharma63 (approved):

hmhealey (approved):
I have a couple questions, but they don't need to block this.

1. Does this have any effect on what types of files can be linked to from the web app or a web app plugin? I'm thinking of if someone had a custom plugin that opened a file on a network drive or something, but I don't think we supported that anyway.
2. Similarly, does this mean that the web app could open `mattermost-desktop` links? I'm not sure if these changes affect the web app or just the outer Electron processes

src/main/app/initialize.ts
```diff
@@ -291,6 +297,24 @@ function initializeInterCommunicationEventListeners() {
...
 async function initializeAfterAppReady() {
+    protocol.handle('mattermost-desktop', (request: Request) => {
+        const url = parseURL(request.url);
+        if (!url) {
+            return new Response('bad', {status: 400});
+        }
+
+        // Including this snippet from the handler docs to check for path traversal
+        // https://www.electronjs.org/docs/latest/api/protocol#protocolhandlescheme-handler
+        const pathToServe = path.join(app.getAppPath(), 'renderer', url.pathname);
```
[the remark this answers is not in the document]
devinbinnie: It should be pretty specific to this use case since I'm not really allowing for any other arbitrary paths to be loaded, outside of preload scripts (but those are all included in the bundle and cannot be overridden via normal app usage)

src/main/app/intercom.test.js
devinbinnie: I'm actually thinking I might just do a helper method like I had before with `getLocalURLString`, just so I can also control how things are concatenated. Good call nonetheless :)

devinbinnie:
> 1. Does this have any effect on what types of files can be linked to from the web app or a web app plugin? I'm thinking of if someone had a custom plugin that opened a file on a network drive or something, but I don't think we supported that anyway.

None of these should affect any of the remote code (ie. webapp code), this is only concerned with loading files internal to the Desktop App. Nothing loaded into the Mattermost web app itself should be affected.

> 2. Similarly, does this mean that the web app could open `mattermost-desktop` links? I'm not sure if these changes affect the web app or just the outer Electron processes

This would technically allow any renderer process to use this protocol, however the navigation is blocked via other means, so it will try to open `mattermost-desktop:` links externally, which of course won't work. This protocol is only registered internally to the Desktop App

enzowritescode (approved):

PR #3097, opened 2024-07-18
[MM-59044] Add per-server permission check for screen sharing

devinbinnie:
#### Summary
We currently allow for unscrutinized use of the `getSources` API allowed a malicious server to potentially abuse this API without the user's knowledge. This PR adds a per-server permission check for screen sharing so that the user will always be asked before the screen sharing API is used.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59044

```release-note
Fixed an issue where the screen sharing API could be used without the user's permission
```


src/main/windows/callsWidgetWindow.ts
```diff
@@ -403,6 +404,11 @@ export class CallsWidgetWindow {
             }
         }
 
+        if (!await PermissionsManager.doPermissionRequest(view.webContentsId, 'screenShare', {requestingUrl: view.view.server.url.toString(), isMainFrame: false})) {
```
marianunez, line 407: Is this going to show when you start a screenshare or calls?

devinbinnie: Just a screen share.

marianunez: And the permission is granted once or does it ask each time?

src/main/permissionsManager.ts
```diff
@@ -239,10 +241,12 @@ t('main.permissionsManager.checkPermission.dialog.message.media');
 t('main.permissionsManager.checkPermission.dialog.message.geolocation');
 t('main.permissionsManager.checkPermission.dialog.message.notifications');
 t('main.permissionsManager.checkPermission.dialog.message.openExternal');
+t('main.permissionsManager.checkPermission.dialog.message.screenShare');
```
marianunez, line 244: Will this now show it in the permissions settings window as well?

src/main/windows/callsWidgetWindow.ts
devinbinnie: It's granted once, same as all of the other permissions

src/main/permissionsManager.ts
devinbinnie: Yep it will

marianunez (approved):
Thanks @devinbinnie! LGTM 

cc// @cpoile for visibility on calls

hmhealey (approved):

enzowritescode (approved):

Issue #3102, opened 2024-07-21
[Bug]: No window controls


vitoyucepi:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.9.0-rc.1

### Operating System

Debian 12 x64

### Mattermost Server Version

N/A

### Steps to reproduce

1. Install fresh Debian 12.
   Other distributions should work out of the box, except ubuntu 24.04.
1. Use a wayland or xorg session.
1. Download the [release artifact](https://releases.mattermost.com/desktop/5.9.0-rc.1/mattermost-desktop-5.9.0-rc.1-linux-x64.tar.gz).
1. Extract the archive.
1. Run the `mattermost-desktop` file.

### Expected behavior

The window control buttons should be visible.

### Observed behavior

No window control buttons like minimize, maximize, close.
![Screenshot of the GNOME desktop with the Mattermost application](https://github.com/user-attachments/assets/ecdeb416-469a-48a6-81c9-8173e6697f91)

### Log Output

```shell
[2024-07-21 09:12:56.980] [error] Logger Log level set to: info
[2024-07-21 09:12:57.192] [info]  [App.Config] config.autostart has been configured: true
[2024-07-21 09:12:57.247] [info]  [App.Initialize] Autoupgrade disabled: false
[2024-07-21 09:12:57.280] [info]  [ServerDropdownView] init
[2024-07-21 09:12:57.301] [warn]  [i18nManager] Failed to set new language en-US
[2024-07-21 09:12:57.302] [warn]  [i18nManager] Failed to set new language US
[...]

PR #3104, opened 2024-07-22
[MM-59840] Add Linux title bar buttons using `titleBarOverlay` field of `BrowserWindow`

devinbinnie:
#### Summary
With the upgrade to Electron v30, they seem to have quietly taken away the default window frame for Linux clients, or something else has broken in the configuration. However, they have given us a way to add native window buttons to Linux apps now, so this PR adds that overlay in the same way used by macOS.

This will require some follow-up for the next release to re-work how we do these window buttons for macOS and Windows, but this will fix the immediate issue for Linux users.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/3102
https://mattermost.atlassian.net/browse/MM-59840

```release-note
Changed the window buttons on the Linux client to use the native ones provided by Electron, removed the frame
```

### DIFF
--- a/src/main/windows/mainWindow.ts
+++ b/src/main/windows/mainWindow.ts
@@ -5,7 +5,7 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import type {BrowserWindowConstructorOptions, Event, Input, IpcMainEvent} from 'electron';
+import type {BrowserWindowConstructorOptions, Event, Input} from 'electron';
 import {app, BrowserWindow, dialog, globalShortcut, ipcMain, screen} from 'electron';
 import {EventEmitter} from 'events';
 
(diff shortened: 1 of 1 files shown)


larkox (approved):

src/main/windows/mainWindow.ts
```diff
@@ -246,6 +249,13 @@ export class MainWindow extends EventEmitter {
         return os.platform() === 'darwin' || (os.platform() === 'win32' && Utils.isVersionGreaterThanOrEqualTo(os.release(), '6.2'));
     };
 
+    private getTitleBarOverlay = () => {
+        return {
+            color: Config.darkMode ? '#2e2e2e' : '#efefef',
+            height: TAB_BAR_HEIGHT,
```
M-ZubairAhmed, line 255: Does just by increasing height we have the button? i dont see where is the part of the buttons?

devinbinnie: The `titleBarOverlay` field takes in an object of type `Object | boolean`. So just by having the object there it evaluates to `true` and under the hood Electron uses these fields to construct the buttons.

M-ZubairAhmed: Ah i see

M-ZubairAhmed (approved):

PR #3105, opened 2024-07-22
[MM-59489] Change logged in tracking code for new API, check for `/` when navigating tabs

devinbinnie:
#### Summary
One of the changes to the Desktop API was that individual tabs report their login change directly to the Desktop App instead of relying on local storage to do so. At some point it appears that `localStorage` was not syncing to inactive tabs, which makes sense given that that doesn't happen on Chrome either unless the tab is focused. This caused no issue on web app, but the navigation code on Desktop App was effected in that if you logged out of another tab, logged back in on the main tab and tried to navigate to the other tab using a link of some kind, the other tab would try to load and not be logged in correctly before it tried to navigate, causing a session expiry.

This PR fixes the log in tracking code for the new API, forcing each individual tab to reload if necessary (like the old code, but this time just the individual tab is responsible for forcing the reload), and also checks to make sure that on logout, we go back to the Channels tab, since it seems like that wasn't always happening either.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59489

```release-note
Fixed an issue where logging out from the Boards/Playbooks tabs and trying to navigate after logging back in would force an unexpected logout
```

### DIFF
--- a/src/main/preload/externalAPI.ts
+++ b/src/main/preload/externalAPI.ts
@@ -41,7 +41,7 @@ import {
     GET_DESKTOP_SOURCES,
     UNREADS_AND_MENTIONS,
     LEGACY_OFF,
-    TAB_LOGGED_IN,
+    TAB_LOGIN_CHANGED,
 } from 'common/communication';
 
 import type {ExternalAPI} from 'types/externalAPI';
@@ -73,8 +73,8 @@ const desktopAPI: DesktopAPI = {
     setSessionExpired: (isExpired) => ipcRenderer.send(SESSION_EXPIRED, isExpired),
     onUserActivityUpdate: (listener) => createListener(USER_ACTIVITY_UPDATE, listener),
 
-    onLogin: () => ipcRenderer.send(TAB_LOGGED_IN, true),
-    onLogout: () => ipcRenderer.send(TAB_LOGGED_IN, false),
+    onLogin: () => ipcRenderer.send(TAB_LOGIN_CHANGED, true),
+    onLogout: () => ipcRenderer.send(TAB_LOGIN_CHANGED, false),
 
     // Unreads/mentions/notifications
     sendNotification: (title, body, channelId, teamId, url, silent, soundName) =>
--- a/src/main/preload/externalAPI.ts
+++ b/src/main/preload/externalAPI.ts
@@ -41,6 +41,7 @@ import {
     GET_DESKTOP_SOURCES,
     UNREADS_AND_MENTIONS,
     LEGACY_OFF,
+    TAB_LOGGED_IN,
 } from 'common/communication';
 
 import type {ExternalAPI} from 'types/externalAPI';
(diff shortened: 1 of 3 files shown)


harshilsharma63 (approved):

hmhealey:
> At some point it appears that localStorage was not syncing to inactive tabs, which makes sense given that that doesn't happen on Chrome either unless the tab is focused. 

Does this also mean the same bug might apply on the web app as well? It sort of makes sense that it wouldn't want to wake up every tab on a given domain whenever `localStorage` changes, but I wasn't aware that Chrome had this behaviour.

I don't know if that would be the case though because the web app sort of implicitly tracks if its logged in based on the `MMUSERID` and `MMAUTHTOKEN` cookies. The `localStorage` method is only used because the web app can't access the cookie directly, and we wanted to give it a kick to have it realize it had been logged in (and we didn't know `postMessage` existed at the time apparently)

src/main/preload/externalAPI.ts
```diff
@@ -72,8 +73,8 @@ const desktopAPI: DesktopAPI = {
     setSessionExpired: (isExpired) => ipcRenderer.send(SESSION_EXPIRED, isExpired),
     onUserActivityUpdate: (listener) => createListener(USER_ACTIVITY_UPDATE, listener),
 
-    onLogin: () => ipcRenderer.send(APP_LOGGED_IN),
-    onLogout: () => ipcRenderer.send(APP_LOGGED_OUT),
+    onLogin: () => ipcRenderer.send(TAB_LOGGED_IN, true),
+    onLogout: () => ipcRenderer.send(TAB_LOGGED_IN, false),
```
hmhealey, line 77: Nitpicking, but this naming is a bit confusing because TAB_LOGGED_IN is sent when you log out. I'd suggest something like TAB_LOGIN_CHANGED if you think it matters enough to change

hmhealey (approved):

src/main/preload/externalAPI.ts
devinbinnie: I just wanted it to be distinct from the other ones and I think just one name works. But you name is better I'll change it real quick.

devinbinnie:
> > At some point it appears that localStorage was not syncing to inactive tabs, which makes sense given that that doesn't happen on Chrome either unless the tab is focused.
> 
> Does this also mean the same bug might apply on the web app as well? It sort of makes sense that it wouldn't want to wake up every tab on a given domain whenever `localStorage` changes, but I wasn't aware that Chrome had this behaviour.
> 
> I don't know if that would be the case though because the web app sort of implicitly tracks if its logged in based on the `MMUSERID` and `MMAUTHTOKEN` cookies. The `localStorage` method is only used because the web app can't access the cookie directly, and we wanted to give it a kick to have it realize it had been logged in (and we didn't know `postMessage` existed at the time apparently)

This won't affect the web app as a bug because the tabs still receive the signal, they just don't seem to act on it until the tab gets focused. That must have changed at some point (not sure if it was us or Chrome that did that) but it should affect the web app since there's no forwarding logic between the two.

PR #3108, opened 2024-07-23
[MM-59480] Fix crash on Windows caused by titlebar overlay

devinbinnie:
#### Summary
Adding the `titleBarOverlay` for just Linux crashes Windows when the app tries to load (apparently calling the `set` function will crash if it's not enabled :(). This PR makes sure we're on Linux and that it's enabled before we try to set it.

```release-note
NONE
```

### DIFF
--- a/src/main/windows/mainWindow.ts
+++ b/src/main/windows/mainWindow.ts
@@ -560,7 +560,9 @@ export class MainWindow extends EventEmitter {
     };
 
     private handleUpdateTitleBarOverlay = () => {
-        this.win?.setTitleBarOverlay?.(this.getTitleBarOverlay());
+        if (process.platform === 'linux') {
+            this.win?.setTitleBarOverlay?.(this.getTitleBarOverlay());
+        }
     };
 }
 


marianunez (approved):
Thanks @devinbinnie! What options do we have for test coverage to catch these kinda of issues in the future? Would the E2E catch this crash in Windows?

devinbinnie:
> Thanks @devinbinnie! What options do we have for test coverage to catch these kinda of issues in the future? Would the E2E catch this crash in Windows?

E2E would have caught this for sure - it just hadn't been run yet.

marianunez:
> > Thanks @devinbinnie! What options do we have for test coverage to catch these kinda of issues in the future? Would the E2E catch this crash in Windows?
> 
> E2E would have caught this for sure - it just hadn't been run yet.

Should we be running the E2E tests in the PRs?

devinbinnie:
> > > Thanks @devinbinnie! What options do we have for test coverage to catch these kinda of issues in the future? Would the E2E catch this crash in Windows?
> > 
> > 
> > E2E would have caught this for sure - it just hadn't been run yet.
> 
> Should we be running the E2E tests in the PRs?

Usually we run them before releases, for each PR would take a long time and a lot of processing power we don't generally need to spend. We also already run RF tests on nightly so that should catch this as well.

marianunez:
> 
> Usually we run them before releases, for each PR would take a long time and a lot of processing power we don't generally need to spend. We also already run RF tests on nightly so that should catch this as well.

@yasserfaraazkhan can we check if the nightly E2E tests are running and if it caught this crash?

hmhealey (approved):

PR #3109, opened 2024-07-24
[MM-51514] Allow Calls widget to open user settings

streamer45:
#### Summary

As we are implementing persistent Calls preferences and settings, we'd like a way to open the modal from within a call. PR implements a new `openCallsUserSettings` method to allow triggering this functionality from the global widget window.

#### Example

https://github.com/user-attachments/assets/c3521aac-69de-4daf-8b06-d8d7d7bbdff1

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51514

#### Release Note

```release-note
NONE
```

### DIFF
--- a/api-types/lib/index.d.ts
+++ b/api-types/lib/index.d.ts
@@ -63,5 +63,7 @@ export type DesktopAPI = {
     onOpenThreadForCalls: (listener: (threadID: string) => void) => () => void;
     openStopRecordingModal: (channelID: string) => void;
     onOpenStopRecordingModal: (listener: (channelID: string) => void) => () => void;
+    openCallsUserSettings: () => void;
+    onOpenCallsUserSettings: (listener: () => void) => () => void;
     unregister: (channel: string) => void;
 };
--- a/api-types/package.json
+++ b/api-types/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mattermost/desktop-api",
-  "version": "5.9.0-1",
+  "version": "5.9.0-2",
   "description": "Shared types for the Desktop App API provided to the Web App",
   "keywords": [
     "mattermost"
--- a/api-types/package.json
+++ b/api-types/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mattermost/desktop-api",
-  "version": "5.9.0-2",
+  "version": "5.10.0-1",
   "description": "Shared types for the Desktop App API provided to the Web App",
   "keywords": [
     "mattermost"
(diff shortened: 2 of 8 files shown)


devinbinnie:
Adding @enzowritescode since this changes the Desktop API

api-types/lib/index.d.ts
```diff
@@ -63,5 +63,7 @@ export type DesktopAPI = {
     onOpenThreadForCalls: (listener: (threadID: string) => void) => () => void;
     openStopRecordingModal: (channelID: string) => void;
     onOpenStopRecordingModal: (listener: (channelID: string) => void) => () => void;
+    openCallsUserSettings: () => void;
+    onOpenCallsUserSettings: (listener: () => void) => () => void;
```
devinbinnie, line 67: @streamer45 Can you run `npm run build` and increment the `api-types` version as part of this PR?

streamer45: @devinbinnie Done in https://github.com/mattermost/desktop/pull/3109/commits/0353d98a0381319a15cf3b6d60439d06d116a87f

api-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mattermost/desktop-api",
-  "version": "5.9.0-1",
+  "version": "5.9.0-2",
```
devinbinnie, line 3: @streamer45 This should actually be v5.10 now
```suggestion
  "version": "5.10.0-1",
```

streamer45: Too bad

api-types/lib/index.d.ts
devinbinnie: I also just realized, these are missing from `api-types/index.ts`, which is what we should be editing. Can we add them there?

streamer45: Yeah, got confused with all these files.

src/types/externalAPI.ts
```diff
@@ -24,4 +24,5 @@ export interface ExternalAPI {
     createListener(event: 'desktop-sources-modal-request', listener: () => void): () => void;
     createListener(event: 'calls-widget-open-thread', listener: (threadID: string) => void): () => void;
     createListener(event: 'calls-widget-open-stop-recording-modal', listener: (channelID: string) => void): () => void;
+    createListener(event: 'calls-widget-open-user-settings', listener: () => void): () => void;
```
enzowritescode, line 27: For my understanding, how come this file does not use the constants in `src/common/communication.ts`?

streamer45: I'll defer that to @devinbinnie

devinbinnie: Because it's a types file and TypeScript does not recognize those constants as the type of `const` where the value is that string. In order to make the typing actually work, you need to put it the actual string as the type, not the value.

enzowritescode (approved):

cpoile (approved):

streamer45:
@devinbinnie Do you think we could push the new package version?

devinbinnie:
> @devinbinnie Do you think we could push the new package version?

Done

PR #3120, opened 2024-08-06
[MM-59552] Remove `unsafe-inline` from internal CSP, replace with nonce and rework some dynamic styling

devinbinnie:
#### Summary
`unsafe-inline` is definitely not recommended for production apps, so I've removed it and made some changes so that we can safely load dynamic styles:
- Added a randomly generated nonce that is passed to the renderer process (only once) and given to the frameworks that need it (`react-beautiful-dnd` and `react-select`) so they can still load their custom styles
- Created a SCSS file that contains the styles for dark mode for modals, and instead of toggling the dynamic import, we just toggle the `body` level class.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59552

```release-note
NONE
```


hmhealey (approved):

enzowritescode (approved):

streamer45 (approved):
:+1:

PR #3136, opened 2024-08-29
[MM-60226] Allow plugins to open blank popup windows

streamer45:
#### Summary

Slightly more than the single line envisioned, but here we go :sweat_smile: 

PR makes some changes to allow plugins (or any script that runs in Mattermost's main view) to open blank (i.e., `about:blank` URL) popup windows.

What we are doing is mostly allowing the browser side to open a window natively, as it would in a browser context. This means:

- [`window.open`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) returning a reference to the child window
- the child window (popup) has a reference to the parent window through [`window.opener`](https://developer.mozilla.org/en-US/docs/Web/API/Window/opener)

> [!NOTE]
> I had to take a slightly different route than what I first thought (creating the popups natively using [`new BrowserWindow()`](https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions)) because (of course) there was a reason we did it differently in Calls as well. The main limitation was that in such case `window.open` returns `null` which makes it more or less useless as the parent side cannot control it whatsoever.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60226

#### Release Note

```release-note
Allowed plugins to open `about:blank` popup windows using `window.open()`
```


src/main/views/pluginsPopUps.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export class PluginsPopUpsManager {
+    popups: Record<number, PluginPopUp>;
+
+    constructor() {
+        this.popups = {};
+    }
+
+    public handleCreateWindow(win: BrowserWindow, details: Electron.DidCreateWindowDetails) {
```
devinbinnie, line 28: Nit: Arrow functions are preferred for these classes.

src/main/views/webContentEvents.ts
```diff
@@ -169,6 +170,11 @@ export class WebContentsEventManager {
                 return {action: 'allow'};
             }
 
+            // Allow plugins to open blank popup windows.
+            if (parsedURL.toString() === 'about:blank') {
```
devinbinnie, line 174: We are missing the check for plugin URLs:
```suggestion
            if (isPluginUrl(serverURL, parsedURL) && parsedURL.toString() === 'about:blank') {
```

This should also probably be done further down.

src/main/views/webContentEvents.ts
```diff
@@ -352,6 +358,12 @@ export class WebContentsEventManager {
         const newWindow = this.generateNewWindowListener(contents.id, spellcheck);
         contents.setWindowOpenHandler(newWindow);
 
+        // Defer handling of new popup windows to PluginsPopUpsManager. These still need to be
+        // previously allowed from generateNewWindowListener through PluginsPopUpsManager.handleNewWindow.
+        contents.on('did-create-window', (win: BrowserWindow, details: Electron.DidCreateWindowDetails) => {
+            PluginsPopUpsManager.handleCreateWindow(win, details);
```
devinbinnie, line 364: Won't this create two windows using the existing allowance for plugin popups? See https://github.com/mattermost/desktop/pull/3136/files#diff-daeb8475a075c7e3efcef7ebc2b325e61c5275b03fdc4911c4d48e6b064bcfa5L233

devinbinnie (changes requested):
Design looks good, just a few comments

src/main/views/webContentEvents.ts
streamer45: @devinbinnie I think the point is to allow creating popouts from the webapp, not necessarily a plugin path. If we add this check then it won't really work the way I am thinking.

src/main/views/pluginsPopUps.ts
```diff
@@ -0,0 +1,95 @@
...
+            log.warn(`prevented popup window from redirecting to: ${ev.url}`);
+            ev.preventDefault();
+        });
+        win.webContents.on('will-navigate', (ev: Event<WebContentsWillNavigateEventParams>) => {
+            if (ev.url === details.url) {
+                return;
+            }
+
+            log.warn(`prevented popup window from navigating to: ${ev.url}`);
+            ev.preventDefault();
```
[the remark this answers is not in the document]
streamer45: Yeah, given the required use case I've started super strict but happy to look into adding back some basic functionality :+1:

src/main/views/webContentEvents.ts
streamer45: It shouldn't interfere since the new manager only deals with `about:blank` targets at the moment. Eventually, I'd like to unify both of these popup handlers.

src/main/views/webContentEvents.ts
devinbinnie: Is there a reason we can't forward them to a link in the `plugins` route that would open this?
The reason I'd rather have that is that I wouldn't want any part of the app opening these windows, for example, if someone was able to post a message that was able to execute some JS, they could force open a malicious popup.

streamer45: > Is there a reason we can't forward them to a link in the plugins route that would open this?

Mostly it's the requirement to open basic `about:blank` popups and not loading any server side content under different paths. 

> The reason I'd rather have that is that I wouldn't want any part of the app opening these windows, for example, if someone was able to post a message that was able to execute some JS, they could force open a malicious popup.

This can happen today in a browser context, the plugin can open whatever. But in Desktop at least we'd be ensuring they could only open `about:blank` which I can't see being any more malicious than the plugin (and webapp as a consequence) already is.

src/main/views/pluginsPopUps.ts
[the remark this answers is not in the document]
streamer45: @devinbinnie To confirm, 

- Are we looking to use `shell.openExternal()` to open external URLs in browser? 
- Should we handle internal links in a particular way? (e.g. through `ViewManager.handleDeepLink`) or should we ignore them for the time being?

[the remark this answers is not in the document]
devinbinnie: `shell.openExternal` is preferred for opening in the browser, yes. Internal links would need to be verified that they are internal to a server, and then yes `ViewManager.handleDeepLink` is the correct method.

src/main/views/webContentEvents.ts
devinbinnie: Test it for me if you would please, if it works then yeah this is fine for now.

streamer45: Yeah, I gave it a try to confirm I wasn't completely wrong :p

src/main/views/pluginsPopUps.ts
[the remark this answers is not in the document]
streamer45: Okay, so I added some basic level of handling:

- Internal (team URLs) get properly routed by the respective main view.
- Anything else we open externally in web browser.

We can review further of course but this should already provide a basic level of functioning.

devinbinnie (approved):

src/main/views/pluginsPopUps.ts
```diff
@@ -0,0 +1,137 @@
...
+        });
+        win.webContents.on('did-start-navigation', (ev: Event<WebContentsDidStartNavigationEventParams>) => {
+            if (ev.url === details.url) {
+                return;
+            }
+
+            log.warn(`prevented popup window from navigating to: ${ev.url}`);
+            ev.preventDefault();
+        });
+        win.webContents.setWindowOpenHandler(({url}): {action: 'deny' | 'allow'} => {
```
enzowritescode, line 70: Just so I understand, is it intended that this function never returns {action: 'allow'}?

streamer45: Yes, we prevent popups from opening more child windows. We still want to allow to open links to either browser (external) or redirect in the main app (e.g. if it's a link to a post/channel).

streamer45: @enzowritescode That's just to make typescript happy. Since we never return `allow` I suppose we could remove that.

> Also, are there any other attributes in the details object that we should be checking aside from URL?

You tell me :p We are denying and only using the URL so we don't really care about the rest I suppose. This is not differing from how we handle this event in other places.

enzowritescode: @streamer45 I think we should remove `allow`. 
I will have to look further into the other stuff to see if it's of any use.

streamer45: Done

cpoile (approved):
Approving (relying on other reviews for security details). Thanks Claudio, and nice tests!

enzowritescode (approved):

cwarnermm:
@devinbinnie - Are developer docs updates needed for this PR from your perspective?

PR #3144, opened 2024-09-11
[MM-60308] Add a set of "Developer Mode" settings that allow the user to turn off systems or force the app to behave a certain way

devinbinnie:
#### Summary
This PR creates a Developer Mode, hidden behind an environment variable that when set, will unlock certain options allowing users to fiddle with the behaviour of the Desktop App. This mode is designed to aid developers in debugging performance issues with the Desktop App, where we believe certain systems with the Desktop App may be the culprit.

The settings added here are:
- Browser Mode Only: Completely disables the preload script, stops web app from knowing it's in the Desktop App
	- Should be the best indicator of if something we're doing with the web app is causing it to slow down or retain memory
	- NOTE: Disables notifications, cross-tab navigation, unread/mentions badges, calls widget, breaks resizing on macOS
- Disable Notification Storage: Turns off the maps that hold references to the notifications until they've been clicked
	- Good for debugging if we're just holding onto too many references to unused notifications
- Disable User Activity Monitor: Turns off the interval that checks if the user is AFK or not
	- Good for debugging if that is affecting the user status negatively, or causing other strange behavior
- Disable Context Menu: Turns off the context menu attached to the BrowserViews
	- Just a sanity check around if that library is causing anything odd
- Force Legacy API: Forces the app to revert back to the old messaging API instead of the newer `contextBridge` API
	- Another sanity check to see if the new API is responsible for holding onto memory
- Force New API: Forces the app to use the `contextBridge` API and completely disables the legacy one
	- Currently the application runs some listeners for the legacy API side-by-side, so this forces them off entirely.

This mode should not be turned on by anyone who isn't a developer, or was instructed by a Mattermost developer to turn this on. When the feature is on, an indicator will show up on the top bar showing that you are in Developer Mode:

Screen Recording 2024-09-11 at 11 26 00 AM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60308

```release-note
Add `Developer Mode` settings to help debug performance issues
```


src/main/app/initialize.ts
```diff
@@ -430,6 +434,7 @@ async function initializeAfterAppReady() {
     }
 
     handleUpdateMenuEvent();
+    DeveloperMode.on(DEVELOPER_MODE_UPDATED, handleUpdateMenuEvent);
```
larkox, line 437: Shouldn't we remove line 436?

larkox (approved):
LGTM, just one small detail.

src/main/app/initialize.ts
devinbinnie: The event only emits when we toggle a developer tool, we still need to create the menu in the first place which is what line 436 does.

src/main/UserActivityMonitor.ts
```diff
@@ -82,6 +82,7 @@ export class UserActivityMonitor extends EventEmitter {
    */
     stopMonitoring() {
         clearInterval(this.systemIdleTimeIntervalID);
+        this.systemIdleTimeIntervalID = -1;
```
marianunez, line 85: Is this change related to Developer Mode? Just wondering

marianunez (approved):

src/main/UserActivityMonitor.ts
devinbinnie: Not directly, it fixed an issue where if we wanted to start monitoring again, we check if the ID is <0. But we never reset it after we stopped monitoring, so it wouldn't be able to restart.

src/main/menus/app.ts
```diff
@@ -176,6 +177,58 @@ export function createTemplate(config: Config, updateManager: UpdateManager) {
         });
     }
 
+    if (DeveloperMode.enabled()) {
+        devToolsSubMenu.push(separatorItem);
```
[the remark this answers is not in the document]
devinbinnie: Each call to `push()` is for a unique item, with different actions for each one. Currently it's all just `toggle()` but we could have others.

Let's say we never do that: We'd have to create an array of values that include the translation string, the default translation and the id anyways. To me this decreases readibility, because then you have to backtrack through past the call to `push` to the array and figure out what that's going.

This seems pretty clear to me to see a menu item added for each menu item, and is also how the rest of the menu items work (outside of servers which aren't hardcoded).

One thing that could be made nicer is calling `push()` on an array of `MenuItem[]` instead of `push()` each time, so I'll change that.

src/main/notifications/index.ts
```diff
@@ -22,13 +23,23 @@ import MainWindow from '../windows/mainWindow';
 const log = new Logger('Notifications');
 
 class NotificationManager {
-    private mentionsPerChannel: Map<string, Mention> = new Map();
-    private allActiveNotifications: Map<string, Notification> = new Map();
+    private mentionsPerChannel?: Map<string, Mention>;
+    private allActiveNotifications?: Map<string, Notification>;
```
hmhealey, line 27: Are you nulling these out to ensure that none of this logic runs without having to change every method to check some `this.isEnabled()` before running its logic? If so, I'm not a huge fan of making these nullable and implicitly disabling this logic by deleting them, but I'm 0/5 on that

src/main/developerMode.ts
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    switchOff = (
+        setting: keyof DeveloperSettings,
+        onStart: () => void,
+        onStop: () => void,
+    ) => {
+        if (!this.get(setting)) {
+            onStart();
+        }
```
hmhealey, line 50: (0/5 and kind of ranty) I wonder if we might want to also call `onStop` when this method is called with the feature disabled to help reduce the number of states that the app is in. Currently, features that use this can be in one of three states:

1. `onStart` was called last because `switchOff` was called with the setting enabled
2. `onStop` was called last because `switchOff` was called with the setting enabled and then the setting was disabled
3. Neither `onStop` not `onStart` have been called because `switchOff` was called with the setting disabled

The third state is potentially the weird one if `onStop` puts the feature into a different state than its initial state. While it might seem a bit odd to stop something that was never started, always calling `onStop` gets rid of that rare third case so we never have to think about it

abhijit-singh (commented):
@devinbinnie Not sure how to test this, which env variable do I change and how? 

But I'm also not sure if this needs any UX testing. We should just make sure it works as expected on Windows too if we haven't already checked that. Other than that, feel free to remove me as a reviewer and go ahead with this.

devinbinnie:
@abhijit-singh 
> Not sure how to test this, which env variable do I change and how?

The variable is `MM_DESKTOP_DEVELOPER_MODE` and it should be set to `true`. On Mac, you can run `export MM_DESKTOP_DEVELOPER_MODE=true` and then run the build from the command line.

> But I'm also not sure if this needs any UX testing. We should just make sure it works as expected on Windows too if we haven't already checked that. Other than that, feel free to remove me as a reviewer and go ahead with this.

The reason I asked for UX was to confirm that the indicator (screenshotted above) looked alright, and that the descriptions of the menu items made sense. If that's all okay, you can approve (or remove yourself) and we'll go from there

abhijit-singh (approved):
Looks good to me! 

One quick suggestion — For `Force Legacy API` and `Force New API`, we can make it clearer what it is about by changing them to something like `Force Legacy Messaging API` and `Force New Messaging API`. I'm 0/5 on it though and it might be obvious enough for the right audience as is.

Issue #3143, opened 2024-09-09
Mattermost desktop app instance cannot be resized or moved on Ubuntu via UI


mascarion:
#### Summary
User is unable to resize or move Mattermost desktop app in Ubuntu. It remains as maximised and no way to change the monitor it appears on either.

#### Steps to reproduce
1. Open Mattermost desktop app (multiple monitors in use, 3 external and laptop one)
2. Attempt to resize or move Mattermost via UI

#### Operating System
Ubuntu Linux 22.04.4 LTS x64

#### Mattermost Versions
Desktop app: 5.9.0
Mattermost: 9.11.0
Schema version: 124

#### Expected behavior
The app can be moved around freely across current and other screens.

#### Observed behavior
Can't find any way to move the app. Frustrating as now stuck with it on a specific monitor which is not ideal for productivity.

devinbinnie:
@mascarion Did the user accidentally press F11 at some point? It sounds like they're stuck in full screen mode. F11 toggles that on/off.

mascarion:
Interesting, that resolves it. At some point F11 must have been used. However, that was a previous app instance and there have been multiple system restarts in between. 
Also the buttons top-right to minimise or get user out of fullscreen mode did not have any effect, only F11 would resolve the issue.

devinbinnie:
The fullscreen setting is persisted between app restarts, as some users prefer to work with the app in that mode consistently. Unfortunately the top buttons do not work in full screen mode and I don't know if there's a way to hide them on Linux when the app is full screened.

My personal opinion: we should not be supporting full screen on Linux as it behaves different between window managers and causes all sorts of issues.

PR #3148, opened 2024-09-19
[MM-60605] Fix the Download button being hidden on Windows/Linux

devinbinnie:
#### Summary
When we made the change to switch to the titleBarOverlay for Windows and Linux, the Downloads dropdown button has disappeared. This is because it was hidden behind the new overlay.

This PR adds a `span` to force the Downloads dropdown in front of the title bar buttons.

NOTE: Will have to update this when cherry-picking to v5.9, as Windows doesn't use these new buttons on v5.9.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60605

```release-note
Fix the Download button being hidden on Windows/Linux
```

### DIFF
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -483,6 +483,7 @@ class MainPage extends React.PureComponent<Props, State> {
                         developerMode={this.state.developerMode}
                     />
                     {downloadsDropdownButton}
+                    <span style={{width: `${window.innerWidth - (window.navigator.windowControlsOverlay?.getTitlebarAreaRect().width ?? 0)}px`}}/>
                     {window.process.platform !== 'darwin' && this.state.fullScreen &&
                         <span className='title-bar-btns'>
                             <div
--- a/src/renderer/components/MainPage.tsx
+++ b/src/renderer/components/MainPage.tsx
@@ -483,7 +483,9 @@ class MainPage extends React.PureComponent<Props, State> {
                         developerMode={this.state.developerMode}
                     />
                     {downloadsDropdownButton}
-                    <span style={{width: `${window.innerWidth - (window.navigator.windowControlsOverlay?.getTitlebarAreaRect().width ?? 0)}px`}}/>
+                    {window.process.platform !== 'darwin' &&
+                        <span style={{width: `${window.innerWidth - (window.navigator.windowControlsOverlay?.getTitlebarAreaRect().width ?? 0)}px`}}/>
+                    }
                     {window.process.platform !== 'darwin' && this.state.fullScreen &&
                         <span className='title-bar-btns'>
                             <div
--- a/src/types/window.ts
+++ b/src/types/window.ts
@@ -155,4 +155,10 @@ declare global {
             };
         };
     }
+
+    interface Navigator {
+        windowControlsOverlay?: {
+            getTitlebarAreaRect: () => DOMRect;
+        }
+    }
 }


marianunez:
> NOTE: Will have to update this when cherry-picking to v5.9, as Windows doesn't use these new buttons on v5.9.

What are these new buttons? Just trying to get more context here

devinbinnie:
> > NOTE: Will have to update this when cherry-picking to v5.9, as Windows doesn't use these new buttons on v5.9.
> 
> What are these new buttons? Just trying to get more context here

They're not really new - just replacements for what we had, using native Electron APIs instead of the custom ones we had ourselves.

marianunez (commented):
Anything we can test to avoid this issue in the future?

src/renderer/components/MainPage.tsx
```diff
@@ -483,6 +483,9 @@ class MainPage extends React.PureComponent<Props, State> {
                         developerMode={this.state.developerMode}
                     />
                     {downloadsDropdownButton}
+                    {window.process.platform !== 'darwin' &&
+                        <span style={{width: `${window.innerWidth - (window.navigator.windowControlsOverlay?.getTitlebarAreaRect().width ?? 0)}px`}}/>
+                    }
```
marianunez, line 488: Would there be any overlap in full screen due to the condition below?

devinbinnie:
> Anything we can test to avoid this issue in the future?

If we had front-end tests, absolutely. Never was able to add those.
Add to the pile of things to do for Desktop App 🙃

src/renderer/components/MainPage.tsx
devinbinnie: Ugh...why does full screen mode have to exist...
Seems like it's causing some weird behaviour, i'll fix that.

devinbinnie: Okay I've fixed it for Windows and "fixed" it for Linux.
Unfortunately full screen on Linux really doesn't work very well, so we'll need to address that separately. See: https://github.com/mattermost/desktop/issues/3143

src/renderer/components/MainPage.tsx
```diff
@@ -438,6 +438,22 @@ class MainPage extends React.PureComponent<Props, State> {
...
+                titleBarSpacing = (
+                    <div
+                        className={`button full-screen-button${this.props.darkMode ? ' darkMode' : ''}`}
+                        onClick={this.handleExitFullScreen}
+                    >
+                        <i className='icon icon-arrow-collapse'/>
+                    </div>
+                );
+            } else {
+                titleBarSpacing = (<span style={{width: `${window.innerWidth - (window.navigator.windowControlsOverlay?.getTitlebarAreaRect().width ?? 0)}px`}}/>);
```
marianunez, line 453: So, does this mean that Linux doesn't need this special case for full screen?

src/renderer/components/MainPage.tsx
```diff
@@ -438,6 +438,22 @@ class MainPage extends React.PureComponent<Props, State> {
...
+        if (window.process.platform !== 'darwin') {
+            if (this.state.fullScreen && window.process.platform !== 'linux') {
+                titleBarSpacing = (
+                    <div
+                        className={`button full-screen-button${this.props.darkMode ? ' darkMode' : ''}`}
+                        onClick={this.handleExitFullScreen}
+                    >
+                        <i className='icon icon-arrow-collapse'/>
+                    </div>
+                );
```
marianunez, line 451: In full-screen does the download button is shown with this case for Windows?

src/renderer/components/MainPage.tsx
devinbinnie: At this point, no. I spent some time trying to work around it but full screen under Linux refuses to behave nicely. So i'm going to create a separate PR that removes full screen support from Linux.

marianunez (approved):

Rajat-Dabade (approved):

Issue #3122, opened 2024-08-14
[Bug]: Handelable exception thrown as runtime error. (Server selection)


fredx30:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

Github "master" at time of writing

### Operating System

Windows_NT 10.0.19045 x64

### Mattermost Server Version

n/a

### Steps to reproduce

n/a - Dont know.

### Expected behavior

A working mattermost client with a servers list should not log uncaughtExceptions due to lack of current server.

### Observed behavior

"uncaughtExceptions" are logged to disk.

### Log Output

```shell
Application: Mattermost 5.8.1
Platform: Windows_NT 10.0.19045 x64
Error: No server set as current
    at Object.getCurrentServer (C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:47496)
    at C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:235771
    at Array.map (<anonymous>)
    at C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:235605
    at C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:237537
    at Te (C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:237544)
    at C:\Users\<win-username>\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:213456
[...]

PR #3152, opened 2024-09-20
[MM-60201] Fixed a potential crash on removing server

devinbinnie:
#### Summary
Wasn't able to reproduce the exact crash, other than this possible case where it's possible that the current server ID isn't set before the menu regenerates, causing a crash where the app menu might be re-rendered before the last server is removed. This PR just reverses those statements so that it's less likely that the `currentServerId` isn't set.

#### Ticket Link
Closes https://github.com/mattermost/desktop/issues/3122
https://mattermost.atlassian.net/browse/MM-60201

```release-note
Fix a potential crash where the app menu might regenerate when `currentServerId` isn't set.
```

### DIFF
--- a/src/app/serverViewState.ts
+++ b/src/app/serverViewState.ts
@@ -208,11 +208,11 @@ export class ServerViewState {
                     this.currentServerId = remainingServers[0].id;
                 }
 
+                ServerManager.removeServer(server.id);
+
                 if (!remainingServers.length) {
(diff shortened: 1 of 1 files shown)


M-ZubairAhmed (approved):

pvev (approved):
LGTM! Good catch 🚀 , thanks @devinbinnie

PR #3165, opened 2024-10-10
[MM-60086][MM-60610] Implement performanceMonitor, collect CPU/memory usage data and send via API

devinbinnie:
#### Summary
This PR contains a new module called `performanceMonitor`, which is responsible for:
- Collect metrics from the main process and various utility processes
- Dispatch requests to the renderer processes for their respective metrics
- Collect all metrics and make any transformations necessary
- Send those metrics to the servers that are able to accept and display them at a set interval

This monitoring will take CPU and memory usage data and send it to servers that have elected to subscribe to the Desktop App for performance data. This data will then be available to administrators via Prometheus. 

The idea behind this is to get a better idea of how hard users are getting hit in terms of system resources by various parts of our app. The data is divided up by unique process, so that should give us some idea of which part of the app is causing issues if there is a lot of system resources being dedicated to a process.

Not in scope for this change are load times, which we will calculate and send as part of a separate API created later.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60610
https://mattermost.atlassian.net/browse/MM-60086

#### Screenshots
image

```release-note
Implement a performanceMonitor to collect and send anonymous usage data to server dashboards.
```

### DIFF
--- a/src/main/performanceMonitor.ts
+++ b/src/main/performanceMonitor.ts
@@ -126,7 +126,13 @@ export class PerformanceMonitor {
                 view.webContents.send(METRICS_REQUEST, view.name, view.serverId);
             });
         });
-        await Promise.all(viewPromises);
+
+        // After 5 seconds, if all the promises are not resolved, resolve them so we don't block the send
+        // This can happen if a view doesn't send back metrics information
+        setTimeout(() => {
+            [...viewResolves.values()].forEach((value) => value());
+        }, 5000);
+        await Promise.allSettled(viewPromises);
         ipcMain.off(METRICS_RECEIVE, listener);
         return metricsMap;
     };
@@ -146,7 +152,7 @@ export class PerformanceMonitor {
             }
 
             const serverMetricsMap = new Map([...metricsMap].filter((value) => !value[1].serverId || value[1].serverId === view.serverId));
-            view.webContents?.send(METRICS_SEND, serverMetricsMap);
+            view.webContents.send(METRICS_SEND, serverMetricsMap);
         }
     };
 
--- a/src/main/performanceMonitor.ts
+++ b/src/main/performanceMonitor.ts
@@ -1,0 +1,171 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {app, ipcMain, type IpcMainEvent, powerMonitor, type WebContents} from 'electron';
+
+import {EMIT_CONFIGURATION, METRICS_RECEIVE, METRICS_REQUEST, METRICS_SEND} from 'common/communication';
+import Config from 'common/config';
+import {Logger} from 'common/log';
+
+const METRIC_SEND_INTERVAL = 60000;
+const log = new Logger('PerformanceMonitor');
+
+type MetricsView = {
+    name: string;
+    webContents: WebContents;
+    serverId?: string;
+}
+
+type Metrics = {
+    serverId?: string;
+    cpu?: number;
+    memory?: number;
+}
+
+export class PerformanceMonitor {
+    private updateInterval?: NodeJS.Timeout;
+    private views: Map<number, MetricsView>;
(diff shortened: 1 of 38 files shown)


yasserfaraazkhan:
verified e2e tests locally. No blocking failures.

yasserfaraazkhan (approved):

src/main/performanceMonitor.ts
```diff
@@ -0,0 +1,171 @@
...
+            viewResolves.get(event.sender.id)?.();
+        };
+        ipcMain.on(METRICS_RECEIVE, listener);
+        const viewPromises = [...this.views.values(), ...this.serverViews.values()].map((view) => {
+            return new Promise<void>((resolve) => {
+                viewResolves.set(view.webContents.id, resolve);
+                view.webContents.send(METRICS_REQUEST, view.name, view.serverId);
+            });
+        });
+        await Promise.all(viewPromises);
```
pvev, line 129: maybe here would be good to add some preventive code:
1) use `allSettled` instead of `all` so you can handle individual promise rejections (in case one of the promise rejects).
2) would be good to wrap this line and the following in a `try..finally`, so you can make sure that no matter what the removal of the listener will happen.
3) should there be a timeout maybe?

src/main/performanceMonitor.ts
```diff
@@ -0,0 +1,171 @@
...
+                continue;
+            }
+
+            if (!view.webContents) {
+                log.error(`Cannot send metrics for ${view.name}  - missing web contents`);
+                continue;
+            }
+
+            const serverMetricsMap = new Map([...metricsMap].filter((value) => !value[1].serverId || value[1].serverId === view.serverId));
+            view.webContents?.send(METRICS_SEND, serverMetricsMap);
```
pvev, line 149: no need for optional chaining `view.webContents?.send` since it occurs after checking `if (!view.webContents)`

src/main/views/serverDropdownView.ts
```diff
@@ -83,6 +84,7 @@ export class ServerDropdownView {
             // @ts-ignore
             transparent: true,
         }});
+        performanceMonitor.registerView('ServerDropdownView', this.view.webContents);
```
pvev, line 87: nit: for this and all the other registered views, maybe put the names in global constant values.
```suggestion
        performanceMonitor.registerView(SERVER_DROPDOWN_VIEW, this.view.webContents);
```

pvev (approved):
LGTM! thanks @devinbinnie

src/main/views/serverDropdownView.ts
devinbinnie: This would be worth doing I'd say if we were using these constants in multiple places, but right now they're just labels that get passed along with the data and will only matter once it hits the dashboard, so I'm okay to leave this one as is.

src/main/performanceMonitor.ts
devinbinnie: i don't see a case where I wouldn't want the app to be loud about the `Promise` failing, so the `try/catch` doesn't seem as necessary. However I did want to add the timeout, thank you for reminding me :)

abhijit-singh (approved):

devinbinnie:
@amyblais Am I okay to cherry-pick this to v5.10 to include in the next RC?

amyblais:
> Am I okay to cherry-pick this to v5.10 to include in the next RC?

I think this would be better for v5.11 unless this is considered to be urgent for v5.10. I thought you were planning only bug fixes for RC-2.

cwarnermm:
@devinbinnie - Are there any product or dev docs updates needed for this PR from your perspective?

devinbinnie:
> @devinbinnie - Are there any product or dev docs updates needed for this PR from your perspective?

Dev docs would be helpful, but I likely won't be able to get to it any time soon.
Product doc wise - maybe we can add something for the upcoming dashboard? https://mattermost.atlassian.net/browse/MM-60613

PR #3166, opened 2024-10-11
[MM-61040] Handle custom protocols in plugins popouts

streamer45:
#### Summary

Unfortunately, I am reminded once more that not all operating systems are created equal.

PR adds the additional check for custom URL protocols for new windows created by plugin popouts. These will be handled via `allowProtocolDialog`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61040

#### Release Note

```release-note
NONE
```

### DIFF
--- a/src/main/views/webContentEventsCommon.ts
+++ b/src/main/views/webContentEventsCommon.ts
@@ -7,6 +7,10 @@ import type {Event} from 'electron';
 import type {Logger} from 'common/log';
 import {getLevel} from 'common/log';
 
+import {protocols} from '../../../electron-builder.json';
+
+const scheme = protocols && protocols[0] && protocols[0].schemes && protocols[0].schemes[0];
+
 enum ConsoleMessageLevel {
     Verbose,
     Info,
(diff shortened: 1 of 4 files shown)


devinbinnie (approved):
Thanks for the quick fix @streamer45!

src/main/views/webContentEventsCommon.ts
```diff
@@ -7,6 +7,10 @@ import type {Event} from 'electron';
 import type {Logger} from 'common/log';
 import {getLevel} from 'common/log';
 
+import {protocols} from '../../../electron-builder.json';
+
+const scheme = protocols && protocols[0] && protocols[0].schemes && protocols[0].schemes[0];
```
enzowritescode, line 12: For my understanding, what does this do exactly? Not sure if it's worth a helper function and some tests?

streamer45: @enzowritescode It simply extracts the protocol scheme we define for our app (i.e., `mattermost`). Then we check against that and if it doesn't match then it must be custom. It's an exception essentially. 

> Not sure if it's worth a helper function and some tests?

Coverage for that is already at 100% through higher level tests. That said, I moved that definition inside `isCustomProtocol` since it's only used there and it makes sense to be close.

enzowritescode (approved):

esarafianou (approved):

PR #3173, opened 2024-10-21
[MM-61058] Upgrade `electron-log`, turn on async logging

devinbinnie:
#### Summary
This PR turns on async logging for the Desktop App, as the default behaviour of `electron-log` is to write synchronously and wait on the main thread. Also I have upgraded `electron-log` to the latest released version.

This should address issues for those with endpoint security/AV software running, or users on a weaker/slower disk that are experiencing slowness in the app, especially while on a higher logging level.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61058

```release-note
Upgrade `electron-log`, turn on async logging
```

### DIFF
--- a/src/common/log.ts
+++ b/src/common/log.ts
@@ -6,6 +6,10 @@ import log from 'electron-log';
 
 import Util from 'common/utils/util';
 
+// Turn off sync logging to prevent blocking the main thread
+// One downside to this is that some logs may not be written to the log file when the app closes
+log.transports.file.sync = false;
+
 export const setLoggingLevel = (level: string) => {
     if (log.transports.file.level === level) {
         return;
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "electron-context-menu": "3.6.1",
         "electron-extension-installer": "1.2.0",
         "electron-is-dev": "2.0.0",
-        "electron-log": "5.1.1",
+        "electron-log": "5.2.0",
         "electron-updater": "6.3.0",
         "joi": "17.12.2",
         "macos-notification-state": "3.0.0",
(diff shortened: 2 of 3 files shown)


src/common/log.ts
```diff
@@ -6,6 +6,10 @@ import log from 'electron-log';
 
 import Util from 'common/utils/util';
 
+// Turn off sync logging to prevent blocking the main thread
+// One downside to this is that some logs may not be written to the log file when the app closes
```
devinbinnie, line 10: To expand on this, any logs that need to be written after the `handleAppBeforeQuit` listener fires seem to be missed. Most of those are debug logs though, so it won't affect many cases, but it's important to be aware in case we need to debug some issue with the app on quit.

nickmisasi: Could you re-enable sync logs when before-quit is fired?

devinbinnie: Tried that - unfortunately didn't help. At the time the Node process is quitting any logs are already in the queue to write to file, and they get lost when the process ends. You'd think they'd have a built-in way to wait for those to finish :P

nickmisasi: Yeah you'd think there'd be some sort of `.flush()`. Too bad!

nickmisasi (approved):
LGTM

crspeller (approved):

devinbinnie:
@amyblais Would we be okay to cherry-pick this into v5.10? It's a pretty big performance improvement for people with weak disks/have certain AV solutions installed.

amyblais:
> Would we be okay to cherry-pick this into v5.10? It's a pretty big performance improvement for people with weak disks/have certain AV solutions installed.

@devinbinnie Sounds good.

PR #3174, opened 2024-10-21
[MM-59483] Remove legacy preload and custom login code

devinbinnie:
#### Summary
This PR removes code in the Desktop App that was left to support older versions of the Mattermost Server/Web App:
- Removed the legacy preload listeners using `window.postMessage`, needed for server versions <v9.4
- Removed the custom login code that allows the app to go out to another site and login with the app, needed for versions <v9.2.

What this does mean is that as of this change, Mattermost servers running versions under v9.4 will now have a reduced experience, and may or may not function at all. We will be dropping support for those server versions in the Desktop App officially as of v5.11. 

This is being done to both reduce complication, improve security (in the case of custom logins) and potentially improve performance (in the case of the preload script changes)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59843

```release-note
Remove legacy code for older unsupported Mattermost servers
```


devinbinnie:
@yasserfaraazkhan We can test this whenever, but since we're removing support for older versions, we should double check that the versions we support are all working well :)
I ran E2E tests on my local machine and everything went okay - but a bit more thorough testing seems applicable here.

yasserfaraazkhan (approved):
Tested this with servers

- v9.5 ✅ 
- v9.4 ✅ 
- v9.2 & v9.0 the login flow was stuck and in logs we can see error `[TAB_MESSAGING] Cannot reach server: Error: net::ERR_CONNECTION_REFUSED`

gabrieljackson (approved):
Nice clean up. I am assuming the failing `Electron Playwright Tests` is expected or being worked on separately?

PR #3185, opened 2024-10-25
[MM-37379] Add "Clear Data" options in the menu to allow users to force Electron to blow away session data.

devinbinnie:
#### Summary
A common problem that seems to happen when users upgrade their Desktop App is that when the new app loads, they can't access their server and it appears to be stuck at a white screen. The only solution we've found to this is to have the user manually reset their app by removing their config and app data folder.

We haven't been able to exactly reproduce this issue, but the feeling is that it's likely to do with some session or cache data being corrupted during the upgrade process. However, since we can't reproduce it, I'm not able to exactly pinpoint where the issue is and to fix it exactly.

So instead, this PR adds two menu options: Clear Data for Current Server and Clear All Data. These options will force the `session` to clear its data for either a single server, or clear for all, which shouldn't hopefully be an easier, in-app way that does not force the user to reconfigure their application/servers and should still help alleviate this problem.

It's possible this might not end up doing anything - as we know Clear Cache and Reload doesn't work, but this is the next step to forcing the app to blow out session data that is likely the culprit here.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37379

#### Screenshots
image
image

```release-note
Added two menu items to help users forcibly clear out cache and session data.
```


hanzei (approved):
Great work! I love seeing more debugging tool for the desktop app :rocket:

svelle (approved):
LGTM!

I think this is a great addition.

PR #3200, opened 2024-11-11
[MM-22555] Auto-fill server URLs when deep linking into the Desktop App and the server isn't configured

devinbinnie:
#### Summary
This PR add a quality-of-life change to the Desktop App, where now users are able to click on deep links for server they do not have configured in the app, and have the app automatically pre-fill the Add Server modal (or Welcome Screen) for them with the server information. From there, the user can simply click Connect.

On top of this, permalinks will be automatically redirected to under this flow, so you can get to the message you want to see quicker.

Done for 2024 November Hackathon.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22555

```release-note
Auto-fill server URLs when deep linking into the Desktop App and the server isn't configured
```

PR #3201, opened 2024-11-11
Fixed issues with loading the app from cold when deep linking

devinbinnie:
#### Summary
While working on https://github.com/mattermost/desktop/pull/3200 I found an issue with deep linking when booting the app from cold. The application would fail to load the initial server, and the dropdown menu would also not load. This was caused by the main window creation call happening too early, before a bunch of other setup code could be run.

This PR fixes the issue by having the application wait for the main window to be created normally, before attempting to deep link.

```release-note
Fixed issues with loading the app from cold when deep linking
```

### DIFF
--- a/src/main/app/utils.ts
+++ b/src/main/app/utils.ts
@@ -8,7 +8,7 @@ import type {BrowserWindow, Rectangle} from 'electron';
 import {app, Menu, session, dialog, nativeImage, screen} from 'electron';
 import isDev from 'electron-is-dev';
 
-import {APP_MENU_WILL_CLOSE} from 'common/communication';
+import {APP_MENU_WILL_CLOSE, MAIN_WINDOW_CREATED} from 'common/communication';
 import Config from 'common/config';
 import JsonFileManager from 'common/JsonFileManager';
 import {Logger} from 'common/log';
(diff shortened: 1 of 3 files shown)


devinbinnie:
@amyblais This might be a case for a dot release - as the app is somewhat unusable when users get into this state.

src/main/app/utils.ts
```diff
@@ -38,8 +38,12 @@ const log = new Logger('App.Utils');
 
 export function openDeepLink(deeplinkingUrl: string) {
     try {
-        MainWindow.show();
-        ViewManager.handleDeepLink(deeplinkingUrl);
+        if (MainWindow.get()) {
+            MainWindow.show();
+            ViewManager.handleDeepLink(deeplinkingUrl);
+        } else {
+            MainWindow.on(MAIN_WINDOW_CREATED, () => ViewManager.handleDeepLink(deeplinkingUrl));
```
lieut-data, line 45: Are there any race conditions at play? (Could `MainWindow` get created between our `get` and `on` registration?

lieut-data (approved):
Thanks for the chance to take a peek! One question below.

src/main/app/utils.ts
devinbinnie: Well I went to go double-check, I found another instance where we're calling `show()` before `app.ready()` which is really not good.

However, other callers of this are also waiting for the `ready` event, so as long as we're calling it after that there shouldn't be any race conditions as that's when we expect the window to be registered.

The `ready` event docs if you're interested: https://www.electronjs.org/docs/latest/api/app#event-ready

larkox, line 45: This route doesn't call `MainWindow.show`. Is that ok? Is `MainWindow.show` called by other side effect?

larkox (approved):
LGTM, just one small thing to verify.

yasserfaraazkhan (approved):

Issue #3140, opened 2024-09-03
[Bug]: Downloading an SVG leads to error "Failed to get thumbnail from local thumbnail cache reference"


bjoernmartin:
### Checks before filing an issue

- [X] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [X] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing.
- [X] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [X] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [X] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [X] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.9.0

### Operating System

Windows 11 x64

### Mattermost Server Version

EE 9.5.6

### Steps to reproduce

* Navigate to a chat with SVG images attached
* Click on the `Download` icon of that SVG image

![image](https://github.com/user-attachments/assets/d32fd633-339d-42b4-bf62-8dc2471af82b)

### Expected behavior

* SVG is successfully downloaded 
* No error message is shown and Mattermost does not need to restart

### Observed behavior

* SVG is successfully downloaded (yes, actual download works)
* An error message is shown (see below) and Mattermost has to be restarted

![image](https://github.com/user-attachments/assets/cf2ad89e-e7c4-4b79-9148-f7eb858afe1e)

### Log Output

```shell
Application: Mattermost 5.9.0
Platform: Windows_NT 10.0.22631 x64
Error: Failed to get thumbnail from local thumbnail cache reference
    at C.formatDownloadItem (C:\Users\martin\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:258206)
    at C.upsertFileToDownloads (C:\Users\martin\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:255614)
    at C.doneEventController (C:\Users\martin\AppData\Local\Programs\mattermost-desktop\resources\app.asar\index.js:2:256970)
[...]

PR #3224, opened 2024-11-25
[MM-61919] Wrap potential crash in try/catch for thumbnails, fall back to `createFromPath`

devinbinnie:
#### Summary
There seems to be a Windows-only crash case where calling `createThumbnailFromPath` would throw an error failing to get the thumbnail from a local cache (see https://github.com/electron/electron/blob/main/shell/common/api/electron_api_native_image_win.cc#L69). I'm not certain but I'm guessing it has something to do with files that do not have thumbnails pre-generated by the OS, and Electron expects that they are always there. This causes a hard crash since that method throws an error.

This PR simply adds an error boundary to fallback to `createFromPath` in those situations where a thumbnail cannot be retrieved, to avoid having the app crash hard.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61919
Closes https://github.com/mattermost/desktop/issues/3140

```release-note
Fixed an issue where the app could crash loading a thumbnail on Windows
```

### DIFF
--- a/src/main/downloadsManager.ts
+++ b/src/main/downloadsManager.ts
@@ -635,10 +635,11 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
                 if (item.getReceivedBytes() < 1000000) {
                     thumbnailData = (await nativeImage.createFromPath(overridePath ?? item.getSavePath())).toDataURL();
                 }
-            }
+            };
+
             // Linux doesn't support the thumbnail creation so we have to use the base function
             if (process.platform === 'linux') {
-                await fallback()
+                await fallback();
             } else {
                 // This has been known to fail on Windows, see: https://github.com/mattermost/desktop/issues/3140
                 try {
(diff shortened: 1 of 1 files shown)


larkox (approved):

src/main/downloadsManager.ts
```diff
@@ -630,14 +630,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+
+            // Linux doesn't support the thumbnail creation so we have to use the base function
+            if (process.platform === 'linux') {
+                await fallback();
             } else {
-                thumbnailData = (await nativeImage.createThumbnailFromPath(overridePath ?? item.getSavePath(), {height: 32, width: 32})).toDataURL();
+                // This has been known to fail on Windows, see: https://github.com/mattermost/desktop/issues/3140
+                try {
+                    thumbnailData = (await nativeImage.createThumbnailFromPath(overridePath ?? item.getSavePath(), {height: 32, width: 32})).toDataURL();
+                } catch {
```
larkox, line 647: I wonder if it is worth it to have some log here, just so it is not completely silent. 0/5

hanzei (approved):
LGMT :+1:

src/main/downloadsManager.ts
devinbinnie: Since we don't know when this happens, or what to do with the log when we see it it's probably fine to leave as is.

PR #3231, opened 2024-11-27
[MM-61821] Automatically allow permission checks for supported permission types through for GPO configured servers

devinbinnie:
#### Summary
This is a small QoL improvement for enterprise users who have their Desktop App configured for them by their system administrator. Normally the application will ask for permission to send notifications, access the camera/microphone, and other system related permissions on a per-server basis. However, if the server is configured using Group Policy on Windows, we can make an assumption that the server is trusted since it is configured at the registry level.

This PR just removes the permission check for those servers.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61821

```release-note
Skip per-server permission checks for GPO-configured servers on Windows
```

### DIFF
--- a/src/main/permissionsManager.ts
+++ b/src/main/permissionsManager.ts
@@ -22,6 +22,7 @@ import {
     OPEN_WINDOWS_MICROPHONE_PREFERENCES,
     UPDATE_PATHS,
 } from 'common/communication';
+import Config from 'common/config';
 import JsonFileManager from 'common/JsonFileManager';
 import {Logger} from 'common/log';
 import type {MattermostServer} from 'common/servers/MattermostServer';
@@ -141,7 +142,7 @@ export class PermissionsManager extends JsonFileManager<PermissionsByOrigin> {
             return false;
         }
 
-        let serverURL;
+        let serverURL: URL | undefined;
         if (CallsWidgetWindow.isCallsWidget(webContentsId)) {
             serverURL = CallsWidgetWindow.getViewURL();
         } else {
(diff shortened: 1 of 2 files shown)


src/main/permissionsManager.ts
```diff
@@ -152,6 +153,11 @@ export class PermissionsManager extends JsonFileManager<PermissionsByOrigin> {
             return false;
         }
 
+        // For GPO servers, we always allow permissions since they are trusted
+        if (Config.registryData?.servers?.some((s) => parseURL(s.url)?.href === serverURL.href)) {
```
davidkrauser, line 157: Can `serverURL` be `undefined` here? Is it safe to access the `href` property?

devinbinnie: Nope, see the above falsy check: https://github.com/mattermost/desktop/pull/3231/files#diff-fc4629b25400b7a5efa825a58b4db1a3caaba2c20770948a5ba12f6b8de0b8eaR152

davidkrauser: Ahh - sorry for the noise. Was hidden behind a collapsed region :+1:

devinbinnie: Though for some reason `tsc` agreed with you :P

src/main/permissionsManager.ts
```diff
@@ -152,6 +153,12 @@ export class PermissionsManager extends JsonFileManager<PermissionsByOrigin> {
             return false;
         }
 
+        // For GPO servers, we always allow permissions since they are trusted
+        const serverHref = serverURL.href;
+        if (Config.registryData?.servers?.some((s) => parseURL(s.url)?.href === serverHref)) {
```
larkox, line 158: Just sanity check question. At `Config.registryData?.servers?` we only have GPO servers, right?

larkox (approved):
LGTM

davidkrauser (approved):
Looks good to me :+1:

src/main/permissionsManager.ts
devinbinnie: Correct, they are servers that come from the Windows registry. I was considering doing all predefined servers (including ones specified by the build), but I thought that might expose people to be tricked into downloading an app with a pre-configured server that has access to everything.

devinbinnie:
@enzowritescode Gentle ping for review :)

enzowritescode (approved):

PR #3233, opened 2024-11-29
[MM-61744] Import GenericModal, convert Remove Server to new design

devinbinnie:
#### Summary
This PR adds the `GenericModal` to the Desktop App that is used by our web app, and its related stylings. This will be the stepping stone used to refresh the modals in the Desktop App that are still using old Bootstrap.

Additionally, this PR serves as a template for converting the existing modals to the new design, using a proof-of-concept of converting the Remove Server modal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61744

#### Screenshots
|  before  |  after  |
|----|----|
| image | image |

#### Release Note
```release-note
Refresh Remove Server Modal using new designs
```


src/renderer/components/GenericModal/_css_variables.scss
```diff
@@ -0,0 +1,113 @@
+:root {
+    // Default Hex values to be replaced by applyTheme()
+    // these are define to prevent flashes of unstyled content
+    --icon-opacity: 0.64;
+    --icon-opacity-hover: 0.8;
+    --away-indicator: #ffbc42;
```
hmhealey, line 6: Does the desktop app need to define its own values for these so that the new modal follows its theme?

src/renderer/components/GenericModal/_buttons.scss
hmhealey, line 1: Is it worth adding a "do not change" comment to these files that are copied from the web app?

hmhealey (approved):
I have a couple questions, but this looks decent to me. I'm a bit surprised how much needs to be copied over for just GenericModal, but I really shouldn't be

src/renderer/components/GenericModal/_css_variables.scss
devinbinnie: The theme is actually set by a separate module called `bootstrap-dark` that injects dark mode styles when it's enabled, so leaving these as-is should be fine.

src/renderer/components/GenericModal/_buttons.scss
devinbinnie: 0/5, but probably not. I doubt anyone else will be changing these other than the contributor working on the modals - and I can just tell him directly.

src/renderer/components/GenericModal/_css_variables.scss
hmhealey: Will it override the values of these variables though? Bootstrap may have its own theme variables, but they could clash with these ones from the web app

devinbinnie: It won't change our variables, no. I forget exactly what it does but I think it's something to do with just lazy loading the style sheets. The way it works for us is that when the `darkMode` class is applied it will use styles that use the variables we have.

M-ZubairAhmed:
@devinbinnie i will not be having time to review this one sorry. removing my assignment

abhijit-singh (changes requested):
Looks great @devinbinnie. Requesting a few small tweaks:
1. There is some extra space on the right and below the buttons. And there's some extra space between the title and the body text. Can we [match them to the screenshot](https://www.figma.com/design/bUtJ0WJJNNwNEOCw6guQGa/Desktop-App-Settings-UI?node-id=1682-14558&t=pXj9zq4VcdfGFE5b-4) below? 
image

2. On hover, the `Remove` button turns blue, whereas it should just become a bit darker (an additional 0.08 opacity of `#000000` on the base red color as per [figma](https://www.figma.com/design/bUtJ0WJJNNwNEOCw6guQGa/Desktop-App-Settings-UI?node-id=1682-14558&t=pXj9zq4VcdfGFE5b-4), not sure if we have another color variable for it).
image

3. In the dark mode, the background is a bit light which reduces the contrast with the text, especially the `Cancel` button. Can we change it from `#424242` to `#191B1F` (which is the background-color in one of our standard dark themes). I believe this can be done for all desktop app specific modals in dark mode unless you see any issues. 
image

abhijit-singh (approved):
Looks great, thank you @devinbinnie!

devinbinnie:
@jespino Gentle ping for review :)

jespino (approved):
It looks good to me, I'm expecting here that most of it is copy and paste from the existing implementation in the mattermost webapp, right?

devinbinnie:
> It looks good to me, I'm expecting here that most of it is copy and paste from the existing implementation in the mattermost webapp, right?

Correct, I just made a few modifications which I've commented for the most part. Hopefully we can just export it from the web app eventually :)

PR #3238, opened 2024-12-03
[MM-60416] Check for the presence of the EXE uninstaller before trying to run it

devinbinnie:
#### Summary
Once of the features of the new MSI installer is to basically replace the EXE one, despite the fact that they use different installation methods and thus different ids, making them able to be installed side by side. What we did was to force the uninstallation of the EXE installer if we install via MSI. However, this was throwing an error (which seemed to be ignored most of the time) when the uninstaller was missing for those using a deployment system.

This PR adds a check for the uninstaller before we try to run it, which avoids the error altogether.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60416

```release-note
Fixed a potential error thrown by the MSI when trying to uninstall the EXE
```


larkox (approved):

M-ZubairAhmed (approved):
what command did you ran to generate this patch?

devinbinnie:
> what command did you ran to generate this patch?

`npx patch-package app-builder-lib`

PR #3249, opened 2024-12-12
[MM-62196] Add support for downgrades using the MSI

devinbinnie:
#### Summary
An issue was noted while testing the MSI and trying to install an older version over top. An error would be thrown and the MSI would technically install, but would miss the main EXE file causing the app to not work at all.

This PR makes a couple template changes to actually support downgrading properly (as of the installer that this is released in) and fixes the error and missing EXE.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62196

```release-note
Add support for downgrading using the MSI installer
```


hmhealey (approved):

marianunez (approved):

devinbinnie:
@yasserfaraazkhan Just to speed things along, I'm going to merge without QA approval here and cut a new dot release that you can test. Feel free to test this fix there.

PR #3253, opened 2024-12-12
[MM-62205] Dynamically toggle widget window visibility when popout opens/closes

streamer45:
#### Summary

PR adds some logic to dynamically change the global Calls widget visibility when the popout window is open. This is to address the inconvenience of having this window always on top even when focusing on the popout.

Hiding the widget completely is also possible, but I still feel it's a more impactful change in behavior. My hope is that the proposed changes are enough to address the reported issue. That said, let me know if you feel strongly about this.

The previous conversation for extra context is at https://hub.mattermost.com/private-core/pl/wbf4jd67g7fnpmon8fafwnxmuy

#### Video

https://github.com/user-attachments/assets/c58e7dee-36a4-4b6c-a043-0acfd4b7da92

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62205

#### Release Note

```release-note
Calls: while the popout window is open, the widget window's visibility will change so that it is not always on top of other windows.
```

### DIFF
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -233,14 +233,14 @@ export class CallsWidgetWindow {
         ev.preventDefault();
     };
 
-    private toggleWidgetVisibility = (flag: boolean) => {
-        log.debug('toggleWidgetVisibility', flag);
+    private setWidgetWindowStacking = ({onTop}: {onTop: boolean}) => {
+        log.debug('setWidgetWindowStacking', onTop);
 
         if (!this.win) {
             return;
         }
 
-        if (flag) {
+        if (onTop) {
             this.win.setVisibleOnAllWorkspaces(true, {visibleOnFullScreen: true, skipTransformProcessType: true});
             this.win.setAlwaysOnTop(true, 'screen-saver');
             this.win.focus();
@@ -257,7 +257,7 @@ export class CallsWidgetWindow {
             return;
         }
 
-        this.toggleWidgetVisibility(true);
+        this.setWidgetWindowStacking({onTop: true});
 
         const bounds = this.win.getBounds();
         const mainBounds = mainWindow.getBounds();
@@ -303,7 +303,7 @@ export class CallsWidgetWindow {
     private onPopOutCreate = (win: BrowserWindow) => {
         this.popOut = win;
 
-        this.toggleWidgetVisibility(false);
+        this.setWidgetWindowStacking({onTop: false});
 
         // Let the webContentsEventManager handle links that try to open a new window.
         webContentsEventManager.addWebContentsEventListeners(this.popOut.webContents);
@@ -321,7 +321,7 @@ export class CallsWidgetWindow {
         this.popOut.on('closed', () => {
             delete this.popOut;
             contextMenu.dispose();
-            this.toggleWidgetVisibility(true);
+            this.setWidgetWindowStacking({onTop: true});
         });
 
         // Set the userAgent so that the widget's popout is considered a desktop window in the webapp code.
--- a/src/main/windows/callsWidgetWindow.ts
+++ b/src/main/windows/callsWidgetWindow.ts
@@ -233,6 +233,23 @@ export class CallsWidgetWindow {
         ev.preventDefault();
     };
 
+    private toggleWidgetVisibility = (flag: boolean) => {
+        log.debug('toggleWidgetVisibility', flag);
+
+        if (!this.win) {
+            return;
+        }
+
+        if (flag) {
+            this.win.setVisibleOnAllWorkspaces(true, {visibleOnFullScreen: true, skipTransformProcessType: true});
+            this.win.setAlwaysOnTop(true, 'screen-saver');
+            this.win.focus();
+        } else {
+            this.win.setAlwaysOnTop(false);
+            this.win.setVisibleOnAllWorkspaces(false);
+        }
+    };
+
     private onShow = () => {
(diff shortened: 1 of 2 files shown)


matthewbirtch:
Nice solve @streamer45. I like this as a solution and I think it will solve the biggest part of the issue. I've triggered a build to test and play with a bit and I'll report back.

matthewbirtch:
I think this feels good to me. While testing a few things I noticed (that may or may not be as a results of this PR)

1. I'm not sure if this is a results of this change, but when I tried to share my screen I was unable to from the popout view. I had not yet provided permissions, it did not trigger the OS permission dialog either. It just did nothing. Once I cliked the share button from the widget, the OS prompt came up and allowed me to share. However, even after allowing the OS permission from there, I was still unable to share my screen from the popout.

2. The widget window shows up in the list of screens to use. I don't believe it does this currently in the desktop app and we should exclude it.
image

streamer45:
> I think this feels good to me. While testing a few things I noticed (that may or may not be as a results of this PR)
> 
> 1. I'm not sure if this is a results of this change, but when I tried to share my screen I was unable to from the popout view. I had not yet provided permissions, it did not trigger the OS permission dialog either. It just did nothing. Once I cliked the share button from the widget, the OS prompt came up and allowed me to share. However, even after allowing the OS permission from there, I was still unable to share my screen from the popout.
> 2. The widget window shows up in the list of screens to share. I don't believe it does this currently in the desktop app and we should exclude it.
> 
> image

Thanks @matthewbirtch . I think both are unrelated to this PR. 

For 1, I am honestly surprised this wasn't reported before but I'll have a look asap. Created https://mattermost.atlassian.net/browse/MM-62241

For 2, from what I remember, that's always been the case since the widget is effectively its own window. Excluding it may require us to do some pattern matching on the window title because (as far as I know) there isn't a great way to uniquely identify that specific window from the sources returned by the desktop capturing engine. I'll see what can be done. Created https://mattermost.atlassian.net/browse/MM-62242

matthewbirtch:
Hmm, interesting. I tried to replicate both of these things on hub with desktop v5.10.1 and didn't get the same result.

streamer45:
> Hmm, interesting. I tried to replicate both of these things on hub with desktop v5.10.1 and didn't get the same result.

On Linux (both 5.9 and 5.10), I can definitely see the widget being among the windows available to share. But yeah, I can't reproduce the sharing from popout issue outside of `master` branch so there could be something there.

devinbinnie (approved):
Nice little QoL fix :)

src/main/windows/callsWidgetWindow.ts
```diff
@@ -233,16 +233,31 @@ export class CallsWidgetWindow {
         ev.preventDefault();
     };
 
+    private toggleWidgetVisibility = (flag: boolean) => {
```
davidkrauser, line 236: **suggestion:** I initially found the `visibility` terminology here confusing, and thought the purpose of this function was to hide/show the widget. Could we use a name more specific to what we're toggling? e.g. it seems the purpose here is to toggle if the widget should be shown on all workspaces on top of other windows.

davidkrauser (approved):
Thank you for tackling this so quickly - this addresses my concerns with the widget :+1: 

Approving, but just have a minor non-blocking suggestion about the name of a function.

src/main/windows/callsWidgetWindow.ts
streamer45: @davidkrauser Thanks. You know what they say about naming things :p

For fun, I asked Copilot for a better function name, but its response was far from convincing (it is still using the word visibility). Do you happen to have a suggestion in mind?

Maybe we could use something like priority since we are essentially mutating the level of the window.

matthewbirtch:
> I can't reproduce the sharing from popout issue outside of `master` branch so there could be something there.

Ok, let me know if you want to look in to this as part of this PR or not. Happy to help test

src/main/windows/callsWidgetWindow.ts
davidkrauser: > _You know what they say about naming things_

I know, right? :sweat_smile: 

I like priority - maybe something like `setWindowHighPriority(bool)`?

matthewbirtch (approved):
Approving as the issues I identified don't appear to be a part of this PR. Thanks @streamer45!

src/main/windows/callsWidgetWindow.ts
cpoile: Priority could be confusing too... Maybe `setWindowHighZIndex` because we're kind of talking about window z-indexes? 

(apologies for the bikeshedding)

cpoile, line 236: Another: `flag` has little meaning... Anything else would be better, imo.  `highIndex` `highVisibility` etc.?

cpoile (approved):
Thanks!

src/main/windows/callsWidgetWindow.ts
streamer45: @cpoile Interestingly, I copied that arg name straight from the Electron API for one of those methods :) I'll see what I can do.

streamer45:
@cpoile Check https://github.com/mattermost/desktop/pull/3253/commits/59a3d939d97aadd492fac3b507d6fcf3c6b8776b and let me know if it's better. I changed the method's name and also made its argument explicit so that it's easier to understand its meaning from the calling point.

cpoile (approved):
Awesome! :)

PR #3255, opened 2024-12-16
Revert "Translations update from Mattermost Weblate"

devinbinnie:
Reverts mattermost/desktop#3254

Needs to be rebased and merged, not squashed.


lieut-data:
@devinbinnie, curious why this needs to be merged vs squashed? We fixed the i18n issues that previously required this, so not necessary from weblate's perspective at least.

devinbinnie:
*sigh* Well that's news to me...

I will un-revert :P

lieut-data:
It's fine either way, just for the future @devinbinnie :)

PR #3258, opened 2024-12-16
[MM-62241] Fix screen sharing from Calls popout window

streamer45:
#### Summary

As of https://github.com/mattermost/desktop/pull/3174, some of the functionality in the Calls popout window (e.g. screen sharing) stopped working since `desktopAPI` is missing on the window object. Using the local preload seems to be the way to go.

Since we are here, I am also adding an extra dynamic menu item to access the developer tools of the popout, which can be helpful in debugging future issues like this one.

/cc @matthewbirtch 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62241

#### Release Note

```release-note
NONE
```


cpoile (approved):

streamer45:
@DHaussermann You should be familiar with this since you hit it before while testing unrelated changes.

streamer45:
@DHaussermann Updated

devinbinnie:
@DHaussermann Can we prioritize QA review on this? Need to cut the RC on Monday.

DHaussermann:
@devinbinnie Yes, happy to make this a priority. 

Perhaps you can assist? I tried testing this yesterday and can't get a Mac build to install. 

- The pipeline is green so I go to the summary of the Details if the build step for Mac here https://github.com/mattermost/desktop/actions/runs/12792438869/job/35662896248?pr=3258
- From there I go to the Summary page here https://github.com/mattermost/desktop/actions/runs/12792438869?pr=3258
- Then I download my artifact from the bottom 
image

When I unzip the build for ARM64 and try to launch it Mac OS tells me the app is damaged. 
image

Is there another way to do this? I feel like this ☝️ is exactly what I did in the past for desktop PRs

devinbinnie:
@DHaussermann You need to add the "Build Apps for PR" label, that creates a signed app that you should be able to run. I've added it for you and you should be able to grab the build from the check when it is finished building.

devinbinnie:
@DHaussermann They're in the artifacts for the GitHub Action: https://github.com/mattermost/desktop/actions/runs/12831481292

DHaussermann:
@streamer45 I don't see this issue solved. 
- Install build from the pipeline that is now in the `build-for-pr` and confirm I'm running a dev build `Version 5.11.0-develop.1 - Start a call and open the expanded view
- Click the screen share option
**Observed**: The share button does not seem responsive and does not open the view were I can select what I want to share.

Can you please advice if you can repro this?

devinbinnie:
@streamer45 @DHaussermann this ideally needs to merged before EOD today so I can cut v5.11-rc.1. Please prioritize or let me know if we plan to push this to a future RC/release.

DHaussermann:
@streamer45 I have done a round of testing on this and the fix seems to work once I have the updated plugin build to support it.
- The screen share works from expanded view now
- The screen share from expanded view will flip back to `Channels` as needed.

But I see a couple issues. 
- The `Leave call` option embedded in the post seems to have stopped working. This works on 5.10.2 I can't find the cause from the browser console. 
image

Can you advise if this could also be related or if it's a completely separate? 
 
 - I also notice that when sharing a screen from the expanded view, the focus does not return to the expanded view so, if you're on the same monitor, it disappears under the desktop app. This does not happen when sharing from the module. 
 This is **not** a regression from 5.10.2 I can open a separate issue.

streamer45:
Thanks @DHaussermann 

1. Looks like another regression on the plugin side. Will send a fix over there.
2. We don't do anything to bring back focus to popout. It's something we can track separately as an improvement but not a blocker here.

DHaussermann (approved):
Tested and passed 
- As mentioned above, the share option now works from the expanded view
- The share option will toggle to channel view when sharing
- Confirmed issue with Leave call option in the post is resolved
- Did a click-through test to ensure all options in the modal and the expanded view are functional when using updated Calls build https://github.com/mattermost/mattermost-plugin-calls/pull/948

 LGTM!
 
 Thanks @streamer45!

PR #3276, opened 2025-01-06
Switch to Xcode 16.2 for CI

devinbinnie:
Apparently GitHub has shifted the Xcode image to 16.2, which broke our workflow again. For some reason there doesn't seem to be a static place to select the Xcode SDK, so we're stuck manually upgrading it everytime they do...

Perhaps they will have an answer for me [here](https://github.com/actions/runner-images/issues/11021#issuecomment-2573225187) but for now we'll use this to get us back up and running.

```release-note
NONE
```


pvev (approved):
LGTM!

mvitale1989 (approved):
Thanks for fixing Devin!

PR #3294, opened 2025-01-27
[MM-62649] Switch DeveloperModeIndicator over to floating-ui

devinbinnie:
#### Summary
Removing the original `OverlayTrigger` usage from `react-bootstrap` and replacing the tooltip with `floating-ui`, as we do in the webapp. I re-used a modified version of the `WithTooltip` component from the web app, that can be replaced if that one is ever exported as a separate package.

Also realized it might make sense to have developer mode on, while in development, so I've set that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62649

```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@emotion/react": "11.11.4",
+        "@floating-ui/react": "0.26.28",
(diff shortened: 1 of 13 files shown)


M-ZubairAhmed (approved):
desktop to monorepo when 😝  So we avoid copying over stuff

PR #3310, opened 2025-01-31
[MM-62846] Fix crash caused by `electron-context-menu`

devinbinnie:
#### Summary
An issue with `electron-context-menu` where it assumes `isDestroyed` is on a window (its not) crashes the app.

This PR patches the library to check the webContents object for `isDestroyed` instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62846

```release-note
NONE
```

### DIFF
--- a/patches/electron-context-menu+4.0.4.patch
+++ b/patches/electron-context-menu+4.0.4.patch
@@ -47,7 +47,7 @@ index 468e48b..e182878 100644
  		event: ElectronEvent
  	) => MenuItemConstructorOptions[];
 diff --git a/node_modules/electron-context-menu/index.js b/node_modules/electron-context-menu/index.js
-index b10daea..ea2f891 100644
+index b10daea..d7d26f3 100644
 --- a/node_modules/electron-context-menu/index.js
 +++ b/node_modules/electron-context-menu/index.js
 @@ -1,7 +1,6 @@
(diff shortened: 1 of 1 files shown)


patches/electron-context-menu+4.0.4.patch
```diff
@@ -103,3 +103,12 @@ index b10daea..ea2f891 100644
  				},
  			}),
  			copyImage: decorateMenuItem({
+@@ -350,7 +349,7 @@ const create = (win, options) => {
```
marianunez, line 106: @devinbinnie can you give a bit on context on what is this file we are changing?

devinbinnie: It's the package `electron-context-menu`, it provides the right-click menu used in some places.

devinbinnie: No this is my patch I made to fix the issue.

marianunez: Is a divergence to the upstream module? I'm just trying to understand the process

devinbinnie: Yes. They have a bug where they assume that the object passed in has an `isDestroyed` method. By consequence, the ones they've defined do, but the `WebContentsView` that we're using does not, it's supposed to be on the `webContents` object. They're misusing the schema of where the function is supposed to be, so I'm correcting it with this patch.

marianunez (approved):
@devinbinnie is there an issue where this was reported in Electron? If no, we should file or contribute back to avoid diverging too much

devinbinnie:
> @devinbinnie is there an issue where this was reported in Electron? If no, we should file or contribute back to avoid diverging too much

This isn't an Electron library, and given the number of headaches I'm thinking we could just build something in-house: https://www.electronjs.org/docs/latest/api/menu

hmhealey (approved):

PR #3316, opened 2025-02-07
Spin server when Tests start

yasserfaraazkhan:
```release-note
NONE
```


devinbinnie (approved):

.github/actions/install-os-dependencies/action.yaml
hmhealey, line 1: Should we be doing anything to pin versions of dependencies? I feel like that's not often done for these sorts of things, so it's probably fine either way, but I thought I'd at least mention it

hmhealey (approved):
Looks great!

Issue #3333, opened 2025-02-14
Updating to the Version 5.11.0 created an issue


edwardboony:
When I updated today to the new version, it provided me with the white screen and nothing else, except basic menus. Couldn't message anyone. The only resolution that worked was rolling back to the previous version.

Steps to recreating the issue:

Download the latest update either from Mattermost itself or their Github page

![Image](https://github.com/user-attachments/assets/d23be1fc-ea92-4311-9b7e-b45c8248fe6a)

tkvisol:
I think it is due to version skew. See table: https://docs.mattermost.com/about/mattermost-desktop-releases.html#latest-releases

maxklock:
Same issue here. clearing all data like @devinbinnie said, didn't work for me.

devinbinnie:
> Same issue here. clearing all data like [@devinbinnie](https://github.com/devinbinnie) said, didn't work for me.

Can you go to View > Developer Tools for Current Server and show me the results from the Console there?

Fukurokudzu:
![Image](https://github.com/user-attachments/assets/f3a68b32-8bce-4c83-bbea-d97443c9c471)
![Image](https://github.com/user-attachments/assets/025a8518-b17e-43e1-adcf-bdcbe811c1ea)

jasonmacer:
Same issue here. I completely wiped all folders and re-installed from scratch and still had this on Windows 10 and Windows 11 machines.

Manually uninstalled all then went back to 5.10.x.

Artem-thenewage:
Same issue here. MacOS 15.3. Using Clear Data for Current Server doesn't solve the issue.

muromtsev:
Same thing. White screen..
Had to roll back to previous version 5.10..

maxklock:
> > Same issue here. clearing all data like [@devinbinnie](https://github.com/devinbinnie) said, didn't work for me.
> 
> Can you go to View > Developer Tools for Current Server and show me the results from the Console there?

Elements:
```
<body class="font--open_sans enable-animations">
	<div id="root" class="app-bar-enabled multi-teams">...</div>
	<div id="root-portal"></div>
	<noscript>To use Mattermost, please enable JavaScript.</noscript>
	<div></div>
[...]

PR #3348, opened 2025-02-20
[MM-63224] Add incompatible server screen

devinbinnie:
#### Summary
To help ease the transition for some users on older server versions that have run into compatibility issues with the newest Desktop App version, we have implement a screen with a check for the server version, that will show up when the user tries to use the Desktop App with an older version. This screen also includes a link to a forum post that can help users or system admins upgrade their servers.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63224
Closes https://github.com/mattermost/desktop/issues/3333

#### Device Information
This PR was tested on: macOS Sonoma

#### Screenshots
Screenshot 2025-02-20 at 2 14 09 PM
Screenshot 2025-02-20 at 2 13 57 PM

#### Release Note
```release-note
Add incompatible version screen
```


esethna (changes requested):
Thanks Devin, looks like it's replacing "Mattermost" with "Electron" incorrectly

image

devinbinnie:
> Thanks Devin, looks like it's replacing "Mattermost" with "Electron" incorrectly
> 
> image

Sorry, that's there for whitelabelling. In the built version it will say "Mattermost". I'll replace the screenshots.

src/common/constants.ts
```diff
@@ -48,3 +48,4 @@ export const DEFAULT_HELP_LINK = 'https://docs.mattermost.com/guides/collaborate
 export const DEFAULT_ACADEMY_LINK = 'https://academy.mattermost.com/';
 export const DEFAULT_TE_REPORT_PROBLEM_LINK = 'https://mattermost.com/pl/report-a-bug';
 export const DEFAULT_EE_REPORT_PROBLEM_LINK = 'https://support.mattermost.com/hc/en-us/requests/new';
+export const DEFAULT_UPGRADE_LINK = 'https://forum.mattermost.com/t/mattermost-desktop-app-5-11-important-compatibility-notice/22599';
```
lieut-data, line 51: Should we use a permalink here instead?

src/main/views/MattermostWebContentsView.ts
```diff
@@ -426,6 +428,14 @@ export class MattermostWebContentsView extends EventEmitter {
 
     private loadSuccess = (loadURL: string) => {
         return () => {
+            const serverInfo = ServerManager.getRemoteInfo(this.view.server.id);
+            if (serverInfo?.serverVersion && semver.lt(serverInfo.serverVersion, '9.4.0')) {
+                MainWindow.sendToRenderer(LOAD_INCOMPATIBLE_SERVER, this.id, loadURL.toString());
+                this.emit(LOAD_FAILED, this.id, 'Incompatible server version', loadURL.toString());
+                this.status = Status.ERROR;
+                return;
+            }
```
lieut-data, line 437: Should this logic default to showing the error /unless/ we successfully get the remote info and confirm it's `>= 9.4.0`? Worried about edge cases with old servers that we won't know about vs. confidence in the server being new enough for this request and check to work reliably.

lieut-data (approved):
Two thoughts, but non-blocking and will defer to you @devinbinnie.

src/main/views/MattermostWebContentsView.ts
devinbinnie: Yeah that makes sense, the only thing I worry about is if we somehow don't get the configuration from the API and then they get the incompatible screen on a seemingly working server. I think though if we're somehow ending up not getting that info into the app that would be a different bug we'd want to expose, so let's go your way.

src/common/constants.ts
esethna: Yes for sure, we should pl so we can redirect as needed in the future

src/main/views/MattermostWebContentsView.ts
esethna: If we default it to show, what happens if the user is on a spotty connection, is there a chance the version check fails and returns the error screen locking the user out? @devinbinnie 

Also for QA, I propose we spot check this on a couple newer release versions to ensure we aren't getting a false positive and displaying the error screen, especially on the v9.5 and v9.11 ESRs cc// @yasserfaraazkhan

devinbinnie: It would be a very small chance, but easily verifiable via refreshing. I think Jesse's point might cover more cases since it's clear we might have users on much much older versions.

esethna: For example: 
image

esethna: Other than a spotty connection are there other cases where we might show this screen incorrectly sending the user down the wrong path (downgrading) to resolution ? 

For example:
- Server is unreachable momentarily
- Server is down for maintenance
etc

Worried we might be showing this screen in way too many cases that result in the user spending energy on unhelpful resolution steps that will make them more frustrated

devinbinnie: So to clarify, the only case where we'd be showing this screen is if with the two requests made to the server that check for version and just hit the homepage, if the version check one fails but the homepage one succeeds and is able to load the app. They should happen within milliseconds of each other.

The app makes somewhere around 50 requests to load the app, and if any one of them were to fail, the app would be in a degraded state. I think we might be trying to cover too many potential cases here, and not just focusing on the main issue of the server upgrade issue.

Remember that this needs to go into a patch release, so we want to minimize the amount of checks needed here. I think what we have is more than sufficient and will be plenty robust for all users.

esethna: > So to clarify, the only case where we'd be showing this screen is if with the two requests made to the server that check for version and just hit the homepage, if the version check one fails but the homepage one succeeds and is able to load the app.

Got it, thanks

esethna (approved):

larkox (approved):
LGTM

devinbinnie:
@yasserfaraazkhan Going to defer QA since this will be tested with the dot release

PR #3356, opened 2025-02-26
Reload server info on retry

devinbinnie:
#### Summary
For the incompatible view, if your server is unavailable when the app loads, after you reload it will show the incompatible screen until you switch back from another server. This PR stops that.

```release-note
NONE
```


src/main/views/MattermostWebContentsView.ts
```diff
@@ -411,7 +412,10 @@ export class MattermostWebContentsView extends EventEmitter {
             getServerAPI(
                 parsedURL,
                 false,
-                () => this.reload(loadURL),
+                async () => {
+                    await updateServerInfos([this.view.server]);
+                    this.reload(loadURL);
+                },
```
larkox, line 418: Do we want to add tests for this behavior?

larkox (approved):

mgdelacroix (approved):

PR #3394, opened 2025-04-14
Fix: Build Mac no dmg job failing due to MACOS_NOTIFICATION_STATE_NO_SDK_CHECK

yasserfaraazkhan:
this PR https://github.com/mattermost/desktop/pull/3393 had the Macos build job failing. 
I had added a temprory Fix to skip check. 

The error we were seeing:

xcodebuild: error: SDK "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" cannot be located.
This is because the active developer path is set to the Command Line Tools, which don’t include full SDKs like MacOSX.sdk.

With this fix the full SDK is used

```release-note
NONE
```


hmhealey (approved):
I don't know much about fighting with Xcode, but this looks like it makes sense to me

PR #3406, opened 2025-04-24
Fix bad path name for user data dir, upgrade node-abi

devinbinnie:
#### Summary
Found an issue causing the Windows E2E tests to not run, this PR should fix it.

```release-note
NONE
```


yasserfaraazkhan:
Thank you Devin this should work now :D

yasserfaraazkhan:
Hi @toninis @saturninoabril 
https://github.com/mattermost/desktop/actions/runs/14749804652/job/41404418955?pr=3406#step:10:272

We been facing this gpu error in linux runners. Tried few things, but nothing fixed it.
any chance you might have seen this error somewhere? 

If there's no solution, I'll disable the linux tests

PR #3409, opened 2025-04-25
[MM-63899] Stop blocking of server loading on contact with each configured server

devinbinnie:
#### Summary
A change made to support the unsupported server screen caused issues when users had servers that were not reachable, or taking a long time to timeout/respond. This was caused by blocking loading any servers until the app had updated server info from every server.

This PR changes the logic to only block on the current server while its loading, while allowing each other server to load in the background. We still block each server to load on its own server info gathering, but not on each others.

NOTE: This should be backported back to v5.11 as well.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63899

```release-note
Stop blocking of server loading on contact with each configured server
```

### DIFF
--- a/src/main/views/viewManager.ts
+++ b/src/main/views/viewManager.ts
@@ -100,11 +100,21 @@ export class ViewManager {
         if (ServerManager.hasServers()) {
             // TODO: This init should be happening elsewhere, future refactor will fix this
             ServerViewState.init();
-            await updateServerInfos(ServerManager.getAllServers());
             LoadingScreen.show();
-            ServerManager.getAllServers().forEach((server) => this.loadServer(server));
+
+            // We need to wait for the current server to be initialized before showing anything
+            // But we can initialize other servers in parallel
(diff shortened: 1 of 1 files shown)


src/main/views/viewManager.ts
```diff
@@ -100,13 +100,23 @@ export class ViewManager {
...
             ServerViewState.init();
-            await updateServerInfos(ServerManager.getAllServers());
             LoadingScreen.show();
-            ServerManager.getAllServers().forEach((server) => this.loadServer(server));
+
+            // We need to wait for the current server to be initialized before showing anything
+            // But we can initialize other servers in parallel
+            const otherServers = ServerManager.getAllServers().filter((server) => server.id !== ServerViewState.getCurrentServer().id);
+            const currentServer = ServerViewState.getCurrentServer();
+            otherServers.forEach((server) => this.initServer(server));
```
agarciamontoro, line 109: Non-blocking nit: I have not tested this, but it seems simpler to me:
```suggestion
            otherServers.forEach(this.initServer);
```

marianunez (approved):

Issue #3377, opened 2025-03-19
[Bug]: Message input loose focus when hovering a link on the channel


Atthis:
### Checks before filing an issue

- [x] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [x] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing. This should include closed issues.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [x] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [x] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

Version 5.11.2 commit : 831f0c7

### Operating System

Ubuntu Linux 24.04 LTS x64

### Mattermost Server Version

10.5.1

### Steps to reproduce

When on a channel with links send, start typing a message into the input.
Hover a link (or its open graph preview).

### Expected behavior

The focus must stay on the message input.

### Observed behavior

As soon as I hover a link/open graph on the channel, the focus on the message input is lost.

This behaviour is not present when hovering an image or a embeded video.

### Log Output

```shell
[2025-03-17 08:48:16.933] [info]  [App.Config] config.autostart has been configured: true
[2025-03-17 08:48:17.127] [info]  [App.Initialize] Autoupgrade disabled: false
[2025-03-17 08:48:17.279] [info]  [ServerDropdownView] init
[2025-03-17 08:48:17.381] [info]  [i18nManager] Set new language fr
[2025-03-17 08:48:17.627] [info]  [App.Config] config.autostart has been configured: true
[2025-03-17 08:48:19.420] [info]  [App.Config] config.autostart has been configured: true
[2025-03-18 09:09:54.347] [error] Logger Log level set to: info
[2025-03-18 09:09:54.833] [warn]  [App.Initialize] Current working directory is /home/cedric, changing into /opt/Mattermost
[2025-03-18 09:09:55.180] [info]  [App.Config] config.autostart has been configured: true
[...]

PR #3453, opened 2025-06-27
MM-62005 Change UrlView to always stay mounted

hmhealey:
#### Summary
Unfortunately, the previous fix isn't enough to stop all cases where the UrlView steals focus when it's added by hovering a link. Instead of doing that though, let's just keep it mounted all of the time and only make it visible when needed.

From my testing in Parallels, it seems to fix the issue on both Windows and Ubuntu

#### Ticket Link
MM-62005
Closes https://github.com/mattermost/desktop/issues/3377

#### Checklist
- [x] Run E2E tests by adding label `Run Desktop E2E Tests`

#### Device Information
This PR was tested on: Windows 11 21H2, Ubuntu 24.04 LTS

#### Screenshots
Comparisons of before/after:
https://github.com/user-attachments/assets/102baa89-15ec-4b47-b2c0-2500888c0042
https://github.com/user-attachments/assets/5f53fb45-79ec-4ceb-bd55-237e6822edc5

#### Release Note
```release-note
Fixed URL preview stealing focus from app
```

### DIFF
--- a/src/main/preload/internalAPI.js
+++ b/src/main/preload/internalAPI.js
@@ -100,6 +100,7 @@ import {
     LOAD_INCOMPATIBLE_SERVER,
     OPEN_SERVER_UPGRADE_LINK,
     OPEN_CHANGELOG_LINK,
+    UPDATE_TARGET_URL,
 } from 'common/communication';
 
 console.log('Preload initialized');
@@ -188,6 +189,7 @@ contextBridge.exposeInMainWorld('desktop', {
     onUpdateDownloadsDropdown: (listener) => ipcRenderer.on(UPDATE_DOWNLOADS_DROPDOWN, (_, downloads, darkMode, windowBounds, item) => listener(downloads, darkMode, windowBounds, item)),
     onAppMenuWillClose: (listener) => ipcRenderer.on(APP_MENU_WILL_CLOSE, () => listener()),
     onFocusThreeDotMenu: (listener) => ipcRenderer.on(FOCUS_THREE_DOT_MENU, () => listener()),
+    onUpdateTargetURL: (listener) => ipcRenderer.on(UPDATE_TARGET_URL, (_, url) => listener(url)),
     updateURLViewWidth: (width) => ipcRenderer.send(UPDATE_URL_VIEW_WIDTH, width),
     openNotificationPreferences: () => ipcRenderer.send(OPEN_NOTIFICATION_PREFERENCES),
     openWindowsCameraPreferences: () => ipcRenderer.send(OPEN_WINDOWS_CAMERA_PREFERENCES),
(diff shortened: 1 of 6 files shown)


pvev (approved):
Code-wise looks good to me, but I'm not sure how exactly reproduce the issue for validating it does what is intended in the code without missing anything or adding a regression, but the code looks solid enough that I approve without having actually reproduced the bug and confirmed this code solves it.

src/main/preload/internalAPI.js
```diff
@@ -188,6 +189,7 @@ contextBridge.exposeInMainWorld('desktop', {
     onUpdateDownloadsDropdown: (listener) => ipcRenderer.on(UPDATE_DOWNLOADS_DROPDOWN, (_, downloads, darkMode, windowBounds, item) => listener(downloads, darkMode, windowBounds, item)),
     onAppMenuWillClose: (listener) => ipcRenderer.on(APP_MENU_WILL_CLOSE, () => listener()),
     onFocusThreeDotMenu: (listener) => ipcRenderer.on(FOCUS_THREE_DOT_MENU, () => listener()),
+    onUpdateTargetURL: (listener) => ipcRenderer.on(UPDATE_TARGET_URL, (_, url) => listener(url)),
```
devinbinnie: Can we change the name of this and use a new constant? This one is used by the WebContentsView object to tell the ViewManager that the URL is updated when the user actually hovers the URL. This might change in the new architecture, but I like to keep these relatively atomic (despite some evidence to the contrary :P)

Something like `SET_URL_FOR_URL_VIEW` maybe?

src/main/views/viewManager.ts
```diff
@@ -110,13 +111,33 @@ export class ViewManager {
...
+
+        this.initURLView();
     };
 
     private initServer = async (server: MattermostServer) => {
         await updateServerInfos([server]);
         this.loadServer(server);
     };
 
+    private initURLView = () => {
```
devinbinnie, line 124: This is definitely a good case for pulling this out into its own singleton. Not something for you to do here - but just a note for me to do later :P

src/main/views/viewManager.ts
```diff
@@ -372,35 +404,24 @@ export class ViewManager {
...
-            // @ts-expect-error Using an undocumented event that fires last when the URL view pops up
-            urlView.webContents.once('ready-to-show', () => {
-                log.debug('URL view focus prevented');
-                this.getCurrentView()?.focus();
-            });
-
-            MainWindow.get()?.contentView.addChildView(urlView);
+
+            if (this.urlView && !this.isViewInFront(this.urlView)) {
+                log.debug('moving URL view to front');
```
devinbinnie: Make this a `silly`, this will definitely get in the way of `debug` logs in the future, since this can write a LOT of logs to the file.

hmhealey: While testing, this didn't seem to actually trigger very often. It only happens once after changing servers or opening a menu. I can change that though

devinbinnie:
We can have Yasser test this at release time.

PR #3457, opened 2025-07-10
MM-62838: Desktop app not preventing invalid URLs from being added as a server

catalintomai:
#### Summary
Desktop app not preventing invalid URLs from being added as a server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62838

#### Release Note
```release-note
NONE
```


devinbinnie (changes requested):
@catalintomai I don't think this is the correct fix. The issue seems to be that the app doesn't recognize `**` to be an invalid URL component, instead trying to add `https://` to the front and saying it's not a Mattermost server.

We don't necessarily want to block a server that can't be verified from being added, for a number of reasons. So the fix here should be to fix the URL validation itself, I'm guessing that's where the issue is.

src/common/utils/url.ts
```diff
@@ -11,6 +11,17 @@ export const parseURL = (inputURL: string | URL) => {
     if (inputURL instanceof URL) {
         return inputURL;
     }
+
+    // Check for invalid URL characters that shouldn't be in a valid server URL
```
devinbinnie: Regexes are considered generally unsafe when trying to parse URLs, since they are susceptible to attacks. It looks like this code is also just checking for a specific example of `**`, which doesn't really make sense, since I think we should be checking for the actual invalid characters.

I think we need a deeper dive here, not just a patch job, since there is a library we use to check for invalid URLs here but seems to be failing to catch these characters. Perhaps we aren't using it properly, or it is missing this logic.

devinbinnie: Also, make sure to run `npm run check` before you submit so that it passes the linter/type checker/unit tests ;)

devinbinnie: For reference, this is the invalid URL code: https://github.com/mattermost/desktop/blob/master/src/common/utils/url.ts#L25

Which we use here to validate the URL: https://github.com/mattermost/desktop/blob/ac8b62a446fdbc8666c5cda8bd5f150621659016/src/app/serverViewState.ts#L264

catalintomai: @devinbinnie , re: "which doesn't really make sense" , obv. agree, started experimenting with a different approach. Moved this to draft until I converge to something that should address the issue differently.

At first look, the issue doesn't seem to be with the library though, but with the flow and `parseURL`. Meaning, `isValidURL('**')` returns `false` , but then `parseURL('https://**')` returns a valid URL object. 

Also, looking more into it, since I don't have the historical context (feels like we try to "fix" invalid URLs - like we should prob. prevent having input like `**` converted to `https://**` to begin with), do you know why we are not calling `isValidURL` as the last validation step of the flow?  (and instead use `parseURL` which seems to be too permissive)? Thanks!

devinbinnie: I think what's happening is that `isValidURL('**')` returns `false`, but `isValidURI('**')` returns `true`, which doesn't seem to be accounting for special characters, or it might be that URIs are more lenient here.

We do this to basically see if the base of the URL is valid and then fill in `https://` for the user, but maybe that's a bad assumption to make? If so, then we should modify that flow to be stricter. Maybe the easiest way would be to add the `https://` and then once again check that it's a valid URL and if not we throw the error.

catalintomai: Or we can keep the flow as it is and try to handle the special chars the last step by augmenting the `parseUrl` functionality - since this is the function that validates the final URL after all modifications - smth. like the latest commit. The thing is `isValidURL('**')` is not valid but `isValidURL('http://**'`) is valid (according to `valid-url`, which I presume validates the URL standard), which means that a secondary/final `isValidURL` invocation would not work. Still , we need to understand why (and should) we need to do smth. different from what `valid-url` does.

devinbinnie: In that case, I think I would augment `isValidURL` then instead of `parseURL`, since we don't need to run the validation every time, just on user input entry. I guess it doesn't really hurt to do it everytime, but I do know that `parseURL` is used a lot more.

As an aside, I'm wondering if there's just a better URL validation library out there. I think I looked a year ago, give or take but not more recently.

catalintomai: re: "I'm wondering if there's just a better URL validation library out there", probably is, but I assume they all implement the same standard/RFC. The main q that I have coming out of all this,  is, why do we have to do smth. different that what the library validates?

Going back tot the ticket, it says : "Enter an invalid URL such as `**`"
"a. An error displays `The server URL provided does not appear to point to a valid Mattermost server. Please verify the URL and check your connection.`
b. The Add button is still clickable, and the invalid server is added"

So I see at least two issues:
1) If the user would've entered `http://**` instead of `**`, would that've been ok? Library says `http://**` is valid according to the standard, so would it be ok to MM? If yes, then the issue is that we are trying to "fix" the input and make it "valid", which the library then validates positively. If no, then the issue is that we are going beyond the standard and want to validate some other URLs that "we think" as invalid but are actually valid according to the standard.

2) the ticket seems to call out the behavior delta between the "to point to a valid Mattermost server." and the "Add button is still clickable". The first commit fix was trying to align the two, making sure the url is a valid MM one.
However, what we seem to be saying is that we need to have the URL "valid" (according to the library AND our criteria of "invalidity", which are not well defined). If that is the goal then I guess we'll also need to change the error text, since an invalid URL is also an invalid MM URL, but an invalid MM URL is not necessarily an invalid URL.
cc: @lindalumitchell

lieut-data:
Is it worth just closing out the original Rainforest report? It doesn't feel terribly high impact -- presumably it doesn't work when you add a bad server anyway, so it's at best mildly better to test the URL upfront.

src/common/utils/url.ts
devinbinnie: My confusion is purely is why the `valid-url` library thinks that `https://**` is a valid URL, because if I type that into any major browser, it won't even let me navigate there (just sends me to Google).

Because of this I suspect that the library was kind of just a *best effort* at one time where someone needed a URL validation library, as I'm pretty sure that package is old. I would assume newer ones exist since this isn't exactly a new problem.

Responding as well to @lieut-data's question here, this is honestly really low impact, and is a major edge case where someone might type that in. The worst thing that happens is that you get a server that just doesn't work, the app doesn't crash or anything. This is starting to feel like a HW ticket, so I'd also be okay closing it.

My opinion is definitely that we shouldn't be trying to just check for everything ourselves, there should be a "correct standard" here somewhere, be it in the form of a library, or if we are to implement it ourselves at least some reference point.

lieut-data:
Going to make the call and propose we "Won't fix" this based on the conversation above. Apologies for the thrash, @catalintomai, and appreciate the support :)

PR #3460, opened 2025-07-15
[MM-56389] Switch to `registry-js` native module for Windows registry reading

devinbinnie:
#### Summary
We've had a persistent problem with `winreg`, our library for registry reading in scenarios where organizations may limit use of registry editing or the command prompt via Group Policy. Both these options break `winreg`'s registry reading as they limit the use of `reg.exe`, a tool provided with Windows.

This PR switches us over to use a native library called `registry-js` that implements native calls to read the registry without using other tools.

An interesting side-effect: using this changed the behaviour of Electron's `nativeTheme` module, in which now we have to check after the app is ready.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56389

```release-note
Windows: Switch to `registry-js` native module for Windows registry reading
```


lieut-data (commented):
Looks good, @devinbinnie! I didn't quite follow why this was true:

> An interesting side-effect: using this changed the behaviour of Electron's nativeTheme module, in which now we have to check after the app is ready.

Can you elaborate for posterity?

hmhealey (approved):

PR #3463, opened 2025-07-18
[MM-64617] Add enter key press handler to each of the modals

devinbinnie:
#### Summary
When we added the new modals, we neglected to account for how Enter key presses work with the new Modal component. For most of them, it would just hide the modal and not actually close it.

This PR adds the handlers to bind most of them to just their confirm button options, and also makes a change in the component itself to not hide unless the confirm button is enabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64617

```release-note
Fixed an issue where the Enter key would not confirm most modals
```


hanzei (approved):

yasserfaraazkhan (approved):
hitting enter after entering details will close modal and have the server loaded

PR #3473, opened 2025-07-28
E2e jobs with server

yasserfaraazkhan:
Duplicate of pr https://github.com/mattermost/desktop/pull/3466

```release-note
NONE
```


.github/workflows/e2e-functional.yml
```diff
@@ -26,75 +24,114 @@ on:
...
-        default: "10.5.0"
+        required: true
 
 jobs:
-  determine-platforms:
+  update-initial-status:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        platform: ['linux', 'macOS', 'windows']
```
saturninoabril: Should this come from `matrix.platform`?

e2e/webpack.config.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
         new webpack.DefinePlugin({__IS_MAC_APP_STORE__: false}),
     ],
     externals: {
+        'registry-js': 'commonjs2 registry-js',
```
devinbinnie: I think this can just follow the same pattern as the other ones, though I'm surprised it's needed here since the other ones don't trigger a problem.
```suggestion
        'registry-js': 'require("registry-js")',
```

devinbinnie:
> thank you @devinbinnie I made the suggested change. That registry.node error is back 😢 .
> 
> should I disable cache ? I wonder if it'll cause issue after merge.
> 
> ```
> 
> FAIL src/main/views/serverDropdownView.test.js
>   ● Test suite failed to run
> 
>     \\?\D:\a\desktop\desktop\node_modules\registry-js\build\Release\registry.node is not a valid Win32 application.
>     \\?\D:\a\desktop\desktop\node_modules\registry-js\build\Release\registry.node
> 
>        5 | import {EventEmitter} from 'events';
> ```

Yeah we can disable the cache at least for this PR, I don't know why that is consistently failing for you.

PR #3476, opened 2025-07-30
[MM-64949] Ignore invalid URIs and show a dialog when a user tries to open them

devinbinnie:
#### Summary
When a user tries to open an invalid URL, we used to redirect to the browser. However, doing so now causes the app to crash, likely a change in Electron.

This PR instead just checks for URL validity and shows a dialog box if we can't open the URL.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64949

```release-note
Ignore invalid URIs and show a dialog when a user tries to open them
```

### DIFF
--- a/src/main/views/webContentEvents.test.js
+++ b/src/main/views/webContentEvents.test.js
@@ -148,6 +148,7 @@ describe('main/views/webContentsEvents', () => {
 
         it('should deny and show dialog on invalid URL', () => {
             expect(newWindow({url: 'https://google.com/?^'})).toStrictEqual({action: 'deny'});
+            expect(newWindow({url: 'https://example.com/path}'})).toStrictEqual({action: 'deny'});
             expect(shell.openExternal).not.toBeCalled();
             expect(dialog.showErrorBox).toBeCalled();
         });


harshilsharma63 (approved):

src/main/views/webContentEvents.ts
```diff
@@ -121,6 +122,20 @@ export class WebContentsEventManager {
                 return {action: 'deny'};
             }
 
+            // Check for valid URL
+            // Let the browser handle invalid URIs
```
agarciamontoro: This comment does not seem accurate, if I understood the goal of the PR correctly.

src/main/views/webContentEvents.test.js
```diff
@@ -140,6 +146,12 @@ describe('main/views/webContentsEvents', () => {
             expect(newWindow({url: 'a-bad<url'})).toStrictEqual({action: 'deny'});
         });
 
+        it('should deny and show dialog on invalid URL', () => {
```
enzowritescode, line 149: Can you add a test with `}`

PR #3489, opened 2025-08-20
[MM-65085] Pre auth headers

BenCookie95:
#### Summary
- Modify server configurations to add an advanced field for Pre auth header
- Inject the header for all requests to the saved server url
- Store the value in secureStorage with the server URL as the key
- If there is no keychain, the secret is stored in plain text
- We perform validation on the pre auth header by sending a request with your unsaved secret
- We refresh the server view when you edit the pre auth header and save it

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65573

#### Verifying Pre-Auth header
The header should be sent with every single request to the server it's assigned to.  You can view it in network requests by going to:
1. Click View in the toolbar
2. Click developer tools
3. Click Developer tools for current server
4. In the pop up window click network
5. Click any request to the server
6. Request headers should include a header called `X-Mattermost-Preauth-Secret` with it's assigned value.

#### Checklist

- [ ] Has UI changes
- [ ] Run E2E tests by adding label `E2E/Run`

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Add pre auth headers to the server modal and inject header into all requests to that server
```


src/main/app/initialize.ts
```diff
@@ -307,6 +326,16 @@ async function initializeAfterAppReady() {
     });
 
     ServerManager.reloadFromConfig();
+
+    // Initialize secure storage cache after servers are loaded
+    try {
```
BenCookie95, line 314: Claude analysis of when to initialize

Based on the analysis, the current placement in initializeAfterAppReady is correct. Here's why:

Current Placement Analysis:

Correctly placed in initializeAfterAppReady because:

  1. App lifecycle dependency - secureStorage.initializeCache() calls loadSecrets() which performs file I/O operations that should happen after the app is fully ready
  2. No early access needed - Secure storage isn't accessed during the critical startup path before app ready
  3. Performance benefit - Cache initialization is non-blocking for the main startup sequence
  4. Logical ordering - It happens after ServerManager.reloadFromConfig() (line 328), so servers are loaded before secure storage cache initializes

  Why NOT initializeBeforeAppReady:

  - File I/O risk - Reading encrypted files before app ready could potentially block startup
  - Dependencies - Secure storage uses app.getPath('userData') which is safer to access after app ready
  - No urgency - Secure storage operations are lazy-loaded via getSecretsFromCache()

src/main/app/initialize.ts
```diff
@@ -330,6 +359,59 @@ async function initializeAfterAppReady() {
...
+                        const requestHeaders = {
+                            ...details.requestHeaders,
+                            'X-Mattermost-Preauth-Secret': secret,
+                        };
+
+                        callback({requestHeaders});
+                        return;
+                    }
+                } else {
+                    // Fallback: Check if this URL matches any server using normalized URL comparison
```
BenCookie95: Apparently there are cases where view can be undefined or just not found  so Claude insists on this fallback. It's a bit inefficient because it needs to search through all servers but it's supposed to be an edge case

src/types/config.ts
```diff
@@ -24,6 +24,13 @@ export type ConfigServer = Server & {
 export type UniqueServer = Server & {
```
devinbinnie, line 24: This type is actually for when the server is setup and assigned an id already. I think we might want an intermediary type here, something like `NewServer`. Another option would be to do the update in the same way we do edit server, ie. we do 2 separate calls from the renderer. I'm okay with either.

src/app/serverViewState.ts
```diff
@@ -367,6 +432,18 @@ export class ServerViewState {
         return ServerManager.getOrderedTabsForServer(serverId).map((view) => view.toUniqueView());
     };
 
+    private handleReloadView = (event: IpcMainInvokeEvent, viewId: string) => {
```
devinbinnie: We already have a method for reloading the server. When the server manager updates the URL, it emits a `SERVER_URL_MODIFIED` event which the `MattermostWebContentsView` listens for.

If we want to store the secret in memory (which we probably do), we might want to keep it on the `UniqueServer` type as well and store it on the `MattermostServer` and have it managed by the `ServerManager`. We can then have it have a similar event to reload the view.

src/types/window.ts
```diff
@@ -55,8 +55,9 @@ declare global {
             getLastActive: () => Promise<{server: string; view: string}>;
             getOrderedServers: () => Promise<UniqueServer[]>;
             getOrderedTabsForServer: (serverId: string) => Promise<UniqueView[]>;
+            reloadView: (viewId: string) => Promise<void>;
```
devinbinnie: This doesn't need to be exposed to the front-end

src/types/window.ts
```diff
@@ -160,6 +161,14 @@ declare global {
                     unreads?: Map<string, boolean>,
                 ) => void) => void;
             };
+
+            secureStorage: {
```
devinbinnie: I think we only need to expose the `getSecret` method here. It doesn't need its own section either.

src/renderer/components/NewServerModal.tsx
```diff
@@ -56,14 +60,18 @@ class NewServerModal extends React.PureComponent<Props, State> {
         this.wasShown = false;
         this.mounted = false;
         this.state = {
-            serverName: '',
-            serverUrl: '',
+            serverName: props.server?.name || '',
```
devinbinnie: I'm not sure why these were changed?

src/renderer/components/ConfigureServer.tsx
```diff
@@ -218,10 +238,29 @@ function ConfigureServer({
...
+        setPreAuthSecret(value);
+
+        // Trigger validation when pre-auth secret is entered and URL exists
+        if (url && !validating) {
+            clearTimeout(validationTimeout.current as unknown as number);
+            validationTimeout.current = setTimeout(() => {
+                if (!mounted.current) {
+                    return;
+                }
+                fetchValidationResult(url, value);
```
devinbinnie, line 318: I think this needs `editing.current`

src/main/app/initialize.ts
devinbinnie: We can remove this. The only reason `view.server` should be null would be in the cases where we use `getServerAPI`, but we're already loading the secret manually there.

Funny enough, this is actually made a lot better by my architecture changes, so once those are merged we should be good to just find the server here and avoid even changing `getServerAPI`.

src/main/app/initialize.ts
devinbinnie: Yep, that all makes sense. As long as the servers are load we're good to initialize it.

src/main/secureStorage.ts
```diff
@@ -0,0 +1,211 @@
...
+            encrypted: this.encryptionAvailable,
+            available: true,
+            warning: this.encryptionAvailable ? undefined : ENCRYPTION_UNAVAILABLE_WARNING,
+        };
+    }
+}
+
+let secureStorageInstance: SecureStorage | null = null;
+
+export function getSecureStorage(userDataPath: string): SecureStorage {
```
devinbinnie: This needs to follow the correct pattern in case we change the `userData` path during operation (yes you can do that). Here's an example from `trustedOrigins`:

```
const trustedOriginsStore = new TrustedOriginsStore(trustedOriginsStoreFile);
export default trustedOriginsStore;

ipcMain.on(UPDATE_PATHS, () => {
    log.debug('UPDATE_PATHS');
    trustedOriginsStore.storeFile = trustedOriginsStoreFile;
    if (trustedOriginsStore.data) {
        trustedOriginsStore.load();
    }
});
```

src/main/app/config.ts
```diff
@@ -116,3 +117,60 @@ export function handleDarkModeChange(darkMode: boolean) {
 
     ipcMain.emit(EMIT_CONFIGURATION, true, Config.data);
 }
+
+//
+// secure storage event handlers
+//
+
+export async function handleSecureStorageGet(event: Electron.IpcMainInvokeEvent, serverUrl: string, keySuffix: string): Promise<string | null> {
```
devinbinnie: This should all be done in `secureStorage` as well.

src/main/app/initialize.ts
```diff
@@ -282,6 +294,13 @@ function initializeInterCommunicationEventListeners() {
 
     ipcMain.on(TOGGLE_SECURE_INPUT, handleToggleSecureInput);
 
+    // Secure storage handlers
+    ipcMain.handle(SECURE_STORAGE_GET, handleSecureStorageGet);
```
devinbinnie: These should be handled in the `secureStorage` singleton class.

devinbinnie: Also we can likely remove most of these listeners and constants, since I think only `SECURE_STORAGE_GET` is being used.

src/app/serverViewState.ts
```diff
@@ -289,13 +329,38 @@ export class ServerViewState {
             return {status: URLValidationStatus.URLExists, existingServerName: existingServer.server.name, validatedURL: existingServer.server.url.toString()};
         }
 
+        // Try to retrieve saved pre-auth secret for this URL
+        // Priority logic:
+        // - preAuthSecret === null: Use saved secret (initial validation)
```
devinbinnie: This seems like an incredibly unnecessary amount of input validation. Either the `preAuthSecret` should be there, or not. We can probably just check if it's `string | undefined`, it's not going to be `null` I don't think.

src/app/serverViewState.ts
```diff
@@ -307,7 +372,7 @@ export class ServerViewState {
         if (!remoteInfo) {
             // If the URL provided has a path, try to validate the server with parts of the path removed, until we reach the root and then return a failure
             if (parsedURL.pathname !== '/') {
-                return this.handleServerURLValidation(e, parsedURL.toString().substring(0, parsedURL.toString().lastIndexOf('/')), currentId);
+                return this.handleServerURLValidation(e, parsedURL.toString().substring(0, parsedURL.toString().lastIndexOf('/')), currentId, effectivePreAuthSecret || undefined);
```
devinbinnie: If `effectivePreAuthSecret` is falsy, I don't think we need to force `undefined`.

src/types/config.ts
BenCookie95: I created a `NewServer` type but I feel like UniqueServer was being misused before if you say it's only for when the server is setup and assigned an id already because it was being used in the onSave function of ConfigureServer and NewServerModal which can be before a server has been created/setup. I debated just adding the `preAuthHeader` field to `Server` instead of adding another type but since it's used to construct the `ConfigServer` it didn't feel right even though the `toConfigServer()` function would have left it out of the config files

src/app/serverViewState.ts
BenCookie95: So I ended up adding the preauth header as a field to UniqueServer and the ServerManager now has the secret in memory so it's a bit easier to access and handle the reloading logic. I already had an in-memory cache in the secureStorage class which I decided to keep because while we aren't making much use of it for the preAuth header it could be useful down the line. 

I also ended up renaming `SERVER_URL_MODIFIED` to `SERVER_MODIFIED` instead of creating a second event for the header change. The reason being that it felt weird juggling 2 events that did the exact same thing, they both reloaded the view. The constant value was already `'servers-modified'` so there was some foreshadowing there lol.

src/app/serverViewState.ts
```diff
@@ -289,25 +330,55 @@ export class ServerViewState {
...
-        let remoteInfo = await this.testRemoteServer(secureURL);
-        if (!remoteInfo && insecureURL) {
-            // Try to fall back to HTTP
-            remoteInfo = await this.testRemoteServer(insecureURL);
-            if (remoteInfo) {
-                remoteURL = insecureURL;
+        let remoteInfo;
+        let preAuthRequired = false;
+
+        const httpsResult = await this.testRemoteServer(secureURL, effectivePreAuthSecret);
```
BenCookie95, line 318: I needed to change testRemoteServer to return errors instead of swallowing the error, 403 means a potential pre-auth header issue so we need to work around that

src/renderer/components/NewServerModal.tsx
```diff
@@ -129,22 +154,46 @@ class NewServerModal extends React.PureComponent<Props, State> {
...
+    validateServerURL = (serverUrl: string, preAuthSecret?: string) => {
         clearTimeout(this.validationTimeout as unknown as number);
         this.validationTimeout = setTimeout(() => {
             if (!this.mounted) {
                 return;
             }
             const currentTimeout = this.validationTimeout;
             this.setState({validationStarted: true});
-            window.desktop.validateServerURL(serverUrl, this.props.server?.id).then((validationResult) => {
+            window.desktop.validateServerURL(serverUrl, this.props.server?.id, preAuthSecret).then((validationResult) => {
```
BenCookie95, line 186: This temp value needs to be passed down the chain for the testRemote request

src/renderer/components/ConfigureServer.tsx
```diff
@@ -68,34 +68,63 @@ function ConfigureServer({
...
+    const [showAdvanced, setShowAdvanced] = useState(false);
+    const [preAuthSecret, setPreAuthSecret] = useState('');
+    const [preAuthSecretError, setPreAuthSecretError] = useState<{type: STATUS; value: string}>();
+    const [currentValidationStatus, setCurrentValidationStatus] = useState<string>();
+
+    // Basic form requirements
+    const hasBasicRequirements = name && url && !nameError && !validating;
+
+    // Determine if we can save based on validation status
+    const canSaveBasedOnValidation = () => {
```
BenCookie95, line 81: Validation is a bit more complex now with the extra field

src/types/config.ts
devinbinnie: The reason it was used there was that the `NewServerModal` is also the "edit server modal", meaning it was possible for an id to exist. It wasn't the ideal case, but it did work. This seems like an improvement on that case, so I'm happy we went this direction :+1:

src/app/serverViewState.ts
```diff
@@ -187,16 +201,31 @@ export class ServerViewState {
...
             {server: server.toUniqueServer(), permissions: PermissionsManager.getForServer(server) ?? {}},
             mainWindow);
 
-        modalPromise.then((data) => {
+        modalPromise.then(async (data) => {
             if (!server.isPredefined) {
                 ServerManager.editServer(id, data.server);
+
+                // Handle secure storage persistence separately
+                if ('preAuthSecret' in data.server) {
```
devinbinnie: We have this string validation duplicated - this is also in the ServerManager. We could probably just do it in the ServerManager, or if not we can pull it out into a helper.

src/main/secureStorage.ts
devinbinnie: This is still following a different pattern than a lot of our other modules. Everything in the app is just a singleton, so we instantiate it and export the one instance that's needed. If there is any other setup needed after the app is ready, we use an `init` function. I'd like this module to follow the same pattern.

You can take a look at the ServerManager or some modules to see how we do it.

src/app/serverViewState.ts
```diff
@@ -441,6 +518,21 @@ export class ServerViewState {
 
         if (!server.isPredefined) {
             ServerManager.editServer(server.id, server);
+
+            // Handle secure storage persistence separately
+            if ('preAuthSecret' in server) {
```
devinbinnie: Same here re: validation

src/app/serverViewState.ts
```diff
@@ -289,25 +326,55 @@ export class ServerViewState {
...
+                    }
+                }
+            } else if (httpsIs403) {
+                // No HTTP fallback available, but HTTPS returned 403
+                preAuthRequired = true;
             }
         }
 
+        // If we detected a 403 error, return PreAuthRequired status
+        if (preAuthRequired) {
```
devinbinnie, line 348: I don't know if it's a good idea to say that the pre-auth header is required purely based on a 403 error. There are many other reasons where pinging the server could return a 403 (eg. the server URL is wrong/there is some other system in between the user and Mattermost).

I think it makes sense to show some kind of error message here - but I don't think we should say "the pre-auth is required", since that's way too specific to this use case and may confuse users.

cc @matthewbirtch Can we come up with something that could suggest users to either contact their administrator, or that one possible solution is they need the pre-auth header?

src/main/app/initialize.ts
```diff
@@ -282,6 +285,12 @@ function initializeInterCommunicationEventListeners() {
 
     ipcMain.on(TOGGLE_SECURE_INPUT, handleToggleSecureInput);
 
+    // Secure storage handlers
+    ipcMain.handle(SECURE_STORAGE_GET, (event, serverUrl, keySuffix) => {
```
devinbinnie: This should be in the `secureStorage` module. You can add this listener in the constructor.

src/main/secureStorage.ts
```diff
@@ -0,0 +1,251 @@
...
+const ENCRYPTION_UNAVAILABLE_WARNING = 'Secure storage is not available on this system. Secrets will be stored in plain text. Consider installing keyring services for better security.';
+
+export class SecureStorage {
+    userDataPath: string;
+
+    // ts can't tell that updatePaths() initializes these in the constructor
+    private storageDir!: string;
+    private secretsPath!: string;
+    private plaintextSecretsPath!: string;
+    memoryCache: Record<string, string> | null = null;
```
devinbinnie: Nit: `undefined` vs. `null`?

src/app/serverViewState.ts
BenCookie95: Yeah true, mm could be behind a VPN or something. Right now if it 403s, we expand the advanced dropdown and show an error. The error message is `This server requires a pre-authentication header. Please provide the pre-authentication secret.` right now. It's a difficult balance to strike with prompting the user for this.

Would the following be better? `This server may require a pre-authentication header. Please contact your system administrator.`

The error is slightly different when you provide an invalid pre-auth value but the error could be misleading because their server might not even need the pre-auth header. `The pre-authentication header is invalid. Please check the secret value.`

For that particular case: `The pre-authentication header is invalid. Please contact your system administrator.`

src/app/serverViewState.ts
BenCookie95: Added some helper methods for storing and validating the secret

lindy65 (approved):
Thanks @BenCookie95 - tested on the PR's Windows artifact build as per verification steps in this PR as well as on the associated Jira ticket. Looks good to me 👍

src/app/serverViewState.ts
matthewbirtch: Answered this in this thread https://hub.mattermost.com/private-core/pl/kbhzp366c3gsukqd4cbnr6yx8a

devinbinnie (approved):

matthewbirtch:
A few things I noticed:

1. Let's get rid of 'Connect anyway' - just make it 'Connect'. Adds confusion for me unneccesarily

2. If I open 'advanced' for a server that doesn't require the secret, it still tells me it's a valid secret (regardless of what I enter)
    image

3. When I hover 'Advanced', it disappears in the hover state (only on the initial add server screen in the Light theme). Looks like @yasserfaraazkhan saw this too

4. Do we have a better way to present this if I use an incorrect secret, but still connect and save the server to the desktop app? Are we able to show a nicer page (similar to our could not connect to server screen)?
    image


**Possibly unrelated to this PR**
5. Not sure if this is a problem in this PR, but the error color is not readable here. And it's the wrong error color as compared to the other error color we've used for other error messages.
    image

6. The test build seemed really slow overall to me for some reason and seem to take a lot of CPU when adding servers and initially loading the server. Every desktop interaction feels quite unresponsive.
image

src/main/secureStorage.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    private async saveSecrets(secrets: Record<string, string>): Promise<void> {
+        try {
+            await this.ensureStorageDir();
+            const jsonData = JSON.stringify(secrets, null, 2);
+
+            if (this.encryptionAvailable) {
+                // Save encrypted
+                const encryptedData = safeStorage.encryptString(jsonData);
+                await writeFile(this.secretsPath, encryptedData);
```
enzowritescode: We should explicitly set file permissions

example:
```
await writeFile(this.secretsPath, encryptedData, {
     mode: 0o600
});
```

src/main/secureStorage.ts
```diff
@@ -0,0 +1,251 @@
...
+                log.debug('Secrets saved with encryption');
+            } else {
+                // Save as plaintext with warning
+                if (!this.hasWarnedAboutPlaintext) {
+                    log.warn(ENCRYPTION_UNAVAILABLE_WARNING);
+                    this.hasWarnedAboutPlaintext = true;
+                }
+                await writeFile(this.plaintextSecretsPath, jsonData, 'utf-8');
```
enzowritescode: Same as above.

example:
```
await writeFile(this.plaintextSecretsPath, jsonData, {
     encoding: 'utf-8',
     mode: 0o600
});
```

enzowritescode (changes requested):

enzowritescode (approved):

src/renderer/css/components/NewServerModal.scss
```diff
@@ -29,6 +29,38 @@
...
+        color: var(--button-bg);
+        cursor: pointer;
+        font-family: inherit;
+        font-size: 12px;
+        line-height: 16px;
+        font-weight: 600;
+        padding: 0;
+        
+        &:hover {
+            color: var(--button-bg-hover);
```
matthewbirtch: does this variable exist? I assume not which is why it's disappearing on hover

BenCookie95: oops, that was a mistake. I thought I cleared that out

PR #3491, opened 2025-08-20
[MM-64645][MM-64646][MM-64647][MM-64648][MM-64650][MM-64652][MM-61046] Multi-View Architecture

devinbinnie:
#### Summary
This PR contains a major overhaul to the UX and architecture of the Desktop App. Users are now able to open and manage multiple tabs and windows per server, up to a limit of 15 tabs (which is configurable). Tabs and windows live for the duration of the app, and are closed when the server is removed or they're manually closed. Tabs can be converted to windows and vice-versa via a right click menu. Users must have at least one tab open per server, which acts as the primary view and facilitates notifications.

Under the hood, a few new concepts are introduced. We keep the idea of a `ServerManager` to manage configuration of servers, acting as a single source of truth while the app is running. We now add a `ViewManager` to manage the view state, implying that the existence of a view means that it's open. There is a `TabManager` and `PopoutManager` responsible for tabs and windows respectively, managing their relationships with their corresponding windows. A `WebContentsManger` replaces the existing `ViewManager` to manage the lifecycle of the `WebContentsView` that holds the Mattermost web application itself. More information can be found here: 
https://mattermost.atlassian.net/wiki/spaces/ICU/pages/3524755458/Desktop+App+Multi-View+Architecture

Multiple new E2E and unit tests have also been added. Additionally, a few modules have been moved around to begin the transition towards a more correct Desktop App architecture, in-line with changes described here: https://mattermost-my.sharepoint.com/:w:/r/personal/devin_binnie_mattermost_com/_layouts/15/Doc.aspx?sourcedoc=%7BF95A385D-14C8-5D1A-54F2-41B47D53CF51%7D&file=Desktop%20App%20Architecture%20Refactor.docx&action=default&mobileredirect=true&DefaultItemOpen=1. This is consistent with the "calls up, events down" architecture that we employ here. Further changes will happen in this respect, but

This change massively overhauls the Desktop App's codebase, and will need to be broken down well to be reviews. Developer review should be mostly concerned with the above managers and their relationships, as well as the front-end React changes. 
Security should be taking a look at the new `NavigationManager`, which consolidates some of the tab routing code. Otherwise most navigation restrictions remain intact.
QA can have a look at the new E2E tests, we will need to assign test numbers to those.
UX can have a look at some of the CSS changes if need be.

Feel free to reach out to me on Mattermost for any questions.
This has been an effort years in the making and I am so happy to finally see it realized :rocket:

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64645
https://mattermost.atlassian.net/browse/MM-64646
https://mattermost.atlassian.net/browse/MM-64647
https://mattermost.atlassian.net/browse/MM-64648
https://mattermost.atlassian.net/browse/MM-64650
https://mattermost.atlassian.net/browse/MM-64652
https://mattermost.atlassian.net/browse/MM-61046

#### Device Information
This PR was tested on:
- Windows 11, x64
- macOS Sonoma, M4
- Ubuntu 24.04, x64

#### Screenshots
Screenshot 2025-08-20 at 1 33 34 PM
image
image

#### Release Note
```release-note
Add the ability to open multiple tabs and windows for any given server.
```


src/renderer/components/SettingsModal/components/NumberSetting.tsx
```diff
@@ -0,0 +1,71 @@
...
+    label,
+    value: propValue,
+    subLabel,
+    min,
+    max,
+    step = 1,
+    bottomBorder,
+    defaultValue,
+}: Props) {
+    const [value, setValue] = useState<number>(propValue ?? defaultValue);
```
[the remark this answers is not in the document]
devinbinnie: This works fine as designed.

src/renderer/components/TabBar.tsx
```diff
@@ -138,24 +141,24 @@ class TabBar extends React.PureComponent<Props, State> {
...
-                                            <button
-                                                className='serverTabItem__close'
-                                                onClick={this.onCloseTab(tab.id!)}
-                                            >
-                                                <i className='icon-close'/>
-                                            </button>
-                                        }
+                                        <span>{tab.title}</span>
+                                        {badgeDiv}
+                                        {this.props.tabs.length > 1 && !this.props.tabsDisabled && (
```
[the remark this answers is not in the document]
devinbinnie: Not repeated anywhere.

src/renderer/components/SettingsModal/components/NumberSetting.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const newValue = parseInt(event.target.value, 10);
```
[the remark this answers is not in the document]
devinbinnie: Can't happen with this input

src/renderer/components/MainPage.tsx
```diff
@@ -444,37 +467,31 @@ class MainPage extends React.PureComponent<Props, State> {
         }, false);
 
         const activeServer = this.state.servers.find((srv) => srv.id === this.state.activeServerId);
+        const tabStatus = activeServer && this.getTabViewStatus();
+        if (!tabStatus) {
+            if (this.state.activeTabId) {
+                console.error(`Not tabStatus for ${this.state.activeTabId}`);
```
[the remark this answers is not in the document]
devinbinnie: The front-end doesn't use Logger.

matthewbirtch (changes requested):
This is awesome work @devinbinnie. A few detailed things I found:

1. We do still need the dividers between tabs (noted [here](https://hub.mattermost.com/private-core/pl/87impnncc7nytcagwjx17urzyr)

2. If I add a lot of tabs, we start truncating them, but we don't seem to shrink them all uniformly. So in my screenshot below, we have the smallest tabs on the right and the widest tabs on the left. Additionally, it looks like the x is getting hidden for very small tabs, so there is no way to close it
    
    image

    image

3. I got this "Limit exceeded" view. Is that from the desktop app or from the webapp? Is that caused by the 15-tab limit or something else? It seems if we open too many browser tabs in chrome we get something similar. If I open too many tabs too quickly on hub, I get the login screen somehow and it thinks my session has expired. Sometimes I get the team selection page instead. Very strange inconsistent behavior.
    
    image

    image

    image


4. Can we increase the click area for the x button on tabs? It seems like I need to be very precise. I'm thinking we should add some padding around the actual icon that increases the hit area. 

5. If I have multiple tabs open, command-w should close the current tab, but it appears to close the whole desktop app window. Can we make it close the tab? (unless it's the last tab). If I have a new window open (rather than a tab) and that new window is active, command-w should close that active window (which it does properly). To that end, can we show the shortcuts on the context menus for these actions?

6. Command-T should open a new tab. Right now, it doesn't do anything.

7. Using Command-N to open a new window works well. However, it doesn't seem to have a limit. Should we make this follow the same limit as tabs?

8. In the window menu, we have `Create New Window for Current Server`. Feels a bit verbose. Can we remove 'Create' so it just reads `New Window for Current Server`

9. Command-clicking an internal link in MM works great. However, I can't do it on a channel in the LHS due to the multi-select function. I would vote for deprecating the command-click multi-select function in the LHS (but leave Shift-Click)

10. Command-clicking the `Playbooks` item from the product switcher seemed to fail loading the landing page. `Boards` seemed to work fine though

    image

src/app/menus/app.ts
```diff
@@ -152,22 +153,23 @@ export function createTemplate(config: Config, updateManager: UpdateManager) {
                 return 'Ctrl+Shift+I';
             })(),
             click() {
-                const mainWindow = MainWindow.get();
-                if (!mainWindow) {
+                // TODO: This is a bit unreliable, find a better way to get the focused window
```
enzowritescode: Does this need to be addresses before release?

src/app/serverHub.ts
```diff
@@ -4,134 +4,76 @@
...
 
-        const orderedServers = ServerManager.getOrderedServers();
-        if (orderedServers.length) {
-            if (Config.lastActiveServer && orderedServers[Config.lastActiveServer]) {
-                this.currentServerId = orderedServers[Config.lastActiveServer].id;
-            } else {
-                this.currentServerId = orderedServers[0].id;
-            }
-        }
+    // TODO: Move me somewhere else later
```
enzowritescode, line 64: Does this need to be addressed before release?

src/app/serverHub.ts
```diff
@@ -453,16 +328,23 @@ export class ServerViewState {
...
         log.debug('handleRemoveServer', serverId);
 
-        const remainingServers = ServerManager.getOrderedServers().filter((orderedServer) => serverId !== orderedServer.id);
-        if (this.currentServerId === serverId && remainingServers.length) {
-            this.currentServerId = remainingServers[0].id;
-        } else if (!remainingServers.length) {
-            delete this.currentServerId;
-        }
-
+        // Remove the server from ServerManager
```
enzowritescode, line 392: nitpick: this comment seems unnecessary

matthewbirtch (approved):
Fantastic work @devinbinnie! I love it. It looks like you've resolved the UX items and other than the _really minor_ nit on the spacing on the tab I mentioned in our DM, I think we're ready to rock here.

e2e/modules/environment.js
```diff
@@ -30,7 +30,7 @@ const downloadsFilePath = path.join(userDataDir, 'downloads.json');
 const downloadsLocation = path.join(userDataDir, 'Downloads');
 const boundsInfoPath = path.join(userDataDir, 'bounds-info.json');
 const appUpdatePath = path.join(userDataDir, 'app-update.yml');
-const exampleURL = 'http://example.com/';
+const exampleURL = 'https://mattermost.com/';
```
larkox: Intended change?

e2e/specs/server_management/long_server_name.test.js
```diff
@@ -12,7 +12,7 @@ describe('LongServerName', function desc() {
     this.timeout(30000);
     const config = env.demoConfig;
     const longServerName = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus quis malesuada dolor, vel scelerisque sem';
-    const longServerUrl = 'https://example.org';
+    const longServerUrl = 'https://web.archive.org';
```
larkox: Again, intended change?

e2e/specs/server_management/popout_windows.test.js
```diff
@@ -0,0 +1,223 @@
...
+    };
+
+    const afterFunc = async () => {
+        if (this.app) {
+            await this.app.close();
+        }
+        await env.clearElectronInstances();
+    };
+
+    describe('MM-TXXXX popout window functionality', async () => {
```
larkox, line 37: Is TXXXX intended?

e2e/specs/settings.test.js
```diff
@@ -205,7 +205,6 @@ describe('Settings', function desc() {
                 });
                 await settingsWindow.waitForSelector('#settingCategoryButton-advanced');
                 await settingsWindow.click('#settingCategoryButton-advanced');
-                console.log('balls');
```
larkox, line 208: 🤣

e2e/specs/server_management/tab_management.test.js
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    const afterFunc = async () => {
+        if (this.app) {
+            await this.app.close();
+        }
+        await env.clearElectronInstances();
+    };
+
+    describe('MM-TXXXX should be able to close server tabs', async () => {
```
larkox, line 31: Another TXXXX

src/app/callsWidgetWindow.ts
```diff
@@ -24,29 +29,26 @@ import {
...
-import webContentsEventManager from 'main/views/webContentEvents';
-import MainWindow from 'main/windows/mainWindow';
 
 import type {
     CallsJoinCallMessage,
     CallsWidgetWindowConfig,
 } from 'types/calls';
 
-import ContextMenu from '../contextMenu';
+import ContextMenu from '../main/contextMenu';
```
larkox: Nitpick: going to parent folders can be confusing. Since `main` is a folder in the root, we should probably just reference it from the root, instead of doing the `..`.

src/app/callsWidgetWindow.ts
```diff
@@ -116,18 +118,19 @@ export class CallsWidgetWindow {
     };
 
     getViewURL = () => {
-        return this.mainView?.view.server.url;
+        return this.mainView && WebContentsManager.getServerURLByViewId(this.mainView.id);
```
larkox, line 124: 0/5 nitpick: if we allow `getServerURLByViewId` to receive an undefined value, this could be slightly simpler.

src/app/callsWidgetWindow.ts
```diff
@@ -116,18 +118,19 @@ export class CallsWidgetWindow {
...
     };
 
     private getWidgetURL = () => {
-        if (!this.mainView) {
+        const serverURL = this.getViewURL();
+        if (!serverURL) {
             return undefined;
         }
-        const u = parseURL(this.mainView.view.server.url.toString()) as URL;
+        const u = parseURL(new URL(serverURL)) as URL;
```
larkox, line 136: Out of scope: In theory, `parseURL` can return undefined, so this cast is dangerous. Are we sure we want to keep the cast and not handle the undefined value?

src/app/callsWidgetWindow.ts
```diff
@@ -297,7 +300,11 @@ export class CallsWidgetWindow {
         if (!parsedURL) {
             return {action: 'deny' as const};
         }
-        if (isCallsPopOutURL(this.mainView?.view.server.url, parsedURL, this.options?.callID)) {
+        const serverURL = this.getViewURL();
+        if (!serverURL) {
+            return {action: 'deny' as const};
```
larkox, line 308: Out of scope: If you type the return value, you can probably avoid so many `as const`.

src/app/mainWindow/mainWindow.ts
```diff
@@ -2,66 +2,75 @@
...
+import ViewManager from 'common/views/viewManager';
 import {boundsInfoPath} from 'main/constants';
 import {localizeMessage} from 'main/i18nManager';
 import performanceMonitor from 'main/performanceMonitor';
 
 import type {SavedWindowState} from 'types/mainWindow';
 
-import ContextMenu from '../contextMenu';
-import {getLocalPreload, isInsideRectangle, isKDE} from '../utils';
+import {isInsideRectangle, isKDE} from '../../main/utils';
```
larkox: Double `../` is even worse than just going to the parent 😜

src/app/mainWindow/mainWindow.ts
```diff
@@ -332,17 +264,14 @@ export class MainWindow extends EventEmitter {
             return;
         }
 
-        globalShortcut.unregisterAll();
-
-        this.emit(MAIN_WINDOW_RESIZED, this.getBounds());
-        ipcMain.emit(TOGGLE_SECURE_INPUT, null, false);
```
larkox, line 338: Why do we remove this?

src/app/mainWindow/modals/modalView.ts
```diff
@@ -7,8 +7,8 @@ import {WebContentsView} from 'electron';
 import {Logger} from 'common/log';
 import performanceMonitor from 'main/performanceMonitor';
 
-import ContextMenu from '../contextMenu';
-import {getWindowBoundaries} from '../utils';
+import ContextMenu from '../../../main/contextMenu';
+import {getWindowBoundaries} from '../../../main/utils';
```
larkox: Three levels of `../`...

src/app/menus/app.ts
```diff
@@ -144,30 +145,21 @@ export function createTemplate(config: Config, updateManager: UpdateManager) {
...
-                const mainWindow = MainWindow.get();
-                if (!mainWindow) {
-                    return;
-                }
-
-                if (mainWindow.webContents.isDevToolsOpened()) {
-                    mainWindow.webContents.closeDevTools();
-                } else {
-                    mainWindow.webContents.openDevTools({mode: 'detach'});
-                }
```
larkox, line 164: This seems to have some toggle behavior that we are losing. Is that intended?

src/app/navigationManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    constructor() {
+        this.ready = false;
+
+        ipcMain.handle(REQUEST_BROWSER_HISTORY_STATUS, this.handleRequestBrowserHistoryStatus);
+        ipcMain.on(HISTORY, this.handleHistory);
+        ipcMain.on(BROWSER_HISTORY_PUSH, this.handleBrowserHistoryPush);
+    }
+
+    private openLinkInTab = (url: string | URL, getView: (server: MattermostServer) => MattermostView | undefined) => {
+        if (url) {
```
larkox: In order to make this function less deep, we can follow a more "go-like" approach. Like...

```
if (!url) {
    return;
}

const parsedURL = parseURL(url)!;
const server = ServerManager.loodupServerByURL(parsedURL, true);

if (!server) {
    if (ServerManager.hasServers()) {
        ...
    } else {
        ...
    }
    return;
}

...
```

(Interestingly enough, you follow this pattern inside the `if (server)` block 😅 )

src/app/navigationManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                }
+
+                const webContentsView = WebContentsManager.getView(view.id);
+                if (!webContentsView) {
+                    log.error(`Couldn't find a server for the view ${view.id}`);
+                    return;
+                }
+
+                const urlWithSchema = `${server.url.origin}${getFormattedPathName(parsedURL.pathname)}${parsedURL.search}`;
+                if (webContentsView.isReady() && ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion && Utils.isVersionGreaterThanOrEqualTo(ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion ?? '', '6.0.0')) {
```
larkox: I would add line breaks to make every condition a bit more isolated.

```suggestion
                if (
                    webContentsView.isReady() &&
                    ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion &&
                    Utils.isVersionGreaterThanOrEqualTo(ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion ?? '', '6.0.0')
                ) {
```

I would even consider putting this condition into a variable, just so it has a name and it is easier to reason about.

```
const reallyClearNameAboutWhatWeAreCheckingHere =
    webContentsView.isReady() &&
    ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion &&
    Utils.isVersionGreaterThanOrEqualTo(ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion ?? '', '6.0.0');

if (reallyClearNameAboutWhatWeAreCheckingHere) {
    ...
```

larkox: After rereading what this condition does, you could add a util function "isMinimumServerVersionFromWebContentsView" that does all this hidden in its own function.

(Or any intermediate solution between the two extremes 😜 )

src/app/popoutMenu.ts
```diff
@@ -0,0 +1,91 @@
...
+            click() {
+                const webContentsView = WebContentsManager.getView(viewId);
+                if (!webContentsView) {
+                    return;
+                }
+                const url = webContentsView.getWebContentsView().webContents.getURL();
+                clipboard.writeText(url);
+            },
+        },
+    ] as Array<MenuItemConstructorOptions | MenuItem>;
```
larkox: If you want to be more "type safe", add the type to the variable, instead of casting to a type.
i.e.
```
const template: MyType = [...];
```
vs
```
const template = [...] as MyType;
```

src/app/popoutMenu.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import TabManager from './tabs/tabManager';
+
+function createTemplate(viewId: string) {
+    const view = ViewManager.getView(viewId);
+    if (!view) {
+        return [];
+    }
+    const isWindow = view.type === ViewType.WINDOW;
+    const isLastTab = view.type === ViewType.TAB && TabManager.getOrderedTabsForServer(view.serverId).length === 1;
```
larkox, line 21: What about renaming this one to "onlyOneTabRemaining" or something like that? Last tab could mean the last of many, not the last remaining one.

larkox (commented):
Half through with the review.

So far just nitpicks and 0/5. No problem if you ignore all of the comments.

src/app/windows/baseWindow.ts
```diff
@@ -0,0 +1,248 @@
...
+        this.win.setMenuBarVisibility(false);
+
+        this.win.webContents.once('did-finish-load', () => {
+            this.win.webContents.zoomLevel = 0;
+            this.ready = true;
+        });
+
+        this.win.once('restore', () => {
+            this.win?.restore();
+        });
```
larkox, line 81: Why restore only once? (Maybe I don't have a clear idea of what `restore` means in this context 😅 )

src/common/appState.ts
```diff
@@ -14,37 +21,77 @@ export class AppState extends EventEmitter {
...
     updateMentions = (viewId: string, mentions: number) => {
-        ServerManager.getViewLog(viewId, 'AppState').silly('updateMentions', mentions);
+        ViewManager.getViewLog(viewId, 'AppState').silly('updateMentions', mentions);
 
         this.mentions.set(viewId, mentions);
+        this.reduceNotificationsPerServer(this.mentionsPerServer, this.mentions, (base, value) => (base ?? 0) + (value ?? 0));
         this.emitStatusForView(viewId);
     };
 
     updateUnreads = (viewId: string, unreads: boolean) => {
```
larkox, line 73: When do we want to update something only for the view, and not for all the views that handle that server?

src/common/config/index.ts
```diff
@@ -356,7 +363,7 @@ export class Config extends EventEmitter {
             }
         }
 
-        return configData as ConfigType;
+        return configData as CurrentConfig;
```
larkox, line 366: Out of scope: I wonder if we can avoid the somehow the casting, to be more type safe.

src/common/views/viewManager.ts
```diff
@@ -0,0 +1,206 @@
...
+        return Boolean(Config.viewLimit && this.views.size >= Config.viewLimit);
+    };
+
+    createView = (server: MattermostServer, type: ViewType) => {
+        log.debug('createView', server.id, server.name, type);
+
+        if (this.isViewLimitReached()) {
+            log.warn(`createView: View limit reached for server ${server.id}`);
+            return undefined;
+        }
```
larkox, line 88: I know I saw an alert somewhere else for the view limit reached. Is that triggered somewhere else?

src/renderer/components/BasePage.tsx
```diff
@@ -0,0 +1,77 @@
...
+}: Props) {
+    let errorComponent;
+    if (errorState === ErrorState.FAILED) {
+        errorComponent = (
+            <ConnectionErrorView
+                darkMode={darkMode}
+                errorInfo={errorMessage}
+                url={errorUrl}
+                appName={appName}
+                handleLink={() => window.desktop.openServerExternally()}
```
larkox: The copilot comment is valid. Since we are not using anything from inside of the component, you can define these functions (`openServerExternally` and `openServerUpgradeLink`) outside of the component, so you don't need the useCallback boilerplate.

larkox: Another option is just doing something like...

```suggestion
                handleLink={window.desktop.openServerExternally}
```

The only problem with this approach is that the references are resolved on render time, not on execution time, so the value may be different.

src/renderer/components/MainPage.tsx
```diff
@@ -151,12 +154,60 @@ class MainPage extends React.PureComponent<Props, State> {
...
+        const currentServer = await window.desktop.getCurrentServer();
+        if (currentServer?.id) {
+            this.setState({
+                currentServer,
+                activeServerId: currentServer.id,
+            });
+        }
+    };
+
     async componentDidMount() {
```
larkox, line 180: First time I see a componentDidMount as an async. But well, if it is working, it is working 😅

src/renderer/components/ServerDropdownButton.tsx
```diff
@@ -11,13 +11,15 @@ type Props = {
...
+    currentMentions: number;
+    currentUnread: boolean;
     hasUnreads: boolean;
     isMenuOpen: boolean;
     darkMode: boolean;
 }
 
 const ServerDropdownButton: React.FC<Props> = (props: Props) => {
-    const {isDisabled, activeServerName, totalMentionCount, hasUnreads, isMenuOpen, darkMode} = props;
+    const {isDisabled, activeServerName, totalMentionCount, currentMentions, currentUnread, hasUnreads, isMenuOpen, darkMode} = props;
```
larkox, line 22: If you are going to deconstruct the props, deconstruct them inside the function arguments. That way there are no "two ways to access the same variable" and keep things more consistent.

src/renderer/components/SettingsModal/components/NumberSetting.scss
```diff
@@ -0,0 +1,33 @@
+.NumberSetting {
+    padding-bottom: 24px;
```
larkox, line 2: Not sure if it makes sense in this context, but I see a bottom padding and a top padding of 24px. If this elements are supposed to be in a flexbox, you can use the gap property to make it more simple.

src/renderer/components/SettingsModal/components/NumberSetting.tsx
```diff
@@ -0,0 +1,81 @@
...
+    min,
+    max,
+    step = 1,
+    bottomBorder,
+    defaultValue,
+    allowUndefinedValue,
+}: Props) {
+    const [value, setValue] = useState<number | undefined>(allowUndefinedValue ? propValue : propValue ?? defaultValue);
+
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
larkox: nitpick: New function on every render. Add a `useCallback`.

src/renderer/components/BasePage.tsx
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
```
larkox, line 1: Nothing in the renderer folder seems to have tests.

src/renderer/components/TabBar.tsx
```diff
@@ -105,24 +124,15 @@ class TabBar extends React.PureComponent<Props, State> {
                     key={tab.id}
                     draggableId={`serverTabItem-${tab.id}`}
                     index={index}
+                    isDragDisabled={this.props.tabsDisabled || this.props.isMenuOpen}
```
larkox, line 127: Out of scope:

I put the comment here because it is where you had a change. This whole part should be a completely different `<Tab>` component. So this piece of the code looks like:

```
const tabs = this.props.tabs.map((tab, index) => (
     <Tab
        prop1={...}
        ...
     />
));
```

That way you can separate concerns, and this component becomes less bloated.

src/renderer/components/TabBar.tsx
```diff
@@ -59,6 +59,13 @@ class TabBar extends React.PureComponent<Props, State> {
         };
     };
 
+    onOpenPopoutMenu = (id: string) => {
```
larkox, line 62: This returns a new function on every call, and you are calling it on every render. So... you are passing a new function on every render to the child component that receives this. If you follow my advice of moving the tab logic to a new component, you can circumvent that with a `useCallback` on that new component.

src/renderer/components/TabBar.tsx
larkox, line 127: Again, commenting here because it is where you had a change.

Going over all tabs on every render is not the most performant thing. If this was a function component, I would recommend to put that in a `useMemo`. But for a class component I don't know a good way to handle this.

Either way... probably our bottleneck is not here 😄

src/renderer/components/TabBar.tsx
```diff
@@ -175,16 +185,36 @@ class TabBar extends React.PureComponent<Props, State> {
...
                             ref={provided.innerRef}
                             className={classNames('TabBar', {
                                 darkMode: this.props.isDarkMode,
+                                disabled: this.props.tabsDisabled,
                             })}
                             id={this.props.id}
                             {...provided.droppableProps}
                         >
                             {tabs}
+                            {!this.props.tabsDisabled && !snapshot.draggingFromThisWith && !this.props.isViewLimitReached &&
```
larkox, line 199: For long conditions like this, specially when the thing you are adding is slightly complex to understand what it does, it is nice to add them to a variable, just to work as some kind of documentation. For example...

```
const showNewTabButton = !this.props.tabsDisabled &&
     !snapshot.draggingFromThisWith &&
     !this.props.isViewLimitReached;
```

src/renderer/components/TopBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+        macOS: window.process.platform === 'darwin',
+        darkMode,
+        fullScreen,
+    });
+
+    return (
+        <div
+            className={topBarClassName}
+            onContextMenu={openPopoutMenu}
+            onDoubleClick={() => window.desktop.doubleClickOnWindow()}
```
larkox: Same thing about avoiding arrow functions.

src/renderer/hooks/useConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import {useState, useEffect, useCallback} from 'react';
+
+import type {CombinedConfig} from 'types/config';
+
+export function useConfig() {
+    const [config, setConfig] = useState<CombinedConfig | undefined>(undefined);
+
+    const requestConfig = useCallback(async (exitOnError?: boolean) => {
+        // todo: should we block?
```
larkox, line 12: It is good to add either a block so two request cannot be happening at the same time (a loading ref should be enough) or a way to only accept the last request sent (a lastSent ref should be also enough). If you want to know anything about either of the two approaches, let me know.

src/renderer/hooks/useConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+        try {
+            const configRequest = await window.desktop.getConfiguration() as CombinedConfig;
```
larkox, line 14: We control `getConfiguration`, right? Why do we need to cast it?

If CombinedConfig is just a subset of what it is really returning, we can just add the return type to the function, and typescript should just handle it.

src/renderer/hooks/useConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+        setConfig(configData);
+    }, [requestConfig]);
+
+    useEffect(() => {
+        setInitialConfig();
+
+        window.desktop.onReloadConfiguration(() => {
+            reloadConfig();
+        });
+    }, [setInitialConfig, reloadConfig]);
```
larkox: This effect is doing too much. The way it is implemented right now it is ok. Everything depends on request config, and request config doesn't depend on anything, so in practice, nothing is ever going to change. So "bad things" will not happen. But it is not "future proof". If you add anything to requestConfig that makes it to change between renders, you can start having annoying bugs. I would divide in two useEffect, each one with its own responsibilities.

First one that handles the initialConfig. Since `setInitialConfig` is only used in the effect, you don't need to define the function outside of the effect. I can stay like this:

```
useEffect(() => {
    const configData = await requestConfig(true);
    setConfig(configData);
}, []);
```
Keep in mind the empty dependencies (even if the linter complains). This is because you only want to execute this piece of code on mount. You don't want to set the initial config over and over.

For the other one, it should just add the event:
```
useEffect(() => {
    window.desktop.onReloadConfiguration(() => reloadConfig()};
}, [reloadConfig]);
```
In this case, we want to have the function in the dependencies, because if `reloadConfig` changes, we want to update the callback.

Another thing that bothers me here, is that we are setting the callback (calling `onReloadConfiguration`) but we are not deleting the callback when the component is unmounted. This could potentially lead to memory leaks.

src/renderer/index.tsx
```diff
@@ -4,93 +4,46 @@
...
-        window.desktop.onReloadConfiguration(() => {
-            this.reloadConfig();
-        });
+    useEffect(() => {
+        printVersion();
 
         // Deny drag&drop navigation in mainWindow.
         // Drag&drop is allowed in webview of index.html.
         document.addEventListener('dragover', (event) => event.preventDefault());
         document.addEventListener('drop', (event) => event.preventDefault());
```
larkox, line 30: Shouldn't we remove the listeners on unmount?

src/renderer/index.tsx
```diff
@@ -4,93 +4,46 @@
...
-        return undefined;
-    };
+    }, []);
 
-    openMenu = () => {
+    const openMenu = () => {
         if (window.process.platform !== 'darwin') {
             window.desktop.openAppMenu();
         }
     };
```
larkox: This is passed as a prop, so should be stable between renders. You could add a useCallback, but since this has no dependencies from the component, you can move this out of the component altogether.

src/renderer/index.tsx
```diff
@@ -4,93 +4,46 @@
...
-            <IntlProvider>
-                <MainPage
-                    openMenu={this.openMenu}
-                    darkMode={config.darkMode}
-                    appName={config.appName}
-                />
-            </IntlProvider>
-        );
+    if (!config) {
+        return null;
```
larkox, line 34: Do we want to add a loading state?

src/renderer/index.tsx
```diff
@@ -4,93 +4,46 @@
...
-        this.state = {};
-    }
-
-    async componentDidMount() {
-        await this.setInitialConfig();
-
-        window.desktop.onSynchronizeConfig(() => {
-            this.reloadConfig();
-        });
+function Root() {
```
larkox, line 21: There is one small difference while doing the migration. This happens also on other files you have migrated to function components.

Many of these components extended from PureComponent, which had memoization attached to them. In order to have the exact same thing on function components, you have to wrap the component into a `React.memo` call. For example...

```
function MyComponent() {...}
export default React.memo(MyComponent);
```

That being said, I don't think it is such a big deal.

src/renderer/modals/loadingScreen/index.tsx
```diff
@@ -1,65 +1,60 @@
...
-    };
+        // Cleanup function
+        return () => {
+            // Note: In a real app, you might want to remove event listeners here
+            // but since this is a root component that doesn't unmount, it's not strictly necessary
+        };
+    }, []);
 
-    onFadeOutComplete = () => {
+    const onFadeOutComplete = () => {
```
larkox: Same comment about passing this function later in a prop. Either move it out of the component, or add a useCallback.

src/renderer/modals/loadingScreen/index.tsx
```diff
@@ -1,65 +1,60 @@
...
-    render() {
-        return (
-            <LoadingScreen
-                loading={this.state.showLoadingScreen}
-                darkMode={this.state.darkMode}
-                onFadeOutComplete={this.onFadeOutComplete}
-            />
-        );
+    if (!config) {
+        return null;
```
larkox, line 47: I was going to ask if we need a loading screen, but this is already the loading screen 😅 . 0/5 on this one.

src/renderer/modals/loadingScreen/index.tsx
```diff
@@ -1,65 +1,60 @@
...
-    };
+        initializeApp();
 
-    onToggleLoadingScreenVisibility = (showLoadingScreen: boolean) => {
-        this.setState({showLoadingScreen});
-    };
+        // Cleanup function
+        return () => {
+            // Note: In a real app, you might want to remove event listeners here
+            // but since this is a root component that doesn't unmount, it's not strictly necessary
```
larkox, line 42: It cost you nothing to add it, and it becomes more future proof 😜

src/renderer/modals/loadingScreen/index.tsx
```diff
@@ -1,65 +1,60 @@
...
-        window.addEventListener('click', () => {
-            window.desktop.closeServersDropdown();
-            window.desktop.closeDownloadsDropdown();
-        });
-    }
+            window.addEventListener('click', () => {
+                window.desktop.closeServersDropdown();
+                window.desktop.closeDownloadsDropdown();
+            });
+        };
```
larkox: This lack consistency. You are putting one of the functions (the one for click) inline, while the other, (handle toggle visibility) you put it in a separate function.

I have no strong preference for any of them, but we should be consistent.

src/renderer/popout.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        window.desktop.onLoadIncompatibleServer((_, loadUrl) => {
+            setErrorState(ErrorState.INCOMPATIBLE);
+            setErrorUrl(loadUrl);
+        });
+
+        window.desktop.onUpdatePopoutTitle((viewId, title) => {
+            setTitle(title);
+            setViewId(viewId);
+        });
```
larkox, line 44: Shouldn't we de-register all those callbacks?

src/renderer/popout.tsx
```diff
@@ -0,0 +1,83 @@
...
+            return;
+        }
+        window.desktop.openPopoutMenu(viewId);
+    };
+
+    const openMenu = () => {
+        if (window.process.platform !== 'darwin') {
+            openPopoutMenu();
+        }
+    };
```
larkox: useCallback for both

src/renderer/popout.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!config) {
+        return null;
+    }
```
larkox, line 62: Loading state?

e2e/modules/environment.js
devinbinnie: Hehe nope, forgot to revert that commit :)

e2e/specs/server_management/long_server_name.test.js
devinbinnie: Same here

e2e/specs/server_management/popout_windows.test.js
devinbinnie: For now, I think once we have test case numbers we can add something here.

e2e/specs/settings.test.js
devinbinnie: sus-whos-sus-not-me

e2e/specs/server_management/tab_management.test.js
devinbinnie: Same here, just need test case numbers

src/app/callsWidgetWindow.ts
devinbinnie: Yeah this never should have been relative to begin with. Will fix.

src/app/callsWidgetWindow.ts
devinbinnie: It can't do anything with an undefined value, all it would return is `undefined` itself. I prefer this pattern, to ensure that the data going in is defined by the caller, instead of having each function make that call.

FWIW, if you're asking for this value, the view id should never be undefined.

src/app/callsWidgetWindow.ts
devinbinnie: What would probably be best here is a `copyURL` helper function, because that's all we're doing here. It actually can't be `undefined` if you look at what `parseURL` does, unless this is somehow passed something that isn't a `URL` object. We could probably fix this, but not right now.

src/app/callsWidgetWindow.ts
devinbinnie: We should probably do this for the `webContentsEvents` stuff too, they all need this `as const`. Definitely will do in a different PR.

src/app/mainWindow/mainWindow.ts
devinbinnie: Same as above, this should never have been relative to begin with...

src/app/mainWindow/mainWindow.ts
devinbinnie: It was moved to https://github.com/mattermost/desktop/blob/d00f016c8b723cb0bdc90241c91026f26ad28a67/src/app/windows/baseWindow.ts#L205

src/app/menus/app.ts
devinbinnie: I don't think it was ever needed to be honest, so no huge loss.

src/app/navigationManager.ts
devinbinnie: The top level check for `url` makes sense, but a structured `if-else if` is preferred when checking for a server. Makes it easier to read the program flow IMO.

src/app/navigationManager.ts
devinbinnie: This seems like some cleanup we could do across the app. I'll save it for later since this isn't new code, just moved.

src/app/popoutMenu.ts
devinbinnie: I probably shouldn't have fixed this here, but it was annoying me so I did :P

src/app/popoutMenu.ts
devinbinnie: Don't think it matters that much, you can infer what it does from the code.

src/app/windows/baseWindow.ts
devinbinnie: I think this is a fix for something on Windows from a while ago...it should probably be commented but we didn't do that :P

src/common/appState.ts
devinbinnie: The plan is to eventually support servers that can handle mentions and unreads per view, so for example if you have a conversation open on one tab, and that conversation is what has an unread/mention, we'd just display that.

It's not implemented yet, but in those cases we'd only update and store per view. For all current servers, we're just going to show one dot per server, so we only need to store it as such.

This probably needs to be fleshed out more, but technically we've always supported this, just not properly :P.

src/common/config/index.ts
devinbinnie: I think this is one of those rare cases where it's actually correct to type cast. The reason is that this function is for upgrading the configuration, so what we pass in can technically be a different type. But if we've upgraded, we can be assured it's in the latest type.

In short, if this were to cause a problem, it should be loud and unsafe so we can fix it.

src/common/views/viewManager.ts
devinbinnie: If you deep-link or try to open a link from another server, yes you get an alert. It's done here: https://github.com/mattermost/desktop/blob/d00f016c8b723cb0bdc90241c91026f26ad28a67/src/app/navigationManager.ts#L122

We only want to show it when the user expects navigation to happen.

src/renderer/components/BasePage.tsx
devinbinnie: I swore that wasn't working before...well it is now. Fixed :)

src/renderer/components/MainPage.tsx
devinbinnie: Old code :D. This should be migrated to a functional component and split up. It's on my ever growing tech debt list lol.

src/renderer/components/ServerDropdownButton.tsx
devinbinnie: Sure, as a best practice. But nothing horribly wrong here, will leave and clean-up later.

src/renderer/components/SettingsModal/components/NumberSetting.scss
devinbinnie: The CSS is a mess, and needs to be revamped. Will ignore most of the CSS nitpicks for now since I can agree with most of them likely, and I have little opinion on how CSS should work. I have a plan to at least rework it all into SCSS, and make sure of new themeing. Will do some cleanup there :)

src/renderer/components/SettingsModal/components/NumberSetting.tsx
devinbinnie: Wouldn't be a Daniel review if he didn't recommend this at least once 🤣

src/renderer/components/BasePage.tsx
devinbinnie: Yep. We should make those, at some point.

src/renderer/components/TabBar.tsx
devinbinnie: Good idea. We'll do that, at some point :D

src/renderer/components/TabBar.tsx
devinbinnie: Will do that, at some point.

src/renderer/components/TabBar.tsx
devinbinnie: See all above :D

src/renderer/components/TabBar.tsx
devinbinnie: Will leave for now as it's readable enough and not re-used. But makes sense once we fix up these components a bit.

src/renderer/components/TopBar.tsx
devinbinnie: This one complains because `doubleClickOnWindow` had a parameter. But then I realized it wasn't using that parameter, so fixed :)

src/renderer/hooks/useConfig.ts
devinbinnie: This is a really old comment, but yeah we can look into this later.

src/renderer/hooks/useConfig.ts
devinbinnie: Old code, will look at later.

src/renderer/hooks/useConfig.ts
devinbinnie: Since this is newer-ish code, I will fix this now.

src/renderer/index.tsx
devinbinnie: Probably. In this case though if this component is even unmounted it's because the window is closing, so I think it's moot to remove them. But still a best practice. Will deal with later.

src/renderer/index.tsx
devinbinnie: Easy enough.

src/renderer/index.tsx
devinbinnie: Worst thing is you see a white screen that should be gone in milliseconds. Probably not a huge deal.

src/renderer/index.tsx
devinbinnie: I'll leave for now. We should re-evaluate though once we go to clean-up the front-end.

src/renderer/modals/loadingScreen/index.tsx
devinbinnie: Moved out.

src/renderer/modals/loadingScreen/index.tsx
devinbinnie: Yep, no need.

src/renderer/modals/loadingScreen/index.tsx
devinbinnie: Yes, will clean up later though :P

src/renderer/modals/loadingScreen/index.tsx
devinbinnie: Sure.

src/renderer/popout.tsx
devinbinnie: This needs to be added and redone for all internal listeners. Since the assumption was that they've never registered more than once, and destroyed when the window closes, we didn't set that up for internal stuff.

The externalAPI does this though. I'll look at it at a later date.

src/renderer/popout.tsx
devinbinnie: nahh

yasserfaraazkhan (approved):

src/common/views/MattermostView.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+        if (serverURL.pathname === '/') {
+            const url = parseURL(new URL(serverURL));
+            if (!url) {
+                throw new Error(`URL for server ${this.serverId} is not valid`);
+            }
+            url.pathname = this.initialPath;
+            return url;
+        }
+        const url = parseURL(serverURL.toString() + this.initialPath);
```
sbishel, line 48: Do we need to check for proper separator on this.initialPath?

devinbinnie: If you mean the leading `/`, this component just assumes that it's a properly formatted path. We aren't checking for path traversal here though.

Funny enough, I wrote this parameter to work and it's not even used since we have other code that opens tabs and navigates for us :P. I'm going to leave it since I intend to use it at some point, at which time it's probably a good idea to validate that path as well.

src/app/popoutMenu.ts
```diff
@@ -0,0 +1,91 @@
...
+import {localizeMessage} from 'main/i18nManager';
+
+import TabManager from './tabs/tabManager';
+
+function createTemplate(viewId: string) {
+    const view = ViewManager.getView(viewId);
+    if (!view) {
+        return [];
+    }
+    const isWindow = view.type === ViewType.WINDOW;
```
enzowritescode, line 20: nitpick: I see the `view.type === ...` logic across a few files. Some helper functions could be useful.

src/app/menus/app.ts
devinbinnie: I couldn't think of a better way in that time, so I think for now we can leave it because it *does* work, I've just noticed sometimes it doesn't. Will keep an eye out for what can be done going forward.

src/app/serverHub.ts
devinbinnie: Nahh, we can refactor this later.

src/app/serverHub.ts
devinbinnie: Yeah I think AI added this one - I'll leave it for now since it doesn't deter from anything

src/app/popoutMenu.ts
devinbinnie: I think it's fine - it's just a boolean comparison. If it was more than one comparison we were doing over and over I'd agree with the helper function.

PR #3493, opened 2025-08-25
[MM-65131] Use the correct values and check for explicit undefined with `registry-js`

devinbinnie:
#### Summary
When we switched our Windows registry reading library in #3460, we didn't account for how the new library works with bit values. The new library will give us a `number` instead of a `string`, meaning we have to account for the actual `undefined` value instead of just looking for truthy.

This PR makes some changes to account for that, preserving the same logic otherwise. Also added a bunch of test cases to verify the logic.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65131

```release-note
NONE
```

### DIFF
--- a/src/common/config/RegistryConfig.ts
+++ b/src/common/config/RegistryConfig.ts
@@ -50,7 +50,7 @@ export default class RegistryConfig extends EventEmitter {
             // extract EnableServerManagement from the registry
             try {
                 const enableServerManagement = this.getEnableServerManagementFromRegistry();
-                if (enableServerManagement !== null) {
+                if (enableServerManagement !== undefined) {
                     this.data.enableServerManagement = enableServerManagement;
                 }
             } catch (error) {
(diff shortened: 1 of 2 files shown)


src/common/config/RegistryConfig.ts
```diff
@@ -50,7 +50,7 @@ export default class RegistryConfig extends EventEmitter {
             // extract EnableServerManagement from the registry
             try {
                 const enableServerManagement = this.getEnableServerManagementFromRegistry();
-                if (enableServerManagement !== null) {
+                if (enableServerManagement !== undefined) {
```
hanzei, line 53: Should we check for both `null` and `undefined`?

src/common/config/RegistryConfig.ts
```diff
@@ -118,7 +114,7 @@ export default class RegistryConfig extends EventEmitter {
         for (const hive of REGISTRY_HIVE_LIST) {
             results.push(this.getRegistryEntryValues(hive, key, name));
         }
-        return results.filter((value) => value);
+        return results.filter((value) => value !== undefined);
```
hanzei, line 117: same here

src/common/config/RegistryConfig.ts
devinbinnie: It can't be `null`, just `undefined`.

PR #3501, opened 2025-09-08
[MM-65617] Allow scrolling in modals, fix other modal issues

devinbinnie:
#### Summary
A miss with the new modals was what they looked like under a small viewport.

This PR fixes that by allowing scrolling in modals properly.

A couple other fixes I've tacked on:
- Stop the modals from accidentally closing when clicking and dragging from the modal to the backdrop
- Turn off user selection of the modal text

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65617

```release-note
Fixed various issues with the newer modals
```


abhijit-singh (approved):
Looks good @devinbinnie! This fixes the original bug 👍 

The only other thing to consider is if we can/want to anchor the modal to the top similar to the account settings modal so that the modal doesn't jump up and down while switching tabs. Not a blocker for this PR though.

pvev (approved):
👍

src/renderer/css/components/Modal.scss
```diff
@@ -4,6 +4,20 @@
...
+
+::-webkit-scrollbar-track {
+    background: rgba(0, 0, 0, 0);
+}
+
+::-webkit-scrollbar-thumb {
+    border-radius: var(--radius-s);
+    background: rgba(var(--center-channel-color-rgb), 0.4);
+}
+
```
Rajat-Dabade, line 20: ~Firefox does not support `::-webkit-scrollbar`, `::-webkit-scrollbar-track`, `::-webkit-scrollbar-thumb`~

~https://caniuse.com/?search=%3A%3A-webkit-scrollbar~

Ah, my bad. Electron uses Chromium under the hood which will support.

PR #3508, opened 2025-09-17
[MM-65610] Unify pre-authorization (basic auth, client certs, pre-auth header) methods into a single manager, remove unused permission hooks

devinbinnie:
#### Summary
This PR unifies the three methods of pre-authorization (ie. any method of auth that happens before the web view reaches our app) into one single manager, which is responsible for listening for the auth challenges, popping modals and receiving the results.

The changes include, broadly:
- Merging of the `authManager` and the `certificateManager` into a `preAuthManager` class that handles both of those functions
- Refactoring of the recently adding "pre-auth header" changes by having the flow mimic the same feel of the other two methods, additional encapsulation such that most of the logic is in the `preAuthManager`
- Reworking of the UI for the pre-auth header such that it pops its own modal as well
- The UI will now tell the user which pre-auth method is required and instruct them that they will need to provide this authentication on the next modal
- Removal of the `trustedOriginStore` and the `permissionModal`, which was used to grant basic auth permissions to outside sources. This was only needed for certain installations that had their SAML provider behind basic auth, which is no longer needed since the authentication is handled in the browser now
- Added a bunch of additional, non-data exposing logging to the URL validation, such that we can understand the flow better if anything goes wrong
- Further improvements to the serverAPI module, in which we know explicitly report the type of auth challenge when we get an error

Some notes to consider:
- I had to remove the E2E tests added for the pre-auth header, because they were focused on validating the header via the add/edit server modals which is no longer the case. I think it best to create some E2E tests around all 3 cases, so I've created a ticket to address that as a follow-up: https://mattermost.atlassian.net/browse/MM-65773
- We were unable to support validating the method in the add/edit server modal, due to the design of how the application sends the auth challenge. We could control the pre-auth header ourselves, but basic auth and client cert challenges are sent by the session and the callbacks are what drives fulfilling the auth challenge. In the interest of consistency, we've kept it all the same.

Screenshots of the UX are below. Unfortunately setting up testing environments for these cases is a bit tricky, so I would mostly rely on the screenshots. @yasserfaraazkhan I can assist you if you'd like for QA, and if you are able to set them up publicly perhaps UX can use those to test against. @BenCookie95 is a good resource on the pre-auth header.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65610

#### Screenshots
|Method|Add Server Modal|Challenge Modal|
|-|-|-|
|Basic Auth|image|image|
|Client Certificate|image|image|
|Pre-Auth Header|image|image|

#### Release Note
```release-note
Unify and polish basic authentication, client certificate and pre-auth header authentication methods
```


yasserfaraazkhan (changes requested):
Thanks @devinbinnie 

Few of observations I want to share

1. When the user enters a wrong secret, should there be an error shown?

2. Should `Confirm` button be disabled until user enters something

3. If I am logged in a server that doesn't need auth token. And I try to add another server that needs auth token, from Settings > Servers > Add server

The enter auth token pop-up is hidden by the settings pop up.

https://github.com/user-attachments/assets/1abd592a-ec3d-4a6f-80fb-fa1e7b8b0a4b

BenCookie95 (changes requested):
The code looked good to me but I did some testing and I found a few issues:

- When you enter a url for a server that requires a pre-auth secret, the button says connect anyway which doesn’t seem right since the next screen will prompt you to enter that secret
- We only validate the secret on submit, on typing was a nice feature
- If I submit the wrong secret, there is no error on the modal when it reappears. If I submit an empty secret, the modal doesn’t reappear and the 403 apache2 page appears. The desktop app didn’t refresh after I submitted the correct secret after initially providing an empty secret or if I initially closed the modal without submitting any secret.
- I find the UX for when you edit a server a little confusing. Should there be a way to open the server authentication modal directly from the edit server modal instead of waiting for the user to close or submit the edit server modal.

matthewbirtch (changes requested):
Thanks @devinbinnie. This is working pretty well. A few things I noticed @devinbinnie:



- I know I've mentioned this already, but just flagging the error color again here as it's not legible. 
    image

- The notifications permissions show on top of the new secret modal. Is there a way to defer that permission until after the secret has been correctly entered?
    image
- Also, when the server authentication modal shows, there is a spinner in the background the never stops. Can we defer this spinner until it actually starts loading something?
    image

### Secret-specific feedback:
- If I don't enter in the secret, then navigate away to another server, this 'Server authentication required' modal shows on top of my other server. And it will keep trying if I dismiss it. I wonder if we should stop trying once I've switched to another server and then start trying again when I've switched back. This is an edge case, so maybe something we can improve in a follow-up.
    image

### Certificiate method feedback:
I think we need a radio icon to more clearly highlight the selected certificate. 
image

Something like this:
image

### Basic auth
All looked good to me. No issues here

wiggin77 (approved):

src/renderer/components/ConfigureServer/ConfigureServer.tsx
```diff
@@ -216,9 +179,25 @@ function ConfigureServer({
             };
         }
 
+        if (validationResult?.status === URLValidationStatus.BasicAuthRequired) {
```
enzowritescode, line 182: nitpick: this long sequence of if statements seems like a good use-case for a switch statement.

enzowritescode (approved):

src/renderer/components/ConfigureServer/ConfigureServer.tsx
devinbinnie: I'll look into this when we clean up the front-end components. On my list of tech debt for the Desktop App.

PR #3509, opened 2025-09-18
[MM-65011][MM-65664] Remove extraneous logging of potentially sensitive info, clear server data on removal

devinbinnie:
#### Summary
Our logs have defaulted to recording as much information as possible for debugging purposes, which has not been the best practice. Now if users were to remove servers for example, there would still be a record of that server existing in the log file, which is not ideal. Furthermore, when servers are removed, we still store all of the data for that server and if a user were to re-add it, the same data would still exist (ie. cookies, cache, IndexedDB, etc.)

This PR both removes a lot of extra logging information that might be sensitive, and ensures that we clear all data we can when a server is removed. On top of this, I've reformatted much of the logging to be a bit easier to read.

As a future follow-up, auditing our logging to be more descriptive when needed, and more verbose, should still happen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65011
https://mattermost.atlassian.net/browse/MM-65664

```release-note
Remove extraneous logging of potentially sensitive info, clear server data on removal
```

### DIFF
--- a/src/app/callsWidgetWindow.ts
+++ b/src/app/callsWidgetWindow.ts
@@ -190,7 +190,7 @@ export class CallsWidgetWindow {
         this.win?.loadURL(widgetURL, {
             userAgent: composeUserAgent(),
         }).catch((reason) => {
-            log.error(`failed to load: ${reason}`);
+            log.error('failed to load', {reason});
         });
     };
 
@@ -245,12 +245,12 @@ export class CallsWidgetWindow {
         if (url === this.getWidgetURL()) {
             return;
         }
-        log.warn(`prevented widget window from navigating to: ${url}`);
+        log.warn('prevented widget window from navigating');
         ev.preventDefault();
     };
 
     private setWidgetWindowStacking = ({onTop}: {onTop: boolean}) => {
-        log.debug('setWidgetWindowStacking', onTop);
+        log.debug('setWidgetWindowStacking', {onTop});
 
         if (!this.win) {
             return;
@@ -319,7 +319,7 @@ export class CallsWidgetWindow {
             };
         }
 
-        log.warn(`onPopOutOpen: prevented window open to ${url}`);
+        log.warn('onPopOutOpen: prevented window open');
         return {action: 'deny' as const};
     };
 
@@ -365,7 +365,7 @@ export class CallsWidgetWindow {
                     userAgent: composeUserAgent(),
                 });
             } catch (e) {
-                log.error('did-frame-finish-load, failed to reload with correct userAgent', e);
+                log.error('did-frame-finish-load, failed to reload with correct userAgent', {e});
             }
         });
     };
@@ -375,7 +375,7 @@ export class CallsWidgetWindow {
      ************************/
 
     private handleResize = (ev: IpcMainEvent, width: number, height: number) => {
-        log.debug('handleResize', width, height);
+        log.debug('handleResize', {width, height});
 
         if (!this.win) {
             return;
@@ -420,7 +420,7 @@ export class CallsWidgetWindow {
     };
 
     private handleGetDesktopSources = async (event: IpcMainInvokeEvent, opts: Electron.SourcesOptions) => {
-        log.debug('handleGetDesktopSources', opts);
+        log.debug('handleGetDesktopSources');
 
         // For Calls we make an extra check to ensure the event is coming from the expected window (main view).
         // Otherwise we want to allow for other plugins to ask for screen sharing sources.
@@ -446,7 +446,7 @@ export class CallsWidgetWindow {
                 }
                 this.missingScreensharePermissions = true;
             } catch (err) {
-                log.error('failed to reset screen sharing permissions', err);
+                log.error('failed to reset screen sharing permissions', {err});
             }
         }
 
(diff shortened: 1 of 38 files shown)


src/app/callsWidgetWindow.ts
```diff
@@ -190,7 +190,7 @@ export class CallsWidgetWindow {
         this.win?.loadURL(widgetURL, {
             userAgent: composeUserAgent(),
         }).catch((reason) => {
-            log.error(`failed to load: ${reason}`);
+            log.error('failed to load', {reason});
```
larkox, line 193: I see we are changing this pattern in many places. Is this intended?

src/app/mainWindow/modals/modalManager.ts
```diff
@@ -145,7 +145,7 @@ export class ModalManager {
     };
 
     handleResizeModal = (bounds: Electron.Rectangle) => {
-        log.debug('handleResizeModal', {bounds, modalQueueLength: this.modalQueue.length});
+        log.debug('handleResizeModal', {modalQueueLength: this.modalQueue.length});
```
larkox: Why removing the bounds here? I think I saw them in a different file and they were not removed.

src/main/app/utils.ts
```diff
@@ -249,7 +249,7 @@ export async function updateServerInfos(servers: MattermostServer[]) {
                 map.set(srv.id, data);
             }).
             catch((error) => {
-                log.warn('Could not get server info for', srv.name, error);
+                log.warn('Could not get server info', srv.id, error);
```
larkox: We are not following here the pattern of returning things as an object. Intended or forgotten?

src/app/views/webContentEventsCommon.ts
```diff
@@ -22,14 +23,22 @@ export const generateHandleConsoleMessage = (log: Logger) => (event: Event<WebCo
...
 };
 
+function sanitizeMessage(sourceURL: string, message: string) {
+    const parsedURL = parseURL(sourceURL);
+    if (!parsedURL) {
+        return message;
+    }
+    return message.replace(parsedURL.host, '<host>');
+}
+
```
marianunez, line 41: Unit test for this would be nice, even though seems this whole file would benefit from that as well

src/app/serverHub.ts
```diff
@@ -400,6 +406,19 @@ export class ServerHub {
...
+        const server = ServerManager.getServer(serverId);
+        if (!server) {
+            return;
+        }
+
+        session.defaultSession.clearData({
+            origins: [server.url.origin],
+        });
+    };
+
```
marianunez, line 421: Same with this regarding tests

src/main/notifications/index.ts
```diff
@@ -81,15 +81,15 @@ class NotificationManager {
         };
 
         if (!await PermissionsManager.doPermissionRequest(webcontents.id, 'notifications', {requestingUrl: server.url.toString(), isMainFrame: false})) {
-            log.verbose('permissions disallowed', webcontents.id, serverName, server.url.toString());
+            log.verbose('permissions disallowed', webcontents.id, server.id);
```
larkox: Here you have a few other places where you don't wrap the arguments in objects.

src/main/security/permissionsManager.ts
```diff
@@ -119,7 +119,7 @@ export class PermissionsManager extends JsonFileManager<PermissionsByOrigin> {
         permission: string,
         details: PermissionRequestHandlerHandlerDetails,
     ) => {
-        log.debug('doPermissionRequest', permission, details);
+        log.debug('doPermissionRequest', webContentsId, permission);
```
larkox: And another

src/main/server/serverAPI.ts
```diff
@@ -43,11 +43,11 @@ export async function getServerAPI(url: URL, isAuthenticated: boolean, onSuccess
 
     if (onSuccess) {
         req.on('response', (response: Electron.IncomingMessage) => {
-            log.silly('response', response);
+            log.silly('response');
             if (response.statusCode === 200) {
                 let raw = '';
                 response.on('data', (chunk: Buffer) => {
-                    log.silly('response.data', `${chunk}`);
+                    log.silly('response.data');
```
larkox, line 50: Are these logs useful anymore?

src/app/serverHub.ts
devinbinnie: This is just a function that calls another function so not much to test here.

src/app/views/webContentEventsCommon.ts
devinbinnie: Good call. Added tests for the whole file :)

src/app/callsWidgetWindow.ts
devinbinnie: Yes, it makes things easier to read.

src/app/mainWindow/modals/modalManager.ts
devinbinnie: AI was a bit overzealous on this one, will restore here.

src/main/app/utils.ts
devinbinnie: Forgotten :P

src/main/server/serverAPI.ts
devinbinnie: Good to know if you got a response or an error, so yeah. That's why they're marked silly.

PR #3517, opened 2025-09-24
[MM-65620] Add clear cache and reload link to the error screen

devinbinnie:
#### Summary
This PR adds a `Clear Cache and Reload` link to the error screen to clarify which steps a user can take when they are having trouble loading their server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65620

#### Screenshots
image

```release-note
Add clear cache and reload link to the error screen
```


esethna (changes requested):
Thanks Devin. Can't really test this fully, but assuming the link executes a hard refresh I think it looks good. Minor text proposal:

Propose the into paragraph is slightly updated to remove repetitive refresh instructions (since that's now the first bullet point with link to execute it):

```
"We're having trouble connecting to this {appName} server. We'll keep trying to establish a connection automatically. 

Please try the following steps if this issue persists:
```

PR #3521, opened 2025-09-29
[MM-64651] Add parent-child popout window support, with listeners to facilitate communication

devinbinnie:
#### Summary
This PR allows pop-out windows to be opened via the Web App, with a reference to those pop-outs passed back to the web-app to allow for communication between the pop-out and its parent view. Pop-outs are allowed to provide a serializable set of customization options which can affect how the pop-outs are sized and where they're rendered. Other options may become available as needed.

The following API is provided to the web app, allowing for the set up and communication flow for these pop-outs:
- `canPopout()`: Determines whether the Desktop App in-use has the ability to create pop-out windows at all. If this returns false, the pop-out button should be disabled.
- `openPopout(path, options): string`: Send a request to open a pop-out window in the Desktop App at the given path, and with a set of options assigned to the internal view that might modify how the Desktop App behaves. The options object must be serializable, as for security reasons no compiled code or binary can be sent up. This returns an ID in the form of a string that identifies the pop-out to the Desktop App.
- `canUsePopoutOption(optionName): boolean`: Asks the Desktop App if a given optionName is supported, in cases where an older Desktop App version is used on a newer server, and fall back logic is required.
- `sendToParent(channel, ...args)`: Send a message to the parent window of a pop-out via a certain channel. Note that the parent view can change.
- `sendToPopout(id, channel, …args)`: Send a message to a pop-out with the given ID. The ID must represent a pop-out view that is a child of the current tab or window.
- `onMessageFromParent(listener): () => void`: Attaches listeners for certain messages from the parent window. Returns a remove listener function.
- `onMessageFromPopout(listener): () => void`: Attaches listeners for certain messages for any child pop-outs to the parent window. Returns a remove listener function.
- `onPopoutClosed(listener): () => void`: Attaches a listener that will fire when the pop-out is closed, such that other listeners can be cleaned up. Returns a remove listener function.

The pop-outs coming from the web app are rate limited such that a malicious server cannot create infinite numbers of windows in a short period of time.

These changes require a version of the Mattermost server/web app to be taken advantage of, you can test using the server attached to the PR here: https://github.com/mattermost/mattermost/pull/34106

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64651

```release-note
Add parent-child popout window support, with listeners to facilitate communication
```

### DIFF
--- a/src/app/windows/popoutManager.ts
+++ b/src/app/windows/popoutManager.ts
@@ -309,6 +309,16 @@ export class PopoutManager {
             return undefined;
         }
 
+        const existingView = ViewManager.getViewsByServerId(view.serverId).
+            find((v) => v.type === ViewType.WINDOW && v.initialPath === path);
+        if (existingView) {
+            const window = this.popoutWindows.get(existingView.id);
+            if (window) {
+                window.browserWindow.show();
+            }
+            return existingView.id;
+        }
+
         this.debouncePopout = true;
         setTimeout(() => {
             this.debouncePopout = false;
--- a/src/app/views/MattermostWebContentsView.ts
+++ b/src/app/views/MattermostWebContentsView.ts
@@ -429,7 +429,7 @@ export class MattermostWebContentsView extends EventEmitter {
     private handleUpdateTarget = (e: Event, url: string) => {
         this.log.silly('handleUpdateTarget');
         const parsedURL = parseURL(url);
-        if (parsedURL && isInternalURL(parsedURL, this.view.getLoadingURL())) {
+        if (parsedURL && isInternalURL(parsedURL, ServerManager.getServer(this.view.serverId)?.url ?? this.view.getLoadingURL())) {
             this.emit(UPDATE_TARGET_URL);
         } else {
             this.emit(UPDATE_TARGET_URL, url);
--- a/src/app/windows/popoutManager.test.js
+++ b/src/app/windows/popoutManager.test.js
@@ -97,6 +97,7 @@ jest.mock('common/views/viewManager', () => {
         createView: jest.fn(),
         getViewTitle: jest.fn(),
         isViewLimitReached: jest.fn(),
+        getViewsByServerId: jest.fn(),
         mockViewManager,
     };
 });
@@ -175,6 +176,7 @@ describe('PopoutManager', () => {
         BaseWindow.mockImplementation(() => mockBaseWindow);
         WebContentsManager.createView.mockReturnValue(mockWebContentsView);
         ViewManager.getView.mockReturnValue(mockView);
+        ViewManager.getViewsByServerId.mockReturnValue([mockView]);
         ServerManager.getServer.mockReturnValue(mockServer);
     });
 
@@ -782,6 +784,7 @@ describe('PopoutManager', () => {
         };
         const mockNewView = {
             id: 'new-popout-id',
+            initialPath: '/test/path',
             serverId: 'test-server-id',
             type: ViewType.WINDOW,
         };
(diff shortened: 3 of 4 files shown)


src/app/windows/popoutManager.ts
```diff
@@ -251,6 +288,72 @@ export class PopoutManager {
...
+        const server = ServerManager.getServer(view.serverId);
+        if (!server) {
+            return undefined;
+        }
+
+        this.debouncePopout = true;
+        setTimeout(() => {
+            this.debouncePopout = false;
+        }, POPOUT_RATE_LIMIT);
+
```
M-ZubairAhmed, line 326: can you explain why is this needed?

src/app/windows/popoutManager.ts
```diff
@@ -251,6 +288,72 @@ export class PopoutManager {
...
+        return ViewManager.createView(server, ViewType.WINDOW, path, view.id, props)?.id;
```
enahum, line 327: shouldn't we validate the path and props?

src/app/windows/popoutManager.ts
```diff
@@ -251,6 +288,72 @@ export class PopoutManager {
...
+        const view = ViewManager.getView(webContentsView.id);
+        if (!view?.parentViewId) {
+            return;
+        }
+        const parentView = WebContentsManager.getView(view.parentViewId);
+        if (!parentView) {
+            log.warn('handleSendToParent: no parentView found', {parentViewId: view.parentViewId});
+            return;
+        }
+        parentView.sendToRenderer(MESSAGE_FROM_POPOUT, view.id, channel, ...args);
```
enahum, line 355: same question about validation, in this case channel and args

src/app/windows/popoutManager.ts
```diff
@@ -251,6 +288,72 @@ export class PopoutManager {
...
+    };
+
+    private handleSendToPopout = (_: IpcMainEvent, id: string, channel: string, ...args: unknown[]) => {
+        const view = WebContentsManager.getView(id);
+        if (!view) {
+            log.debug('handleSendToPopout: no view found', {id});
+            return;
+        }
+        view.sendToRenderer(MESSAGE_FROM_PARENT, channel, ...args);
```
enahum, line 364: here to

src/app/windows/popoutManager.ts
devinbinnie: There is a limit to the number of windows you can create, however, you can remove that limit. This ensures that a malicious server cannot create pop-out windows fast enough to DoS your computer.

devinbinnie: As for your other question - the pop-out limit was in another PR: https://github.com/mattermost/desktop/pull/3491

src/app/windows/popoutManager.ts
devinbinnie: The URL is validated when we go to load it: https://github.com/mattermost/desktop/blob/master/src/common/views/MattermostView.ts#L61

We could do it here but I don't see a reason to.

src/app/windows/popoutManager.ts
devinbinnie: All of this stuff is just relayed back down to the server that sent it up. The aim is to create an open channel between the two windows, facilitated by the Desktop App.

Nothing that comes through can be executed (it all must be serializable) so I don't think it's unsafe - but i'll let @enzowritescode chime in.

src/app/windows/popoutManager.ts
enzowritescode: Adding some validation in these handlers is a good idea since it is at the boundary of the IPC. It would be a good defense-in-depth practice and will help future-proof any changes.

devinbinnie: @enzowritescode What kind of validation do we want here? I guess I'm failing to see what's important to validate when we're already doing it elsewhere.

enzowritescode: Internal followup summary: current location of validation is fine. Moving forward with merging.

Issue #3526, opened 2025-10-05
[Bug]: Bump Electron to mitigate MacOS Tahoe performance


freak12techno:
### Checks before filing an issue

- [x] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [x] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing. This should include closed issues.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [x] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [x] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

Version 5.13.1 (176149789981)

### Operating System

Mac OS Tahoe 26.0.1

### Steps to reproduce

So MacOS Tahoe has a change that makes all Electron-based apps extremely laggy, see here https://github.com/electron/electron/pull/48376
I see that the electron version used in Mattermost desktop seems to not include that fix yet, which might cause the app to go laggy. Can it be bumped to include the fix?

### Expected behavior

App isn't lagging on Tahoe

### Observed behavior

App is lagging on Tahoe

### Log Output

```shell
-
```

PR #3529, opened 2025-10-06
[MM-66102] Upgrade to Electron v38.2.1, fix small vulnerabilities

devinbinnie:
#### Summary
Upgrade to Electron v38.2.1

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66102
Fixes https://github.com/mattermost/desktop/issues/3526

```release-note
Upgrade to Electron v38.2.1
```


devinbinnie:
> I wonder if dependabot should take care of security updates like this one. What do you think?

We could probably do that for minor/patch versions, but major versions can have breaking changes, so we have to do those manually.

hanzei:
Depenabot can be configured to only pull security updates, which in general means major versions only: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/customizing-dependabot-security-prs#example-1-configuration-for-security-updates-only

PR #3531, opened 2025-10-10
Synchronize Desktop App theme with the web app theme

devinbinnie:
#### Summary
Reviving an old Hackathon project of mine, in which I demonstrated that we could read the CSS variables from servers and pass that information to the Desktop App to influence its appearance. However, the front-end componentry was far from being able to support that easily at the time. Now however, all of the front-end components use the same CSS variable structure as the web app, making this change very easy.

This PR adds a feature in which when a server is focused, all elements in the Desktop App will adhere to the selected theme. In the case of pop-out windows, each popout window will have the theme of the server it is displaying. The functionality can be turned off and the application can return to its default behaviour if the user prefers.

This gives a sense of a much more integrated application, and allows for some customization to reach beyond the borders of the web app view. This has long been something I've wanted to do for the app and I hope for some users it allows our app to stand out a bit more.

Additionally, I've exposed an API for allowing the web app to listen for and read the "dark mode" status from the OS, so that in the future, we can have the web app synchronize with the OS theme.

#### Screenshots
image

#### Release Note
```release-note
Added default theme syncing with the web app
```

### DIFF
--- a/api-types/index.ts
+++ b/api-types/index.ts
@@ -25,7 +25,6 @@ export type Theme = {
     awayIndicator: string;
     dndIndicator: string;
     mentionBg: string;
-    mentionBj: string;
     mentionColor: string;
     centerChannelBg: string;
     centerChannelColor: string;


i18n/en.json
```diff
@@ -286,6 +286,8 @@
   "renderer.components.settingsPage.spellCheckerSetting.language": "Spell Checker Languages",
   "renderer.components.settingsPage.startAppOnLogin": "Start app on login",
   "renderer.components.settingsPage.startAppOnLogin.description": "If enabled, the app starts automatically when you log in to your machine.",
+  "renderer.components.settingsPage.themeSyncing": "Synchronize Desktop App theme with server",
+  "renderer.components.settingsPage.themeSyncing.description": "If enabled, the Desktop App theme will be synchronized with the server theme. For older server versions, a refresh of the server may be required to update the theme.",
```
marianunez, line 290: >  a refresh of the server may be required to update the theme."

Does this mean a reload on the Desktop app? I'm assuming yes but could be misinterpreted as a server restart

src/app/mainWindow/modals/modalManager.ts
```diff
@@ -159,14 +156,13 @@ export class ModalManager {
...
 
-    handleEmitConfiguration = (event: IpcMainEvent, config: CombinedConfig) => {
+    handleEmitConfiguration = () => {
         if (this.modalQueue.length) {
             log.debug('handleEmitConfiguration');
         }
 
         this.modalQueue.forEach((modal) => {
             modal.view.webContents.send(RELOAD_CONFIGURATION);
-            modal.view.webContents.send(DARK_MODE_CHANGE, config.darkMode);
```
marianunez, line 169: How was this `DARK_MODE_CHANGE` used before?

i18n/en.json
devinbinnie: Yeah it just means a reload of the server. @matthewbirtch Can you think of better verbiage here?

src/app/mainWindow/modals/modalManager.ts
devinbinnie: Basically any time the OS light/dark theme changed we emit this event. It's still happening, just better encapsulated: https://github.com/mattermost/desktop/pull/3531/files#diff-4144aecdf1e8e2adb71b981af6d4e34774a380d746d15737750d51f8c8929782R70

matthewbirtch:
> Yes these bugs are out of scope of this PR since it has to do with the checkbox setting component. Can you create a separate bug for that? I can fix it in another PR.

https://mattermost.atlassian.net/browse/MM-66211

i18n/en.json
matthewbirtch: Can we use the word 'workspace' instead? @cwarnermm in the end-user docs, we seem to use the term 'workspace' more, so I lean towards that. Also, maybe we don't need to state 'for older servers'.

**Sync Desktop App theme with workspace theme**
When enabled, the Desktop App theme will match the connected Mattermost workspace. A refresh may be required to apply changes.

api-types/index.ts
```diff
@@ -11,7 +11,35 @@ export type DesktopCaptureSource = {
...
+    sidebarTextActiveBorder: string;
+    sidebarTextActiveColor: string;
+    sidebarHeaderBg: string;
+    sidebarTeamBarBg: string;
+    sidebarHeaderTextColor: string;
+    onlineIndicator: string;
+    awayIndicator: string;
+    dndIndicator: string;
+    mentionBg: string;
+    mentionBj: string;
```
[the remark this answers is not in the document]
devinbinnie: I have no control over this - it is provided by the web app.

src/renderer/utils.ts
```diff
@@ -95,11 +97,119 @@ const printVersion = () => {
...
+    document.body.style.removeProperty('--new-message-separator');
+    document.body.style.removeProperty('--new-message-separator-rgb');
+    document.body.style.removeProperty('--link-color');
+    document.body.style.removeProperty('--link-color-rgb');
+    document.body.style.removeProperty('--button-bg');
+    document.body.style.removeProperty('--button-bg-rgb');
+    document.body.style.removeProperty('--button-color');
+    document.body.style.removeProperty('--button-color-rgb');
+    document.body.style.removeProperty('--error-text');
+    document.body.style.removeProperty('--error-text-color');
```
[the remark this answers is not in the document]
devinbinnie: AI'ed :rofl:

src/main/themeManager.ts
```diff
@@ -0,0 +1,169 @@
...
+        if (!this.popoutViews.has(serverId)) {
+            this.popoutViews.set(serverId, new Set());
+        }
+        this.popoutViews.get(serverId)?.add(webContents);
+        webContents.on('destroyed', () => {
+            this.popoutViews.get(serverId)?.delete(webContents);
+        });
+    };
+
+    private handleEmitConfiguration = (event: IpcMainEvent, config: CombinedConfig) => {
```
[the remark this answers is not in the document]
devinbinnie: Not needed.

src/renderer/utils.ts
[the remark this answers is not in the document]
devinbinnie: Actually it did point out one thing - the AI tool that helped me generate this generated an extra variable that we don't use, I'll remove that.

src/main/themeManager.ts
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        this.mainWindowViews.forEach((view) => {
+            view.send(DARK_MODE_CHANGE, config.darkMode);
+        });
+        this.popoutViews.forEach((views) => {
+            views.forEach((view) => {
+                view.send(DARK_MODE_CHANGE, config.darkMode);
+            });
+        });
```
marianunez, line 89: What case is this? Is this resetting the server theme synched and changing to dark OS setting?

src/main/themeManager.ts
```diff
@@ -0,0 +1,169 @@
...
+                view.send(RESET_THEME);
+            });
+            return;
+        }
+        if (!server.theme.isUsingSystemTheme) {
+            const themeSource = isLightColor(server.theme.centerChannelBg) ? 'light' : 'dark';
+            if (nativeTheme.themeSource !== themeSource) {
+                nativeTheme.themeSource = themeSource;
+            }
+        }
```
marianunez, line 126: And just to understand, what is this doing?

src/main/themeManager.ts
devinbinnie: This is for when the OS theme changes (dark/light mode). This still needs to be communicated to the front-end in case we disable (or have disabled) the theme syncing from the web app.

marianunez: So we do support Desktop app updating to dark based on OS setting? I thought that was a pending feature

src/main/themeManager.ts
devinbinnie: The system theme still affects the traffic light buttons (on macOS) and the top bar min/close buttons (on Windows) using the system theme. This overrides it based on the theme you've chosen from the browser, using a function I found online to check the relative brightness.

src/main/themeManager.ts
devinbinnie: No that's been there for years now.

api-types/index.ts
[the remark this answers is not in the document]
hmhealey: You can probably remove this field. This is an old typo, and while theoretically that could exist in some user's custom theme still, the web app corrects it to `mentionBg` in the `getTheme` selector

api-types/lib/index.d.ts
```diff
@@ -11,6 +11,34 @@ export type DesktopCaptureSource = {
...
+    sidebarTextActiveBorder: string;
+    sidebarTextActiveColor: string;
+    sidebarHeaderBg: string;
+    sidebarTeamBarBg: string;
+    sidebarHeaderTextColor: string;
+    onlineIndicator: string;
+    awayIndicator: string;
+    dndIndicator: string;
+    mentionBg: string;
+    mentionBj: string;
```
pvev, line 28: should this be part of the type?

pvev (approved):
🎉

api-types/lib/index.d.ts
devinbinnie: It needs to get rebuilt. I'll do that in a follow-up (I'm just about to publish the types :P)

PR #3535, opened 2025-10-15
Update to v6.0.0-1 api types

devinbinnie:
#### Summary
Update to v6.0.0-1 api types

```release-note
NONE
```

### DIFF
--- a/api-types/package-lock.json
+++ b/api-types/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mattermost/desktop-api",
-  "version": "5.10.0-2",
+  "version": "6.0.0-1",
   "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


api-types/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@mattermost/desktop-api",
-  "version": "5.10.0-2",
+  "version": "6.0.0-1",
```
marianunez, line 3: Curious what is the dash notation?

marianunez (approved):

api-types/package-lock.json
devinbinnie: Just a subversion. Usually noted as pre-release, but helps in case we mess something up :P

PR #3544, opened 2025-10-28
Easy login Support

enahum:
#### Summary
this PR introduces the EasyLogin for Guests accounts. This is part of the project of Build 2025

#### Ticket Link
TDB

#### Release Note
```release-note
NONE
```

### DIFF
--- a/src/app/navigationManager.ts
+++ b/src/app/navigationManager.ts
@@ -62,6 +62,10 @@ export class NavigationManager {
             if (webContentsView.isReady() && ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion && Utils.isVersionGreaterThanOrEqualTo(ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion ?? '', '6.0.0')) {
                 const formattedServerURL = `${server.url.origin}${getFormattedPathName(server.url.pathname)}`;
                 const pathName = `/${urlWithSchema.replace(formattedServerURL, '')}`;
+
+                // When navigating to the magic link url, we are hitting a special route
+                // in the server that is not handled by the react routers. Therefore,
+                // we need to load the url directly instead of using the browser history push.
                 if (isMagicLinkUrl(server.url, parsedURL)) {
                     webContentsView.load(urlWithSchema);
                 } else {


src/app/navigationManager.ts
```diff
@@ -45,6 +45,12 @@ export class NavigationManager {
 
         const parsedURL = parseURL(url)!;
         const server = ServerManager.lookupServerByURL(parsedURL, true);
+
+        let trimmedPathname = getFormattedPathName(parsedURL.pathname);
```
devinbinnie: I don't think this is necessary, `getFormattedPathName` basically just adds a `/` to the end of the pathname:

```
export const getFormattedPathName = (pn: string) => (pn.endsWith('/') ? pn : `${pn}/`);
```

So if we want to remove it, we should just use `parsedURL.pathname`

src/common/utils/url.ts
```diff
@@ -113,3 +113,25 @@ const equalUrlsIgnoringSubpath = (url1: URL, url2: URL, ignoreScheme?: boolean)
...
+        const parsed = parseURL(urlToTest);
+
+        // Check pathname and search params
+        if (parsed?.pathname === '/login/one_time_link') {
+            const t = parsed.searchParams.get('t');
+            return Boolean(t && (/^[A-Za-z0-9]{64}$/).test(t));
+        }
+
+        // Also allow for URLs where /login/sso/easy is not at root (e.g., subpath)
+        const match = url.match(/\/login\/sso\/easy\?t=([A-Za-z0-9]{64})/);
```
devinbinnie: We should avoid a regex here, they are susceptible to path traversal. Instead, we should use the `isUrlType` function above and match on that. It's been reviewed by security so it should be safe :)

src/common/utils/url.ts
```diff
@@ -113,3 +113,25 @@ const equalUrlsIgnoringSubpath = (url1: URL, url2: URL, ignoreScheme?: boolean)
...
+export const isEasySSOLoginURL = (url: string): boolean => {
+    try {
+        let urlToTest = url;
+        if (!url.startsWith('http://') && !url.startsWith('https://')) {
+            urlToTest = `https://${url}`;
+        }
+        const parsed = parseURL(urlToTest);
+
+        // Check pathname and search params
+        if (parsed?.pathname === '/login/one_time_link') {
```
devinbinnie: This should also use `isUrlType`. We should also check for subpath on this one too.

src/app/navigationManager.ts
```diff
@@ -70,10 +80,11 @@ export class NavigationManager {
                 webContentsView.load(urlWithSchema);
             }
         } else if (ServerManager.hasServers()) {
-            ServerHub.showNewServerModal(`${parsedURL.host}${getFormattedPathName(parsedURL.pathname)}${parsedURL.search}`);
+            ServerHub.showNewServerModal(`${parsedURL.host}${trimmedPathname}${parsedURL.search}`);
```
devinbinnie: Was `getFormattedPathName` not necessary here anymore? I think we needed that to ensure that we matched on the correct `URL` object (since it always appends it when doing `toString()`)

src/app/serverHub.ts
```diff
@@ -191,8 +191,13 @@ export class ServerHub {
         e: IpcMainInvokeEvent,
         url?: string,
         currentId?: string,
+        originalUrl?: string,
```
devinbinnie, line 194: This doesn't look like it's being used anywhere, can we remove it?

devinbinnie (changes requested):

src/app/navigationManager.ts
larkox: Not my code, but my reasoning around this is...

We don't use `parsedURL.toString()` because we want to remove the protocol (don't know why, but I see no other reason).

And we are joining host + path + searchquery. If path has a slash at the end it feels weird (`www.website.com/something/?a=b` instead of `www.website.com/something?a=b`).

No strong feelings here, so happy revert.

src/app/navigationManager.ts
```diff
@@ -57,11 +58,15 @@ export class NavigationManager {
                 return;
             }
 
-            const urlWithSchema = `${server.url.origin}${getFormattedPathName(parsedURL.pathname)}${parsedURL.search}`;
+            const urlWithSchema = `${server.url.origin}${parsedURL.pathname}${parsedURL.search}`;
             if (webContentsView.isReady() && ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion && Utils.isVersionGreaterThanOrEqualTo(ServerManager.getRemoteInfo(webContentsView.serverId)?.serverVersion ?? '', '6.0.0')) {
                 const formattedServerURL = `${server.url.origin}${getFormattedPathName(server.url.pathname)}`;
                 const pathName = `/${urlWithSchema.replace(formattedServerURL, '')}`;
-                webContentsView.sendToRenderer(BROWSER_HISTORY_PUSH, pathName);
+                if (isMagicLinkUrl(server.url, parsedURL)) {
```
devinbinnie, line 69: A comment here as to why we make this exception could be helpful :)

I'll probably forget why we did this otherwise 🤣

larkox (approved):

larkox:
@lindalumitchell This is ready to review, to get it into the latest version of desktop 😄

larkox:
@amyblais Not sure if this has to be cherrypicked to the release branch.

amyblais:
> Not sure if this has to be cherrypicked to the release branch.

@larkox No cherry-pick needed.

PR #3545, opened 2025-10-29
[MM-66396] Reset and clear server theme when the app logs out

devinbinnie:
#### Summary
When the application logs out and the user is using theme syncing, the theme will change to the default light theme that the web app uses, which for dark mode users can be a bit jarring.

This PR forces the theme to reset when the user logs out so that it will continue to respect dark mode for those using it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66396

```release-note
NONE
```

### DIFF
--- a/src/common/servers/serverManager.ts
+++ b/src/common/servers/serverManager.ts
@@ -101,9 +101,15 @@ export class ServerManager extends EventEmitter {
         if (!server) {
             return;
         }
+        if (!loggedIn) {
+            server.theme = undefined;
+        }
         server.isLoggedIn = loggedIn;
         this.servers.set(serverId, server);
         this.emit(SERVER_LOGGED_IN_CHANGED, serverId, loggedIn);
+        if (!loggedIn) {
+            this.emit(SERVER_THEME_CHANGED, serverId);
+        }
     };
 
     private createServer = (server: Server, isPredefined: boolean, initialLoadURL?: URL) => {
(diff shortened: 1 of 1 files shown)


src/common/servers/serverManager.ts
```diff
@@ -101,9 +101,15 @@ export class ServerManager extends EventEmitter {
...
             return;
         }
+        if (!loggedIn) {
+            server.theme = undefined;
+        }
         server.isLoggedIn = loggedIn;
         this.servers.set(serverId, server);
         this.emit(SERVER_LOGGED_IN_CHANGED, serverId, loggedIn);
+        if (!loggedIn) {
+            this.emit(SERVER_THEME_CHANGED, serverId);
```
larkox, line 111: Why not emit this event inside the same if as setting the theme to undefined? That way is more tightly coupled and easier to read.

src/common/servers/serverManager.ts
larkox, line 1: Do we want to add tests for this?

src/common/servers/serverManager.ts
devinbinnie: It's a timing thing, we emit events after all of the variables have been updated.

It's a bit confusing I know but it avoids any possible data conflicts.

src/common/servers/serverManager.ts
devinbinnie: I will revisit testing for this module at some point, seeing as we're missing tests for most of the methods.

PR #3546, opened 2025-10-29
[MM-66211] Fixed UX issues with the checkbox component

devinbinnie:
#### Summary
The checkbox component had two issues: horizontal shifting when toggling due to the icon element changing from 0px to 16px width, and the label not being clickable because it was only connected to a disabled input element instead of the actual click handler.

This PR fixes both issues by constraining the checkbox button with `max-width`, `min-width`, and `flex-shrink: 0` to prevent horizontal shifting while preserving the 0px-to-16px icon animation, and adds an `onClick` handler to the label element to make it clickable for toggling the setting.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66211

```release-note
Fixed UX issues with the checkbox component
```

### DIFF
--- a/src/renderer/components/SettingsModal/components/CheckSetting.scss
+++ b/src/renderer/components/SettingsModal/components/CheckSetting.scss
@@ -63,6 +63,7 @@
         .CheckSetting__label {
             font-size: 14px;
             line-height: 20px;
+            cursor: pointer;
 
             .CheckSetting__sublabel {
                 margin-top: 4px;


src/renderer/components/SettingsModal/components/CheckSetting.scss
matthewbirtch, line 63: @devinbinnie can we add `cursor: pointer` to the label?

matthewbirtch (approved):
Fix looks good. Just one minor comment about the cursor.

src/renderer/components/SettingsModal/components/CheckSetting.scss
devinbinnie: Done :)

PR #3584, opened 2025-11-20
[MM-66719] Fix issues with server manager and predefined servers

devinbinnie:
#### Summary
Architectural changes to the `ServerManager` caused some issues, especially with predefined servers (including GPO configured ones).

This PR fixes:
- If the last selected server was a predefined server, the `lastActiveServer` in the config would be set to `-1` and would fail validation when the app reloads.
- Clearing data for a predefined server would do nothing
- Removing the current server if user configured would not reliably bring you to the previous server

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66719

```release-note
Fixed an issue where quitting the app while the last server set was a GPO server would corrupt the config file
```

### DIFF
--- a/src/common/servers/serverManager.ts
+++ b/src/common/servers/serverManager.ts
@@ -321,7 +321,7 @@ export class ServerManager extends EventEmitter {
         // Set the current server based on config if the user last used a local server
         // Otherwise, just use the first server
         let currentServerIndex = Config.lastActiveServer ?? 0;
-        if (currentServerIndex >= initialServers.length - 1) {
+        if (currentServerIndex > initialServers.length - 1) {
             currentServerIndex = initialServers.length - 1;
         }
 


src/common/servers/serverManager.ts
```diff
@@ -313,10 +321,8 @@ export class ServerManager extends EventEmitter {
         // Set the current server based on config if the user last used a local server
         // Otherwise, just use the first server
         let currentServerIndex = Config.lastActiveServer ?? 0;
-        if (Config.localServers.length && Config.predefinedServers.length) {
-            currentServerIndex += Config.predefinedServers.length;
-        } else if (Config.predefinedServers.length) {
-            currentServerIndex = 0;
+        if (currentServerIndex >= initialServers.length - 1) {
```
wiggin77: Should this be ">=" and "-1"?  Feels like an off by one.

wiggin77: Also, the PR description suggests "Config.lastActiveServer" can be -1.   In that case the conditional equates to false, and currentServerIndex remains -1.

wiggin77 (changes requested):

src/common/servers/serverManager.ts
devinbinnie: I should clarify: `Config.lastActiveServer` CANNOT be -1. See: https://github.com/mattermost/desktop/blob/723c626b55ddd72a49958b087c2892ba5d93eda5/src/common/Validator.ts#L179

The validation would fail and (what I neglected to mention) the config would be reset and the value would be `0`.

devinbinnie: Also the `currentServerIndex` is 0-indexed, so there's no bug here, but the extra `=` is unnecessary, so I'll remove that.

PR #3598, opened 2025-12-02
[MM-66759] Fixes to browser push logic

devinbinnie:
#### Summary
I've made a few fixes around the browser push logic that broke in v6:
- The calls widget was not account for, so links in the calls widget did not open in an external tab
- When popping out a thread from a window, redirection would not focus the popout window

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66759

```release-note
Fixed an issue where links in the calls widget did not open in an external tab
Fixed an issue when popping out a thread from a window, redirection would not focus the pop-out window
```

### DIFF
--- a/src/app/navigationManager.ts
+++ b/src/app/navigationManager.ts
@@ -136,18 +136,18 @@ export class NavigationManager {
         log.debug('handleBrowserHistoryPush', {webContentsId: e.sender.id});
 
         const callsViewId = CallsWidgetWindow.isCallsWidget(e.sender.id) && CallsWidgetWindow.mainViewId;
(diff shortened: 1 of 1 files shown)


src/app/navigationManager.ts
```diff
@@ -152,12 +156,23 @@ export class NavigationManager {
             cleanedPathName = pathName.replace(server.url.pathname, '');
         }
 
-        if (currentView.parentViewId) {
-            if (!MainWindow.get()?.isFocused()) {
-                MainWindow.get()?.focus();
+        if (currentView.parentViewId || callsViewId) {
+            currentView = currentView.parentViewId ? WebContentsManager.getView(currentView.parentViewId) : currentView;
```
davidkrauser: **suggestion:** we're making `currentView` do a lot of work, and it's hard to follow. Could we break it up into a few different variables with more specific names?

PR #3609, opened 2025-12-04
[MM-66704] Remove MAS migration logic

devinbinnie:
#### Summary
We added logic to migrate users over to the Mac App Store build in v5.1, back when we we introduced that released. This was done to make sure users didn't need to re-add their servers when they installed via Mac App Store vs the original DMG. However, some current users still see this prompt pop up if they've ever installed the DMG build despite not needing a migration, since they've been using the MAS version for years.

This PR removes the dialog and logic to migrate the config to Mac App Store.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66704

```release-note
Remove MAS migration logic
```


i18n/ar.json
```diff
@@ -40,10 +40,6 @@
   "main.app.app.handleAppCertificateError.certNotTrusted.dialog.title": "الشهادة غير موثقة",
   "main.app.app.handleAppCertificateError.dialog.extraDetail": "الشهادة مختلفة عن السابقة.\n\n",
   "main.app.initialize.downloadBox.allFiles": "كل الملفات",
-  "main.app.utils.migrateMacAppStore.button.dontImport": "لا تستورد",
```
lieut-data, line 43: The contract with weblate is that we don't touch non-`en.json` and "they" don't touch `en.json` -- this avoids the merge conflicts issues that we previously encountered. 

I don't know what might happen if we merge this. It /might/ be fine? But in general we should ignore the non-English translation files entirely and leave the source of truth for those to weblate.

lieut-data (approved):

i18n/ar.json
devinbinnie: Yeah I was wondering about that. If we've said that we shouldn't touch those files though I'll remove the changes from those.

larkox (approved):

PR #3610, opened 2025-12-04
[MM-66809] Add Sentry integration to the Desktop App

devinbinnie:
#### Summary
This PR integrates Sentry error tracking into the Electron application to capture exceptions from the `CriticalErrorHandler`. This will track crashes in production and prerelease builds and allow us to track them via Sentry. 

The change also includes a user setting in the Settings Modal to enable or disable Sentry reporting (defaulting to enabled) for privacy reasons.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66809

```release-note
Add Sentry error tracking
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,8 @@
       "__IS_NIGHTLY_BUILD__": false,
       "__IS_MAC_APP_STORE__": false,
       "__DISABLE_GPU__": false,
-      "__SKIP_ONBOARDING_SCREENS__": false
+      "__SKIP_ONBOARDING_SCREENS__": false,
+      "__SENTRY_DSN__": "https://test@sentry.io/123"
     },
     "setupFiles": [
       "./src/jest/jestSetup.js"
@@ -160,6 +161,7 @@
     "@emotion/react": "11.11.4",
     "@floating-ui/react": "0.26.28",
     "@mattermost/compass-icons": "0.1.45",
+    "@sentry/electron": "7.4.0",
     "auto-launch": "5.0.6",
     "classnames": "2.5.1",
     "electron-context-menu": "4.0.4",
--- a/src/common/config/defaultPreferences.ts
+++ b/src/common/config/defaultPreferences.ts
@@ -52,6 +52,7 @@ const defaultPreferences: ConfigV4 = {
     startInFullscreen: false,
     logLevel: 'info',
     enableMetrics: true,
+    enableSentry: true,
     viewLimit: 15,
     themeSyncing: true,
 };
--- a/src/main/sentryHandler.ts
+++ b/src/main/sentryHandler.ts
@@ -1,0 +1,99 @@
+// Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import os from 'os';
+
+import * as Sentry from '@sentry/electron/main';
+import {app} from 'electron';
+
+import Config from 'common/config';
+import {Logger} from 'common/log';
+
+const log = new Logger('SentryHandler');
+
+export class SentryHandler {
+    private sentryInitialized = false;
+
+    init = () => {
+        if (!this.shouldSendToSentry()) {
+            return;
+        }
+
+        // eslint-disable-next-line no-undef
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        const sentryDsn = __SENTRY_DSN__;
+        if (!sentryDsn || typeof sentryDsn !== 'string' || sentryDsn.length === 0) {
+            log.warn('Sentry is enabled, but not configured (no DSN provided)');
+            return;
+        }
+
+        const isPrerelease = this.isPrereleaseBuild();
+        Sentry.init({
+            dsn: sentryDsn,
+            sendDefaultPii: false,
+            environment: isPrerelease ? 'prerelease' : 'stable',
+            attachStacktrace: true,
+        });
+
+        this.addSentryContext();
+        this.sentryInitialized = true;
+        log.info('Sentry initialized for error tracking', {isPrerelease});
+    };
+
+    captureException = (error: Error) => {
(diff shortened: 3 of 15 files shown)


src/main/sentryHandler.ts
```diff
@@ -0,0 +1,99 @@
...
+            return;
+        }
+
+        Sentry.captureException(error);
+    };
+
+    private addSentryContext = () => {
+        if (!this.shouldSendToSentry()) {
+            return;
+        }
```
hanzei, line 60: At this point, we know this check is false. Can we remove it?

src/main/sentryHandler.ts
```diff
@@ -0,0 +1,99 @@
...
+    private sentryInitialized = false;
+
+    init = () => {
+        if (!this.shouldSendToSentry()) {
+            return;
+        }
+
+        // eslint-disable-next-line no-undef
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
hanzei, line 24: Do we really need the linter exception, or can we declare `__SENTRY_DSN__` somewhere?

src/main/sentryHandler.ts
```diff
@@ -0,0 +1,99 @@
...
+        const sentryDsn = __SENTRY_DSN__;
+        if (!sentryDsn || typeof sentryDsn !== 'string' || sentryDsn.length === 0) {
+            log.warn('Sentry is enabled, but not configured (no DSN provided)');
+            return;
+        }
+
+        const isPrerelease = this.isPrereleaseBuild();
+        Sentry.init({
```
hanzei, line 32: Do we track the app version somewhere? If not, should we?

src/common/config/defaultPreferences.ts
```diff
@@ -52,6 +52,7 @@ const defaultPreferences: ConfigV4 = {
     startInFullscreen: false,
     logLevel: 'info',
     enableMetrics: true,
+    enableSentry: true,
```
hanzei, line 55: I know we would like to get the data, but we should follow the principle of "gain trust" here. Requesting sentry data (including system information!) by default doesn't build trust.

I suggest we ask users on the first start if they want to provide the data similar to what other apps, e.g., Firefox do 

Image

src/main/sentryHandler.ts
devinbinnie: This is unfortunately how the webpack variables work, you can see other examples in the application.

We could throw them all in a constants file perhaps, but that's outside the scope of this PR. Maybe I should do it in a follow up :)

src/main/sentryHandler.ts
devinbinnie: See below the `Sentry.setContext` calls, that sets the app version.

src/common/config/defaultPreferences.ts
devinbinnie: I don't think there's a huge issue of trust here - we're not gathering anything remotely identifiable.

Plus - I don't think we'll end up receiving any data if the setting isn't turned on for people by default because it's not particularly discoverable and there's no immediate incentive for them to do so.

In an ideal world where everyone is aware that crash data is important to us and would happily turn on the setting, I would totally agree to have it default off, but most of our users won't look here unless we specifically instruct them to do so, so I think we're okay to leave as-is.

src/main/sentryHandler.ts
devinbinnie: Technically it can change while the app is running (even though the settings modal says otherwise) because it's dependent on the Config variable.

I don't see a huge harm in leaving it here.

src/main/sentryHandler.ts
hanzei: I would love to see a holistic follow-up that moves them into a const file :+1:

src/main/sentryHandler.ts
hanzei: Thanks! Do we also track the MM server version?

src/common/config/defaultPreferences.ts
hanzei: What information do the stack traces include? Do they include full file path, which might include user names and host names?

> because it's not particularly discoverable

I would love to change that. A startup dialogue could greatly enhance discoverability. Do you know if there is bandwidth to implement that in a follow-up ticket?

devinbinnie: > What information do the stack traces include? Do they include full file path, which might include user names and host names?

Nope, just paths with relative path names to the application.

Here's a screenshot of what we get:
image

> I would love to change that. A startup dialogue could greatly enhance discoverability. Do you know if there is bandwidth to implement that in a follow-up ticket?

We could do it, but I don't think we should. Dialogs for settings that again, most people don't care about because there's no incentive for them, will just annoy people. They just want to read their messages. I think nagging them would probably reduce the number of people that turn on the setting.

src/main/sentryHandler.ts
devinbinnie: We don't - the exceptions always come from the main process which are independent of a specific server. Usually anything coming from these exceptions should be reasonably actionable in the Desktop App itself.

package.json
```diff
@@ -89,7 +89,8 @@
       "__IS_NIGHTLY_BUILD__": false,
       "__IS_MAC_APP_STORE__": false,
       "__DISABLE_GPU__": false,
-      "__SKIP_ONBOARDING_SCREENS__": false
+      "__SKIP_ONBOARDING_SCREENS__": false,
+      "__SENTRY_DSN__": "https://test@sentry.io/123"
```
pavelzeman, line 93: ignore my previous comment, I now understand that DSN stands for "Data Source Name" - no typo here.

pavelzeman:
This is great! Thank you.

With more of us running nightly builds, we should be able to proactively spot, catch & fix crashing issues well before a release to customers.

Let's give this some more audience (maybe a demo in Thu dev meeting) when it's merged, so we get QA, SDETs and other engineers looking at Sentry more frequently.

PR #3656, opened 2026-01-19
[MM-67210] Add in-app notice for auto-update deprecation

devinbinnie:
#### Summary
This PR adds a deprecation notice for users of the auto-updater, that it will be removed in v6.1. This will be pushed out with a v6.0 dot release. See screenshots below for what the notice looks like.

NOTE: This will all be removed along with the auto-updater in v6.1, so as a note for reviewers, don't focus too much on the structure here as it's more important that it works and gets out quickly for our users.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67210

#### Screenshots
image
image

```release-note
Add in-app notice for auto-update deprecation for users who have auto updates enabled
```


yasserfaraazkhan (approved):

devinbinnie:
> Nice but unfortunately the microsoft store seems to be completely outdated and just downgraded my mattermost right away :-(

We'll be fixing this ASAP :)

PR #3664, opened 2026-01-26
[MM-67027][MM-67028][MM-54413] Remove NSIS installer, replace auto update with in-app notifications

devinbinnie:
#### Summary
This PR switches away from the `electron-updater` package that provides automatic updates within the application for the Windows EXE installer and the Linux AppImage, and replaces the system with an in-app notification system that provides the same functionality, but requires manual download.

When updates are available, our application will display a familiar pop-up box in the Downloads dropdown that will provide links on where to download new versions of the app. A user may choose to "Skip This Version" if they don't want to see the updates anymore, however they can get the prompt back if they manually check for updates via the menu or Settings modal. More information for each OS is in the Screenshots table below.

This approach has some advantages to it:
- Allows us to remove the less secure Windows EXE installer that also conflicts with our MSI, allowing for one single, stable Windows installer. This also avoids some headaches that users experienced with the EXE installer either not closing the app properly, corrupting the configuration, or not updating at all.
- Provides automatic update notifications for ALL delivery systems, using our website as a single source of truth. This includes the Mac App Store, and the Windows MSI builds (note: Admins deploying the MSI can continue to turn off the update notifications via GPO)

There are a couple downsides to this approach:
- Users will have to manually run the installer themselves, as it is delivered via a web browser. However, we feel this is a common practice that shouldn't create too many headaches.
- Linux users may have different ways of deploying the application which we're unable to detect, so the current plan is to direct them to the GitHub release. We plan to work on this in the future to make this much easier.

In addition to this change, we will stop shipping the Windows EXE installer altogether to avoid conflicting installations, which supports the change made here: https://github.com/mattermost/desktop/pull/3659

Another small change: We have changed the name of the macOS DMG for M-series chips to `arm64` to match the architecture.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54413
https://mattermost.atlassian.net/browse/MM-67207
https://mattermost.atlassian.net/browse/MM-67208

#### Device Information
This PR was tested on:
- Windows 11 25H2
- macOS Tahoe 26 (Mac App Store + DMG)
- Ubuntu 24.04 LTS

#### Screenshots
|Mac (DMG)|Mac (Mac App Store)|Windows|Linux|
|-|-|-|-|
|The basic DMG installation directs the user to click the Download Update button, which opens their web browser and downloads the update for them from our website.|The Mac App Store installation is detectable by Electron, so we can have the user "Open Mac App Store" and update from there.|We cannot detect how the MSI is installed for Windows users, however we will direct them to use the Windows Store when possible. However, there is also an option to Download Manually which will open a web browser and download the update from our website.|Linux users may have their app deployed in a few different ways, so we direct them to the GitHub release page so they can choose the release that suits them best (see above)|
|Screenshot 2026-01-27 at 11 54 03 AM|Screenshot 2026-01-27 at 11 54 45 AM|Screenshot 2026-01-27 at 11 51 22 AM|Screenshot 2026-01-27 at 11 52 17 AM|

Dropdown menu:
Screenshot 2026-01-27 at 11 54 08 AM

#### Release Note
```release-note
Remove the Windows EXE installer
Changed the name of the macOS ARM64 DMG from `m1` to `arm64`
Added in-app notifications for new versions, including for Mac App Store and the Windows MSI
```


src/main/updateNotifier.ts
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        downloadsManager.removeUpdateBeforeRestart();
+        ipcMain.emit(NO_UPDATE_AVAILABLE);
+        log.info('Version skipped', {version: versionToSkip});
+    };
+}
+
+const updateNotifier = new UpdateNotifier();
+export default updateNotifier;
```
marianunez, line 272: Should we guard here for possible errors at the time of initialization of this singleton? It could crash the desktop app if something fails here

src/main/updateNotifier.ts
```diff
@@ -0,0 +1,258 @@
...
+        const currentVersion = app.getVersion();
+        const baseURL = buildConfig.updateNotificationURL;
+
+        let filename = 'latest.txt';
+        const versionMatch = currentVersion.match(/^[0-9]+\.[0-9]+\.[0-9]+-([^.]+)/);
+        if (versionMatch) {
+            filename = `${versionMatch[1]}.txt`;
+        }
+
+        const url = `${baseURL}/${filename}`;
```
marianunez, line 147: Should we do any url validation here in case of a configuration error?

src/main/updateNotifier.ts
```diff
@@ -0,0 +1,258 @@
...
+
+    constructor() {
+        ipcMain.on(CHECK_FOR_UPDATES, () => this.checkForUpdates(true));
+        ipcMain.on(OPEN_WINDOWS_STORE, this.openWindowsStore);
+        ipcMain.on(DOWNLOAD_UPDATE_MANUALLY, this.downloadUpdateManually);
+        ipcMain.on(OPEN_UPDATE_GUIDE, this.openUpdateGuide);
+        ipcMain.handle(GET_IS_MAC_APP_STORE, () => this.isMacAppStore());
+        ipcMain.on(OPEN_MAC_APP_STORE, this.openMacAppStore);
+        ipcMain.on(SKIP_VERSION, this.skipVersion);
+    }
```
marianunez, line 58: Should we add a destroy for this singleton to clean up ? Cursor suggest mostly for testing that recreates the `UpdateNotifier` and the handlers add up 0/5

src/main/updateNotifier.ts
devinbinnie: This happens at boot time, so it's very unlikely this will crash the app. Other singletons are instantiated the same way.

The ideal case here is to introduce dependency injection, but that's a long way away at this point.

src/main/updateNotifier.ts
devinbinnie: Again, this should live for the entire time the app is running, and it will get cleaned up when it closes. It should never be reinstantiated.

src/main/updateNotifier.ts
devinbinnie: There isn't really much to validate here, if there's an issue it will simply return a bad response from the `net.fetch` and fail with a log message, which I would assume is ideal if the file is missing. It's also wrapped in try/catch to avoid crashing :)

.github/workflows/nightly-main.yml
```diff
@@ -181,10 +179,7 @@ jobs:
           zsh -c 'echo -n $APPLE_API_KEY_RAW > ./key.p8'
           mkdir -p ./build/macos-release
           npm run package:mac-with-universal
-          bash -x ./scripts/patch_updater_yml.sh
           bash -x ./scripts/cp_artifacts.sh release ./build/macos-release
-      - name: nightly/rename-arm64-to-m1
-        run: rename 's/arm64/m1/' ./build/macos-release/$(jq -r .version package.json)/*
```
marianunez, line 187: Would this name change break anything in our release pipelines?

src/common/config/buildConfig.ts
```diff
@@ -33,7 +33,12 @@ const buildConfig: BuildConfig = {
     academyLink: DEFAULT_ACADEMY_LINK,
     upgradeLink: DEFAULT_UPGRADE_LINK,
     enableServerManagement: true,
-    enableAutoUpdater: true,
+    enableUpdateNotifications: true,
+    updateNotificationURL: 'https://releases.mattermost.com/desktop',
+    macAppStoreUpdateURL: 'macappstore://apps.apple.com/us/app/mattermost-desktop/id1614666244',
+    windowsStoreUpdateURL: 'ms-windows-store://pdp/?productid=XP8BR8MH3LPKLT',
+    linuxUpdateURL: 'https://docs.mattermost.com/deployment-guide/desktop/linux-desktop-install.html',
+    linuxGitHubReleaseURL: 'https://github.com/mattermost/desktop/releases/tag/v',
```
marianunez, line 41: We should include these also as default constants as the other links above to be consistent

src/renderer/components/DownloadsDropdown/Update/UpdateAvailable.tsx
```diff
@@ -1,45 +1,116 @@
...
     item: DownloadedItem;
     appName: string;
 }
 
-const UpdateAvailable = ({item, appName}: OwnProps) => {
-    const [isProcessing, setIsProcessing] = useState(false);
-    const onButtonClick = (e: React.MouseEvent<HTMLButtonElement>) => {
-        if (isProcessing) {
-            return;
+const UpdateAvailable = ({item, appName, activeItem}: OwnProps) => {
```
marianunez, line 18: Any tests we can add for this component and the button actions?

src/renderer/components/DownloadsDropdown/Update/UpdateAvailable.tsx
```diff
@@ -49,35 +120,42 @@ const UpdateAvailable = ({item, appName}: OwnProps) => {
...
                     className='primary-button DownloadsDropdown__Update__Details__Button'
-                    onClick={onButtonClick}
-                    disabled={isProcessing}
+                    onClick={handleMainButtonClick}
                 >
-                    <LoadingWrapper
-                        loading={isProcessing}
-                        text={(
+                    {getMainButtonText()}
+                </button>
```
marianunez, line 138: Can we include aria elements for accessibility of this notice?

.github/workflows/nightly-main.yml
devinbinnie: It might, I see a reference to it here: https://github.com/mattermost/delivery-platform/blob/6d012f6e40621046b1485eeec64dda16c71aee6d/.github/actions/install-url-tool/main.go#L51

cc @NARSimoes

src/common/config/buildConfig.ts
devinbinnie: I'm kind of back and forth on how I want to approach this. On one hand, we want our app to be easy to custom brand so it's good to have a single point of change to do that. On the other, a lot of these are somewhat core to the application, so they went into constants.

I think I'll figure this out and clean it up in an upcoming change.

src/renderer/components/DownloadsDropdown/Update/UpdateAvailable.tsx
devinbinnie: No front-end tests in the Desktop App yet :( that still needs to be set up.

.github/workflows/nightly-main.yml
NARSimoes: Thank you for checking it. If I understood correctly in the current PR we change `$(print_link "${BASE_URL}/mattermost-desktop-${VERSION}-mac-m1.dmg")` to `$(print_link "${BASE_URL}/mattermost-desktop-${VERSION}-mac-arm64.dmg")` as generated output - [here](https://github.com/mattermost/desktop/blob/5e6164e6177766b9dc8484399a8c7f96b6896088/scripts/generate_release_markdown.sh#L32).

The `map` mentioned above is used to filter the files that we have in the bucket. For `dmg` we support all the generated outputs - `-mac-universal-dmg`, `-mac-x64-dmg`, `-mac-m1-dmg` and `-amd64-deb`. So, in this case we already support the new suffix and if we don't find any file with `-mac-m1-dmg` we just continue (we can consider updating the suffixes list in a follow up).

larkox (approved):

PR #3683, opened 2026-02-04
Update package.json

amyblais:
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,7 +84,7 @@
     },
     "api-types": {
       "name": "@mattermost/desktop-api",
-      "version": "6.1.0-1",
+      "version": "6.0.0-1",
       "dev": true,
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -84,7 +84,7 @@
     },
     "api-types": {
       "name": "@mattermost/desktop-api",
-      "version": "6.0.0-1",
+      "version": "6.1.0-1",
```
amyblais: I wasn't sure if I'm expected to update this and whether I updated it correctly.

devinbinnie: I don't think we've changed this one since v6.0. I usually don't update it unless we're making a change to it, so we can revert this one.

amyblais:
Assigning to Devin as I don't have merge permissions.

PR #3692, opened 2026-02-11
[MM-46859] Refactor RegistryConfig into policyConfigLoader, include macOS CFPref support

devinbinnie:
#### Summary
This PR introduces the `policyConfigLoader` to replace the `RegistryConfig` module, in which there is a single point of truth for organization level configuration management items. On top of this, we've added support for macOS MDM configuration using `cf-prefs`, a native module that is able to read preference files added by the user's organization to preconfigure servers, and set the same configuration items we use today with GPO.

Included is a sample .plist file and instructions on how to install it for testing.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46859

```release-note
Added support for macOS MDM
```


patches/cf-prefs+2.0.1.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/cf-prefs/index.js b/node_modules/cf-prefs/index.js
+index b86c7ba..e816a58 100644
+--- a/node_modules/cf-prefs/index.js
++++ b/node_modules/cf-prefs/index.js
+@@ -1,6 +1,8 @@
+-const preferences = require('bindings')('cf-prefs.node');
++const preferences = process.platform === 'darwin'
```
devinbinnie, line 7: We unforutunately need to do the same thing here that we do for `macos-notification-state`, since both of them assume that the machine they're loaded on is macOS instead of checking beforehand.

I think we should probably come up with a better way to export these modules, perhaps via a specific module that lazy loads them for us, but for now we'll just mimic what we've done in the past.

enahum: Perhaps a ticket to look into it?

src/common/config/policyConfigLoader.ts
```diff
@@ -0,0 +1,112 @@
...
+        // The first one is the local machine hive, which should take priority
+        const results = this.getWindowsValues()?.filter((r) => r.name === valueName);
+        return results?.pop()?.data;
+    };
+
+    private getWindowsValues = (key?: string) => {
+        const results: RegistryValue[] = [];
+        for (const hive of WINDOWS_REGISTRY_HIVE_LIST) {
+            try {
+                results.push(...enumerateValues(hive, key ? `${WINDOWS_REGISTRY_PATH}\\${key}` : WINDOWS_REGISTRY_PATH));
```
enahum, line 93: Ouch, but I guess this is the way the lib works so trust internals I guess

enahum (approved):
Lgtm

patches/cf-prefs+2.0.1.patch
devinbinnie: https://mattermost.atlassian.net/browse/MM-67580

src/common/config/policyConfigLoader.ts
devinbinnie: Yep, windows registry and such lol

larkox (approved):

yasserfaraazkhan (approved):

PR #3695, opened 2026-02-11
Fix Linux cross-compiler scripts

devinbinnie:
#### Summary
Our Linux build scripts are set up to run cross-compile for ARM64, but not on x64. If you're like me and run Linux in a VM on an Apple Silicon Mac, this means you have to go edit the package.json to get the compiler to work properly.

However, since the flags are ignored if you're on the same arch, this PR just forces them on regardless of the arch so that it works without editing.

```release-note
NONE
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "package:linux:arch-arm64": "cross-env CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ electron-builder --linux",
     "package:linux-tar": "npm run build-prod && npm run package:linux:arch-x64 -- tar.gz --x64 --publish=never && npm run package:linux:arch-arm64 -- tar.gz --arm64 --publish=never",
     "package:linux-pkg": "npm run build-prod && npm run package:linux:arch-x64 -- deb rpm flatpak --x64 --publish=never && npm run package:linux:arch-arm64 -- deb rpm flatpak --arm64 --publish=never",
-    "package:linux-appImage": "npm run build-prod && npm run package:linux-arch-x64 -- tar.gz appimage --x64 --publish=never && npm run package:linux-arch-arm64 -- tar.gz appimage --arm64 --publish=never"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -52,9 +52,11 @@
...
-    "package:linux-appImage": "npm run build-prod && electron-builder --linux tar.gz appimage --x64 --publish=never && cross-env CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ electron-builder --linux tar.gz appimage --arm64 --publish=never"
+    "package:linux:arch-x64": "cross-env CC=x86_64-linux-gnu-gcc CXX=x86_64-linux-gnu-g++ electron-builder --linux",
+    "package:linux:arch-arm64": "cross-env CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ electron-builder --linux",
+    "package:linux-tar": "npm run build-prod && npm run package:linux:arch-x64 -- tar.gz --x64 --publish=never && npm run package:linux:arch-arm64 -- tar.gz --arm64 --publish=never",
+    "package:linux-pkg": "npm run build-prod && npm run package:linux:arch-x64 -- deb rpm flatpak --x64 --publish=never && npm run package:linux:arch-arm64 -- deb rpm flatpak --arm64 --publish=never",
+    "package:linux-appImage": "npm run build-prod && npm run package:linux-arch-x64 -- tar.gz appimage --x64 --publish=never && npm run package:linux-arch-arm64 -- tar.gz appimage --arm64 --publish=never"
```
hmhealey: There's a typo here in `package:linux:arch-arm64` because there's some disagreement here between colons and hyphens.

Should we perhaps change the new targets to start with `linux-` to be consistent and avoid the chance of that in the future?

devinbinnie: New targets all do start with `linux-`. I missed one since that's a "modifier" effectively. Good catch :)

PR #3708, opened 2026-02-24
feat(e2e): Run E2E test on GPO/MDM configured github runners

yasserfaraazkhan:
### E2E Policy Test Suite

**Location:** `e2e/specs/policy/policy.test.js`

Added a new E2E suite to validate **Windows GPO** and **macOS MDM** policy enforcement.

Coverage includes:

* **Baseline behavior** with no policies applied
* **`EnableServerManagement=false` (Windows GPO)** blocking server add/edit/remove
* **`RestrictServerManagement` (macOS MDM)** enforcing equivalent restrictions

---

### Security Hardening – Block Dangerous Protocols

**File:** `allowProtocolDialog.ts`

Introduced a **blocked protocol list** to prevent unsafe URL scheme handling.

Blocked protocols include:
`file:`, `javascript:`, `data:`, `vbscript:`, `ms-msdt:`, `search-ms:`

Behavior:

* Requests using these protocols are **silently rejected**
* Prevents protocol execution or OS shell invocation

Tests added in `allowProtocolDialog.test.js`.

---

### External Link Routing Security

**Files:** `webContentEvents.ts`, `url.ts`

Improved external navigation safety by:

* Adding **protocol consent gating** in the `will-navigate` handler
* Strengthening **URL validation and error handling**
* Removing overly broad validation that blocked valid URLs with special characters

```release-note
NONE
```


e2e/specs/policy/policy.test.js
```diff
@@ -0,0 +1,415 @@
...
+    }
+}
+
+function setupMacOSPolicy({servers = [], enableServerManagement, enableAutoUpdater} = {}) {
+    // Use `defaults write` instead of writing raw plist XML so that changes go
+    // through the CFPreferences daemon — this avoids cache staleness issues where
+    // CFPreferencesCopyAppValue returns the old (empty) value even after the plist
+    // file has been written to disk.
+
+    const APP_ID = 'com.github.Electron';
```
devinbinnie: Nit: put this constant with the others

e2e/specs/policy/policy.test.js
```diff
@@ -0,0 +1,415 @@
...
+    try {
+        execFileSync('defaults', ['delete', APP_ID], {stdio: 'ignore'});
+    } catch (err) {
+        // Ignore — domain may not exist yet.
+    }
+
+    // Write DefaultServerList as an XML-format array using `defaults write`.
+    // Each server dict is written as a plist fragment accepted by `defaults write -array`.
+    if (servers.length > 0) {
+        const escapeXml = (value) => String(value).replace(/[&<>"']/g, (ch) => {
```
devinbinnie: This seems like a utility function. Perhaps we should add it to a utils file - or even just environment.js?

Another thought - is there a library that does this? Feels like we're reinventing the wheel here.

src/app/views/webContentEvents.ts
```diff
@@ -111,7 +111,22 @@ export class WebContentsEventManager {
                 return;
             }
 
-            this.log(webContentsId).info('Prevented desktop from navigating to external URL');
+            if (isMattermostProtocol(parsedURL)) {
```
devinbinnie: How come this was added? Do we have a bug in this flow?

e2e/specs/policy/policy.test.js
yasserfaraazkhan: I could use `xml-escape` but for just 1 thing I felt avoid it. I can move it to the top with constant as its just used here.

src/app/views/webContentEvents.ts
```diff
@@ -160,7 +168,9 @@ export class WebContentsEventManager {
 
             // Check for other custom protocols
             if (isCustomProtocol(parsedURL)) {
-                allowProtocolDialog.handleDialogEvent(parsedURL);
+                allowProtocolDialog.handleDialogEvent(parsedURL).catch((err) => {
```
devinbinnie, line 171: Is this throwing an error in your tests? If so, should we try and figure out the root cause of the error?

src/main/security/allowProtocolDialog.ts
```diff
@@ -16,6 +16,20 @@ import {allowedProtocolFile} from '../constants';
 
 const log = new Logger('AllowProtocolDialog');
 
+// Protocols that must never be opened via shell.openExternal regardless of user choice.
+// These are known attack vectors for remote code execution or local file access.
+const BLOCKED_PROTOCOLS = new Set([
```
devinbinnie, line 21: Good change, definitely compliments the work we did for the URL validation :)

Just to verify - this doesn't break anything under normal operation right?

yasserfaraazkhan: It didn't Devin
I did a manual check on mac and windows

src/app/views/webContentEvents.ts
yasserfaraazkhan: Sorry for the confusing diff! The duplicate call was a mistake that's been fixed in a subsequent commit — there's now only a single call with `.catch()`.

The `.catch()` itself is intentional: `handleDialogEvent()` is async and returns a `Promise<void>`. Without it, any rejection (e.g. `shell.openExternal` failing because no app is registered for the protocol) surfaces as an unhandled promise rejection, which in Node 15+ terminates the process. The catch logs the error gracefully instead of crashing.

.github/workflows/e2e-functional.yml
```diff
@@ -98,3 +98,125 @@ jobs:
...
+            runner: macos-26
+          - platform: windows
+            runner: windows-2025
+      fail-fast: false
+    runs-on: ${{ matrix.runner }}
+    defaults:
+      run:
+        shell: bash
+    env:
+      AWS_S3_BUCKET: "mattermost-cypress-report"
```
esarafianou, line 117: Just curious, is this the same bucket cypress in `mattermost` uses?

yasserfaraazkhan: @esarafianou Yes. The same is being used here.

Issue #3525, opened 2025-10-03
[Bug]: Handling MS OneNote deeplinks


Tabiskabis:
### Checks before filing an issue

- [x] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [x] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing. This should include closed issues.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [x] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [x] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

5.13.1

### Operating System

Windows 10

### Mattermost Server Version

10.11.3

### Steps to reproduce

Copy a Microsoft OneNote (classic offline version) link - specifically a deep link that references a certain page or paragraph - using the OneNote context-menu "Copy Link to Page"/Paragraf.
You don't need OneNote to reproduce this. Just go with my hypothetical examples below.

Paste the clipboard in a plain text editor. See something like this:
`onenote:///D:\OneNote\Apps\Mattermost.one#Schöne%20neue%20Seite&page-id={840EDD0C-B6FB-481E-A342-E39AEDA50EE6}`
Note the windows path delimiter backslashes.
Note that OneNote did URL-encode the spaces, but not the umlaut nor the curly brackets. 
Note that the URI does not contain a ?query, and the #fragment part is not rightmost.
It looks all wrong, but it is valid in Windows/OneNote.

Paste the link in Mattermost, yielding this markdown: 
`[Schöne neue Seite](onenote:///D:/OneNote/Apps/Mattermost.one#Schöne%20neue%20Seite&page-id={840EDD0C-B6FB-481E-A342-E39AEDA50EE6})` 
Notice the backslashes became regular (forward) slashes.
This is because the clipboard has an HTML formatted verison, that essentially contains a regular link with exactly this text and URI.

Also paste the previously seen plaintext version.
[...]

PR #3685, opened 2026-02-05
MM-65645: Allow parsing of custom or non-standard URLs to support Microsoft product links

wiggin77:
#### Summary
This PR allows custom protocol URLs (OneNote, MS Teams, Spotify, etc.) to work properly by handling non-RFC-3986-compliant URLs more leniently while maintaining security.

- Fixes handling of URLs that don't strictly comply with RFC 3986 but are valid in practice (e.g., Microsoft Teams, SharePoint, OneNote links)
- Adds lenient URL parsing for custom protocols that normalizes Windows backslash paths and encodes curly braces
- Replaces strict RFC validation with security-focused pattern detection for custom protocol URLs
- For http/https URLs that fail strict validation but can still be parsed, allows them through with security checks

#### Ticket Link
fixes https://mattermost.atlassian.net/browse/MM-65645
fixes https://github.com/mattermost/desktop/issues/3525

#### Checklist

- [X] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [X] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [X] completed [Mattermost Contributor Agreement](https://mattermost.com/contribute/)
- [X] executed `npm run lint:js` for proper code formatting
- [ ] Run E2E tests by adding label `E2E/Run`

#### Device Information
This PR was tested on: Ubuntu 24.04

#### Release Note
```release-note
Fixed an issue where Microsoft Teams, SharePoint, and OneNote links were incorrectly rejected as "Invalid Link" due to special characters in their URLs.
```

PR #3716, opened 2026-03-04
[MM-65645] Remove unnecessary URL validation for external links, focus only on where security issues exist and allow special characters to be handled as needed

devinbinnie:
#### Summary
The initial fix for MM-65645 ([#3685](https://github.com/mattermost/desktop/pull/3685)) addressed Microsoft product links by normalizing specific characters before strict RFC 3986 validation, but this approach required adding new workarounds for each non-standard URI pattern.

This PR replaces that approach by not enforcing the `isValidURI` check for external links, since it doesn't directly present a security issue. Instead, some additional changes were made that addresses where security issues could crop up while also allowing the URLs to work correctly:
- Fixed the `parseURL` duplicate-slash regex to preserve `scheme:///` patterns
- Always passes the serialized URL to `shell.openExternal` (percent-encoding dangerous characters like backticks, spaces, and quotes)
- Rejects URLs with null bytes

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65645
Closes https://github.com/mattermost/desktop/issues/3525

```release-note
Fixed further issues around URL parsing with special characters
```

## Change Impact: 🔴 High

**Reasoning:** The changes modify a shared URL utility (parseURL), remove normalization logic used across the codebase, and change a public API signature (allowProtocolDialog.handleDialogEvent), impacting navigation, protocol handling, and external-open flows across multiple modules and processes.

**Regression Risk:** High. Widespread parsing and external-opening paths are altered (including handling of special characters and null bytes), there's a breaking API change, and edge cases—especially interactions with native apps and non-standard URIs—may be insufficiently covered by tests.

** QA Recommendation:** Require extensive manual QA plus expanded automated tests. Prioritize validation of standard and non-standard protocols (http/https, mailto, onenote, custom), navigation/new-window/popup flows, shell.openExternal behavior with special and non-ASCII characters, null-byte rejection, and compatibility of protocol-dialog integrations. Do not skip manual QA.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/views/pluginsPopUps.ts
+++ b/src/app/views/pluginsPopUps.ts
@@ -93,7 +93,7 @@ export class PluginsPopUpsManager {
                 NavigationManager.openLinkInNewTab(parsedURL);
             } else {
                 // We allow to open external links through browser.
-                shell.openExternal(url);
+                shell.openExternal(parsedURL.toString());
             }
 
             log.warn('prevented popup window from opening window');
--- a/src/app/views/pluginsPopUps.test.js
+++ b/src/app/views/pluginsPopUps.test.js
@@ -141,7 +141,7 @@ describe('PluginsPopUpsManager', () => {
 
         // Verify opening external links is allowed through browser
         expect(handlers['window-open']({url: 'https://www.example.com'})).toEqual({action: 'deny'});
-        expect(shell.openExternal).toHaveBeenCalledWith('https://www.example.com');
+        expect(shell.openExternal).toHaveBeenCalledWith('https://www.example.com/');
 
         // Simulate render process gone
         handlers['render-process-gone'](null, {reason: 'oom'});


src/app/views/pluginsPopUps.ts
edgarbellot: @devinbinnie Should we do `shell.openExternal(parsedURL.toString());` here to make sure we are using the serialized form and not the raw URL?

devinbinnie: Probably not a concern since it should be caught by the other checks first, but definitely won't hurt to do so so let's do it.
```suggestion
                shell.openExternal(parsedURL.toString());
```

src/app/views/pluginsPopUps.ts
```diff
@@ -93,7 +93,7 @@ export class PluginsPopUpsManager {
                 NavigationManager.openLinkInNewTab(parsedURL);
             } else {
                 // We allow to open external links through browser.
-                shell.openExternal(url);
+                shell.openExternal(parsedURL.toString());
```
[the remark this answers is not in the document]
devinbinnie: @coderabbit This is not necessary. Resolve.

wiggin77 (approved):

Issue #3712, opened 2026-03-03
[Bug]: Mattermost Windows v6.0.4 Stuck on Blue Loading Screen (Server 11.3.0)


da5nsy:
### Checks before filing an issue

- [x] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [x] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing. This should include closed issues.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [x] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [x] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

6.0.4

### Operating System

Windows 10

### Mattermost Server Version

11.4.1

### Steps to reproduce

Open Mattermost

### Expected behavior

Mattermost opens and a page loads

### Observed behavior

It gets stuck on the loading screen

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/1d0db090-42b0-4f7d-a3b9-457c611df275" />

### Log Output

```shell
[older logs available]

[2026-02-20 10:07:10.504] [info]  [App.Config] config.autostart has been configured: { autostart: false }
[2026-02-20 10:07:10.504] [info]  [App.Config] config.autostart has been configured: { autostart: false }
[2026-02-20 10:07:11.563] [info]  [App.Config] config.autostart has been configured: { autostart: false }
[2026-02-20 10:07:14.718] [info]  [UpdateManager] Checking for update
[2026-02-20 10:07:15.248] [info]  [UpdateManager] Found version 6.0.4 (url: 6.0.4/mattermost-desktop-setup-6.0.4-win.exe)
[2026-02-20 10:07:15.252] [info]  [UpdateManager] New version available: 6.0.4
[2026-02-20 10:07:15.984] [info]  [App.Config] config.autostart has been configured: { autostart: false }
[...]

PR #3717, opened 2026-03-04
[MM-67761] Ensure loading screen doesn't attempt to re-add itself if already fading or removing

devinbinnie:
#### Summary
The loading screen can intermittently get stuck on first launch due to a race condition between `show()` and `fade()` in `LoadingScreen`. When `show()` is called while the loading screen HTML is still loading, it defers adding the view until `did-finish-load`. If the server view finishes loading and `fade()` fires before that callback, the deferred `did-finish-load` handler re-adds the loading screen view after it has already started fading, leaving it visible until the user switches servers or tabs.

This PR adds a state guard in the `did-finish-load` callback that skips showing the loading screen if the state is no longer `VISIBLE` (i.e., `fade()` has already been called).

I wasn't able to directly reproduce this issue since it's very intermittent, so I've added debug logging to the `show()` and `fade()` methods to help diagnose any other loading screen dismissal issues that may surface.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67761
https://github.com/mattermost/desktop/issues/3712

```release-note
Fix an issue where the loading screen might get stuck over top of the app
```

## Change Impact: 🟢 Low

**Reasoning:** The changes are isolated to the LoadingScreen component (a single UI module) and implement a defensive state guard to prevent a specific race condition. A new test directly validates this fix, with no modifications to public APIs, shared dependencies, or critical business logic paths.

**Regression Risk:** Minimal. The change adds a state check (`if (this.state !== LoadingScreenState.VISIBLE) { return; }`) in the did-finish-load callback that only exits early if the state is no longer VISIBLE—preventing the problematic re-add scenario. This guard does not alter the normal loading screen flow and purely prevents an edge case. Debug logging additions have no functional impact. Existing happy-path behavior is completely unaffected.

**QA Recommendation:** Light manual QA recommended. Focus on first-launch scenarios and rapid server/tab switching to verify the loading screen appears and dismisses properly under normal conditions. Given the addition of dedicated test coverage for the race condition and the isolated, defensive nature of the changes, skipping manual QA carries acceptable risk if standard first-launch verification is performed.

### DIFF
--- a/src/app/views/loadingScreen.ts
+++ b/src/app/views/loadingScreen.ts
@@ -51,7 +51,13 @@ export class LoadingScreen {
 
         if (this.view.webContents.isLoading()) {
             this.view.webContents.once('did-finish-load', () => {
+                if (this.state !== LoadingScreenState.VISIBLE) {
+                    return;
+                }
+
                 this.view.webContents.send(TOGGLE_LOADING_SCREEN_VISIBILITY, true);
+
+                log.debug('show: did-finish-load fired, adding loading screen view');
 
                 // Electron does a weird thing where even if the index is undefined, it will not add the view on top properly
                 if (condition?.()) {
@@ -62,6 +68,7 @@ export class LoadingScreen {
             });
         } else {
             this.view.webContents.send(TOGGLE_LOADING_SCREEN_VISIBILITY, true);
+            log.debug('show: not loading, adding loading screen view');
             if (condition?.()) {
                 this.parent.contentView.addChildView(this.view, 1);
             } else {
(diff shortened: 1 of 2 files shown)


src/app/views/loadingScreen.ts
```diff
@@ -51,8 +51,14 @@ export class LoadingScreen {
 
         if (this.view.webContents.isLoading()) {
             this.view.webContents.once('did-finish-load', () => {
+                if (this.state !== LoadingScreenState.VISIBLE) {
+                    return;
+                }
```
marianunez, line 56: Should we add a log here as well here that it's skipping to help see in the logs if the race condition happened?

devinbinnie: Missed this before merge :P

I think we're okay not to, I'm more interested going forward in when the child view is added and removed, which should help us diagnose if this issue happens for other reasons.

PR #3728, opened 2026-03-15
feat(e2e): Migrate Desktop E2E to Playwright and add merged cross-OS reporting

yasserfaraazkhan:
This PR completes the Mattermost Desktop E2E migration away from the legacy Electron test stack and standardizes on Playwright.

**It does three main things:**

- Migrates and stabilizes the Electron E2E suite around Playwright fixtures, shared readiness helpers, isolated userDataDirs, and embedded WebContentsView discovery for Mattermost server tabs.
- Cleans up legacy E2E infrastructure by removing the active electron-mocha/robotjs path and modernizing CI/reporting around Playwright artifacts.
- Improves test correctness and infra reliability by switching menu tests back to native Electron menu/edit actions and replacing shell-based orphan cleanup with PID-based teardown.

**Highlights**

- Playwright is now the active E2E runner.
- CI now publishes one merged cross-OS Playwright HTML report while keeping 3 OS-specific PR checks.
- Menu-bar coverage is stronger:
   - view_menu uses real View menu items
   - edit_menu uses native Electron edit commands instead of a page-local shim
- Test teardown is more robust:
  - test-mode Electron main processes register their PIDs
  - global teardown kills only registered test PIDs
- Downloads, startup, popup, server-management, and several menu-bar specs were stabilized and verified.

**Why this is better**

- More deterministic than the old OS-input-driven approach
- Better artifacts for debugging failures: Playwright HTML report, trace, video, screenshot behavior
- Better CI ergonomics with merged reporting across macOS, Linux, and Windows
- Less fragile cleanup and fewer orphaned Electron processes
- Better support for testing embedded Mattermost views that Playwright cannot see as normal windows

```release-note
NONE
```


src/main/notifications/Mention.ts
```diff
@@ -36,7 +36,8 @@ export class Mention extends Notification {
             Reflect.deleteProperty(options, 'icon');
         }
         const isWin7 = (process.platform === 'win32' && !Utils.isVersionGreaterThanOrEqualTo(os.release(), '6.3') && DEFAULT_WIN7);
-        const customSound = String(!options.silent && ((options.soundName !== 'None' && options.soundName) || isWin7));
+        const soundValue = !options.silent && ((options.soundName !== 'None' && options.soundName) || isWin7);
```
yasserfaraazkhan, line 39: The code built customSound like this conceptually:

`String(!options.silent && ((options.soundName !== 'None' && options.soundName) || isWin7))`

That forces the result through [String(...)](app://-/index.html?hostId=local#) no matter what.
So falsy values become strings like:
- ''
- 'false'
potentially other coerced string values depending on the expression result


**Why that was a bug:**

- customSound is used as a truthy/falsy indicator for whether a real custom sound exists.
- String coercion hides the original type and can produce misleading truthy values.
- In JavaScript, 'false' is truthy, which is not the same as false.
- That makes the code harder to reason about and can lead to subtle behavior mistakes around:
   - **if (customSound)** ...
   - what gets returned from getNotificationSound()

**What changed:**

The logic now does:

- compute soundValue
- set customSound = soundValue || ''

That preserves the intended semantics:

- real sound name if present
- DEFAULT_WIN7 when that fallback applies
- empty string when no sound should be used

After:

- No unwanted stringification of falsy values.
- “No sound” is represented consistently as ''.
- **if (customSound)** now means what it looks like: a real sound exists.

src/app/tabs/tabManager.ts
```diff
@@ -146,6 +146,7 @@ export class TabManager extends EventEmitter {
             this.tabOrder.set(serverId, viewIds);
         }
         this.emit(TAB_ORDER_UPDATED, serverId, viewIds);
+        MainWindow.window?.sendToRenderer(TAB_ORDER_UPDATED, serverId, viewIds);
```
yasserfaraazkhan, line 151: ### Before

* `TabManager.updateTabOrder()` updated main-process state and emitted `TAB_ORDER_UPDATED` **internally only**
* Renderer was **not notified**
* `MainPage` refreshed UI only on:

  * server switch
  * tab added
  * tab removed
  * some view/title events

➡️ **Tab reordering alone did not trigger a UI refresh**

---

### Why this was a bug

* Tab order source of truth = **main process**
* Renderer relied on **cached state**
* No renderer event → **UI could become stale**

**Result:**

* Reorder succeeded internally
* UI stayed outdated until another event triggered refresh

---

### What changed

* **`tabManager.ts`**

  * Sends `TAB_ORDER_UPDATED` to renderer via `MainWindow.window?.sendToRenderer(...)`

* **`internalAPI.js`**

  * Exposes `onTabOrderUpdated(...)`

* **`window.ts`**

  * Adds API typing

* **`MainPage.tsx`**

  * Subscribes to:

    ```ts
    onTabOrderUpdated(this.updateServers)
    ```

---

### After

* Reordering updates main-process state ✅
* Renderer is immediately notified ✅
* UI refresh happens instantly ✅

➡️ **Visible tab order now always matches actual state**

---

### Result

* Eliminates stale UI after reorder
* Keeps renderer in sync with main process
* Removes reliance on unrelated refresh triggers

.github/workflows/e2e-functional.yml
```diff
@@ -78,145 +85,169 @@ jobs:
...
+          name: playwright-report-merged
+          path: e2e/playwright-report-merged
+          if-no-files-found: error
+          retention-days: 7
 
-      - name: e2e/install-os-dependencies
-        uses: ./.github/actions/install-os-dependencies
+  publish-e2e-report:
+    name: Publish E2E report
+    runs-on: ubuntu-22.04
```
saturninoabril: (nit and can be done separately) Let's use ubuntu-24.04 in all places

.github/workflows/e2e-functional.yml
```diff
@@ -78,145 +85,169 @@ jobs:
...
+            echo "No report changes detected; reusing existing Pages content." >> "$GITHUB_STEP_SUMMARY"
+            REPORT_URL="${PAGES_URL}"
+            PUBLISHED=true
+          else
+            git commit -m "Publish E2E report for run ${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+            if git push origin HEAD:gh-pages; then
+              PUBLISHED=true
+              REPORT_URL="${PAGES_URL}"
+              echo "Published merged Playwright report: ${REPORT_URL}" >> "$GITHUB_STEP_SUMMARY"
+            else
```
saturninoabril: Why upload to GH pages?

.github/workflows/nightly-builds.yaml
```diff
@@ -42,3 +42,24 @@ jobs:
...
+      - name: Install root dependencies
+        run: npm ci
+      - name: Build test bundle
+        run: npm run build-test
+      - name: Install E2E dependencies
+        run: cd e2e && npm ci
+      - name: Run E2E full suite
+        run: cd e2e && npx playwright test
+        env:
+          CI: 'true'
```
saturninoabril: Can this nightly-e2e reuse .github/workflows/e2e-functional-template.yml?

e2e/specs/deep_linking/deeplink.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import {electronBinaryPath, appDir, demoConfig} from '../../helpers/config';
+import {waitForLockFileRelease} from '../../helpers/cleanup';
+import {buildServerMap} from '../../helpers/serverMap';
+
+test.describe('application', () => {
+    test('MM-T1304/MM-T1306 should open the app on the requested deep link', {tag: ['@P2', '@win32']}, async ({}, testInfo) => {
+        if (process.platform !== 'win32') {
+            test.skip(true, 'Windows only deep link test');
+            return;
+        }
```
saturninoabril: When skipping test at runtime, it can be `test.skip(condition, description)`

e2e/specs/deep_linking/deeplink.test.ts
```diff
@@ -0,0 +1,75 @@
...
+// See LICENSE.txt for license information.
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+import {test, expect} from '../../fixtures/index';
+import {waitForAppReady} from '../../helpers/appReadiness';
+import {electronBinaryPath, appDir, demoConfig} from '../../helpers/config';
+import {waitForLockFileRelease} from '../../helpers/cleanup';
+import {buildServerMap} from '../../helpers/serverMap';
```
saturninoabril, line 14: It would be great for these helper functions to be part of base.extend so no need import every time and may be able to use like `pw.waitForAppReady`.

e2e/specs/deep_linking/deeplink.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        fs.mkdirSync(userDataDir, {recursive: true});
+        fs.writeFileSync(path.join(userDataDir, 'config.json'), JSON.stringify(demoConfig));
+
+        const {_electron: electron} = await import('playwright');
+        const app = await electron.launch({
+            executablePath: electronBinaryPath,
+            args: [appDir, `--user-data-dir=${userDataDir}`, '--no-sandbox', '--disable-gpu', 'mattermost://github.com/test/url'],
+            env: {...process.env, NODE_ENV: 'test'},
+            timeout: 60_000,
+        });
```
saturninoabril: I see this in several places and may be good to extract and make reusable.

e2e/specs/deep_linking/deeplink.test.ts
```diff
@@ -0,0 +1,75 @@
...
+                );
+                return view ? view.webContents.getURL() : null;
+            }, webContentsId);
+            expect(isActive).toBe('https://github.com/test/url/');
+            const dropdownButtonText = await mainWindow.innerText('.ServerDropdownButton');
+            expect(dropdownButtonText).toBe('github');
+        } finally {
+            await app.close();
+            await waitForLockFileRelease(userDataDir);
+        }
```
saturninoabril: How about the typical test lifecycle using `test.after*`?

src/main/app/initialize.ts
```diff
@@ -197,6 +202,15 @@ function initializeBeforeAppReady() {
...
     }
+
+    if (process.env.NODE_ENV === 'test') {
+        try {
+            fs.appendFileSync(E2E_PROCESS_REGISTRY, `${process.pid}\n`, 'utf8');
+        } catch (error) {
+            log.warn('Failed to register E2E main process PID', {error});
+        }
+    }
+
```
saturninoabril: Is there other option to get app's PID at runtime in Playwright?

.github/workflows/e2e-functional.yml
yasserfaraazkhan: @saturninoabril As now we have playwright and need to combine and showcase. Earlier we used mochaaweseome and merged the reports.  I'll remove this now and look for another way.

.github/workflows/nightly-builds.yaml
yasserfaraazkhan: i'll remove this, there's a bit more logic needed to trigger matterwick

src/app/tabs/tabManager.ts
devinbinnie: This seems like a good change, however, I am not understanding how it was broken though before? On the front-end we just assumed that the re-ordering succeeded so that we wouldn't have to re-render again after the tabs were updated.

Was there a specific bug you encountered that triggered this?

src/app/system/badge.ts
```diff
@@ -132,4 +157,9 @@ export function setUnreadBadgeSetting(showUnreadBadge: boolean) {
 
 export function setupBadge() {
     AppState.on(UPDATE_APPSTATE_TOTALS, showBadge);
+
+    if (process.env.NODE_ENV === 'test') {
+        (global as TestGlobal).__testTriggerBadge = showBadge;
```
devinbinnie: We have a lot of messy code here around these global settings. I imagine this is probably the only way to allow the test framework to hear about these things though.

Instead, could we set up each method that needs to update this stuff in a cleaner way? I'm thinking something like:

```suggestion
        setTestField("testTriggerSetUnreadBadgeSetting", setUnreadBadgeSetting);
```

Elsewhere we could define the above method to set up those variables for the E2E test.
That way the rest of the code continues to remain typed strictly, while still allowing for this functionality. Thoughts?

i18n/de.json
```diff
@@ -301,7 +301,7 @@
   "renderer.components.settingsPage.checkSpelling.editSpellcheckUrl": "Verwende eine alternative Wörterbuch-URL",
   "renderer.components.settingsPage.checkSpelling.preferredLanguages": "Wähle bevorzugte Sprache(n) aus",
   "renderer.components.settingsPage.checkSpelling.revertToDefault": "Zurück auf Standard",
-  "renderer.components.settingsPage.checkSpelling.specifyURL": "Definiere die Url von der Wörterbuch-Definitionen geholt werden können",
```
devinbinnie, line 304: This probably isn't supposed to be here - revert?

e2e/CLAUDE.md
```diff
@@ -0,0 +1,212 @@
...
+- helper reuse over copy-pasted launch/login code
+
+Avoid:
+- arbitrary sleeps when a real readiness signal exists
+- introducing new global state coupling between tests
+- hiding real failures with `skip`, broad retries, or weakened assertions
+- depending on window focus unless the behavior specifically requires it
+- duplicating server-discovery or app-readiness logic already handled by fixtures/helpers
+
+## Fixture Rules
```
devinbinnie: `CLAUDE.md` works better when you're able to give it concrete examples I find. Can you add some to this?

Also, let's rename to AGENTS.md so as to be in-line with the rest of our repos (I'm going to do the same to the main md files don't worry :D)

src/main/notifications/Mention.ts
devinbinnie: We should probably actually move away from using our own internal sounds, and just use whatever the app does. That or allow for custom sounds in the Desktop App, but that's another idea :P

src/app/system/badge.ts
yasserfaraazkhan: Sure Devin, I'll clean it up this way.

```
function setTestField(name: string, value: unknown): void {
    if (process.env.NODE_ENV === 'test') {
        (global as TestGlobal)[name] = value;
    }
}
...
    setTestField('__testTriggerBadge', showBadge);

```

src/app/tabs/tabManager.ts
yasserfaraazkhan: @devinbinnie you are right. On frontend we don't see it. 
in the file `server_management/drag_and_drop.test.ts:277`

```
await mainWindow.evaluate(async ({viewOrder}) => {                                                                                                         
      const currentServer = await window.desktop.getCurrentServer();                                                                                         
      window.desktop.updateTabOrder(currentServer.id, viewOrder);                                                                                            
  }, {viewOrder: [visibleTabOrder[1], visibleTabOrder[0], visibleTabOrder[2]]});                                                                             
                                                                                                                                                             
  await expect.poll(async () => {                                                                                                                            
      const firstTab = await mainWindow.waitForSelector('.TabBar li.serverTabItem:nth-child(1)');                                                            
      return firstTab.innerText();                                                                                                                           
  }, {timeout: 10_000}).toContain('Off-Topic');
```

the setState in MainPage.tsx was not being called and the state was not updated.
in production this works, just for e2e it was not working and failing 1. of the test case. So I made this adjustment.

devinbinnie: Okay, I suppose it doesn't harm anything, so let's keep this.

src/main/app/initialize.ts
```diff
@@ -271,6 +276,16 @@ function initializeInterCommunicationEventListeners() {
 }
 
 async function initializeAfterAppReady() {
+    if (process.env.NODE_ENV === 'test') {
+        (global as any).__e2eTestRefs = {
```
devinbinnie: Can we use that helper function here as well? It should probably be in a utils file somewhere so we use `setTestField` everywhere.

src/main/app/intercom.ts
```diff
@@ -82,9 +82,15 @@ export function handleMainWindowIsShown() {
     log.debug('handleMainWindowIsShown', {showWelcomeScreen, showNewServerModal, mainWindow: Boolean(mainWindow)});
     if (mainWindow?.isVisible()) {
         handleShowOnboardingScreens(showWelcomeScreen(), showNewServerModal(), true);
+        if (process.env.NODE_ENV === 'test') {
```
devinbinnie: Here too.

PR #3732, opened 2026-03-16
Add browser detection and "Open Link in Browser" context menu

svelle:
#### Summary
This PR adds functionality to detect installed browsers on macOS, Windows, and Linux, and provides a context menu option to open external links in the user's preferred browser. The implementation includes:

- New `browserManager` module that detects installed browsers per platform using platform-specific methods (mdfind on macOS, registry on Windows, which/desktop files on Linux)
- Caching of detected browsers to avoid repeated system calls
- Integration with the context menu to show "Open Link in Browser" submenu for external links
- Comprehensive unit tests covering all platforms and error scenarios
- i18n support for the new menu item

#### Changes
- **src/main/browserManager.ts**: New module exporting `getInstalledBrowsers()`, `openLinkInBrowser()`, and `clearBrowserCache()` functions with platform-specific browser detection logic
- **src/main/browserManager.test.js**: Complete test suite covering macOS, Windows, and Linux browser detection, caching behavior, and link opening with fallback to `shell.openExternal`
- **src/app/views/MattermostWebContentsView.ts**: Integrated browser detection and added context menu items for opening external links in detected browsers
- **i18n/en.json**: Added localization string for the new menu option

#### Checklist
- [x] Added unit tests (comprehensive test coverage for all platforms)
- [x] Has UI changes (new context menu option)
- [x] read and understood our Contributing Guidelines
- [x] completed Mattermost Contributor Agreement

#### Release Note
```release-note
Added ability to open external links in installed browsers via context menu option on macOS, Windows, and Linux.
```

https://claude.ai/code/session_01WWHsdZzBXhrUtXo9WYEwY4

## Change Impact: 🟡 Medium

**Reasoning:** Adds a new cross-platform BrowserManager module and context-menu integration into MattermostWebContentsView, spanning main and renderer layers with platform-specific system calls; unit tests cover the manager extensively but integration/UI timing remains untested.

**Regression Risk:** Moderate — Browser detection logic is well-covered by unit tests, but changes touch a widely-instantiated UI component (MattermostWebContentsView) and introduce asynchronous initialization (cachedBrowsers) that can cause timing-dependent context-menu inconsistencies early after startup; fallback behavior reduces severity but integration paths lack tests.

** QA Recommendation:** Perform manual QA on macOS, Windows, and Linux to verify the submenu appears consistently after app start and page loads, that detected browsers launch correctly, and that failures fall back to the system default; add integration/unit tests for MattermostWebContentsView menu population and startup timing to reduce regressions.

*Generated by CodeRabbitAI*


src/main/browserManager.ts
```diff
@@ -0,0 +1,343 @@
...
+async function getWindowsBrowsers(): Promise<BrowserInfo[]> {
+    const seenNames = new Set<string>();
+    const found: BrowserInfo[] = [];
+
+    const results = await Promise.all(
+        WINDOWS_REGISTRY_ROOTS.flatMap((root) =>
+            KNOWN_WINDOWS_BROWSERS.map(async (browser) => {
+                try {
+                    const {stdout} = await exec(
+                        `reg query "${root}\\${browser.registryKey}\\shell\\open\\command" /ve`,
```
devinbinnie: We can avoid all of this code by instead using the `registry-js` module that is already built in to our Desktop App. See `https://github.com/mattermost/desktop/blob/master/src/common/config/policyConfigLoader.ts` for an example.

My approach would be to try and read the registry at start-up as well and pre-load all of the browsers (can't imagine hot-reloading needs to matter here) and store them here. One thing to try and make sure of is that the registry operations don't interfere with each other (I don't think that happens with the new module, but would be worth verifying)

src/main/browserManager.ts
```diff
@@ -0,0 +1,343 @@
...
+
+    const found: BrowserInfo[] = [];
+    for (const result of knownResults) {
+        if (result && !seenNames.has(result.name)) {
+            found.push(result);
+            seenNames.add(result.name);
+        }
+    }
+
+    // Also discover browsers from .desktop files that handle http(s)
```
devinbinnie: I know this is also a way to capture Linux browsers, but is `which` not enough here?

Reason I ask is that I'd like to avoid the advanced matching and parsing of desktop files here, this is quite a bit more complicated that I would expect this to be. The regex pieces have me a bit worried that something could be exploited here.

src/main/browserManager.ts
```diff
@@ -0,0 +1,343 @@
...
+        log.error('execFile failed to open link, falling back to default browser', {browser: browser.name, error: execError});
+        try {
+            await shell.openExternal(url);
+        } catch (fallbackError) {
+            log.error('Fallback shell.openExternal also failed', {error: fallbackError});
+        }
+    }
+}
+
+export class BrowserManager {
```
devinbinnie: I think this class is missing the above state `cachedBrowsers`. Anything that doesn't require state doesn't need a manager, but this does. We should encapsulate the state within the class and put the functions in here, same as the other examples.

This also makes it more testable.

src/main/browserManager.ts
```diff
@@ -0,0 +1,343 @@
...
+    init = async () => {
+        await getInstalledBrowsers();
```
devinbinnie: This should not be awaited, I/O operations can take a while and we don't want to hold up starting the application on this functionality. Once the browsers are populated then the menu items should work.

src/main/browserManager.ts
devinbinnie: Also, let's call this something more descript to the problem, maybe `ExternalBrowserManager`? Would probably be worth having a comment on it as to what it does too.

devinbinnie (changes requested):
Cool feature design, just needs a few adjustments :)

src/main/externalBrowserManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            values.find((value) => value.name === '(Default)') ||
+            values.find((value) => typeof value.data === 'string');
+    };
+
+    private getLinuxBrowsers = async (): Promise<BrowserInfo[]> => {
+        const results: Array<BrowserInfo | null> = await Promise.all(
+            KNOWN_LINUX_BROWSERS.map(async (browser) => {
+                for (const command of browser.commands) {
+                    try {
+                        // eslint-disable-next-line no-await-in-loop
```
devinbinnie, line 253: Can you add a comment why this eslint rule was overridden? I think in this case it makes sense since we want to exit early (and we're only running it once on init) but would be good to document.

devinbinnie (approved):
Thanks @svelle, one quick thing to add but otherwise LGTM.

PR #3734, opened 2026-03-16
[MM-67797] Expose ability to update the title template for popouts

devinbinnie:
#### Summary
Popout windows currently receive a `titleTemplate` only at creation time via `openPopout`. Once a popout is open, there's no way for the web app to change how the window title is formatted — for example, to update the template when the user navigates to a different context within the same popout.

This PR adds an `updatePopoutTitleTemplate` method to the Desktop API. When called from a popout view, it updates the stored title template and immediately re-renders the window title using the existing title components (channel name, team name, server name).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67797

#### Release Note
```release-note
NONE
```

## Change Impact: 🟡 Medium

**Reasoning:** The change introduces a new public API method and IPC channel and updates view management to support dynamic popout title templates; these are additive but span multiple modules (preload API, IPC, popout manager, view manager, and types), increasing surface area.

**Regression Risk:** Moderate. Although behavior is additive and unit tests were added, the change touches shared communication channels and view lifecycle code that could affect UI title updates or event propagation if edge cases are missed.

**QA Recommendation:** Perform focused manual QA: smoke-test popout creation and normal title rendering, verify calling updatePopoutTitleTemplate from a popout updates the title immediately across scenarios (popout vs tab, missing view props), and run related automated tests. Skipping manual QA is not recommended due to cross-module impact.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/windows/popoutManager.ts
+++ b/src/app/windows/popoutManager.ts
@@ -393,6 +393,9 @@ export class PopoutManager {
     };
 
     private handleUpdatePopoutTitleTemplate = (event: IpcMainEvent, titleTemplate: string) => {
+        if (typeof titleTemplate !== 'string') {
+            return;
+        }
         const webContentsView = WebContentsManager.getViewByWebContentsId(event.sender.id);
         if (!webContentsView) {
             return;
--- a/src/app/windows/popoutManager.test.js
+++ b/src/app/windows/popoutManager.test.js
@@ -1090,6 +1090,17 @@ describe('PopoutManager', () => {
             expect(ViewManager.updateViewTitleTemplate).not.toHaveBeenCalled();
         });
 
+        it('should not call updateViewTitleTemplate when titleTemplate is not a string', () => {
+            const mockEvent = {
+                sender: {id: 456},
+            };
+
+            popoutManager.handleUpdatePopoutTitleTemplate(mockEvent, 123);
+
+            expect(WebContentsManager.getViewByWebContentsId).not.toHaveBeenCalled();
+            expect(ViewManager.updateViewTitleTemplate).not.toHaveBeenCalled();
+        });
+
         it('should not call updateViewTitleTemplate when view is a tab', () => {
             const mockWebContentsView = {
                 id: 'tab-view-id',


src/app/windows/popoutManager.ts
```diff
@@ -392,6 +392,14 @@ export class PopoutManager {
         view.sendToRenderer(MESSAGE_FROM_PARENT, channel, ...args);
     };
 
+    private handleUpdatePopoutTitleTemplate = (event: IpcMainEvent, titleTemplate: string) => {
+        const webContentsView = WebContentsManager.getViewByWebContentsId(event.sender.id);
+        if (!webContentsView) {
+            return;
+        }
+        ViewManager.updateViewTitleTemplate(webContentsView.id, titleTemplate);
+    };
```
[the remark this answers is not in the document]
devinbinnie: @coderabbit This is resolved, and so is the other comment around the validation, since it's not necessary. Please re-review.

src/app/windows/popoutManager.ts
```diff
@@ -392,6 +392,20 @@ export class PopoutManager {
         view.sendToRenderer(MESSAGE_FROM_PARENT, channel, ...args);
     };
 
+    private handleUpdatePopoutTitleTemplate = (event: IpcMainEvent, titleTemplate: string) => {
+        if (typeof titleTemplate !== 'string') {
+            return;
+        }
```
Willyfrog, line 398: nitpick: why wouldn't titleTemplate be a string if the function signature requires it being a string?

src/common/views/viewManager.ts
```diff
@@ -132,6 +132,20 @@ export class ViewManager extends EventEmitter {
         this.emit(VIEW_TITLE_UPDATED, view.id);
     };
 
+    updateViewTitleTemplate = (viewId: string, titleTemplate: string) => {
+        log.debug('updateViewTitleTemplate', {viewId});
+
+        const view = this.views.get(viewId);
```
Willyfrog, line 138: should this check if the view is actually a window? as in the other file?

src/app/windows/popoutManager.ts
devinbinnie: There's no enforcement for the client to send up something that conforms to the type, this is a security hardening measure to make sure that we don't consume a `number` and it causes a `TypeError` when we try to run `replace` on it.

src/common/views/viewManager.ts
devinbinnie: Regular views make use of `titleTemplate` so I think we're okay here. Worst case it would just update the title in the same way, which isn't a huge deal.

PR #3765, opened 2026-04-08
Separate update latest version script from the normal release script

devinbinnie:
#### Summary
Separate update latest version script from the normal release script

```release-note
NONE
```

## Change Impact: 🟡 Medium

**Regression Risk:** The change isolates the "update latest version" flow into a new manual `workflow_dispatch` workflow and removes the automated generation/upload step from the release workflow. Code regression risk is low because production code is untouched and the generate script is unchanged; however, process/regression risk is moderate since failing to manually trigger the new workflow after releases can leave S3-hosted version file(s) stale and affect consumers relying on latest-version endpoints.

**QA Recommendation:** Light manual QA recommended: verify the new workflow input validation, manually trigger `Update Latest Version` with representative refs (master and a release-X.Y) to confirm the script generates the expected filename and uploads to the correct S3 path, and validate AWS role assumption and bucket permissions. Also update release runbooks/checklists to ensure the manual step is executed post-release.

*Generated by CodeRabbitAI*

### DIFF
--- a/.github/workflows/update-latest-version.yml
+++ b/.github/workflows/update-latest-version.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        description: 'Branch to read the version from (e.g. release-6.3)'
+        description: 'Branch to read the version from: master or release-X.Y'
         required: true
         type: string
 
@@ -21,7 +21,18 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    env:
+      REF: ${{ inputs.ref }}
     steps:
+      # This is a lightweight guard to ensure the input ref is valid.
+      - name: Verify input ref
+        run: |
+          if ! [[ "$REF" == "master" || "$REF" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
+            echo "Error: Input ref must be master or release-X.Y (got '${REF}')"
+            exit 1
+          fi
+          echo "OK: Input ref is valid ($REF)"
+
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/update-latest-version.yml
+++ b/.github/workflows/update-latest-version.yml
@@ -1,0 +1,44 @@
+name: Update Latest Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-latest-version.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Setup AWS credentials
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.MM_DESKTOP_RELEASE_AWS_ROLE_TO_ASSUME }}
+
+      - name: Generate latest version file
+        id: generate-latest
+        run: bash -x ./scripts/generate_latest_version.sh
+
```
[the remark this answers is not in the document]
devinbinnie: The original one works fine so we'll just keep using that.

amyblais (approved):

.github/workflows/update-latest-version.yml
```diff
@@ -0,0 +1,44 @@
+name: Update Latest Version
+
+on:
+  workflow_dispatch:
```
NARSimoes, line 4: **Clarification:** How are we planning to trigger this workflow and who is allowed to? Also, relates with the checkout `${{ inputs.ref }}` and execution of the generator script from that revision.

Depending on the above we could also add a lightweight guard too
```
on:
  workflow_dispatch:

jobs:
  update-latest-version:
    # just a lightweight guard here since workflow_dispatch doesn't allow protections / validations. 
    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release-')
    runs-on:...
```

NARSimoes (changes requested):
Thanks @devinbinnie, just added a comment to do a minor clarification.

.github/workflows/update-latest-version.yml
devinbinnie: I assume it would be manual and that @amyblais would do it as part of the release playbook, after we've confirmed all update channels are up to date.

I don't think the other trigger for the release have this - if you think it's necessary though we can add it.

NARSimoes: Updated the branch https://github.com/mattermost/desktop/pull/3765/commits/fb7441cafa7291e4db6be3cd7acde1aeb37cf1eb as we discussed in DM.

NARSimoes (approved):

PR #3766, opened 2026-04-09
[MM-68250] Fix stuck mention badge when removing a server with unreads

devinbinnie:
#### Summary
When a server with unreads or mentions was removed, the badge got stuck because `AppState` never cleaned up its per-server maps on removal, and the renderer kept stale entries in its local state indefinitely.

This PR adds a `SERVER_REMOVED` listener in `AppState` that clears the removed server's state and emits zeroed values via `emitStatusForServer`, flowing through the existing `onUpdateMentionsForServer` path in the renderer. The renderer now deletes entries at 0/false instead of preserving them.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68250

#### Release Note
```release-note
Fixed an issue where removing a server with unreads or mentions caused the badge indicator to become stuck until the app was restarted.
```

## Change Impact: 🟡 Medium

**Regression Risk:** The changes introduce deletion of state entries (`delete newMentionsPerServer[serverId]`) when values are falsy in the renderer, which is a semantic shift from the previous behavior of preserving entries with default values. While this is a defensive change that only affects the server removal flow, there is potential for regression if other components depend on these keys always being present in the state objects. The AppState deletion logic is straightforward (deleting from Maps and re-emitting), but the state flow spans multiple layers (AppState → MainWindow → Renderer) with limited test coverage specifically for the cleanup scenario after server removal. The nullish coalescing operators (`??`) in ServerDropdownButton provide fallbacks, but edge cases around concurrent state updates during server removal are not explicitly tested.

**QA Recommendation:** Manual QA should focus on: (1) adding and removing a server with active mentions/unreads while monitoring badge state for stale indicators, (2) removing servers while unreads are being received, (3) switching between servers before/after removal, and (4) verifying badge counts return to 0/empty state immediately after server removal. Automated test coverage should be added for the complete cleanup flow (server removal → badge state zeroing → badge UI update) to prevent regression.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/common/appState.ts
+++ b/src/common/appState.ts
@@ -8,9 +8,11 @@ import {
     UPDATE_APPSTATE_TOTALS,
     UPDATE_APPSTATE_FOR_VIEW_ID,
     SERVER_LOGGED_IN_CHANGED,
+    SERVER_REMOVED,
     UPDATE_APPSTATE_FOR_SERVER_ID,
 } from 'common/communication';
 import {Logger} from 'common/log';
+import type {MattermostServer} from 'common/servers/MattermostServer';
 import ServerManager from 'common/servers/serverManager';
 import ViewManager from 'common/views/viewManager';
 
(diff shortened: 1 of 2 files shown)


davidkrauser (approved):
LGTM 👍

PR #3787, opened 2026-04-16
[MM-67767] Add priority modal queueing so that some modals can show above others

devinbinnie:
#### Summary
The modal queue is strictly sequential — only the front modal is shown. When no servers are configured, the Welcome screen sits at the front of the queue and is uncloseable. Opening Settings enqueues it behind Welcome, where it never gets displayed.

This PR adds priority modal support to `ModalManager` so that certain modals (like Settings) can be inserted at the front of the queue, suspending the current modal without destroying it. When the priority modal closes, the previously-showing modal resumes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67767

#### Release Note
```release-note
Fixed an issue where the Settings modal could not be opened when no servers were configured.
```

## Change Impact: 🟡 Medium

**Regression Risk:** The changes introduce a new modal suspension/resumption mechanism that modifies the lifecycle of queued modals. Previously, modals beyond the first in the queue were fully hidden (closing devtools, unregistering from performance monitor, disposing context menu, closing web contents). Now, non-visible modals call the new `suspend()` method which only detaches the view without these cleanup operations. This lifecycle change could leak resources if modals remain suspended indefinitely or cause unexpected state if callers rely on the previous cleanup behavior. Additionally, `ModalView.suspend()` lacks unit test coverage despite being a critical new lifecycle method. The Settings modal is the only current consumer of `addPriorityModal()`, limiting immediate exposure, but the infrastructure is now in place for broader adoption.

**QA Recommendation:** Manual QA should verify modal state transitions, particularly: (1) opening Settings modal when no servers are configured; (2) queuing multiple modals sequentially to ensure suspension/resumption works correctly; (3) long-lived scenarios where modals may remain suspended to check for resource leaks; (4) closing suspended modals to confirm proper cleanup. E2E tests for welcome screen and other modal-driven flows should be executed. Consider adding unit tests for `ModalView.suspend()` before merge.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/mainWindow/modals/modalManager.test.js
+++ b/src/app/mainWindow/modals/modalManager.test.js
@@ -94,11 +94,52 @@ describe('main/views/modalManager', () => {
         });
     });
 
+    describe('addPriorityModal', () => {
+        const modalManager = new ModalManager();
+
+        beforeEach(() => {
+            modalManager.modalQueue = [];
+            modalManager.modalPromises = new Map();
+            modalManager.showModal = jest.fn();
+        });
+
+        it('should not add modal with the same key, should return the existing promise', () => {
+            modalManager.modalQueue.push({key: 'existing_key'});
+            const promise = Promise.resolve();
+            modalManager.modalPromises.set('existing_key', promise);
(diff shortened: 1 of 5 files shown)


mgdelacroix (approved):
LGTM! 👍

PR #3805, opened 2026-04-29
workflows: only upload artifacts to new bucket (#3710)

NARSimoes:
#### Summary
Cherry-pick https://github.com/mattermost/desktop/pull/3710

#### Ticket Link
https://github.com/mattermost/desktop/pull/3710

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/nightly-main.yml
+++ b/.github/workflows/nightly-main.yml
@@ -210,8 +210,8 @@ jobs:
         uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
         with:
           aws-region: us-east-1
-          aws-access-key-id: ${{ secrets.MM_DESKTOP_RELEASE_AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-main.yml
```diff
@@ -222,20 +222,32 @@ jobs:
         run: |
           mkdir -p ./aws-s3-dist
           cp -r --backup=numbered ./build/{macos-release,win-release,linux}/* ./aws-s3-dist
+      - name: nightly/generate-latest-version
```
devinbinnie, line 225: Should also clean up the nightly too?

devinbinnie (changes requested):

.github/workflows/nightly-main.yml
NARSimoes: Yep, but we still have that logic in master https://github.com/mattermost/desktop/blob/06488e179f720853302bbb3b4dba92553d4a9a2f/.github/workflows/nightly-main.yml#L224

Should we address in a follow up?

devinbinnie: It shouldn't matter since we don't build from here, we should clean it up but in the interest of getting this in we can leave it.

devinbinnie (approved):

Issue #90, opened 2016-03-21
Right clicking icon and choosing 'Quit' just minimizes


DivineOmega:
Running Ubuntu 14.04 and Mattermost Desktop (electron-mattermost) v1.0.7, steps to reproduce:
1. Start up Mattermost Desktop in the normal manner.
2. Right click the icon representing the running program in Unity.
3. Click the 'Quit' menu option.

Expected behaviour:
- Mattermost Desktop exits.

Actual behaviour:
- Mattermost Desktop minimizes if not already minimized and does nothing if it is already minimized.

The only way to quit Mattermost Desktop in this instance is to use the File -> Quit menu option or to kill the process.

yuya-oc:
On Linux, the application performs 'minimize' instead of 'close' in order to keep the window. This feature is working in this issue. Unfortunately, I don't know what is the difference between Unity's 'Quit' and normal 'close'.

brumle80:
same "issue" in Fedora 23, click the x and it minimizes, but doesn't close.

yuya-oc:
@brumle80 Did you mean about window's close button? If you did, the 'x' behavior is by design as I had written at the previous post. I think that the point of this issue is the handling for 'Quit' operation caused by desktop environment like Unity/GNOME.

jasonblais:
Thanks @DivineOmega and @brumle80 for the feedback,

A related ticket #178 has been resolved and closed, which should fix the issue presented here.

However, if there is a further improvement you would like to see, I invite you to help contribute your suggestion in Mattermost [feature idea forum](https://mattermost.uservoice.com/forums/306457-general) so it can be discussed, upvoted and considered for a ticket accepting pull requests?

If you do add it to the [feature idea forum](https://mattermost.uservoice.com/forums/306457-general), please include a link back to this GitHub Issue. If you're interested in implementing, please say so and we'll prioritize the review.

jasonblais:
I realized there actually is still an issue with right clicking icon and choosing 'Quit'. I have added it as a known issue [to the desktop changelog](https://github.com/mattermost/desktop/pull/206)

jasonblais:
@RaphaelKimmig I think that's fair

amyblais:
Ticket for tracking: https://mattermost.atlassian.net/browse/MM-14130

echosa:
[...]

Issue #72, opened 2016-03-13
Windows 7: Leaving "Name" blank creates unusable team site tab


it33:
Repro:  

- Create a team, then create an additional team using this procedure: 
- In **Settings Page** enter a valid **URL**, leave **Name** field blank then click **Add** and then **Save**

Observed:  

- Blank tab created for team, Blank tab cannot be clicked. 

Expected:   

- **Add** button responds with an error that **Name** is required when **Name** field is empty

yuya-oc:
It will be fixed at next version by 97808373ede7aa333a4dd21de062da8024af5890.

it33:
Perfect, thanks!

PR #3829, opened 2026-05-25
ci(e2e): CMT scheduling + macOS E2E stabilization + CI/security hardening

yasserfaraazkhan:
## Summary

Enables matterwick-driven scheduled/CMT E2E provisioning for Desktop and lands a broad round of E2E reliability and CI/security hardening. (Companion to `mattermost/matterwick`#90.)

### CMT / scheduled provisioning (matterwick-driven)
- **`cmt-provisioner.yml`** → lightweight trigger (no inputs/curl) that now fires on **push to `release-v*`** (the branch cut + subsequent pushes), replacing the previous monthly schedule; matterwick hears the `workflow_run` and dispatches `compatibility-matrix-testing.yml`. A `release-v*` push also runs the normal release E2E (whole suite, latest server), so the latest server is exercised by both flows — intended.
- **`compatibility-matrix-testing.yml`** → dropped the in-workflow `/cleanup_e2e` call (matterwick now destroys provisioned servers on `workflow_run` completion, keyed by commit SHA); corrected the stale `cmt_run_id` input description.
- **`e2e-nightly-trigger.yml`** → schedule-only (pushes to `main`/`release-*` are handled by matterwick's native push handler).

### macOS E2E stabilization (the previously-red macOS suite)
- **`intercom.ts`**: fixed the `__e2eAppReady` race (the window `show` event firing before the listener attaches, or `MainWindow.get()` returning `undefined`, on slower macOS runners) — the single error behind nearly all macOS failures. Added unit tests.
- Suppressed the macOS Resume / crash-reporter / Gatekeeper dialogs, stopped sending `SIGKILL` in global teardown, and bumped app-ready / minimize timeouts (`global-setup.ts`, `global-teardown.ts`, `appReadiness.ts`).
- **Per-OS reporting**: each platform uploads its own HTML report; removed the cross-OS merge that produced misleading aggregate counts. Flaky-test accounting fixed so a retried-then-passed test isn't counted as a failure (`analyze-flaky-test.js`, `github-actions.js`).
- Targeted spec fixes: popout (was capturing the child URLView instead of the popout window), copy-link `role="menuitem"` selector, window-menu/minimize timeout, deeplink WebContentsView traversal (Windows), bad-servers cert reload, remove-server-modal partial-config parse, badge, full-screen, tray-restore, startup window.

### CI & security hardening
- **`e2e-functional.yml`** least-privilege permissions (top-level `contents: read`, per-job grants) so untrusted PR code (`npm ci` / Playwright) no longer runs with `pull-requests: write`.
- **Flaky-CI fix:** mocked `electron` in the two unit suites (`UserActivityMonitor`, `diagnostics`) that imported the real module — they only passed when the electron binary happened to be installed.
- **`popoutManager`**: `registerMainWindowCloseHandler` / `closeAllPopouts` + unit tests.

### Cursor-automation removal
Removed the "Server for Cursor Automation" PR-body read/write feature (`resolveMmTestServerUrlFromPr`, the PR-body sync job, and the helper fns). `MM_TEST_SERVER_URL` is supplied only via inputs/env.

### Companion PRs
`mattermost/matterwick`#90, `mattermost/mattermost-mobile`#9800, `mattermost/gitops-platform`#72.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

#### Release Note

```release-note
NONE
```


yasserfaraazkhan:
@coderabbitai resume

yasserfaraazkhan:
@coderabbit review

yasserfaraazkhan:
@coderabbit review

src/main/app/intercom.ts
```diff
@@ -78,18 +78,113 @@ export function handleMainWindowIsShown() {
...
+        }
+
+        // The `show` event fires when `browserWindow.show()` is called and the
+        // window becomes visible. We deliberately do NOT listen to
+        // `ready-to-show` here: that event fires *before* the window is shown
+        // (so the renderer can call `.show()` without a white flash), and using
+        // it would set `__e2eAppReady=true` while the window is still hidden.
+        mainWindow.once('show', onShow);
+        mainWindow.once('closed', onMainWindowClosed);
+        app.once('before-quit', onBeforeQuit);
```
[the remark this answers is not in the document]
yasserfaraazkhan: Already addressed — `src/main/app/intercom.test.js` mocks `electron.app` with `once: jest.fn()` (and `removeListener`), so the `app.once is not a function` regression is gone. The `build-linux`, `build-mac-no-dmg`, `build-win-no-installer`, and Test Results checks are all green on the latest commit. Marking this as resolved.

.github/workflows/e2e-functional.yml
```diff
@@ -253,13 +185,22 @@ jobs:
 
   e2e-policy-tests:
     name: policy-tests-${{ matrix.platform }}
+    # Runs untrusted PR code (npm ci / Playwright); grant only the commit-status write its
+    # check-for-failures step needs, never pull-requests: write.
+    permissions:
+      contents: read
+      statuses: write
```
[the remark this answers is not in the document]
yasserfaraazkhan: Intentionally not doing the job-split in this PR. The dangerous scope — `pull-requests: write` — was already removed from `e2e-policy-tests`; it now holds only `contents: read` + `statuses: write`. `statuses: write` is a minimal capability (it can only set a commit status — it cannot modify code, PR state, or merge), so untrusted test code running with it is low risk. Splitting `e2e/check-for-failures` into a separate `needs: e2e-policy-tests` post-processing job requires uploading the test-results artifact and a dependent job to consume it — a substantial refactor of a matrix job that can't be validated outside CI, disproportionate to the residual risk. Capturing it as a possible follow-up rather than blocking this PR.

src/app/windows/popoutManager.ts
```diff
@@ -80,12 +81,48 @@ export class PopoutManager {
...
     createNewWindow = (serverId: string) => {
         this.handleCreateNewWindow(serverId);
     };
 
+    private registerMainWindowCloseHandler = () => {
+        const win = MainWindow.get();
+        if (!win) {
+            return;
+        }
+        win.once('closed', this.closeAllPopouts);
```
devinbinnie: I don't think this is actually something we want - the idea behind multi-window is that the non-function-specific pop-out windows are somewhat independent, so you could feasibly close the main window and use just a pop-out for your server.

What's the rationale behind doing this?

src/main/app/intercom.ts
```diff
@@ -78,18 +78,113 @@ export function handleMainWindowIsShown() {
...
+    const markReady = (mainWindowIsVisible: boolean) => {
+        if (done) {
+            return;
+        }
+        done = true;
+        handleShowOnboardingScreens(showWelcomeScreen(), showNewServerModal(), mainWindowIsVisible);
+        setTestField('__e2eAppReady', true);
+    };
+
+    const attachWindowListeners = (mainWindow: BrowserWindow) => {
```
devinbinnie: Is this only relevant for E2E testing? If so we should gate it behind the environment variable and maybe we should move it to a separate function so we're not bloating this function.

Also, is there a reason we need to do all this polling? I thought the original functionality of having a listener on the E2E set up was better since we don't have to just guess when the windows are set up.

devinbinnie (changes requested):
Thanks for continuing to work to stabilize the Desktop E2E tests :)

I do have some concerns around changing we need to make to the code base to make this work - especially the polling addition, since that definitely can affect normal app usage.

yasserfaraazkhan:
Thank you Devin
I made the changes to only have it behind E2E variable.

src/main/diagnostics/index.test.js
```diff
@@ -3,6 +3,32 @@
 
 import Diagnostics from '.';
 
+// Mock electron so this suite (and the diagnostic steps it imports) doesn't load the real
+// `electron` module, which needs the downloaded binary and otherwise fails with
+// "Electron failed to install correctly" in CI. Only the surface used by index.ts and the
+// imported steps is stubbed.
+jest.mock('electron', () => ({
+    shell: {
+        showItemInFolder: jest.fn(),
```
devinbinnie: You can revert these test changes and merge master - they were fixed in https://github.com/mattermost/desktop/pull/3836

src/main/UserActivityMonitor.test.js
```diff
@@ -3,6 +3,17 @@
 
 import {UserActivityMonitor} from './UserActivityMonitor';
 
+// Mock electron so the suite doesn't load the real `electron` module (which needs the
+// downloaded binary and otherwise fails with "Electron failed to install correctly" in CI).
+jest.mock('electron', () => ({
+    app: {
+        isReady: jest.fn(() => true),
```
devinbinnie: These ones too

devinbinnie (approved):
Thanks @yasserfaraazkhan :)

PR #3842, opened 2026-06-05
[MM-68985] Add collection module to allow Session Attributes to be passed to the requesting server

devinbinnie:
#### Summary
Mattermost servers need client and session context (client IP, VPN status, MDM enrollment, network interface, etc.) to drive session-based access-control policies, but the desktop app had no mechanism to advertise this information on its requests to the server.

This PR adds the framework for delivering Session Attributes to servers that are set up to enforce them. On servers where `FeatureFlagSessionAttributes` is enabled, the attribute manifest is fetched and a new `SessionAttributesManager` injects a `X-MM-Session-Attributes` header on outbound server requests — gated on an active session cookie and per-attribute TTLs from the manifest. 

The existing `onBeforeSendHeaders` handler is refactored into composable injectors (`PreAuthManager.injectPreAuthSecret` and `SessionAttributesManager.injectHeader`), and IPC/preload methods (`invalidateSessionAttributeManifest`, `resendSessionAttributes`) let the webapp refresh the manifest and force a resend. 

The per-OS native collectors are stubbed here and deferred to a follow-up changeset, so only a few platform-agnostic attributes are enabled by this change.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68985

#### Release Note
```release-note
Enable client-collected Session Attributes to be sent to enabled servers
```

## Change Impact: 🟠 Medium

**Regression Risk:** The change adds a new SessionAttributesManager, a refactor of the onBeforeSendHeaders flow into composable injectors, server manifest fetching, and new header injection behavior — all of which touch request authentication/session header handling. Recent commits added unit tests covering SessionAttributesManager behavior and direct tests for PreAuthManager and serverInfo positive-path manifest fetching, reducing risk. Some base collector methods remain lightly tested, and header injection still affects a critical request path, so regressions are possible but constrained by the server-side feature flag.

**QA Recommendation:** Moderate manual QA is advised. Validate manifest fetch/storage when the feature flag is enabled, verify PreAuth header injection behavior, test TTL-based resend logic across manifests (including ttl_seconds: 0), and ensure header injection doesn't break existing headers or request flows. Adding tests for remaining collector base methods would further reduce manual QA scope.

*Generated by CodeRabbitAI*


src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -0,0 +1,198 @@
...
+    private lastSentAt = new Map<string, Map<string, number>>();
+
+    constructor() {
+        ServerManager.on(SERVER_URL_CHANGED, this.handleServerChanged);
+        ServerManager.on(SERVER_PRE_AUTH_SECRET_CHANGED, this.handleServerChanged);
+        ServerManager.on(SERVER_REMOVED, this.handleServerRemoved);
+
+        ipcMain.on(SESSION_ATTRIBUTES_MANIFEST_INVALIDATED, this.handleManifestInvalidated);
+        ipcMain.on(SESSION_ATTRIBUTES_RESEND_REQUESTED, this.handleResendRequested);
+    }
```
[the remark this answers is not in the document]
devinbinnie: That's not consistent with the architecture at all in the Desktop App, many managers have IPC handled registered in their constructors.

src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -0,0 +1,198 @@
...
+            sentMap = new Map<string, number>();
+        }
+
+        const payload: Record<string, unknown> = {};
+        for (const field of fieldsToSend) {
+            const value = this.collectAttribute(field.name, server.id);
+            if (value) {
+                payload[field.name] = value;
+            }
+            sentMap.set(field.name, now);
```
pvev, line 88: nit: shouldn't this set also go in the if (value) condition above?

src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+    };
+
+    private hasAuthCookie = (headers: Record<string, string | string[]>): boolean => {
+        const cookie = headers.Cookie ?? headers.cookie;
+        if (!cookie) {
+            return false;
+        }
+        const cookieStr = Array.isArray(cookie) ? cookie.join(';') : cookie;
+        return cookieStr.includes(`${COOKIE_NAME_AUTH_TOKEN}=`);
```
pvev, line 151: nit: wondering if this includes is "too permisive"?

pvev (approved):
great work @devinbinnie 🎉 🚀

src/main/sessionAttributes/sessionAttributesManager.ts
devinbinnie: Actually no, if the field is blank or null, we don't want to actually put it in the JSON payload, but we don't want to send it again either. This might mean that if there's an error on the Desktop App getting the info that it won't retry until asked to, but I think this is acceptable since we fail-secure if the value is missing. From a UX perspective, the user could just restart the app to try and re-load it.

src/main/sessionAttributes/sessionAttributesManager.ts
devinbinnie: This isn't really a permission check, it's more of a filter to make sure we're not sending the attributes unnecessarily when we know there's no active session, since it won't do anything with them.

src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -0,0 +1,198 @@
...
+import type {SAField} from 'types/sessionAttributes';
+
+import SessionAttributeCollector from './collector';
+
+const log = new Logger('SessionAttributesManager');
+
+export class SessionAttributesManager {
+    private lastSentAt = new Map<string, Map<string, number>>();
+
+    constructor() {
```
edgarbellot, line 31: The constructor clears `lastSentAt` on `SERVER_URL_CHANGED`, `SERVER_PRE_AUTH_SECRET_CHANGED` and `SERVER_REMOVED`, but not on `SERVER_LOGGED_IN_CHANGED`. When a new session starts, fields with a long TTL won't be re-sent until that TTL expires.

Is that intentional - i.e. the server is expected to handle re-population for new sessions on its side - or would it make sense to also hook into `SERVER_LOGGED_IN_CHANGED` here to force a fresh send at login?

edgarbellot (approved):
@devinbinnie great work! I only have one question, but def non-blocking :)

src/main/sessionAttributes/sessionAttributesManager.ts
devinbinnie: So normally I would rely on `SERVER_LOGGED_IN_CHANGED` but it usually doesn't fire quickly enough to be as reliable. What I ended up going with was a call from the web app when the websocket connects to force it to resend the session attributes immediately:
https://github.com/mattermost/desktop/pull/3842/changes#diff-ad24a33537ba3fb5343ad5bfbb432491d28c429385083a6c0b3ea26b2f768c7eR87
https://github.com/mattermost/mattermost/pull/36934/changes#diff-70620ee1dcea95454c34d5953d7030b314230b5c382ee8bc9cd1ee543a11be3bR420

PR #3855, opened 2026-06-19
test(e2e): Playwright harness and main-process hooks for migration stack

yasserfaraazkhan:
This is a small set of changes to get CI running Playwright. This will put in the main hooks required by e2e tests specs to run tests from these pr. ([#3857](https://github.com/mattermost/desktop/pull/3857)–[#3864](https://github.com/mattermost/desktop/pull/3864)).

## Summary

Playwright’s Electron driver gives us a `BrowserWindow` pages for the app shell. Each Mattermost server renders in a **`WebContentsView`**, which is not a page we get from normal playwright execution. Finding and driving a specific server means going through main-process singletons like ViewManager, WebContentsManager, etc and not the page.click() alone.

To automate these flow we are using hooks. The `app.evaluate()` (main-process context) plus `global.__e2e*` refs to reach `ViewManager`, `WebContentsManager`, tray actions, etc. 

Its the same case with tray menus and native certificate dialogs. They are OS/main-process and not in the DOM. We need the `global.__e2e` refs.

---

In master we already have  ___e2eTestRefs_ (MainWindow, ServerManager, TabManager, ViewManager, WebContentsManager), ___e2eAppReady_ (intercom.ts), dock/taskbar badge hooks (badge.ts).

In this PR we are add more helper to test the scenarios mentioned in the beginning. 

> Why gate on NODE_ENV === 'test' (vs webapp-style always-on testdata)?

We do not want to push hooks to production app. In webapp we attache test fields to DOM. Here we're publishing handles on the Node main-process global. NODE_ENV === 'test' keeps normal user installs from getting those globals.

```release-note
NONE
```

## Change Impact: 🟡 Medium

**Regression Risk:** Refactors and expands the cross-platform Playwright/Electron E2E harness (worker/global Electron main-process lifecycle, PID-registry reaping, teardown/“fast teardown” termination behavior, and evaluation/timeout/retry logic), plus CI workflow/caching and report-tag generation. While production runtime behavior is gated to `NODE_ENV=test` via `setTestField`/test-only stubs (message box, tray/deep-link/menu click wiring, flash effects exposure), the teardown and main-process evaluation/timeout paths are sensitive and could cause flaky E2E behavior, orphaned processes, or teardown timing regressions—especially on macOS.

**QA Recommendation:** Rely primarily on automated CI E2E coverage; minimal manual QA. If any failures occur, manually validate on the failing OS job(s) that the existing specs run under Playwright, the failing scenario reproduces reliably, and Electron shuts down cleanly with no orphan processes (and that behavior differs as expected between normal vs “fast teardown”), paying extra attention to macOS defaults snapshot/restore.
*Generated by CodeRabbitAI*

### DIFF
--- a/.github/workflows/e2e-functional-template.yml
+++ b/.github/workflows/e2e-functional-template.yml
@@ -188,7 +188,9 @@ jobs:
         run: |
           RUNNER_OS=$(echo "${{ runner.os }}" | tr '[:upper:]' '[:lower:]')
           echo "RUNNER_OS=${RUNNER_OS}" >> $GITHUB_ENV
-          echo "CI_ENVIRONMENT_NAME=@${RUNNER_OS}" >> $GITHUB_ENV
+          # Report-only tag for Playwright blob/HTML output (@ci-* avoids colliding with
+          # platform selection grep tokens @linux/@darwin/@win32 in playwright.config.ts).
+          echo "CI_ENVIRONMENT_NAME=@ci-${RUNNER_OS}" >> $GITHUB_ENV
           
           # Define build type and suffix
           # inputs.TYPE takes precedence when provided by the caller (e.g. RELEASE, MASTER, PR).
@@ -235,8 +237,19 @@ jobs:
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: '22.x'
-          cache: "npm"
-          cache-dependency-path: package-lock.json
+          package-manager-cache: false
+          # node_modules is cached in e2e/cache-node-modules; disable setup-node's
+          # npm cache so actions/cache v5 does not hit legacy entries.
+
+      - name: e2e/use-gnu-tar-macos
+        if: runner.os == 'macOS'
+        run: |
+          # BSD tar on macOS can corrupt actions/cache v5 archives; prefer GNU tar when present.
+          if [ -x /opt/homebrew/opt/gnu-tar/libexec/gnubin/gtar ]; then
+            echo "/opt/homebrew/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
+          elif [ -x /usr/local/opt/gnu-tar/libexec/gnubin/gtar ]; then
+            echo "/usr/local/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
+          fi
 
       - name: e2e/cache-node-modules
         id: cache-node-modules
@@ -244,17 +257,41 @@ jobs:
         with:
           path: |
             node_modules
-            C:\Users\runneradmin\.electron-gyp
-          key: ${{ runner.os }}-build-node-modules-${{ hashFiles('**/package-lock.json') }}
+            e2e/node_modules
+          enableCrossOsArchive: false
+          # v7: exact restore only — avoids probing legacy v6 blobs that warn on macOS
+          key: ${{ runner.os }}-build-node-modules-v7-${{ hashFiles('**/package-lock.json') }}
+
+      - name: e2e/cache-electron-gyp-windows
+        if: runner.os == 'Windows'
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: C:\Users\runneradmin\.electron-gyp
+          enableCrossOsArchive: false
+          key: ${{ runner.os }}-electron-gyp-v7-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
-            ${{ runner.os }}-build-node-modules
-            ${{ runner.os }}-build-
-            ${{ runner.os }}-
+            ${{ runner.os }}-electron-gyp-v7-${{ hashFiles('**/package-lock.json') }}
+
+      # Clear pip's HTTP cache on macOS-26 runners before setup-python. The pre-baked
+      # image cache contains entries serialized by an older pip that the newer pip
+      # shipped with Python 3.10 cannot deserialize, producing the noisy
+      # "WARNING: Cache entry deserialization failed, entry ignored" lines in CI.
+      # See actions/setup-python#1317.
+      - name: e2e/clear-pip-cache-macos
+        if: runner.os == 'macOS'
+        run: |
+          rm -rf "${HOME}/Library/Caches/pip" || true
+          python3 -m pip cache purge >/dev/null 2>&1 || true
 
       - name: e2e/setup-python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        env:
+          # Bypass pip's on-disk HTTP cache entirely so a stale/corrupt entry on the
+          # runner image cannot trigger deserialization warnings.
+          PIP_NO_CACHE_DIR: "1"
         with:
           python-version: "3.10"
+          # Omit `cache` — setup-python only accepts pip/pipenv/poetry, not `false`.
 
       - name: e2e/install-os-dependencies
         uses: ./.github/actions/install-os-dependencies
@@ -287,7 +324,7 @@ jobs:
           cd e2e
           export PW_CHROMIUM_ARGS="--disable-gpu --no-sandbox --disable-dev-shm-usage"
           set +e
-          xvfb-run --auto-servernum --server-args='-screen 0 1024x768x24' npm test
+          xvfb-run --auto-servernum --server-args='-screen 0 1024x768x24' npm test -- --project=linux
           echo "PLAYWRIGHT_EXIT_CODE=$?" >> $GITHUB_ENV
         env:
           SERVER_VERSION: ${{ inputs.MM_SERVER_VERSION }}
@@ -303,7 +340,7 @@ jobs:
           echo "Running E2E tests…"
           cd e2e
           set +e
-          npm test
+          npm test -- --project=darwin
           echo "PLAYWRIGHT_EXIT_CODE=$?" >> $GITHUB_ENV
         env:
           SERVER_VERSION: ${{ inputs.MM_SERVER_VERSION }}
@@ -316,7 +353,7 @@ jobs:
           npm run build-test
           cd e2e
           set +e
-          npm test
+          npm test -- --project=win32
           echo "PLAYWRIGHT_EXIT_CODE=$?" >> $GITHUB_ENV
         env:
           SERVER_VERSION: ${{ inputs.MM_SERVER_VERSION }}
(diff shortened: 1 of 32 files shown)


lieut-data (commented):
@yasserfaraazkhan, I'll be stepping in to help review the product side of these changes. Could I ask you to update the pull request with a summary for reviewers like myself to understand the context going in, and then annotate the code with your own comments to  help guide the reviewer towards a thorough understanding?

As a quick example, I see there's new `setTestField` invocations, but it's not obvious to me what this is or why we need it.

Submitting this comment-only review for now, but please re-request when ready!

yasserfaraazkhan:
> @yasserfaraazkhan, I'll be stepping in to help review the product side of these changes. Could I ask you to update the pull request with a summary for reviewers like myself to understand the context going in, and then annotate the code with your own comments to help guide the reviewer towards a thorough understanding?


Thank you @lieut-data , I added PR description to the share more about the changes. Please let me know if I can add more information about any change

lieut-data (commented):
Thanks, @yasserfaraazkhan! Appreciate the updates -- I think I'm still missing something that helps me as a reviewer appreciate the story behind the changes.

For context, I've been steering my own PRs away from agent summaries towards keeping the reviewer audience in mind. For example, when I first submitted https://github.com/mattermost/mattermost/pull/37268, I got the technically accurate but impersonal summary as follows:

> The Makefile already declares setup-go-work as a prerequisite on most targets that need it. Several Cl jobs were calling make setup-go-work explicitly before the actual target, making those steps redundant.
> This PR: 
> * Removes all redundant explicit make setup-go-work steps from server-ci. yml 
> * Adds the missing setup-go-work prerequisite to the four targets that lacked it: modules-tidy, vet-api, default-roles-permissions, and a new check-go-fix target 
> * Adds check-go-fix (runs go fix -diff ./... , exits non-zero if any fixes would be applied) as a prerequisite of check-style, replacing the former standalone check- go-fix Cl job 
> * Removes the standalone check-go-fix Cl job entirely

but before I sent this to Alejandro, I realized I needed to give some practical context and rewrote as the following:

> In anticipation of some rework to Cl to support single-PR Go version bumps, I'm patching `server-ci.yml`'s handling of `make setup-go-work` and `go fix ./...`. The `Makefile` already mostly does `setup-go-work` for us, so we don't need to repeat it (though I've added a few missing cases), and `go fix ./...` is better handled as a `-diff` check inside `make check-style`.

Would you be open to helping me make sure I have right context in mind? For example, this PR makes heavy use of `setTestField`, and I see the new section `Why main-process hooks`, but I'm not familiar with the existing pattern in `master` or what the tradeoffs are if we only do this when `NODE_ENV === 'test'`. (In webapp, at least, I think we unconditionally attach `testdata` fields to nodes, so I'm guessing there's a reason we don't do there here?)

This handwritten "story" or "context" will hopefully help me provide feedback that's relevant and actionable for you and other reviewers :)

yasserfaraazkhan:
> Thanks, @yasserfaraazkhan! Appreciate the updates -- I think I'm still missing something that helps me as a reviewer appreciate the story behind the changes.

Sure @lieut-data, Thank you. I've updated the PR summary as well.

I have been trying to migrate desktop Rainforest tests to playwright.  This is the first part of having some test helpers and hooks for playwright to help cover such flows.

src/main/app/initialize.ts
```diff
@@ -294,6 +300,39 @@ async function initializeAfterAppReady() {
...
+        TrayIcon: Tray,
+        Diagnostics,
+        PopoutManager,
+    });
+
+    setTestField('__e2eOpenDeepLink', (url: string) => {
+        openDeepLink(url);
+    });
+
+    setTestField('__e2eClickTrayMenuItem', (label: string) => {
```
saturninoabril, line 313: Should this be set on `process.env.NODE_ENV === 'test'` only?

package.json
```diff
@@ -138,6 +138,7 @@
     "eslint-plugin-no-only-tests": "3.1.0",
     "eslint-plugin-react": "7.34.0",
     "eslint-plugin-react-hooks": "4.6.0",
+    "fast-xml-parser": "^5.8.0",
```
saturninoabril, line 141: Let's pin this to exact version.

.github/workflows/e2e-functional-template.yml
```diff
@@ -235,26 +237,61 @@ jobs:
...
+
+      - name: e2e/use-gnu-tar-macos
+        if: runner.os == 'macOS'
+        run: |
+          # BSD tar on macOS can corrupt actions/cache v5 archives; prefer GNU tar when present.
+          if [ -x /opt/homebrew/opt/gnu-tar/libexec/gnubin/gtar ]; then
+            echo "/opt/homebrew/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
+          elif [ -x /usr/local/opt/gnu-tar/libexec/gnubin/gtar ]; then
+            echo "/usr/local/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
+          fi
```
saturninoabril: macos-26 has gnu tar installed. How about simply:
```
- name: e2e/use-gnu-tar-macos
  if: runner.os == 'macOS'
  run: echo "/opt/homebrew/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
```

.github/workflows/e2e-functional-template.yml
```diff
@@ -235,26 +237,61 @@ jobs:
...
-            ${{ runner.os }}-build-node-modules
-            ${{ runner.os }}-build-
-            ${{ runner.os }}-
+            ${{ runner.os }}-electron-gyp-v7-${{ hashFiles('**/package-lock.json') }}
+
+      # Clear pip's HTTP cache on macOS-26 runners before setup-python. The pre-baked
+      # image cache contains entries serialized by an older pip that the newer pip
+      # shipped with Python 3.10 cannot deserialize, producing the noisy
+      # "WARNING: Cache entry deserialization failed, entry ignored" lines in CI.
+      # See actions/setup-python#1317.
```
saturninoabril: How about use later version of Python included in macos-26 runner (https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md#python)? Would this workaround not necessary?

e2e/helpers/notificationEffects.ts
```diff
@@ -0,0 +1,21 @@
...
+import type {ElectronApplication} from 'playwright';
+
+/**
+ * Invoke the production flashFrame() helper from src/main/notifications/index.ts.
+ *
+ * OS notification delivery is unreliable in headless CI (Electron's Notification
+ * often emits `failed` without `show`), so flash_taskbar and dock_bounce tests
+ * exercise the same flashFrame() gate the notification `show` handler calls.
+ */
+export async function triggerFlashEffects(app: ElectronApplication, flash = true): Promise<void> {
```
saturninoabril: It would be nice to align the name to notifications purpose detection, like `triggerNotificationEffects`, and `__e2eNotificationEffects`, if these are not use by any other purpose.

saturninoabril (changes requested):
Thanks @yasserfaraazkhan, looking good! Few comments left. And some from AI reviews, may be worth looking at:
```
Correctness bugs:

1. electronApp.ts:269 — app.close().catch(()=>{}).then(()=>{ closed=true; }) always sets closed=true even when close failed. Zombie processes never get the SIGTERM fallback.
2. electronApp.ts:167 — reapPids() on macOS sends only SIGTERM with no SIGKILL escalation. A blocked process survives the 5s poll and stays alive across test runs.
3. initialize.ts:322 — __e2eClickTrayMenuItem truncates the test-supplied label (not itemLabel) to match against native menu items. Any natively-truncated menu item will never match.
4. testMessageBoxStub.ts:11 — Module-level restoreMessageBox is unsafe across sequential install → install calls in the same worker; the second install skips saving the real original, permanently leaking the stub.

Efficiency issue:

5. electronApp.ts:230 — drainPlaywrightClose re-races the same already-expired closePromise for 3 more seconds after attemptClose's 10s budget, making worst-case teardown 13s not 10s.
```

src/main/app/initialize.ts
yasserfaraazkhan: Its already there at https://github.com/mattermost/desktop/blob/c436e1fe39dc66264c365a49800cc5c7070850aa/src/common/utils/util.ts#L76-L80

package.json
yasserfaraazkhan: done

.github/workflows/e2e-functional-template.yml
yasserfaraazkhan: done

.github/workflows/e2e-functional-template.yml
yasserfaraazkhan: done

e2e/helpers/notificationEffects.ts
yasserfaraazkhan: done

saturninoabril (approved):
Thanks @yasserfaraazkhan, LGTM! (Just need to address failing build)

src/main/app/initialize.ts
```diff
@@ -339,6 +378,17 @@ async function initializeAfterAppReady() {
     ServerManager.on(SERVER_URL_CHANGED, updateServerInfo);
     ServerManager.on(SERVER_PRE_AUTH_SECRET_CHANGED, updateServerInfo);
 
+    if (process.env.NODE_ENV === 'test') {
```
lieut-data, line 386: Looks like setTestField already checks `process.env.NODE_ENV === 'test'` -- can we simplify this?

src/main/app/initialize.ts
```diff
@@ -294,6 +300,39 @@ async function initializeAfterAppReady() {
...
+    });
+
+    setTestField('__e2eOpenDeepLink', (url: string) => {
+        openDeepLink(url);
+    });
+
+    setTestField('__e2eClickTrayMenuItem', (label: string) => {
+        const menu = createTrayMenu();
+        const stack = [...menu.items];
+        while (stack.length > 0) {
```
lieut-data: It took me a while to realize what was going on here -- flattening the menu in a kind of manual depth-first-search.

Would we be open to just writing this recursively, ala something like:
```ts
setTestField('__e2eClickTrayMenuItem', (label: string) => {
    const truncated = label.length > 50 ? `${label.slice(0, 50)}...` : label;

    function clickItem(items: Electron.MenuItemConstructorOptions[]): boolean {
        for (const item of items) {
            const itemLabel = typeof item.label === 'string' ? item.label : '';
            if (
                (itemLabel === label || itemLabel === truncated) &&
                item.enabled !== false &&
                item.visible !== false &&
                typeof item.click === 'function'
            ) {
                item.click();
                return true;
            }
            if (item.submenu?.items && clickItem(item.submenu.items)) {
                return true;
            }
        }
        return false;
    }

    if (!clickItem(createTrayMenu().items)) {
        throw new Error(`Tray menu item not found: ${label}`);
    }
});
```

lieut-data (approved):
Thanks, @yasserfaraazkhan! A few comments below, but not strictly blocking -- I don't need to re-review after you've decided next steps.

PR #3866, opened 2026-06-22
Increase max log file size to 10 MB

hanzei:
##  Summary
During a custom investigation, we enabled debug logs. Given the relatively small max log size, the log file did overflow quite quickly. This PR increases the max size to 10 MB.

## AI Summary

- Increases the `electron-log` max file size from the default ~1 MB to 10 MB, giving more log history for debugging
- Adds `countLogLevels()` to the diagnostics utils — streams the log file line-by-line without accumulating lines in memory, so the diagnostics step stays efficient at the larger file size
- Updates `step8.logHeuristics` to use `countLogLevels` instead of `readFileLineByLine`; the payload shape is unchanged

## Test plan

- [ ] Run the app and confirm logs are written to `main.log` as normal
- [ ] Run the diagnostics flow and confirm Step 8 completes without errors and reports correct log level counts
- [ ] Verify that `main.log` is not rotated until it exceeds 10 MB

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Change Impact: 🟡 Medium

**Regression Risk:** The log sizing tweak is isolated, but the new exported `countLogLevels()` changes diagnostics behavior by introducing streaming I/O + regex matching (including `lastIndex` handling). With missing unit tests for this logic, there’s risk of incorrect counts or edge-case failures on unusual/malformed log lines or boundary conditions, which would affect Step 8 diagnostics output (not core app workflows).  

**QA Recommendation:** Do light manual QA (recommended) on Step 8 end-to-end with representative logs (empty, small, near/over 10MB) to confirm accurate level counts and correct rotation timing; also verify diagnostics completion and payload shape in the UI/consumer. *Generated by CodeRabbitAI*


hanzei:
The e2e tests seems unrelated. Ignoring them.

src/main/diagnostics/steps/internal/utils.ts
```diff
@@ -183,11 +183,44 @@ function parseLogFileLine(line: string, lineMatchPattern: RegExp): LogFileLineDa
     };
 }
 
-/**
- * The current setup of `electron-log` rotates the file when it reaches ~1mb. It's safe to assume that the file will not be large enough to cause
- * issues reading it in the same process. If this read function ever causes performance issues we should either execute it in a child process or
- * read up to X amount of lines (eg 10.000)
- */
+export async function countLogLevels(path: fs.PathLike, lineMatchPattern = REGEX_LOG_FILE_LINE): Promise<{logLevelAmounts: LogLevelAmounts; linesCount: number}> {
```
devinbinnie, line 186: We seem to be adding a lot of custom code here just to increase the log file size. My question here is - does it matter if the max size is 1mb? I don't think there's too many cases where it's going to create that many files. Even on silly logging it takes a while to get to 1mb.

hanzei: The PR was prompted by a customer case where we did use debug logs enabled and wanted to check for an event that happened 4 weeks ago. They had debug log enabled (as we told them), but by the time we talked with the customer, the logs already rotate out fully.

Are there downsides with changing the log size? The disk usage is still low.

devinbinnie: I'm fine with the size but I was concerned about this extra code. However this is just diagnostics which is fine (sorry I'm just back from vacation 😅).

devinbinnie (approved):

PR #3872, opened 2026-06-26
MM-69421 - Scope server view network requests by origin and handle subframe navigation

pvev:
#### Summary

Adds a session-level request filter that scopes network requests from loaded server views to their configured origins. It also extends navigation handling to cover subframes in addition to main frame navigation.

* Includes unit and end-to-end test coverage.

## High-Level Architecture

The implementation uses two independent enforcement layers:

1. A session-level `onBeforeRequest` filter that:

   * Classifies each request by its owning view.
   * Applies the configured-origin policy.
   * Covers HTTP/HTTPS, WebSocket, and worker-originated requests.

2. Separate navigation handlers for:

   * Main-frame navigation.
   * Subframe navigation.

This ensures embedded content follows the same navigation policy as the top-level frame.

```text
                 Loaded server view (WebContentsView)
                         │                         │
                    navigation               network request
                         │                         │
              ┌──────────┴──────────┐              ▼
              ▼                     ▼      session onBeforeRequest
       will-navigate       will-frame-navigate             │
        (main frame)           (subframes)         identify owning view
              │                     │                       │
       main-frame policy     subframe policy        ┌───────┴───────┐
                                                 configured?       other
                                                     │              │
                                                   allow          apply
                                                               origin policy
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69241

#### Checklist
- [ x] Added or updated unit tests (required for all new features)
- [ x] Has UI changes
- [x ] read and understood our [Contributing Guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md)
- [x ] completed [Mattermost Contributor Agreement](https://mattermost.com/contribute/)
- [ x] executed `npm run lint:js` for proper code formatting
- [ x] Run E2E tests by adding label `E2E/Run`

#### Device Information
Apple M4 MAX 64 GB Tahoe

#### Release Note
```release-note
Improved how network requests and embedded content from connected servers are handled, keeping requests scoped to your configured servers
```

## Change Impact: Medium 🟡

**Regression Risk:** Medium likelihood of security/policy regressions because the change introduces a session-wide `webRequest.onBeforeRequest` cancellation filter and extends navigation gating to `will-frame-navigate`, impacting HTTP/HTTPS, WebSocket, redirects, and service-worker/worker-originated requests; coverage includes both unit tests and a dedicated local-network E2E spec, but behavior could still differ across Electron view ownership/webContents contexts.

**QA Recommendation:** Limited manual QA recommended: verify in an Electron run that (1) configured-origin requests still load, (2) local/private targets are blocked (including subframe embeds, redirects, and service-worker fetch behavior), and (3) subframe navigation handling matches expectations for `about:blank`/`about:srcdoc` while disallowed protocols are prevented.
*Generated by CodeRabbitAI*


e2e/helpers/localNetworkServers.ts
```diff
@@ -0,0 +1,100 @@
...
+            return;
+        }
+
+        // CORS so an allow-path fetch succeeds regardless of requesting origin.
+        res.setHeader('Access-Control-Allow-Origin', '*');
+        res.setHeader('Cache-Control', 'no-store');
+        res.writeHead(200, {'Content-Type': 'text/html'});
+        res.end(FAKE_SERVER_HTML);
+    });
+    const fakePort = await listen(fakeServer);
```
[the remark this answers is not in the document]
pvev: Good catch — wrapping the startup in try/catch and closing any already-bound server before rethrowing

src/app/views/webContentEvents.ts
```diff
@@ -30,13 +35,15 @@ import ViewManager from 'common/views/viewManager';
 import ContextMenu from 'main/contextMenu';
 import {localizeMessage} from 'main/i18nManager';
 
-import {generateHandleConsoleMessage, isCustomProtocol, isMattermostProtocol} from './webContentEventsCommon';
+import {generateHandleConsoleMessage, isAllowedSubframeNavigation, isCustomProtocol, isMattermostProtocol} from './webContentEventsCommon';
 
 import allowProtocolDialog from '../../main/security/allowProtocolDialog';
 import {composeUserAgent} from '../../main/utils';
```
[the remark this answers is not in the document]
pvev: This import and its casing/order are pre-existing and outside this PR's scope

src/main/app/initialize.test.js
```diff
@@ -323,5 +326,75 @@ describe('main/app/initialize', () => {
...
+            const SERVER_WEBCONTENTS_ID = 1;
+
+            const getRegisteredHandler = async () => {
+                const ServerManager = jest.requireMock('common/servers/serverManager');
+                const WebContentsManager = jest.requireMock('app/views/webContentsManager');
+                ServerManager.getAllServers.mockReturnValue([{url: new URL('http://127.0.0.1:8065')}]);
+                WebContentsManager.getViewByWebContentsId.mockImplementation((id) => (id === SERVER_WEBCONTENTS_ID ? {id} : undefined));
+                await initialize();
+                return session.defaultSession.webRequest.onBeforeRequest.mock.calls[0][0];
+            };
```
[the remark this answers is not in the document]
pvev: Strictly speaking, this is safe today because `afterEach` runs `jest.resetAllMocks()` (clears `mock.calls`), so there's only ever one registration per test. But taking the suggestion anyway—reading the last registration is more robust and clearer about intent.

src/main/app/initialize.ts
```diff
@@ -345,6 +346,34 @@ async function initializeAfterAppReady() {
...
+                ServerManager.getAllServers().map((server) => server.url),
+                (webContentsId) => Boolean(WebContentsManager.getViewByWebContentsId(webContentsId)),
+            );
+
+            if (shouldCancel) {
+                log.warn('Blocked server content from accessing local or private network URL', {
+                    resourceType: details.resourceType,
+                    url: details.url,
+                    webContentsId: details.webContentsId || details.webContents?.id,
+                });
```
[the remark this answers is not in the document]
pvev: Agreed — switching to log only the origin

edgarbellot (approved):
LGTM, happy path + edge cases perfectly covered 🚀 thank you!

src/main/app/initialize.ts
```diff
@@ -345,6 +346,34 @@ async function initializeAfterAppReady() {
 
     app.setAppUserModelId('Mattermost.Desktop'); // Use explicit AppUserModelID
     const defaultSession = session.defaultSession;
+    defaultSession.webRequest.onBeforeRequest(async (details, callback) => {
+        try {
+            const shouldCancel = await shouldCancelLocalNetworkRequest(
+                {
+                    url: details.url,
+                    webContentsId: details.webContentsId,
```
devinbinnie: Passing `webContents` and `webContentsId` is redundant, `webContentsId` is the same as `webContents.id`

src/main/app/initialize.ts
```diff
@@ -345,6 +346,34 @@ async function initializeAfterAppReady() {
...
+                    webContents: details.webContents,
+                },
+                ServerManager.getAllServers().map((server) => server.url),
```
devinbinnie: The method that we're calling can pull these values in themselves, we don't need to pass them in from here.

src/app/views/webContentEvents.ts
```diff
@@ -30,13 +35,15 @@ import ViewManager from 'common/views/viewManager';
...
 
 const log = new Logger('WebContentsEventManager');
 
+type NavigationEvent = Event<WebContentsWillNavigateEventParams | WebContentsWillFrameNavigateEventParams>;
```
devinbinnie: I don't think we need this type? The events are different and are shaped differently. Since we're creating different functions for each event type, we don't need to combine the types here.

src/app/views/webContentEvents.ts
```diff
@@ -128,6 +136,23 @@ export class WebContentsEventManager {
...
+            // so the policy (and any protocol dialog) does not run twice.
+            if (event.isMainFrame) {
+                return;
+            }
+
+            if (isAllowedSubframeNavigation(event.url)) {
+                return;
+            }
+
+            this.log(webContentsId).warn('Prevented subframe from navigating to a blocked protocol');
```
devinbinnie: Nit: this can be a `debug`, we don't need to fill up the logs with these in case someone does have a malicious iframe.

src/main/security/localNetworkAccess.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    if (isIP(address) === 6) {
+        return isLocalOrPrivateIPv6(address);
+    }
+
+    return false;
+}
+
+function isLocalOrPrivateIPv4(address: string): boolean {
```
devinbinnie: The implementations for checking specific IPv4 and IPv6 addresses here and below feel very janky and one thing I think we should be avoiding is bitwise operations in JS.

I checked with Claude and found a potential alternative that seems cleaner even if it's not perfect, it uses Node's `net.BlockList` object that allows us to specify CIDRs we want to block on and check those. Here's an example:

```
import {BlockList, isIP} from 'net';

// Local, loopback, and private ranges we refuse to let server content reach.
// Declarative CIDR data; matching is handled natively by BlockList.
const LOCAL_NETWORK_BLOCKLIST = new BlockList();

// IPv4
LOCAL_NETWORK_BLOCKLIST.addSubnet('0.0.0.0', 8, 'ipv4');      // unspecified / "this network"
LOCAL_NETWORK_BLOCKLIST.addSubnet('10.0.0.0', 8, 'ipv4');     // RFC 1918
LOCAL_NETWORK_BLOCKLIST.addSubnet('127.0.0.0', 8, 'ipv4');    // loopback
LOCAL_NETWORK_BLOCKLIST.addSubnet('169.254.0.0', 16, 'ipv4'); // link-local
LOCAL_NETWORK_BLOCKLIST.addSubnet('172.16.0.0', 12, 'ipv4');  // RFC 1918
LOCAL_NETWORK_BLOCKLIST.addSubnet('192.168.0.0', 16, 'ipv4'); // RFC 1918

// IPv6
LOCAL_NETWORK_BLOCKLIST.addSubnet('::1', 128, 'ipv6');        // loopback
LOCAL_NETWORK_BLOCKLIST.addSubnet('fc00::', 7, 'ipv6');       // unique local
LOCAL_NETWORK_BLOCKLIST.addSubnet('fe80::', 10, 'ipv6');      // link-local

export function isLocalOrPrivateIPAddress(address: string): boolean {
    const family = isIP(address);
    if (family === 4) {
        return LOCAL_NETWORK_BLOCKLIST.check(address, 'ipv4');
    }
    if (family === 6) {
        // Passing 'ipv6' lets BlockList match IPv4-mapped addresses
        // (::ffff:127.0.0.1) against the IPv4 rules above automatically.
        return LOCAL_NETWORK_BLOCKLIST.check(address, 'ipv6');
    }
    return false;
}
```

We should confirm with Prod Sec, but I think this would be a cleaner way to do this. Avoiding rolling our own IP detection here would be preferred.

devinbinnie (changes requested):
Thanks for the work here @pvev, I think I have a few worthwhile improvements we can make to make this more correct. We'll probably want @edgarbellot to re-review afterwards.

devinbinnie:
FYI this needs to be cherry-picked to v6.3 as well.

src/main/security/localNetworkAccess.ts
```diff
@@ -0,0 +1,125 @@
...
+
+LOCAL_NETWORK_BLOCKLIST.addSubnet('0.0.0.0', 8, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('10.0.0.0', 8, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('127.0.0.0', 8, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('169.254.0.0', 16, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('172.16.0.0', 12, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('192.168.0.0', 16, 'ipv4');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('::1', 128, 'ipv6');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('fc00::', 7, 'ipv6');
+LOCAL_NETWORK_BLOCKLIST.addSubnet('fe80::', 10, 'ipv6');
```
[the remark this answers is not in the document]
pvev: Good catch, added `fec0::/10` to the block list and covered it in the IPv6 range test.

pvev:
> DryRun Security
> 
> This pull request introduces a low-severity DNS rebinding vulnerability in the `isLocalOrPrivateHostname` function, where the non-atomic nature of the DNS check and subsequent network request could allow an attacker to bypass SSRF protections. Although the issue is flagged as risky, it is not currently set to block the merge.
> 
> DNS Rebinding in Hostname Validation in `src/main/security/localNetworkAccess.ts` (drs_026233b0)
> 
> **Vulnerability**	DNS Rebinding in Hostname Validation
> **Description**	The `isLocalOrPrivateHostname` function performs a DNS lookup to check if a hostname resolves to a private IP address. However, this check is performed independently of the actual network request made by the Electron application. Because the check and the subsequent network request are not atomic, an attacker could potentially use DNS rebinding (e.g., by providing a short TTL for the DNS record) to cause the check to pass for a benign IP address while the actual request is made to a private or loopback IP address. This is a known limitation of using `webRequest.onBeforeRequest` for SSRF protection without pinning the IP address or using other network-level mitigations.
> https://github.com/mattermost/desktop/blob/edcce10b9b01261f2ca3635a4a2ce8df90c8931d/src/main/security/localNetworkAccess.ts#L100-L103
> 
> Comment to provide feedback on these findings.
> Report false positive: `@dryrunsecurity fp [FINDING ID] [FEEDBACK]` Report low-impact: `@dryrunsecurity nit [FINDING ID] [FEEDBACK]`
> 
> Example: `@dryrunsecurity fp drs_90eda195 This code is not user-facing`
> 
> _All finding details can be found in the **[DryRun Security Dashboard](https://app.dryrun.security/pull-requests/13f24302-5d53-439a-b422-52632e117505?org=mattermost)**._

@dryrunsecurity nit drs_026233b0 This is a valid known limitation of doing hostname classification before Chromium performs the actual request resolution. It is not introduced by the `BlockList` change, and Electron does not expose the final resolved IP in `onBeforeRequest`, so closing this fully would require a different network-layer design. It is safe to ignore and out of the scope of this change.

src/main/app/initialize.ts
pvev: Good call, simplified this to rely on `webContentsId` only.

src/main/app/initialize.ts
pvev: Updated this so the request details are passed through directly and the helper owns the extraction logic.

src/app/views/webContentEvents.ts
pvev: Agreed, removed the combined type and typed the handlers independently.

src/app/views/webContentEvents.ts
pvev: Makes sense, changed this from `warn` to `debug`.

src/main/security/localNetworkAccess.ts
pvev: Agreed, replaced the custom IP range parsing with Node’s `BlockList` and kept the existing coverage for the expected ranges.

src/main/app/initialize.ts
```diff
@@ -331,6 +332,30 @@ async function initializeAfterAppReady() {
 
     app.setAppUserModelId('Mattermost.Desktop'); // Use explicit AppUserModelID
     const defaultSession = session.defaultSession;
+    defaultSession.webRequest.onBeforeRequest(async (details, callback) => {
+        try {
+            const shouldCancel = await shouldCancelLocalNetworkRequest(
+                details,
+                ServerManager.getAllServers().map((server) => server.url),
```
devinbinnie: Sorry, what I was saying before is that `ServerManager.getAllServers().map((server) => server.url)` can be called from `shouldCancelLocalNetworkRequest`. Actually, same with the function that gets the view by webcontentsid.

This call should just pass in the details (which I believe contains the webcontentsid as well)

src/main/security/localNetworkAccess.ts
```diff
@@ -0,0 +1,126 @@
...
+export type LocalNetworkRequestDetails = {
+    url: string;
+    webContentsId?: number;
+}
+
+type IsServerWebContents = (webContentsId: number) => boolean;
+
+const defaultLookup: LookupFunction = (hostname: string) => dns.lookup(hostname, {all: true, verbatim: true});
+
+const LOCAL_NETWORK_BLOCKLIST = new BlockList();
```
devinbinnie, line 25: Nit: maybe a quick comment here explaining what this list is (ie. local IP addresses)

devinbinnie (changes requested):
Thanks @pvev, huge improvement. I have two more concerns then we should be good to go.

edgarbellot:
Removing myself from the reviewers list since I still see there are pending changes. Please add me again after Devin has approved everything and I'll re-review :)

cc @pvev @devinbinnie

pvev:
@dryrunsecurity nit drs_508aff52 . This is a partially valid low severity residual limitation of the pre-request DNS check, rather than a new vulnerability introduced by the PR. Fully addressing it requires connection level enforcement or Chromium’s native Local Network Access controls. I suggest accepting this as a residual very low risk for this fix and perhaps tracking native enforcement separately, this to avoid introducing a high regression custom networking layer completely out of scope of this change.

devinbinnie (approved):
Thanks @pvev :) LGTM

edgarbellot (approved):
LGTM 🚀 thank you both!

PR #3877, opened 2026-06-30
fix(cmt): trigger on RC tag cut instead of release branch push

yasserfaraazkhan:
**Summary:**

The CMT Provisioner was set to fire on every push to `release-v*` branches, but desktop release branches are named `release-X.Y` (no `v` prefix) so the trigger never matched in practice. This PR fixes that.

**What changed:**

- **Trigger corrected:** CMT Provisioner now fires on RC tag cuts (`v*-rc*`, e.g. `v6.2.0-rc.1`) instead of release branch pushes. This matches how the release process actually works. An RC tag is cut when QA is ready to test, not on every branch push.

```release-note
NONE
```

## Change Impact: 🟢 Low

**Regression Risk:** Isolated GitHub Actions workflow changes with a narrow blast radius. The main risk is missed or unexpected workflow triggering on RC tags, but runtime behavior of the underlying job logic is largely unchanged.

**QA Recommendation:** Minimal manual QA is sufficient—verify the workflow triggers on a `v*-rc*` tag and still runs via manual dispatch.

### DIFF
--- a/.github/workflows/cmt-provisioner.yml
+++ b/.github/workflows/cmt-provisioner.yml
@@ -1,44 +1,25 @@
 name: CMT Provisioner
 
-# Lightweight trigger workflow for Compatibility Matrix Testing (CMT).
-#
-# Matterwick listens for this workflow's workflow_run:requested event, then provisions one
-# Mattermost cloud instance per version in its hardcoded CMT version set (Matterwick's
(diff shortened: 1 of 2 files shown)


saturninoabril (approved):
Thanks @yasserfaraazkhan! (Failed E2E tests to address separately)

PR #3883, opened 2026-07-06
[MM-69013] Disallow server adding via deep link when server management is disabled

devinbinnie:
#### Summary
The `enableServerManagement` policy is meant to stop users from adding new servers in locked-down deployments, but the deep link handler never consulted it. As a result, opening a `mattermost://` link to an unconfigured server would still surface the **Add Server** dialog and let the server be persisted, even with the policy set to `false`.

This PR adds the policy check to the deep link path in `NavigationManager`. When a deep link points to a server that isn't already configured and `enableServerManagement` is disabled, we log and bail out instead of opening the new-server modal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69013

#### Release Note
```release-note
Fixed an issue where a mattermost:// deep link could open the Add Server dialog
```

## Change Impact: 🟡 Medium

**Regression Risk:** Moderate risk: this introduces a policy-based early-exit in the desktop app’s deep-link/navigation handling for missing/unconfigured servers. While it only gates UI flows (dialogs/modals) and doesn’t affect auth/session/data integrity, it changes user-visible behavior across both primary-tab and new-tab link paths and may have edge-case variations depending on server/welcome-screen state.  
**QA Recommendation:** Light manual QA. Validate `mattermost://` deep links in both primary tab and new tab for: (1) an already configured server, and (2) a non-configured/missing server with `enableServerManagement` toggled true vs false, confirming the expected error/dialog behavior.
*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/navigationManager.test.js
+++ b/src/app/navigationManager.test.js
@@ -215,6 +215,21 @@ describe('app/navigationManager', () => {
             expect(handleWelcomeScreenModal).toHaveBeenCalledWith('server-2.com/deep/link?thing=yes');
         });
 
+        it('should not handle welcome screen modal when server management is disabled by policy', () => {
+            ModalManager.removeModal.mockClear();
+            handleWelcomeScreenModal.mockClear();
+            ServerManager.hasServers.mockReturnValue(false);
+            ServerManager.lookupServerByURL.mockReturnValue(null);
+            Config.enableServerManagement = false;
+
+            navigationManager.openLinkInPrimaryTab('mattermost://server-2.com/deep/link?thing=yes');
+
+            expect(ModalManager.removeModal).not.toHaveBeenCalled();
+            expect(handleWelcomeScreenModal).not.toHaveBeenCalled();
+
+            Config.enableServerManagement = true;
+        });
+
         it('should handle null URL gracefully', () => {
             navigationManager.openLinkInPrimaryTab(null);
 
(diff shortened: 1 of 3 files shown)


src/app/navigationManager.test.js
```diff
@@ -73,6 +74,10 @@ jest.mock('common/servers/serverManager', () => ({
     getServer: jest.fn(),
 }));
 
+jest.mock('common/config', () => ({
+    enableServerManagement: true,
+}));
```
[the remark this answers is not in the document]
devinbinnie: This is incorrect and unnecessary.

src/app/navigationManager.ts
```diff
@@ -89,8 +90,16 @@ export class NavigationManager {
                 webContentsView.load(urlWithSchema);
             }
         } else if (ServerManager.hasServers()) {
+            if (!Config.enableServerManagement) {
+                log.warn('Ignoring deep link to add a new server because server management is disabled by policy');
+                return;
+            }
```
davidkrauser: Should we show the user something when this happens? I'm assuming the UX will be less-than-ideal in this situation: they click a `mattermost://` link, it opens the app, then the app does... nothing?

davidkrauser (approved):
LGTM, just had a question about the UX

src/app/navigationManager.ts
devinbinnie: Good call, added a dialog box.

PR #3884, opened 2026-07-06
[MM-69253] Restrict call leave to the view that created it and the widget

devinbinnie:
#### Summary
The `leaveCall` desktop API did not validate the sender, so a server view could affect a call owned by a different server. This PR adds sender validation so the request is only honored when it comes from the server that owns the call.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69253

#### Release Note
```release-note
Fixed an issue where other views could end an active call
```

## Change Impact: 🟠 Medium

**Regression Risk:** This changes sender validation in the call-leave IPC flow, which is a user-facing call-control path and introduces authorization-like checks. The affected code is localized to the Calls widget window, and the new tests cover the main allowed/denied cases, but there is still risk around edge cases where sender identity or window state differs.

**QA Recommendation:** Recommend targeted manual QA for leaving calls from the widget and from the owning view, plus a quick negative check that other views cannot end the call.
*Generated by CodeRabbitAI*


edgarbellot (approved):
LGTM, thank you Devin!

PR #3885, opened 2026-07-07
e2e(ci): bootstrap public links and flaky-test reporting for Playwright CI

yasserfaraazkhan:
Rainforest migration needs ephemeral E2E servers to allow public file links before Playwright runs, and CI needs clearer per-OS flaky test reporting.

This PR adds a pre-test script that enables `FileSettings.EnablePublicLink` when the cloud server has it off, wires it into the E2E workflow template, and extends the flaky-test analyzer used when posting results.

It also includes the hooks slice from #3886 (tray menu test hook and history/view menu fixes) because that PR merged into this branch.

Slice of #3853 for easier review. Merge order: #3885 → #3887 → #3888 → #3889.

#### Release Note

```release-note
NONE
```


src/main/e2e/trayMenu.ts
```diff
@@ -1,32 +1,112 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import type {Menu} from 'electron';
+import type {Menu, MenuItem} from 'electron';
```
devinbinnie, line 4: Shouldn't this file and its test be in the `e2e` root folder? I don't see it being called from the main code.

devinbinnie (changes requested):
Everything looks good, just one thing we should do.

src/main/e2e/trayMenu.ts
yasserfaraazkhan: @devinbinnie `register.ts:61` uses it. We wire `createClickTrayMenuItem(createTrayMenu)` into `__e2eClickTrayMenuItem`.

I tried moving it locally to the `e2e/` folder, but that meant changing ESLint/webpack so the main bundle could compile stuff from `e2e/`, which felt like extra overhead ffor just a test hook.

Should I add a short comment in the file explaining why we keep it here?

devinbinnie: Okay, so this is all new stuff I wasn't aware of - guess it was merged while I was away.

I don't think we need a comment, but I would like to avoid E2E specific code in the main process. I guess since this is self-contained we can leave it for now, but if there's a way to pull it out into the `e2e` root folder we should.

I'll approve for now but we should think about that going forward.

devinbinnie (approved):

PR #3887, opened 2026-07-07
e2e(helpers): Playwright helpers for server views and desktop

yasserfaraazkhan:
Migrating Rainforest cases to Playwright surfaced repeated patterns: waiting for a server view to be interactive, opening native menus, copying public links, and recovering from blank channel states.

This PR adds shared helpers and fixture wiring so specs stay focused on behavior instead of re-implementing Electron glue in every file.

Slice of #3853. E2E coverage validated on #3853. Merge after #3885.

#### Release Note

```release-note
NONE
```

## Change Impact: 🟡 Medium

**Regression Risk:** The changes are concentrated in shared E2E helpers and fixture wiring, so regressions would mostly affect desktop Playwright test stability and coverage rather than product runtime behavior. Risk is elevated because several widely reused helpers were refactored or added, but the blast radius is still limited to test code.

**QA Recommendation:** Light manual QA is sufficient; prioritize running the affected E2E suites and a few representative desktop flows (menu navigation, startup, settings, tray, and login) to confirm the new helpers behave consistently across platforms.
*Generated by CodeRabbitAI*


e2e/helpers/channelMenu.ts
```diff
@@ -28,10 +28,12 @@ export const COPY_LINK_SELECTORS = [
  * The webapp migrated from #channelHeaderDropdownButton to Menu.Button
  * with an aria-label like "off-topic channel menu".
  */
-export async function openChannelHeaderMenu(win: ServerView): Promise<void> {
-    await win.waitForSelector(CHANNEL_HEADER_MENU_TRIGGER, {state: 'visible', timeout: 15_000});
+export async function openChannelHeaderMenu(win: ServerView, timeout = 20_000): Promise<void> {
+    const menuTimeout = Math.min(Math.max(Math.floor(timeout * 0.25), 500), 5_000);
+    const triggerTimeout = Math.max(timeout - menuTimeout, 500);
```
saturninoabril, line 33: Why these timeouts has to be computed?

saturninoabril (approved):

e2e/helpers/channelMenu.ts
yasserfaraazkhan: @saturninoabril this was mostly an AI suggestion while I was fixing repeated CI flakiness around the channel header menu, especially in enableBookmarksBar.

The idea is timeout should mean total wait for the helper, not per step. Earlier we had fixed 15s for the trigger + 5s for the menu. When enableBookmarksBar retries and passes remaining, each attempt could still burn the full 15s + 5s. So even with a 15s outer deadline, one retry could overshoot, eat into the test timeout, and fail in a confusing way.

With the split, if you pass 20s, trigger gets ~15s and menu gets ~5s — together ~20s, not 40s. That way we stay within the budget and it’s clearer whether it failed waiting for the trigger or for the menu to open.

PR #3890, opened 2026-07-07
[MM-69724] Allow WS updates to update session attribute manifest rather than rely on refetch every time

devinbinnie:
#### Summary
The current session attribute implementation reacts to a websocket event by having the Desktop App refetch the entire manifest. This works, but as session attribute configuration can change while a server is live, repeatedly refetching the whole manifest is unnecessary work and could become a performance concern over time.

This PR adds an `updateSessionAttribute` bridge on the external API so the web app can forward a single changed property field over IPC. The `SessionAttributesManager` maps that payload into a manifest field and upserts it in place.

Web App PR: https://github.com/mattermost/mattermost/pull/37392

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69724

#### Release Note
```release-note
NONE
```

## Change Impact: 🟡 Medium

**Regression Risk:** Medium risk. The change spans the public Desktop API type surface, IPC contract (new channel + payload), validation, and main-process session-attribute manifest upsert logic. While behavior is well-scoped to session-attribute updates and covered by new Jest cases (add/replace/remove/disabled-platform behavior and ignoring invalid payloads), malformed payloads or platform/enabled filtering mistakes could affect request header generation.

**QA Recommendation:** Limited manual QA is recommended—verify end-to-end session attribute updates via the Desktop App (IPC delivery, enable/disable behavior, and desktop-platform targeting) and confirm subsequent request headers reflect the updated manifest. *Generated by CodeRabbitAI*

### DIFF
--- a/src/main/sessionAttributes/sessionAttributesManager.ts
+++ b/src/main/sessionAttributes/sessionAttributesManager.ts
@@ -9,6 +9,7 @@ import {
     SERVER_PRE_AUTH_SECRET_CHANGED,
     SERVER_REMOVED,
     SERVER_URL_CHANGED,
+    SESSION_ATTRIBUTES_FIELD_UPDATED,
     SESSION_ATTRIBUTES_MANIFEST_INVALIDATED,
     SESSION_ATTRIBUTES_RESEND_REQUESTED,
 } from 'common/communication';
@@ -18,9 +19,10 @@ import {Logger} from 'common/log';
 import type {MattermostServer} from 'common/servers/MattermostServer';
 import ServerManager from 'common/servers/serverManager';
 import {parseURL} from 'common/utils/url';
+import {ipcValidate, sessionAttributeFieldSchema} from 'common/Validator';
 import {updateServerInfos} from 'main/app/utils';
 
-import type {SAField} from 'types/sessionAttributes';
+import type {SAField, SAPropertyField} from 'types/sessionAttributes';
 
 import SessionAttributeCollector from './collector';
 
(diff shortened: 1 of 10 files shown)


src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -98,6 +101,8 @@ export class SessionAttributesManager {
             return undefined;
         }
 
+        log.info('Sending session attributes', {serverId: server.id, payload});
+
```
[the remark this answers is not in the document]
devinbinnie: I didn't realize I'd left that in - can you create a new suggestion to remove that line entirely?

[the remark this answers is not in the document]
devinbinnie:
```suggestion
```

src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -197,6 +200,38 @@ export class SessionAttributesManager {
         }
         this.lastSentAt.delete(serverId);
     };
+
+    private handleFieldUpdated = (event: IpcMainEvent, field: SAPropertyField) => {
+        log.debug('handleFieldUpdated', {name: field.name});
```
marianunez, line 205: Do we want to keep this debug log?

src/types/sessionAttributes.ts
```diff
@@ -9,4 +9,15 @@ export type SAField = {
...
+export type SAPropertyField = {
+    name: string;
+    type: string;
+    attrs: {
+        enabled: boolean;
+        ttl_seconds: number;
+        grace_period_seconds: number;
+        platforms: string[];
+    };
+};
```
marianunez, line 21: Any opportunity to only keep one definition for session attributes? We have 2 in this file and then the ones in the API definition as well.

src/main/sessionAttributes/sessionAttributesManager.ts
devinbinnie: Yeah this one I'm good with. We shouldn't see this happening a lot, but it will be useful when this feature launches if we need to ensure that the updates are coming through.

src/types/sessionAttributes.ts
devinbinnie: Yeah, we could do some type deriving, but I don't think there's much value since we'll just be using the two types defined here anyways. I think there's some value in keeping them explicit like this so that we know the exact structure each method uses.

marianunez (approved):

PR #3891, opened 2026-07-07
chore(e2e): Add TSIO reporting to desktop repo

yasserfaraazkhan:
## Summary

1. Desktop E2E runs used to surface several small, easy-to-miss status checks on a PR, and there was no single place to see that tests were still in progress. Reviewers had to open multiple jobs to see what passed or failed. Playwright E2E reports are uploaded to Test System IO (TSIO). Each run now publishes one commit status `e2e-test/desktop-playwright`. 

2. Similar reporting has been applied to `CMT` and `master` runs. PR and master share `e2e-test/desktop-playwright`. CMT uses `e2e/compatibility-matrix-testing` status check. The TSIO report name (desktop-pr vs desktop-master vs cmt-desktop) is what tells them apart inside Test System IO.

3. E2E app-readiness is registered earlier in main-window startup to avoid flaky waitForAppReady timeouts, Linux worker teardown no longer hangs after a green run, and new Playwright coverage was added for tray restore and the Help menu “Learn More” flow.

#### Release Note

```release-note
NONE
```

## Change Impact: Medium 🟠

**Regression Risk:** Medium—primarily CI/workflow contract changes: TSIO opt-in gating/config plumbing, consolidated final status reporting via new OIDC-based polling, removal of previous per-OS flaky/JUnit analysis, Playwright reporter switch (JUnit→JSON), and a shared commit-status utility behavior change (single status context). These can break or misreport status checks/links without affecting app runtime, but the blast radius spans multiple E2E workflows and job paths.

**QA Recommendation:** Validate end-to-end on the affected workflow variants (desktop functional + policy, and PR trigger) for both TSIO-enabled and TSIO-disabled legs. Specifically confirm: pending→final commit status transitions, correct commit-status context name, TSIO report link/rollup for pass/fail (including upstream-failure scenarios), and that Playwright non-zero exits fail only under the intended TSIO-disabled conditions. Only light manual QA is recommended beyond reviewing CI logs/artifacts and resulting commit statuses.
*Generated by CodeRabbitAI*


saturninoabril (dismissed):
@yasserfaraazkhan Since this is using Playwright, I'd recommend to use actions at `mattermost/mattermost-test-system-io/.github/actions/test-system-io-*`, modify as needed so we don't need to maintain separately its interface to test-system-io. It would be good to have the same flow of `begin` -> `dispatch-run` -> `summary` so that we'll see the in progress status while the test is running. But of course, it can only use the `summary` as well but the report will only show up when all tests are done. We discuss the details offline.

yasserfaraazkhan:
> @yasserfaraazkhan Since this is using Playwright, I'd recommend to use actions at `mattermost/mattermost-test-system-io/.github/actions/test-system-io-*`, modify as needed so we don't need to maintain separately its interface to test-system-io. It would be good to have the same flow of `begin` -> `dispatch-run` -> `summary` so that we'll see the in progress status while the test is running. But of course, it can only use the `summary` as well but the report will only show up when all tests are done. We discuss the details offline.

@saturninoabril In desktop we want one TSIO report per run covering all platform (and CMT OS×server legs). Report-upload fits our case here. The orchestration fits spec sharding against one server. Here (and in mobile) we use 3 servers for different platform. There are also separate jobs verifying Policies on MacOS, windows. It'll be a little over engineering to make TSIO fit this case and filter spec based one `@linux`, `@win` etc. 

I prefer we keep the upload summary at the end of the run. TSIO give better reporting by gathering screenshots,  stack trace and combining tests across platform into 1 Report for easy access. 

Coderabbit's suggestions made me try some over engineering work, but I've reset it and it works.
image


1 thing in TSIO needs to change is that if same test file fails across platforms it only shows it as 1 failure on the table header. I'm working on few more fixes to test locally on https://github.com/mattermost/mattermost-test-system-io/pull/76 and will push and update.

yasserfaraazkhan:
@coderabbit approve

devinbinnie (approved):

PR #3927, opened 2026-07-31
[MM-68968] Guard redirects and subframe navigations alongside will-navigate

devinbinnie:
#### Summary

Our navigation checks were only registered on `will-navigate`, which Electron emits for top-level, renderer-initiated navigations. The other navigation events it exposes — `will-redirect` and `will-frame-navigate` — were never wired up, so navigations arriving through those paths were never evaluated against the same rules.

This PR registers the existing checks on those events too, covering server views, the Calls widget window, and plugin popups. Redirects are dispatched by frame type so that subframes continue to use the subframe check rather than the main-frame one, since ordinary embeds redirect as a matter of course. 

The subframe guard moved into `webContentEventsCommon` so all three call sites can share a single implementation.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-68968

#### Release Note
```release-note
Improved handling of page redirects and embedded content navigation across Desktop App windows.
```

## Change Impact: 🟡 Medium

**Regression Risk:** Changes affect shared navigation handling across multiple Electron windows. Automated tests cover the updated main-frame and subframe paths.

**QA Recommendation:** Manual QA can be skipped because automated coverage is complete and rollback is straightforward.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/callsWidgetWindow.test.js
+++ b/src/app/callsWidgetWindow.test.js
@@ -646,6 +646,33 @@ describe('main/windows/callsWidgetWindow', () => {
         expect(ev.preventDefault).toHaveBeenCalledTimes(1);
     });
 
+    it('onRedirect', () => {
+        const callsWidgetWindow = new CallsWidgetWindow();
+        callsWidgetWindow.getWidgetURL = () => 'http://localhost:8065';
+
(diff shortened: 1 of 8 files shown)


edgarbellot (approved):
LGTM 👍 main path + edge cases covered, thank you!

PR #3932, opened 2026-08-05
Fix Linux desktop entry Exec path handling

nivaas219:
#### Summary

Fix Linux desktop entry generation by preserving proper Exec path quoting.

This prevents incorrect parsing of executable paths containing spaces in generated Linux desktop files.

#### Ticket Link

N/A

#### Checklist

-  Added or updated unit tests (required for all new features)
-  Has UI changes
-  read and understood our Contributing Guidelines
-  completed Mattermost Contributor Agreement
-  executed `npm run lint:js` for proper code formatting
-  Run E2E tests by adding label `E2E/Run`

#### Device Information

This PR was tested on: Windows 11 (development environment)

#### Screenshots

Not applicable - no UI changes.

#### Release Note

```release-note
Fixed Linux desktop entry Exec path handling.
```

## Change Impact: 🟢 Low

**Regression Risk:** The change is isolated to Linux desktop entry generation. It does not affect shared code, APIs, data, or critical flows.

**QA Recommendation:** Manual QA can be skipped because the change has low risk and automated test coverage is available.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/assets/linux/create_desktop_file.sh
+++ b/src/assets/linux/create_desktop_file.sh
@@ -9,7 +9,7 @@ cat <<EOS > Mattermost.desktop
 [Desktop Entry]
 Name=Mattermost
 Comment=Mattermost Desktop application for Linux
-Exec=${FULL_PATH}/mattermost-desktop %U
+Exec="${FULL_PATH}/mattermost-desktop" %U
 Terminal=false
 Type=Application
 MimeType=x-scheme-handler/mattermost
--- a/src/assets/linux/create_desktop_file.sh
+++ b/src/assets/linux/create_desktop_file.sh
@@ -15,7 +15,7 @@ Exec="${FULL_PATH}/mattermost-desktop" %U
 Terminal=false
 Type=Application
 MimeType=x-scheme-handler/mattermost
-Icon=${FULL_PATH}/app_icon.png
+Icon="${FULL_PATH}/app_icon.png"
 Categories=Network;InstantMessaging;
 EOS
 
--- a/src/assets/linux/create_desktop_file.sh
+++ b/src/assets/linux/create_desktop_file.sh
@@ -1,10 +1,12 @@
 #!/bin/sh
 set -e
-WORKING_DIR=`pwd`
-THIS_PATH=`readlink -f $0`
-cd `dirname ${THIS_PATH}`
-FULL_PATH=`pwd`
-cd "${WORKING_DIR}"
+
+WORKING_DIR=$(pwd)
+THIS_PATH=$(readlink -f "$0")
+cd "$(dirname "$THIS_PATH")"
+FULL_PATH=$(pwd)
+cd "$WORKING_DIR"
+
 cat <<EOS > Mattermost.desktop
 [Desktop Entry]
 Name=Mattermost
@@ -16,4 +18,5 @@ MimeType=x-scheme-handler/mattermost
 Icon=${FULL_PATH}/app_icon.png
 Categories=Network;InstantMessaging;
 EOS
+
 chmod +x Mattermost.desktop
--- a/src/assets/linux/create_desktop_file.sh
+++ b/src/assets/linux/create_desktop_file.sh
@@ -9,7 +9,7 @@ cat <<EOS > Mattermost.desktop
 [Desktop Entry]
 Name=Mattermost
 Comment=Mattermost Desktop application for Linux
-Exec="${FULL_PATH}/mattermost-desktop" %U
+Exec=${FULL_PATH}/mattermost-desktop %U
 Terminal=false
 Type=Application
 MimeType=x-scheme-handler/mattermost


nivaas219:
```release-note
Fixed Linux desktop entry Exec path handling.

**Comment by devinbinnie** [comment]
@nivaas219 Looks like the suggestion from CodeRabbit overwrote your fix. Was that intentional?

**Comment by nivaas219** [comment]
> @nivaas219 Looks like the suggestion from CodeRabbit overwrote your fix. Was that intentional?

@devinbinnie  ,Thanks for checking. No, it is not intentional. The CodeRabbit suggestion overwrote the original fix, so I restored the quoted Exec path handling in commit https://github.com/mattermost/desktop/commit/b6888498f356dfaed88edb9f5f1a4541c942e946. The current change keeps the Exec path properly quoted to support paths containing spaces.

**Comment by devinbinnie** [comment]
@nivaas219 This PR is still showing 0 files changed.

**Comment by nivaas219** [comment]
> @nivaas219 This PR is still showing 0 files changed.

@devinbinnie , I made a small follow-up cleanup to the Linux desktop entry script, mainly improving shell path handling and quoting. The existing quoted Exec path fix is unchanged. I also verified the script with bash -n and tested the generated desktop file.

**Comment by nivaas219** [comment]
@devinbinnie ,The macOS check appears to be failing in the E2E label orchestration step due to a 403 permission error when removing the `E2E/Run` label (`Resource not accessible by integration`). This looks unrelated to the code changes in this PR.

**Comment by devinbinnie** [comment]
@nivaas219 So while I don't have any issues with this change, I couldn't get the .desktop file to work on my Linux VM. There's other errors stopping the app from running when there's a space in the path (a Chromium bug for one). When I turned off the sandbox to get around that, it still wouldn't run. So I'm not sure if this change is actually warranted.

Can you explain where this is useful?

**Comment by nivaas219** [comment]
@devinbinnie  Thanks for testing this.

My understanding was that the generated `.desktop` file should remain valid according to the Desktop Entry specification, which requires quoting when the executable path contains spaces or other reserved characters. The change was intended to ensure the generated `Exec` entry is parsed correctly as a single executable path.

I agree that if Mattermost still cannot launch from such paths due to unrelated issues (for example Chromium or Electron limitations), then this change may not provide a practical user-facing benefit today.

The motivation here was correctness of the generated `.desktop` file rather than resolving all launch issues for installations in paths containing spaces. If that use case is not something we intend to support, I'm happy to defer to the maintainers on whether this change should be kept.

**Comment by devinbinnie** [comment]
> @devinbinnie Thanks for testing this.
> 
> My understanding was that the generated `.desktop` file should remain valid according to the Desktop Entry specification, which requires quoting when the executable path contains spaces or other reserved characters. The change was intended to ensure the generated `Exec` entry is parsed correctly as a single executable path.
> 
> I agree that if Mattermost still cannot launch from such paths due to unrelated issues (for example Chromium or Electron limitations), then this change may not provide a practical user-facing benefit today.
> 
> The motivation here was correctness of the generated `.desktop` file rather than resolving all launch issues for installations in paths containing spaces. If that use case is not something we intend to support, I'm happy to defer to the maintainers on whether this change should be kept.

Thanks for the context. I guess there's no harm in keeping this in case Chromium ever fixes their bug, so I'm okay with letting it through.

File: src/assets/linux/create_desktop_file.sh [hunk 1]
```diff
@@ -16,4 +18,5 @@ MimeType=x-scheme-handler/mattermost
 Icon=${FULL_PATH}/app_icon.png
```
devinbinnie: Should we do this for the icon as well?

devinbinnie (approved):
Thanks @nivaas219, one comment but non-blocking.

nivaas219:
> Thanks @nivaas219, one comment but non-blocking.

@devinbinnie, Good catch. The same quoting consideration could apply to the Icon path as well. I left it unchanged to keep this PR focused on the Exec entry, but I'm happy to update it for consistency if you'd prefer.

devinbinnie:
> > Thanks @nivaas219, one comment but non-blocking.
> 
> @devinbinnie, Good catch. The same quoting consideration could apply to the Icon path as well. I left it unchanged to keep this PR focused on the Exec entry, but I'm happy to update it for consistency if you'd prefer.

If we are going for correctness, I think we should ;)

nivaas219:
> > > Thanks @nivaas219, one comment but non-blocking.
> > 
> > 
> > @devinbinnie, Good catch. The same quoting consideration could apply to the Icon path as well. I left it unchanged to keep this PR focused on the Exec entry, but I'm happy to update it for consistency if you'd prefer.
> 
> If we are going for correctness, I think we should ;)

Thanks @devinbinnie  agreed. I’ve also quoted the Icon path for consistency with the Exec path.

nivaas219:
@devinbinnie , The E2E PR Trigger is failing before the E2E test starts. The workflow receives a 403 Resource not accessible by integration when trying to remove the E2E/Run label from the PR. The build and security checks are passing, so this appears to be a workflow permission issue rather than a code-related failure. Could you please take a look or rerun the E2E workflow?

nivaas219:
@devinbinnie ,
Thank you so much for the review and for merging this! Really appreciate your patience and thorough feedback throughout the process.

PR #3937, opened 2026-08-10
[MM-70187] Show session attribute values in Settings Modal

devinbinnie:
#### Summary
Session attributes are collected by the Desktop App and sent to the server for ABAC policy evaluation, but users have no way to see what values the app is actually reporting for debugging purposes, which is helpful when someone is unexpectedly denied access to a resource.

This PR adds a read-only table under the "Enable session attributes" checkbox in Settings > Advanced, listing the common attributes the app is capable of collecting.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-70187

#### Screenshots
Screenshot 2026-08-10 at 3 09 34 PM

#### Release Note
```release-note
Add a table in Settings that shows the session attribute values currently reported by the app
```

## Change Impact: 🟡 Medium

**Regression Risk:** Changes span the renderer, preload API, IPC communication, and session attribute management. Automated tests cover the main paths, but session attribute collection and the new settings UI affect user-facing behavior.

**QA Recommendation:** Perform targeted manual QA for the Advanced settings table, disabled collection, unavailable values, and active server context.

*Generated by CodeRabbitAI*


src/main/sessionAttributes/sessionAttributesManager.ts
```diff
@@ -104,6 +106,22 @@ export class SessionAttributesManager {
         return Buffer.from(JSON.stringify(payload)).toString('base64');
     };
 
+    // server_fqdn is the only server-specific attribute, so it is reported for the active server
```
harshilsharma63: I think the comment isn't clear. If I understand correctly, we're skipping server FQDN from here because its a server-only attribute. But the comment is confusing.

harshilsharma63 (approved):

src/main/sessionAttributes/sessionAttributesManager.ts
devinbinnie: Agreed, it's actually a pointless comment I must have left there. FWIW, the reason it's not included is that it would be different per server, and the modal is server-agnostic. There's also no point to show it since it's just whatever your configured server URL is. We can document that so users know.

hmohammed-prodsec (approved):
LGTM 👍

PR #3938, opened 2026-08-10
Fix issue with validation where local HTTP servers will fail to validate

devinbinnie:
#### Summary
The local network request policy on the default session cancels requests to local or private addresses that aren't a configured server, so validating a not-yet-added server on `localhost` or a private IP had its probes blocked and came back as not a Mattermost server.

This PR runs the validation ping and config requests on a dedicated `server-validation` session, which the policy isn't registered on. Server views still use the default session, so the policy continues to apply to server content.

#### Release Note
```release-note
Fixed an issue where adding a Mattermost server on the local network, such as localhost or a private IP address, would fail to validate.
```

## Change Impact: 🟡 Medium

**Regression Risk:** The change spans server validation, session management, authentication headers, and shared API code. Automated tests cover the new session flow, but session-related regressions remain possible.

**QA Recommendation:** Manual QA can be skipped because automated coverage is complete and rollback is straightforward.

*Generated by CodeRabbitAI*

### DIFF
--- a/src/app/serverHub.ts
+++ b/src/app/serverHub.ts
@@ -455,10 +455,10 @@ export class ServerHub {
         const serverInfo = new ServerInfo(server);
         try {
             // Ping server first for pre-auth - config endpoint might be whitelisted
-            await serverInfo.pingServer();
+            await serverInfo.pingServer(session.fromPartition('server-validation'));
 
             // Only proceed to fetch config if ping is successful
-            const remoteInfo = await serverInfo.fetchConfigData();
+            const remoteInfo = await serverInfo.fetchConfigData(session.fromPartition('server-validation'));
             return {data: remoteInfo};
         } catch (error) {
             return {error: error as Error & { errorReason?: ErrorReason }};
--- a/src/app/serverHub.test.js
+++ b/src/app/serverHub.test.js
@@ -1,6 +1,8 @@
 // Copyright (c) 2016-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {session} from 'electron';
+
 import MainWindow from 'app/mainWindow/mainWindow';
 import ModalManager from 'app/mainWindow/modals/modalManager';
 import {MattermostServer} from 'common/servers/MattermostServer';
@@ -26,6 +28,7 @@ jest.mock('electron', () => ({
             allowNTLMCredentialsForDomains: jest.fn(),
             clearData: jest.fn(),
         },
+        fromPartition: jest.fn(),
     },
 }));
 
@@ -282,8 +285,10 @@ describe('app/serverViewState', () => {
 
     describe('handleServerURLValidation', () => {
         const serverViewState = new ServerHub();
+        const validationSession = {id: 'validation-session'};
 
         beforeEach(() => {
+            session.fromPartition.mockReturnValue(validationSession);
             MattermostServer.mockImplementation(({url}) => ({url}));
             ServerInfo.mockImplementation(({url}) => ({
                 pingServer: jest.fn().mockImplementation(() => ({
@@ -315,6 +320,22 @@ describe('app/serverViewState', () => {
             const result = await serverViewState.handleServerURLValidation({}, 'server.com');
             expect(result.status).toBe(URLValidationStatus.OK);
             expect(result.validatedURL).toBe('https://server.com/');
+        });
+
+        it('should test the server on a separate session', async () => {
+            const pingServer = jest.fn();
+            const fetchConfigData = jest.fn().mockReturnValue({
+                serverVersion: '7.8.0',
+                siteName: 'Mattermost',
+                siteURL: 'https://server.com/',
+            });
+            ServerInfo.mockImplementation(() => ({pingServer, fetchConfigData}));
+
+            await serverViewState.handleServerURLValidation({}, 'https://server.com');
+
+            expect(session.fromPartition).toHaveBeenCalledWith('server-validation');
+            expect(pingServer).toHaveBeenCalledWith(validationSession);
+            expect(fetchConfigData).toHaveBeenCalledWith(validationSession);
         });
 
         it('should correct typos in the protocol', async () => {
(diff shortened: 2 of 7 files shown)


src/app/serverHub.ts
```diff
@@ -455,10 +455,10 @@ export class ServerHub {
         const serverInfo = new ServerInfo(server);
         try {
             // Ping server first for pre-auth - config endpoint might be whitelisted
-            await serverInfo.pingServer();
+            await serverInfo.pingServer(session.fromPartition('server-validation'));
```
edgarbellot, line 458: Moving the probes to `session.fromPartition('server-validation')` takes them fully outside the local network policy, since `webRequest` filters apply per-session and the policy is only set on the default session ([initialize.ts#L336](https://github.com/mattermost/desktop/blob/1f424ab212eb8a7d97a74c2125a022d5aae4c12f/src/main/app/initialize.ts#L336)). I think the original bug just needed the validated URL allowed; this instead allows every local and private address.

Is this exemption intentional, or should it instead be scoped to just the URL being validated?

edgarbellot (changes requested):
@devinbinnie Just one comment. Not 100% sure if it's intentional - re-request review if it is.

src/app/serverHub.ts
devinbinnie: @edgarbellot This is intentional, because that `webRequest` handler we added is what's blocking the check. The only way to keep that and still allow the URL would be to add some temporary state to the local network check where we could prefill a URL allowlist, but I didn't want to overcomplicate and have to manage adding/removing those URLs, so I thought this was cleaner.

Let me know if you think there's any issue with this.

pvev: One possible issue is that this session also misses the pre-auth header hook and NTLM allowlist configured on defaultSession. Could editing an existing protected server then fail validation or prompt again for credentials?

edgarbellot: @devinbinnie Agreed on the allowlist, managing add/remove would be error-prone. 

There's some context I left out of my first comment that I should have included, because it's the actual reason I'm asking. Validation isn't only triggered by the user typing a URL - a deep link triggers it on its own:

- `mattermost://<host>:<port>/<path>` reaches [showNewServerModal(prefillURL)](https://github.com/mattermost/desktop/blob/1f424ab212eb8a7d97a74c2125a022d5aae4c12f/src/app/navigationManager.ts#L98) when the host doesn't match a configured server.
- `prefillURL` arrives asynchronously via `getModalInfo()`, so it goes from `undefined` to a value and [componentDidUpdate](https://github.com/mattermost/desktop/blob/1f424ab212eb8a7d97a74c2125a022d5aae4c12f/src/renderer/components/NewServerModal/NewServerModal.tsx#L76-L81) calls `validateServerURL` with no typing and no confirmation. The welcome-screen path does the same from `ConfigureServer`'s mount effect.
- The scheme-less `host:port` gets `https://` prepended at [serverHub.ts#L215-L221](https://github.com/mattermost/desktop/blob/1f424ab212eb8a7d97a74c2125a022d5aae4c12f/src/app/serverHub.ts#L215-L221), then probed over HTTPS and HTTP.

So one click on a link makes the client connect to whatever local address the sender picked, like `127.0.0.1:<any port>` or `169.254.169.254`. It's blind so it's not exfiltration, but it is an attacker-chosen outbound connection into the user's local network.

There might be a cheaper fix: don't auto-validate a prefilled URL. Prefill the field and let the user trigger validation, or pass the provenance through so a deep-link-supplied URL isn't probed automatically. Right now the main process can't tell the two apart - the renderer calls `validateServerURL(serverUrl, server?.id)` identically either way.

That would let you keep the dedicated session exactly as it is. Would that work?

Feel free to re-request again if it's still intentional.

---

@pvev The pre-auth part is real, but it won't re-prompt - the secret modal is triggered by [preAuthHeaderOnHeadersReceivedHander](https://github.com/mattermost/desktop/blob/1f424ab212eb8a7d97a74c2125a022d5aae4c12f/src/main/security/preAuthManager.ts#L35-L64), which is also default-session only, so it never runs here.. And it doesn't block saving. Net effect: a misleading warning banner when editing a protected server.

NTLM we're fine on. `allowNTLMCredentialsForDomains` looks per-session but isn't - in Electron 43 it calls `GetNetworkService()->ConfigureHttpAuthPrefs`, which applies to the whole NetworkService, so the existing hardening still covers `server-validation`.

edgarbellot (changes requested):

src/app/serverHub.ts
devinbinnie: @edgarbellot Thanks for the additional context, if I'm understanding correctly it seems the crux of the issue is actually this bit:

> So one click on a link makes the client connect to whatever local address the sender picked, like 127.0.0.1:<any port> or 169.254.169.254. It's blind so it's not exfiltration, but it is an attacker-chosen outbound connection into the user's local network.

I think that does represent a similar attack vector to what we patched before by disallowing the local connections, but does the validation path allow for a similar exploit to happen? The only thing we are returning from those endpoints are configuration values and to check if the endpoint is working, but it doesn't support sending arbitrary data back to the server. The only initiator of the request is internal code. I don't think the issue was actually just sending the user to open a local service, it was that a malicious server could send that request and get information about a user's computer, but I don't think that can happen here.

I get that we could avoid the "one-click" bit by making validation manual, but that would degrade the UX not only making the user potentially click twice, but it means that we couldn't automatically validate servers from links to any part of that server (which is a helpful QoL feature).

@pvev The pre-auth bit is actually a problem here though, because yes we should probably attach the same listener to the validation partition to ensure we're running through pre-auth (and other auth forms) correctly. I'll make that fix.

edgarbellot: @devinbinnie Went back over it and you're right - no return channel, and the attacker can't shape the request either, so the exploit I had in mind doesn't apply here. Agreed the UX tradeoff isn't worth it either. Thanks for pushing back on this one!

edgarbellot (approved):

pvev (approved):
🎉

PR #3951, opened 2026-08-18
[MM-70249] Validate SiteURL from remote config before constructing URLs

lieut-data:
#### Summary

A server's client configuration `SiteURL` was passed directly to `new MattermostServer()` / `new URL()`, both of which threw on an unparseable string.

When the server reports a `SiteURL` that can't be parsed, log a warning and continue using the URL originally specified by the user, rather than crashing or showing an error page.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-70249

#### Checklist

#### Device Information

macOS 15 (unit tests, `tsc`, and `eslint`)

#### Release Note

```release-note
Fixed an issue where a malformed SiteURL in a configured server's client configuration could persistently crash the Desktop App on launch.
```


src/app/serverHub.ts
```diff
@@ -392,32 +392,31 @@ export class ServerHub {
         }
 
         // If the URL doesn't match the Site URL, set the URL to the correct one
-        if (remoteInfo.siteURL && remoteURL.toString() !== new URL(remoteInfo.siteURL).toString()) {
+        const parsedSiteURL = remoteInfo.siteURL ? parseURL(remoteInfo.siteURL) : undefined;
```
lieut-data, line 395: Easiest to see when compacting whitespace, but in short we now work with the results of `parseURL` instead of trying to `new URL` ourselves.

src/common/servers/MattermostServer.ts
```diff
@@ -31,11 +31,12 @@ export class MattermostServer {
         this.preAuthSecret = preAuthSecret;
     }
 
-    updateURL = (url: string) => {
-        this.url = parseURL(url)!;
-        if (!this.url) {
+    updateURL = (url: string | URL) => {
```
lieut-data, line 34: `parseURL` always accepted a `URL`, so just allow it through instead of requiring a `toString()`.

src/common/servers/MattermostServer.ts
```diff
@@ -31,11 +31,12 @@ export class MattermostServer {
...
+        const parsedURL = parseURL(url);
+        if (!parsedURL) {
             throw new Error('Invalid url for creating a server');
         }
+        this.url = parsedURL;
```
lieut-data, line 39: Don't mutate `this.url` on an invalid URL.

src/common/servers/serverManager.ts
```diff
@@ -180,8 +180,9 @@ export class ServerManager extends EventEmitter {
 
         this.remoteInfo.set(serverId, remoteInfo);
 
-        if (remoteInfo.siteURL && server.url.toString() !== new URL(remoteInfo.siteURL).toString() && isSiteURLValidated) {
-            server.updateURL(remoteInfo.siteURL);
+        const siteURL = remoteInfo.siteURL ? parseURL(remoteInfo.siteURL) : undefined;
+        if (siteURL && server.url.toString() !== siteURL.toString() && isSiteURLValidated) {
+            server.updateURL(siteURL);
```
lieut-data, line 185: We now `parseURL` above, and then just pass it directly to `updateURL` which now accepts the `URL`.

e2e/specs/server_management/invalid_site_url.test.ts
```diff
@@ -0,0 +1,187 @@
...
+            // Site URL parses, so it is never treated as misconfigured — but it failed
+            // validation, so the configured URL must stand.
+            await expect.poll(
+                () => getServerURL(app, serverName),
+                {timeout: 15_000, message: 'Server URL should not be rewritten to an unvalidated Site URL'},
+            ).toContain(unreachableSiteURLServer.origin);
+
+            const mainWindow = app.windows().find((w) => w.url().includes('index'));
+            expect(mainWindow).toBeDefined();
+            expect(await mainWindow!.$('.ErrorView')).toBeNull();
```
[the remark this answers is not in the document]
lieut-data: Good catch — the URL assertion was vacuous. `getServerURL` returns the configured URL from the moment the server is registered, and the client-config request only proves the fetch started, not that the Site URL branch resolved.

Now gated on `ServerManager.getRemoteInfo(serverId).siteURL`, which is only populated once `updateServerInfos` has decided what to do with the reported value, so the URL check runs after the validation it is about. The `.ErrorView` check is now `expectNoErrorView()`, which re-samples across a window rather than reading the DOM once — a single sample cannot distinguish "absent" from "not rendered yet" for an IPC-driven view.

Verified by mutation: flipping `updateRemoteInfo(srv.id, data, false)` to `true` in the unreachable branch now fails the spec with `Expected substring: "http://127.0.0.1:58547" / Received string: "http://127.0.0.1:1/"`. It passed before the change.

Fixed in 596d980a.

src/renderer/components/InvalidSiteURLErrorView.tsx
```diff
@@ -0,0 +1,45 @@
...
+    );
+
+    return (
+        <ErrorView
+            header={header}
+            subHeader={subHeader}
+            showOpenInBrowser={false}
+        />
+    );
+}
```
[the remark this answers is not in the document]
lieut-data: Leaving this as-is. The component directory convention applies to components that own styles; this one has none, and it sits directly next to its two peers — `ConnectionErrorView.tsx` and `IncompatibleErrorView.tsx` — which are the same thing: flat files in `components/` that render `ErrorView` with different copy. Moving only the third into a directory with an empty `.scss` would make the trio less consistent, not more.

e2e/helpers/errorView.ts
```diff
@@ -155,5 +171,38 @@ export async function waitForErrorView(
...
+ */
+export async function expectNoErrorView(
+    app: ElectronApplication,
+    fallback: Page,
+    {samples = 5, intervalMs = 500}: {samples?: number; intervalMs?: number} = {},
+): Promise<void> {
+    for (let sample = 0; sample < samples; sample++) {
+        const window = resolveErrorViewHost(app, fallback);
+        const visible = await window.isVisible('.ErrorView').catch(() => false);
+        expect(visible, 'ErrorView should never appear').toBe(false);
```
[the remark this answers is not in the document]
lieut-data: Agreed — swallowing the failure is especially wrong here, since the spec exists to catch a crash. Dropped the `.catch()` so a page that cannot be queried fails the assertion instead of reading as "no error view", with a comment noting the omission is deliberate.

Spec still passes (3/3, `npx playwright test specs/server_management/invalid_site_url.test.ts --workers=1`). Fixed in 6752da49.

lieut-data:
@devinbinnie, thanks for the offline feedback -- converged on just rendering on the configured SiteURL while logging a warning, never saving the server-supplied SiteURL if invalid.

lieut-data:
E2E tests should be fixed with https://github.com/mattermost/desktop/pull/3960.

devinbinnie (approved):
Thanks @lieut-data 🎉

PR #3955, opened 2026-08-20
Attribute Sentry events to a persisted install ID

lieut-data:
#### Summary

Generate a UUID on application start, persist it to `app-state.json` in the user data directory, and pass it to Sentry as the user id so crash reports from the same install can be grouped together without exposing any PII.

Desktop currently reports zero affected users on every issue, and release health tracks no users at all, because `@sentry/electron` only derives user identity from the client IP and that path is gated behind `sendDefaultPii: true`. The mobile SDKs already do exactly what this PR adds, generating a per-install UUID in their native layer; this brings desktop in line. The value is random, holds no account, server, or device information, and is sent as `user.id` only — no email or username.

Three notes for reviewers, since each looks like it could be simplified but cannot:

- The user has to be passed via `initialScope`, not a later `Sentry.setUser` call. A default integration starts the session during `Sentry.init` and reads the distinct id from the scope at that instant, so setting the user afterwards populates issue counts but leaves release health empty.
- The `Validator` schema entry is required. `stripUnknown: true` means that without it `installId` is dropped on every launch, producing a new UUID each start. It is a plain `Joi.string()` rather than `.guid()` because a constrained field fails the whole schema on a bad value, and `AppVersionManager` responds by wiping the file, taking `lastAppVersion` and friends with it.
- A failed write leaves `installId` in memory and reported to Sentry even though it never reached the disk. CodeRabbit flagged this and asked for the value to be withheld until persistence succeeds; it is left as is on purpose. `SentryHandler` reads the id synchronously during `initialize()` — in the same run as the `AppVersionManager` constructor, and long before the write callback can fire — so there is no moment at which withholding it would help. Deferring would only mean every first launch reports with no user at all. The failure is logged and the next launch generates a new id; an install that cannot write `app-state.json` has larger problems than a rotating one.

If no id is available, Sentry initialises exactly as it does today with the key omitted entirely.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-70378

#### Checklist

- [ ] Has UI changes
- [ ] Run E2E tests by adding label `E2E/Run`

#### Device Information

This PR was tested on: macOS 15 (arm64), production build with the DSN pointed at the `mattermost-desktop-test` project.

Verified against a running app rather than only in unit tests:

- First launch generated an id and persisted it to `app-state.json`.
- The id reached the Sentry scope as `user.id` in `sentry/scope_v3.json`. This is the scope the minidump integration persists and attaches to the crash that killed the previous run, so native crashes carry the attribution too, not just clean JS errors.
- The id reached the session as `did` in `sentry/session.json`, which is what release health counts.
- Restarting reused the same `installId` and the same session `did`, with a new `sid` per run — so repeated sessions from one install attribute to a single user.
- Deleting `app-state.json` produced a different id on the next launch, confirming generation rather than a fixture.

Not yet exercised on Windows or Linux.

#### Screenshots

N/A — no UI changes.

#### Release Note

```release-note
Added a randomly generated identifier, stored locally and stable for the lifetime of an installation, to the crash and error reports sent to Sentry when error reporting is enabled. It allows reports from the same installation to be grouped together and contains no personal information.
```

## Summary by CodeRabbit

* **New Features**
  * Added a persistent installation identifier to recognize the same app installation across launches.
  * Included the installation identifier in error-reporting metadata when available.
* **Bug Fixes**
  * Improved identifier generation and reuse during app startup.
  * Preserved the generated identifier in memory if saving it fails.
  * The app continues to start and report errors when no identifier is available.


lieut-data:
E2E tests should be fixed with https://github.com/mattermost/desktop/pull/3960

src/main/AppVersionManager.ts
```diff
@@ -52,6 +56,24 @@ export class AppVersionManager extends JsonFileManager<AppState> {
         }
         return null;
     }
+
+    get installId(): string | undefined {
```
devinbinnie: Is it intentional that the user upgrading should mean we give them a new instance? Or do we want to just filter by version in Sentry as well, since we already send that up with crash reports.

lieut-data: Thanks, @devinbinnie -- this is context I'm not privy to. Does something reset `app-state.json` on upgrade that I should think about? The intended use case is for this to be semi-permanent, surviving on the happy path in most cases but fine if it rotates on edge cases.

devinbinnie: I was just a bit confused why we added it here vs. `config.json`, but thinking about it this is actually less volatile. And no it doesn't reset automatically as far as I can tell so I think this is actually fine.

I was worried that we were resetting it here.

One thing I am noticing though is that we're setting the value instead of a `getter`, can we refactor that? Maybe we can check in the `init` function if it's missing and set it there? Or a separate helper?

lieut-data: Aha, yes! Good suggestion -- done :)

src/main/AppVersionManager.ts
```diff
@@ -23,6 +27,15 @@ export class AppVersionManager extends JsonFileManager<AppState> {
...
             this.setJson({});
         }
+
+        if (!this.getValue('installId')) {
+            this.setValue('installId', uuid()).catch((e) => {
+                // Only expose an install ID that reached the disk, otherwise Sentry would attribute
+                // this launch to an ID that the next launch won't reuse.
+                delete this.json.installId;
+                log.warn('failed to persist install ID, this launch will not be attributed', e);
+            });
```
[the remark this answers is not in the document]
lieut-data: Not taking this one, and I removed the cleanup it was pointing at.

You are right that `setValue` assigns `this.json.installId` before the write resolves. But `SentryHandler` reads the id *synchronously* during `initialize()` — `initialize()` runs at module evaluation, `initializeConfig()`s promise executor runs synchronously, and `Config.init` -> `reload()` -> `emit('update')` is a synchronous EventEmitter chain down to `sentryHandler.init()`. No `await` yields in between, so `Sentry.init` reads the id in the same synchronous run as the `AppVersionManager` constructor, before any `fs.writeFile` callback can fire.

That means the old `delete this.json.installId` in the rejection handler never did anything useful: Sentry already had the id by the time it ran. I confirmed this by driving the real `AppVersionManager` through the real `SentryHandler` with a failing write — `Sentry.init` still received the generated id. So the fix here was to drop the misleading cleanup, not to add more of it.

Withholding the id until the write lands would not help either; it would just mean every first launch reports with no user at all, which is the case this feature exists to cover. An install that cannot write `app-state.json` has larger problems than a rotating id, so the generated value is now left intact and the failure is logged. Documented in the PR description.

src/main/AppVersionManager.ts
```diff
@@ -23,6 +27,15 @@ export class AppVersionManager extends JsonFileManager<AppState> {
         if (!validatedJSON) {
             this.setJson({});
         }
+
+        if (!this.getValue('installId')) {
+            this.setValue('installId', uuid()).catch((e) => {
```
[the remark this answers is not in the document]
lieut-data: Skipping this one. `uuid` v4 on Node fills a pooled buffer via `crypto.randomFillSync`, so the only way it throws is if Node crypto is unavailable — at which point the Electron main process cannot start regardless. The surrounding code in this same `init()` is equally undefended: `Validator.validateAppState()` can throw, and the pre-existing `this.setJson({})` above does not even `.catch()` its write. Adding a guard for an unreachable path here would be inconsistent with the file rather than safer.

lieut-data:
Apologies for the thrashing here, @devinbinnie -- my agent went a bit overboard pushing commits before I risk accepted the edge cases CodeRabbit identified above. Keeping it simple for now.

lieut-data:
@coderabbit fullreview

src/main/AppVersionManager.ts
```diff
@@ -23,6 +27,16 @@ export class AppVersionManager extends JsonFileManager<AppState> {
         if (!validatedJSON) {
             this.setJson({});
         }
+
+        if (!this.getValue('installId')) {
+            // SentryHandler reads the install ID synchronously during startup, well before this
```
devinbinnie: Nit: probably don't need this comment, I think the code pretty much explains itself here.

src/main/sentryHandler.ts
```diff
@@ -45,6 +47,19 @@ export class SentryHandler {
         return Sentry.flush(3000);
     };
 
+    // The install ID must go through initialScope rather than a later Sentry.setUser call: the
```
devinbinnie: Nit: Same with this one - it's not super relevant to us

devinbinnie (approved):
Couple comments about extraneous comments, non-blocking though. Thanks @lieut-data :)

PR #3974, opened 2026-09-03
MM-70547: Simplify network request evaluation logic

edgarbellot:
See MM-70547.

```release-note
Updated network request evaluation for desktop app windows.
```

## Summary by CodeRabbit

* **Security**
  * Strengthened protection against unauthorized local-network requests.
  * Requests to private IP addresses, localhost, IPv6 local addresses, WebSocket endpoints, and private DNS resolutions are blocked unless explicitly permitted.
  * Configured server origins and public destinations remain accessible, while requests to other ports on the server host are blocked.

### DIFF
--- a/src/main/app/initialize.test.js
+++ b/src/main/app/initialize.test.js
@@ -383,11 +383,20 @@ describe('main/app/initialize', () => {
                 expect(callback).toHaveBeenCalledWith({});
             });
 
-            it('does not cancel requests from non-server web contents', async () => {
+            it('cancels requests from non-server web contents to local/private targets', async () => {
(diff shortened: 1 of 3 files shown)


edgarbellot:
Re the note on `src/main/security/localNetworkAccess.test.ts` (~lines 10-16): leaving the mock factories as-is. The plain named-property form is the dominant convention in this repo's suite (23 test files mock `serverManager` this way vs 3 using `__esModule`/`default`), and it resolves correctly through the existing TS interop, so switching just this file would make it the outlier. Out of scope for this change.

PR #3975, opened 2026-09-03
MM-70567: Adjust download finalization handling

edgarbellot:
## Ticket Link

Fixes [MM-70567](https://mattermost.atlassian.net/browse/MM-70567).

```release-note
Updated file download finalization handling for saved downloads.
```

[MM-70567]: https://mattermost.atlassian.net/browse/MM-70567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/downloadsManager.ts
+++ b/src/main/downloadsManager.ts
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
         }
 
         const bookmark = this.bookmarks.get(this.getFileId(item));
+        let finalState = state;
         if (bookmark) {
-            const func = app.startAccessingSecurityScopedResource(bookmark?.bookmark);
-            fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
-            func();
+            try {
+                const func = app.startAccessingSecurityScopedResource(bookmark.bookmark);
+                try {
+                    fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
+                } finally {
+                    func();
+                }
+            } catch (e) {
+                log.warn('could not move downloaded file to its final location', {e});
+                this.bookmarks.delete(this.getFileId(item));
+                finalState = 'interrupted';
+            }
         }
 
-        await this.upsertFileToDownloads(item, state, bookmark?.originalPath);
+        await this.upsertFileToDownloads(item, finalState, bookmark?.originalPath);
         this.fileSizes.delete(item.getFilename());
         this.progressingItems.delete(this.getFileId(item));
         this.shouldAutoClose();
--- a/src/main/downloadsManager.test.js
+++ b/src/main/downloadsManager.test.js
@@ -3,7 +3,7 @@
 import fs from 'fs';
 import path from 'path';
 
-import {shell} from 'electron';
+import {app, shell} from 'electron';
 import {getDoNotDisturb as getDarwinDoNotDisturb} from 'macos-notification-state';
 
 import Config from 'common/config';
@@ -29,6 +29,7 @@ jest.mock('electron', () => {
         app: {
             getAppPath: jest.fn(),
             getPath: jest.fn(() => '/valid/downloads/path'),
+            startAccessingSecurityScopedResource: jest.fn(),
         },
         WebContentsView: jest.fn().mockImplementation(() => ({
             webContents: {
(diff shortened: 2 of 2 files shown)


src/main/downloadsManager.ts
```diff
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
         }
 
         const bookmark = this.bookmarks.get(this.getFileId(item));
+        let finalState = state;
         if (bookmark) {
```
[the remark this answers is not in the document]
edgarbellot: Leaving this as-is. The copy here is unconditional in the existing behavior; gating it on the download state is a behavior change outside the scope of this fix.

src/main/downloadsManager.ts
```diff
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
-            const func = app.startAccessingSecurityScopedResource(bookmark?.bookmark);
-            fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
-            func();
+            try {
+                const func = app.startAccessingSecurityScopedResource(bookmark.bookmark);
+                try {
+                    fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
```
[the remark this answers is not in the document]
edgarbellot: Leaving this as-is; switching this call to the async API is an unrelated change, out of scope for this fix.

src/main/downloadsManager.ts
```diff
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+                } finally {
+                    func();
+                }
+            } catch (e) {
+                log.warn('could not move downloaded file to its final location', {e});
+                this.bookmarks.delete(this.getFileId(item));
+                finalState = 'interrupted';
```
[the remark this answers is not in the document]
edgarbellot: Leaving this as-is. Reordering the notification would mean restructuring the early-return flow above it, which is out of scope for this change.

edgarbellot:
Re the nitpick on `src/main/downloadsManager.test.js` (~line 276): leaving as-is. `jest.mocked()` is a TypeScript typing helper and this suite is plain JS — no other `.test.js` file in this repo uses it, so this matches the surrounding convention.

edgarbellot:
@coderabbitai resolve

edgarbellot:
@coderabbitai review

src/main/downloadsManager.ts
```diff
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
-            func();
+            try {
+                const func = app.startAccessingSecurityScopedResource(bookmark.bookmark);
+                try {
+                    fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
+                } finally {
+                    func();
+                }
+            } catch (e) {
+                log.warn('could not move downloaded file to its final location', {e});
```
devinbinnie, line 569: As a follow-up, we should probably show the error more clearly in the dropdown, users might get confused if this ever actually comes up in practice.

src/main/downloadsManager.ts
```diff
@@ -556,13 +556,23 @@ export class DownloadsManager extends JsonFileManager<DownloadedItems> {
...
+        let finalState = state;
         if (bookmark) {
-            const func = app.startAccessingSecurityScopedResource(bookmark?.bookmark);
-            fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
-            func();
+            try {
+                const func = app.startAccessingSecurityScopedResource(bookmark.bookmark);
+                try {
+                    fs.copyFileSync(path.resolve(app.getPath('temp'), path.basename(bookmark.originalPath)), bookmark.originalPath);
+                } finally {
```
devinbinnie, line 565: Any reason we couldn't move the finally to the outer try/catch? Or maybe we don't actually need to call this if the above copy fails?

devinbinnie (approved):
Approving to unblock since the fix works, though I am curious about the code structure.

src/main/downloadsManager.ts
edgarbellot: moving the `finally` outside would mean a failure in `func()` itself could escape without being caught, since a `finally` block's own errors aren't caught by its own try's catch

src/main/downloadsManager.ts
edgarbellot: Agreed, created this ticket: https://mattermost.atlassian.net/browse/MM-70569

PR #3987, opened 2026-09-09
Add custom notification sound configuration for DMs and channels

antomicx:
#### Summary
      • Adds settings in the Notifications panel to customize sound for Direct Messages.
      • Adds per-channel / per-user sound overrides.
      • Validates new configuration properties in Validator.ts.
      • Includes unit tests in index.test.ts.

#### Checklist
      [✓] Added or updated unit tests
      [✓] Has UI changes
      [✓] Executed npm run lint:js

#### Release Note
```release-note
    Added settings to customize notification sounds for direct messages and specific channels.
```

## Summary by CodeRabbit

- **New Features**
  - Added customizable notification sounds for direct messages.
  - Added per-channel notification sound overrides, including silent notifications.
  - Added channel discovery across configured servers when creating overrides.
  - Added searchable channel selection and saved channel names for override listings.
  - Added Settings controls to add and remove custom sound overrides.

- **Bug Fixes**
  - Improved notification route matching to prevent incorrect sound overrides.
  - Added safer direct-message URL detection.
  - Improved reliability when loading channels and handling request timeouts.


devinbinnie:
@antomicx Thanks for the PR! While I really like the idea and I think there's something useful here, I think we need to look into polishing up the UX a bit.

The one thing that stands out is use of an ID to specify a custom channel likely won't work for the average user, since they won't know what an ID is. It would be better if maybe they could pick from a list of channels based on their servers, perhaps one that is searchable. The behaviour could be to contact the servers directly for a list of channels they're a member of by team, then cache it in the Settings temporarily to populate a searchable dropdown (similar to the one we use for languages I believe). Then they can pick from that list and that will provide the id/username they need. We'd probably need to cache the display name as well.

cc @matthewbirtch @abhijit-singh if you have any thoughts on the UX. From a code perspective I think it works otherwise.

antomicx:
Hi @devinbinnie,
  
  Thanks for the feedback! That makes total sense, the channel ID was definitely not user friendly.
  
  I just pushed an update to address this:
  
  • Searchable dropdown: Replaced the text input with a searchable select in Settings. It queries the active servers directly on /api/v4/users/me/teams and /api/v4/users/me/channels using the session cookies and caches the list in the Settings view.      
  • Friendly names & DMs: Channels show the team and server name (e.g. Town Square (Team - Server)). For 1-to-1 DMs, it resolves the other speaker's name (Name (@username) (Direct Message - Server)) instead of showing raw IDs.
  • Display name cache: Added channelNotificationSoundNames to the config, so the overrides list shows the readable name instead of the ID.
  • Tests: Added unit tests in serverChannels.test.ts, and verified all unit tests, type checks, and linter pass.
  
  Tested it locally and it works nicely. Let me know what you think!

antomicx:
@coderabbitai resume

antomicx:
@coderabbitai review

antomicx:
@coderabbitai full review

antomicx:
@coderabbitai approve

asaadmahmood:
@antomicx @devinbinnie 

I don’t think these changes belong in the desktop app.

Notification sounds are already a web app / server preference. Users can set a global sound in Settings → Notifications, and since v10.1 they can also set a sound per channel or DM from that channel’s Notification Preferences. Those sync across devices, and desktop already plays whatever sound the web app sends.

Putting a second set of sound settings in desktop Settings would give people two places to configure the same thing, and they’d only work in the desktop app. They wouldn’t sync, and they could conflict with what someone already set in Mattermost.

Desktop Settings should stay focused on OS-level stuff (dock bounce, taskbar flash, unread badge). Product notification prefs should stay in the web app.

The one idea that is interesting is a global “use a different sound for DMs” setting. That’s still a web app Notifications change though, not a desktop one. Per-channel sounds are already covered.

I’d suggest we don’t take this in desktop, and if you want to keep going, the better place for a global DM sound setting would be the web app.

Issue #4007, opened 2026-09-22
[Bug]: KDE Plasma notification history does not retain Mattermost notifications due to desktop-entry mismatch


lushndm:
### Checks before filing an issue

- [x] This issue doesn't reproduce on web browsers (such as in Chrome). If it does, [issue reports go to the Mattermost Server repository](https://github.com/mattermost/mattermost-server/issues).
- [x] I have checked the [issue tracker](https://github.com/mattermost/desktop/issues) and have not found an issue that matches the one I'm filing. This should include closed issues.
- [x] This issue is not a troubleshooting question. Troubleshooting questions go here: https://forum.mattermost.com/c/trouble-shoot/16.
- [x] This issue is not a feature request. You can request features and make product suggestions here: https://mattermost.com/suggestions/.
- [x] This issue reproduces on the most recent [stable version](https://github.com/mattermost/desktop/releases/latest), or the most recent [prerelease version](https://github.com/mattermost/desktop/releases) of the Mattermost Desktop App.
- [x] I have read the [contribution guidelines](https://github.com/mattermost/desktop/blob/master/CONTRIBUTING.md).

### Mattermost Desktop Version

6.3.0

### Operating System

Kubuntu 24.04

### Mattermost Server Version

10.11.19

### Steps to reproduce

1. Install Mattermost Desktop 6.3.0 on Kubuntu 24.04 with KDE Plasma 5.27.12.
2. Ensure that the Plasma system-tray notification history is enabled.
3. Trigger a notification in Mattermost, for example by receiving a direct message or mention.
4. Do not interact with the popup and wait until it disappears.
5. Open the Plasma system-tray **Notifications** widget (bell icon).

### Expected behavior

The Mattermost notification should remain in the Plasma notification history, similar to notifications from other applications such as Spectacle.

<img width="709" height="858" alt="Image" src="https://github.com/user-attachments/assets/6a022dae-7443-4ab5-b2d7-68ff4d0f224c" />

### Observed behavior

The Mattermost notification appears as a popup, but it is not added to the Plasma notification history.

<img width="709" height="791" alt="Image" src="https://github.com/user-attachments/assets/8723cd4e-44d5-4ca3-86dc-0ca83c328ffa" />

### Log Output

```shell
-
```

### Additional Information

- Desktop environment: KDE Plasma 5.27.12
- Notification server: `plasmashell` / `org.freedesktop.Notifications`

### DBus investigation

Using:

```bash
dbus-monitor --session \
[...]

PR #4010, opened 2026-09-22
[GH-4007] Use com.Mattermost.Desktop for the Linux executable name and make it consistent for all packages

devinbinnie:
#### Summary
On Linux the app advertised a notification `desktop-entry` hint (and Wayland/X11 app ID) that didn't match the `.desktop` file we actually install, and the identifiers were inconsistent across our packages. As a result KDE Plasma wouldn't retain Mattermost notifications in its history, and newer desktop portals increasingly reject an app identity they can't resolve to an installed `.desktop` file.

This PR standardizes the Linux identity on the ID `com.Mattermost.Desktop` — the executable name, the generated/installed `.desktop` file, and the app ID / `StartupWMClass` — so it lines up across deb, rpm, tar.gz, and Flatpak and resolves correctly for KDE and portal lookups. Having just fixed a related Linux desktop-entry issue in #3932, we decided to correct this properly everywhere rather than patch a single spot.

#### Ticket Link
https://github.com/mattermost/desktop/issues/4007

#### Device Information
Tested on Ubuntu 24.04 LTS and Fedora 44 (KDE).

#### Release Note
```release-note
Fixed Mattermost notifications not being retained in the KDE Plasma notification history on Linux.
```

#### Potential breaking change:
As part of this, the Linux executable is now named `com.Mattermost.Desktop`, so custom launchers, shortcuts, or symlinks that reference the old `mattermost-desktop` binary may need to be updated.

## Summary by CodeRabbit

- **Packaging**
  - Updated the Linux application identifier and executable naming for more consistent desktop integration.
  - Linux desktop launchers now use the updated application name and window-class identifier.

- **Documentation**
  - Updated Linux installation and configuration instructions to reflect the new launcher, executable, custom data directory, and command paths.
  - Revised setup guidance for moving, linking, and launching the application on Linux.


electron-builder.ts
```diff
@@ -94,6 +94,7 @@ const config = {
             'flatpak',
         ],
         appId: 'com.Mattermost.Desktop',
+        executableName: 'com.Mattermost.Desktop',
```
wiggin77, line 97: Since the executable name will be changed after upgrade, and autostart is true, on reboot it will at first fail to load.  A manual launch of the app will overwrite the `~/.config/autostart/Mattermost.desktop` exec entry and future restarts will work.

Not a huge deal as the user just needs to start it manually, but unexpected.

devinbinnie: Good callout. Just like the potential dock shortcut breaking, I think we'll call this out in the release notes. But this would also be expected when changing the name of the app.

PR #4013, opened 2026-09-23
Prevent duplicate desktop onBrowserHistoryPush listener registration

sbishel:
#### Summary
When a plugin bundles its own copy of utils/browser_history.tsx, the module-level side effect registers a second onBrowserHistoryPush listener. Desktop fires one BROWSER_HISTORY_PUSH per navigation but delivers it to every registered listener — both fire and push two identical adjacent entries into window.history. The Back button then steps between the two identical entries: the page freezes, the tab renames to the stale title from the older entry, and a second click is needed to actually move.

collapseDuplicateHistoryEntry() is called from the existing did-navigate-in-page handler. After each navigation it checks whether the active entry and the one behind it share the same URL. If they do, it removes the older one using navigationHistory.removeEntryAtIndex(activeIndex - 1), leaving the current page in place. This converges correctly regardless of how many listeners pushed.

Changes
- src/app/views/MattermostWebContentsView.ts — collapseDuplicateHistoryEntry() + wired into handleDidNavigateInPage
- Tests — four cases: collapse fires, distinct entries untouched, start-of-stack guard, removal failure does not throw

Known limitation:
matches on URL equality. Two adjacent entries sharing a URL but differing in router state (e.g. same channel, RHS open vs closed) would also be collapsed. Not observed in practice.

#### Release Note
```release-note
Fixes a desktop app bug where the Back button required two clicks to navigate when a plugin was installed that bundled its own browser history module.
```
How to reproduce the bug (before this fix)

A plugin triggers this bug by bundling its own copy of the browser history module that registers a listener at load time. The minimal reproduction pattern:

1. Create browser_history.js in the plugin's webapp/src/:
``` javascript
const browserHistory = {
    push(path) { window.history.pushState(null, '', path); },
};

// This side effect is the problem — it registers a second listener
if (window.desktopAPI?.onBrowserHistoryPush) {
    window.desktopAPI.onBrowserHistoryPush((pathName) => {
        browserHistory.push(pathName);
    });
}

export default browserHistory;
```
2. Import and use it in any plugin component:
``` javascript
import browserHistory from 'src/browser_history';
// ...
browserHistory.push('/some/path');
```
3. Test steps:
1. Build and install the plugin on a Mattermost instance
2. Open the instance in the Desktop app
3. Navigate to any channel or page using the plugin's navigation links
4. Press the Desktop Back button (top-left arrow)
5. Before fix: Back requires two clicks — first click lands on a phantom history entry (tab title may freeze or show the wrong name); second click reaches the actual previous page
6. After fix: Back works in a single click

Note: This bug is Desktop-only and does not affect browser usage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate entries from appearing in browser navigation history when navigating to the same URL consecutively in the main frame.
  * Preserved distinct history entries and left history unchanged for subframe navigations.
  * Improved handling when a history entry cannot be removed, and kept history-button state updated after main-frame navigation.


src/app/navigationManager.ts
```diff
@@ -203,7 +203,6 @@ export class NavigationManager {
                 break;
             }
         }
-
```
devinbinnie: Nit: unnecessary white space removal?

src/app/views/MattermostWebContentsView.ts
```diff
@@ -96,7 +96,7 @@ export class MattermostWebContentsView extends EventEmitter {
                 ipcMain.emit(CLOSE_DOWNLOADS_DROPDOWN);
             }
         });
-        this.webContentsView.webContents.on('did-navigate-in-page', () => this.handlePageTitleUpdated(this.webContentsView.webContents.getTitle()));
+        this.webContentsView.webContents.on('did-navigate-in-page', this.handleDidNavigateInPage);
```
devinbinnie: This doesn't at all fix your issue - what this would be doing is trying to catch multiple title updates and only processing one, which a) doesn't address the browser history push event duplication (that's a separate event altogether) and b) doesn't actually stop a duplicate registry.

What you probably want to do is actually modify https://github.com/mattermost/desktop/blob/a230d5b862d2357c3c561ad7068d8ef9e16d75c2/src/app/preload/externalAPI.ts#L100 to only allow one registry and if it's already registered, reject the others. You'll want to make sure that between refreshes we allow it to re-register, and that it should be view-scoped so that each view can still register once.

devinbinnie (changes requested):

src/app/preload/externalAPI.ts
```diff
@@ -97,7 +101,16 @@ const desktopAPI: DesktopAPI = {
     // Navigation
     requestBrowserHistoryStatus: () => ipcRenderer.invoke(REQUEST_BROWSER_HISTORY_STATUS),
     onBrowserHistoryStatusUpdated: (listener) => createListener(BROWSER_HISTORY_STATUS_UPDATED, listener),
-    onBrowserHistoryPush: (listener) => createListener(BROWSER_HISTORY_PUSH, listener),
+    onBrowserHistoryPush: (listener) => {
```
devinbinnie: This is definitely better, but I'm wondering if we should actually augment `createListener` to handle this instead of this one-off.

I don't think we have to for this change, but if not, maybe we should add a comment here about the one-off.

src/app/preload/externalAPI.ts
```diff
@@ -68,6 +68,10 @@ const createListener: ExternalAPI['createListener'] = (channel: string, listener
     };
 };
 
+// Tracks the remover for the single registered onBrowserHistoryPush listener.
+// Module-level so it resets automatically on page reload when the preload re-runs.
+let browserHistoryPushRemover: (() => void) | undefined;
```
devinbinnie: Do we need to hold this reference here? I would thinking just a `boolean` is enough.

